PR #3, opened 2025-08-29
Add InferenceServerSet CRD and go types/client

lionelvillard:
Add minimal InferenceServerSet CRD.


serving/pkg/apis/serving/v1alpha1/types.go
```diff
@@ -0,0 +1,63 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
+// InferenceServer represents set of inference servers (eg. vLLM inference server)
```
MikeSpreitzer, line 29: Copy-and-paste error on the name here.

serving/pkg/apis/serving/v1alpha1/types.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// InferenceServerSetSpec is the desired state of the inference server.
+type InferenceServerSetSpec struct {
+	// Replicas is the number of desired replicas of inference servers
+	Replicas string `json:"replicas"`
+
+	// Template is the object that describes the inference server that will be created
+	// if insufficient replicas are detected.
+	Template *corev1.PodTemplateSpec `json:"template"`
```
MikeSpreitzer, line 47: Why is this a pointer?

serving/pkg/apis/serving/v1alpha1/types.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// InferenceServerSetStatus reports on the inference servers status.
+type InferenceServerSetStatus struct {
```
MikeSpreitzer, line 51: I think we are going to need this to be more like https://github.com/kubernetes/api/blob/v0.33.4/apps/v1/types.go#L883-L918, right out of the gate.

serving/pkg/apis/serving/v1alpha1/doc.go
MikeSpreitzer, line 1: What about stuff like. https://github.com/kubestellar/kubestellar/blob/v0.29.0-alpha.1/api/control/register.go and https://github.com/kubestellar/kubestellar/blob/v0.29.0-alpha.1/api/control/v1alpha1/register.go ?

serving/go.mod
```diff
@@ -0,0 +1,59 @@
+module github.com/llm-d/llm-d-inference-serving
+
+go 1.24.2
```
MikeSpreitzer, line 3: There are security updates in later patch releases.
There is a new minor version.

lionelvillard:
closing as we are prioritizing the dual-pod approach.

PR #5, opened 2025-09-03
PART 1 - Launcher: the HTTP server

diegocastanibm:
**Context:**
First part of the https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/6#issue-3384215955
Initial HTTP server

**Test:**
Running:
```
cd inference_server/launcher
((.venv) ) diego@macbookpro $ uvicorn --port 8000 --log-level info service.launcher:app
INFO:     Started server process [74360]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:54186 - "POST /vllm HTTP/1.1" 201 Created
INFO:     127.0.0.1:54197 - "DELETE /vllm HTTP/1.1" 202 Accepted
```
Sent POST:
Screenshot 2025-09-04 at 2 18 43 PM

Sent GET:
Screenshot 2025-09-03 at 3 27 24 PM

Sent DELETE:
Screenshot 2025-09-04 at 2 17 30 PM


MikeSpreitzer:
I do not follow everything in the cited IBM Issue, and am not sure how much you are thinking transfers here. We need to discuss this in public.

I see the virtue in sending the swap out request to the engine rather than the launcher, as the request is really a request for the engine to exit.

inference_server/launcher/deploy/service.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: Service
+metadata:
+  labels:
+    app: launcher
+  name: launcher
+spec:
+  ports:
+    - port: 8080
+      protocol: TCP
+      targetPort: 8080
```
MikeSpreitzer, line 11: FYI, this does not need to be specified when it equals the `port`.

inference_server/launcher/service/launcher.py
```diff
@@ -0,0 +1,57 @@
+"""
+Account Service
+
+This microservice handles the lifecycle of Accounts
```
MikeSpreitzer, line 4: This looks like it is copy-pasted from an example, needs to be adjusted for the actual use at hand.

MikeSpreitzer:
This is not the real HTTP server, I see that the `/create_vllm` request is not looking at the request body and is not really creating a vLLM process.

MikeSpreitzer:
Why define the `/create_vllm` request body in a way that is hostile to the goal of having one interface with multiple implementation methods available behind it? I think that the most natural way to do that is to define the interface to traffic in a normal `vllm serve` command line (or the command line of whatever inference server software is involved). It seems to be that the alternative is to define some other thing that can be translated to and from command line arguments. If that is too big a challenge for vllm to encompass, then it is even less plausible that we could do it here.

inference_server/launcher/service/common/status.py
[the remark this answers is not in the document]
diegocastanibm: Right, I have changed the code to use `http.HTTPStatus` and removed `status.py`

diegocastanibm:
> Why define the `/create_vllm` request body in a way that is hostile to the goal of having one interface with multiple implementation methods available behind it? I think that the most natural way to do that is to define the interface to traffic in a normal `vllm serve` command line (or the command line of whatever inference server software is involved). It seems to me that the alternative is to define some other thing that can be translated to and from command line arguments. If that is too big a challenge for vllm to encompass, then it is even less plausible that we could do it here.

We can discuss this during the design meeting if you want, but I think the idea is to parsing the CLI and bypass the CLI vLLM.

inference_server/launcher/requirements.txt
```diff
@@ -0,0 +1,9 @@
+blinker==1.9.0
+click==8.2.1
+Flask==3.1.2
+gunicorn==23.0.0
```
lionelvillard, line 4: as discussed let's use [fastAPI](https://fastapi.tiangolo.com) as this is what vLLM is using.

diegocastanibm: Done!

inference_server/launcher/service/launcher.py
```diff
@@ -0,0 +1,54 @@
...
+            version="1.0",
+        ),
+        HTTPStatus.OK,
+    )
+
+
+######################################################################
+# CREATE A NEW ACCOUNT
+######################################################################
+@app.route("/create_vllm", methods=["POST"])
```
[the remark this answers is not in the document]
diegocastanibm: Done!

diegocastanibm:
> @diegocastanibm I suggest to remove the `deploy` subdirectory. This code is k8s independent.

Yes, makes sense.

MikeSpreitzer:
I have tested commit 0e0f395451f25d994873bf758fd4ec0aae9b5548 and it works as designed.

MikeSpreitzer (approved):
LGTM

MikeSpreitzer:
However, after testing this PR and then checking out another branch, I found unexpected stuff remained.

```console
mspreitz@mjs13 llm-d-fast-model-actuation % find inference_server 
inference_server
inference_server/launcher
inference_server/launcher/service
inference_server/launcher/service/__pycache__
inference_server/launcher/service/__pycache__/launcher.cpython-313.pyc
inference_server/launcher/service/__pycache__/__init__.cpython-313.pyc
```

I do not understand why. It looks to me like the first thing in `.gitignore` covers those files.

PR #11, opened 2025-09-09
Launcher design documentation

diegocastanibm:
Design document for launcher implementation

### DIFF
--- a/docs/launcher.md
+++ b/docs/launcher.md
@@ -3,25 +3,25 @@
 
 ## Description
 
-This document describes a vLLM launcher that can be used to achieve model swapping without changes to vLLM. The launcher initializes and manages various subsidiary inference vLLM servers and keeps the ID and the model (in addition to other information like the type of accelerator and the GPU number). The launcher does basic vLLM code loading and initialization work (like import modules) of the inference server so that this work does not have to be done at the startup of the inference server process, reducing the startup latency. In future work, the launcher could be extended to sleep and wake up the inference servers that have been initialize by its POST method, and support mutiple instances.
+This document describes a vLLM launcher that can be used to achieve model swapping without changes to vLLM. The launcher initializes and manages a subsidiary inference vLLM server and responses with the PID. The launcher does basic vLLM code loading and initialization work (like import modules) of the inference server so that this work does not have to be done at the startup of the inference server process, reducing the startup latency. In future work, the launcher could be extended to sleep and wake up the inference servers that have been initialize by its POST request, and support mutiple instances.
 
 ## Launcher Methods and functionalities
 
-The client/user will use a launcher-specific command to run the launcher. To swap a model in, the controller will issue a POST request (to the launcher) that includes the model reference and the model-specific flags.
+The client/user will use a launcher-specific command to run the launcher. To swap a model in, the Client will issue a POST request (to the launcher) that includes the model reference and the model-specific flags. The launcher sends an `OK` response to the client/controller to indicate that queries can be sent to the vLLM instance directly.
 
 Setting up process to create a new vLLM inference instance using POST:
 
 ```mermaid
 sequenceDiagram
-    participant Controller
+    participant Client
     participant Launcher
     participant vLLM instance HTTP server
 
-    Controller->>Launcher: POST create new instance
+    Client->>Launcher: POST create new instance
     Launcher-->>vLLM instance HTTP server: create vLLM instance
-    Launcher->>Controller: HTTP 200 OK, Data
-    Controller->>vLLM instance HTTP server: POST Request
-    vLLM instance HTTP server->>Controller: JSON text-generated response
+    Launcher->>Client: HTTP 200 OK, Data
+    Client->>vLLM instance HTTP server: POST Request
+    vLLM instance HTTP server->>Client: JSON text-generated response
 ```
 
 An example of a CURL command that could be sent to the launcher can be seen below:
@@ -30,39 +30,34 @@ An example of a CURL command that could be sent to the launcher can be seen belo
 curl -X POST \
   -H "Content-Type: application/json" \
   -d 'JSON_EXAMPLE' \
-  http://localhost:8000/vllm
+  http://localhost:8000/v1/vllm
 ```
 
 where the `JSON_EXAMPLE` is:
 
 ```json
 {
-    "common_vllm_flags":{
-        "str_common_flag": "text",
-        "int_common_flag": 1000
-    },
-    "model": "meta-llama/Llama-3.1-8B-Instruct",
-    "model_specific_flags": {
-        "bool_specific_flag": true,
-        "str_specific_flag": "text"
-    }
-
+  "options": "--model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --port 8005",
+  "env_var": {
+    "VLLM_USE_V1": "1",
+    "VLLM_LOGGING_LEVEL": "DEBUG"
+  }
 }
 ```
 
-To swap a model out, the controller will issue a request that does not include the model reference nor the model-specific flags, but it will include an ID to identify the inference instance to be deleted.
+To swap a model out, the Client will issue a request that does not include the model reference nor the model-specific flags, but it will include an ID to identify the inference instance to be deleted.
 
 Deleting process of a vLLM inference instance using DELETE:
 
 ```mermaid
 sequenceDiagram
-    participant Controller
+    participant Client
     participant Launcher
     participant vLLM instance HTTP server
 
-    Controller->>Launcher: DELETE instance_ID
+    Client->>Launcher: DELETE
     Launcher-->>vLLM instance HTTP server: delete vLLM instance
-    Launcher->>Controller: HTTP 200 OK, Data
+    Launcher->>Client: HTTP 200 OK, Data
 ```
 
 Following the same example that we used for POST, a DELETE CURL command could be:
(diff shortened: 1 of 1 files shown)


docs/launcher.md
```diff
@@ -0,0 +1,86 @@
+
+# vLLM Launcher
+
+## Description
+
+This document describes a vLLM launcher to be used in model swapping techniques. The launcher initializes and manages various subsidiary inference vLLM servers and keep an inventory with it's ID's and the model (in addition to other information like the type of accelerator and the GPU number). The launcher does basic vLLM code loading and initialization work (like import modules) of the inference server so that this work does not have to be done at the startup of the inference server process, reducing the startup latency. In future work, the launcher could be extended to sleep and wake up the inference servers that have been initialize by its POST method.
```
MikeSpreitzer, line 6: This launcher is not used in _all_ model swapping techniques that we have discussed and, presumably, want to be able to discuss with others. I would reword the first sentence, perhaps something like the following.

> This document describes a vLLM launcher that can be used to achieve model swapping without changes to vLLM.

MikeSpreitzer, line 6: s/keep/keeps/

MikeSpreitzer, line 6: Since sleep/wake is not included now, why would this launcher have more than one launched server at a time?

docs/launcher.md
```diff
@@ -0,0 +1,86 @@
...
+Setting up process to create a new vLLM inference instance using POST:
+
+```mermaid
+sequenceDiagram
+    participant Controller
+    participant Launcher
+    participant vLLM instance
+
+    Controller->>Launcher: POST create new instance
+    Launcher-->>vLLM instance: create vLLM instance
```
MikeSpreitzer, line 21: The launcher client needs to know what a reply implies about the state of the launched vLLM instance. Please be explicit. Is there a wait for anything in between these two actions by the Launcher?

docs/launcher.md
MikeSpreitzer, line 6: > GPU number

Does this launcher only work with single-GPU servers? I had expected it could support using more than one GPU of the node. I had expected it could work with other sorts of accelerators.

MikeSpreitzer, line 6: The launcher does not really care what its client is, right? I would replace references to "the dual-pod controller" with simply references to the launcher client/user, without specifying what that is.

docs/launcher.md
```diff
@@ -0,0 +1,86 @@
...
+    participant Controller
+    participant Launcher
+    participant vLLM instance
+
+    Controller->>Launcher: DELETE instance_ID
+    Launcher-->>vLLM instance: delete vLLM instance
+    Launcher->>Controller: HTTP 200 OK, Data
+```
+
+Following the same example that we used for POST, a DELETE CURL command could be:
```
MikeSpreitzer, line 68: Why is the instance ID part of the request body instead of the URL (e.g., `/vllm/{instance-id}`)?

docs/launcher.md
```diff
@@ -0,0 +1,86 @@
...
+
+```bash
+curl -X DELETE \
+  -H "Content-Type: application/json" \
+  -d '{"instance_ID": "0EFA"}' \
+  http://localhost:8000/vllm
+```
+
+Finally, the controller will fetch all the inference ID's, the models, and the accelerators using a GET method. The launcher will keep track of the instance ID's and its models. The information will be sent to the controller:
```
MikeSpreitzer, line 77: An illustration of a reply would be helpful.

docs/launcher.md
aavarghese, line 6: We were discussing the idea of having 1:1 mapping between a launcher process running in a pod and the inference server that fits the inference request. Do we need the launcher "to initialize and manage various subsidiary inference vLLM servers"?

diegocastanibm: Sure

diegocastanibm: I changed to support 1 instance only for now

docs/launcher.md
```diff
@@ -0,0 +1,86 @@
...
+curl -X POST \
+  -H "Content-Type: application/json" \
+  -d 'JSON_EXAMPLE' \
+  http://localhost:8000/vllm
+```
+
+where the `JSON_EXAMPLE` is:
+
+```json
+{
```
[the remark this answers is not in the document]
diegocastanibm: They are the ones that you described in your `dual_pods.md` doc [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/ac6cab29059ed9ca4efc6d844f2efed7f3b70dcb/docs/dual-pods.md?plain=1#L19). They are defined in the vLLM repo:

docs/launcher.md
diegocastanibm: OK

diegocastanibm: I just change it. For know, the launcher will be in the same pod that the inference server and it will initialize just one instance for now

docs/launcher.md
MikeSpreitzer: You seem to be talking about something making some HTTP request to the vLLM instance. Please show it in the diagram.

docs/launcher.md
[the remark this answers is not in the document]
MikeSpreitzer: The server-requesting Pod is the one that gets an assignment of accelerators from the kubelet. My question here is about the other Pods, the server-running Pods. These have a lifecycle independent of the server-requesting Pods. My conclusion is that the server-running Pods need access to all the accelerators, and needs to take responsibility for making each inference server use the accelerator set assigned to the associated server-requesting Pod. I was thinking that this would be done through the CUDA_VISIBLE_DEVICES environment variable, which the launcher can presumably control in the process of launching vLLM.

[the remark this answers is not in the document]
MikeSpreitzer: I am not sure. I will mention it if/when/where I find a problem.

docs/launcher.md
```diff
@@ -0,0 +1,90 @@
...
+  -d 'JSON_EXAMPLE' \
+  http://localhost:8000/v1/vllm
+```
+
+where the `JSON_EXAMPLE` is:
+
+```json
+{
+  "options": "--model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --port 8005",
+  "env_var": {
```
MikeSpreitzer, line 41: This field name is singular, but the contents are plural. I would name this "env" or "env_vars".

docs/launcher.md
MikeSpreitzer: While the launcher does not really care, I have realized that for clarity we may want to distinguish between two sorts of clients: management clients and inference clients.

docs/launcher.md
```diff
@@ -0,0 +1,96 @@
...
+Following the same example that we used for POST, a DELETE CURL command could be:
+
+```bash
+curl -X DELETE \
+  -H "Content-Type: application/json" \
+  -d '{"instance_ID": "0EFA"}' \
+  http://localhost:8000/vllm
+```
+
+Finally, the controller will fetch the inference ID, the models, and the accelerator using a GET method. The launcher will keep track of the instance ID and model. The information will be sent to the controller:
```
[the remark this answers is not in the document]
MikeSpreitzer: The document should show that path.

docs/launcher.md
```diff
@@ -0,0 +1,90 @@
...
+sequenceDiagram
+    participant Client
+    participant Launcher
+    participant vLLM instance HTTP server
+
+    Client->>Launcher: POST create new instance
+    Launcher-->>vLLM instance HTTP server: create vLLM instance
+    Launcher->>Client: HTTP 200 OK, Data
+    Client->>vLLM instance HTTP server: POST Request
+    vLLM instance HTTP server->>Client: JSON text-generated response
```
MikeSpreitzer, line 24: This request/response interaction is part of _using_ the vLLM instance rather than _creating_ it, right? If so then it does not belong in this sequence diagram.

docs/launcher.md
MikeSpreitzer: Why not update it in this PR?

docs/launcher.md
```diff
@@ -0,0 +1,90 @@
...
+
+    Client->>Launcher: GET
+    Launcher->>Client: HTTP 200 OK, Data
+```
+
+JSON reply example:
+
+```json
+{
+  "status": <"started" | "stopped" | "terminated">,
```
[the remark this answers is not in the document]
MikeSpreitzer: Do you mean that "terminated" appears only in the response to a DELETE? If so then it should be documented there rather than under GET.

docs/launcher.md
MikeSpreitzer: This doc should be updated too

docs/launcher.md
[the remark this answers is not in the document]
diegocastanibm: yes, you are right. I'll documented in the `DELETE` method

docs/launcher.md
```diff
@@ -0,0 +1,96 @@
...
+{
+  "options": "--model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --port 8005",
+  "env_vars": {
+    "VLLM_USE_V1": "1",
+    "VLLM_LOGGING_LEVEL": "DEBUG"
+  }
+}
+```
+
+To swap a model out, the Client will issue a request that does not include the model reference nor the model-specific flags, but it will include an ID to identify the inference instance to be deleted.
```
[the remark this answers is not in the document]
diegocastanibm: Good catch. I'll fix it

docs/launcher.md
MikeSpreitzer: I saw and merged the PR to the code. This thread is not about the code, it is about the documentation.

docs/launcher.md
MikeSpreitzer: I thought that maybe this document was intended to have both management request/response interactions and inference request/response interactions; in that case it would be useful to make that distinction in this document.

If I understand correctly, this document now has only management interactions.

MikeSpreitzer (approved):
This is good enough to commit. The last comment that I made can be addressed in follow-on.

Issue #18, opened 2025-09-10
"Pool manager" to create a set of pre-warmed/reusable inference server pods


aavarghese:
...these inference server pods will run the launcher process that starts up a VLLM server and puts it to sleep.

What is the pool manager to be implemented in Go responsible for?

- inventory of inference server pods
- makes sure pool always has a min size of pods ready to be woken up
- gives up pod to the dual pod controller when request comes in, and retrieves pod back when done

jimcadden:
> What says what to pre-warm? That sounds like it is requiring guidance from a llm-d component that does not exist (predicting future needs for inference servers). I thought that we were pursuing a lazy approach (creating a new Pod when a need arises).

We may want a way to populate a "fresh" cluster with sets of pre-warmed instances in anticipation of future hot starts. For example, "deploy a sleeping instance of model M to 4 GPUs on node N"

MikeSpreitzer:
> For example, "deploy a sleeping instance of model M to 4 GPUs on node N"

Who/what would say that?

There would need to be an API, between the user/client saying that and the implementation. As far as I know, this API has not been designed/reviewed.

PR #15, opened 2025-09-09
WIP: dual pods design

lionelvillard:
First stab at the overall dual-pods architecture and sequence flows.

PR #21, opened 2025-09-12
Pool manager (sub-component of DP controller) architecture docs

aavarghese:
Issue #18 

1st commit has a diagram, next commit will have README.

Go code in separate PR.

### DIFF
--- a/docs/pool-manager.md
+++ b/docs/pool-manager.md
@@ -1,6 +1,6 @@
 # Pool Manager overview
 
-👉 The Pool Manager is a sub-component of the dual-pod controller system. It provisions and manages a pool of idle pods that run only the Launcher process. When requested, it transitions these pods through different lifecycle states (idle → active → sleeping → awake → deleted). This ensures fast activation of server-running vLLM pods while minimizing cold-start latency.
+👉 The Pool Manager is a sub-component of the dual-pod controller system. It provisions and manages a pool of idle pods that run only the Launcher process. When requested, it transitions these pods through different lifecycle states (idle → active → deleted). This ensures fast activation of server-running vLLM pods while minimizing cold-start latency.
 
 ## Sequence Flow
 
@@ -12,8 +12,8 @@
 - Pool Manager issues POST /v1/vllm to the pod's launcher with server-patch options,
 - Launcher starts the vLLM inference server subprocess inside the same pod,
 - Pool Manager patches labels so the pod joins an IGW InferencePool,
-- Requests may be routed directly to the pod bypassing the launcher,
-- On scale-down, Pool Manager may issue DELETE /v1/vllm to the launcher, terminating vLLM before pod deletion.
+- Requests will be routed directly to the vLLM inference server process bypassing the launcher,
+- On scale-down, the Pool Manager issues DELETE /v1/vllm to the pod's launcher, terminating the vLLM subprocess. The server-running pod remains running with the launcher, and its status set back to idle so it can be reused for future allocations if there aren't a sufficient number of idle pods.
 
 ## Pod States
 


docs/pool-manager.md
```diff
@@ -0,0 +1,74 @@
+# What is the Pool Manager and what does it do?
+
+👉 A component that provisions and manages vLLM pods within the Dual-Pod (DP) Controller architecture. Can be a Kubernetes native service that maintains a pool of idle pods running just the launcher, pre-labeled for IGW’s InferencePool, and transitions them through some different lifecycle states (idle → active → sleeping → deleted) to ensure fast activation of server-running vLLM pods while minimizing cold-start latency.
```
MikeSpreitzer, line 3: A given launcher Pod will not match the label selector for one InferencePool throughout the Pod's lifetime. The whole point of model switching is that this Pod will run different models at different times. Thus, it needs to have its labels switched along with the model, so that it matches the right InferencePool at the right time.

I presume you mean to allow a sleeping to active transition. We have generally been calling that "awake", in alignment with the vLLM "/sleep" and "/wake_up" calls.

docs/pool-manager.md
```diff
@@ -0,0 +1,74 @@
...
+
+Flow may look like this for create:
+
+- EPP creates a ISR pod,
+- DP Controller watches it and asks Pool Manager for a inference server pod,
```
MikeSpreitzer, line 8: Both the user/client-created Pod and the one that actually runs the inference server can be described as "inference sever pods". That is why I use two terms to distinguish them: server-requesting Pod and server-running Pod.

docs/pool-manager.md
```diff
@@ -0,0 +1,74 @@
...
+- Pool Manager allocates an idle pod → sends POST /vllm to its launcher with server patch,
```
MikeSpreitzer, line 9: #15 says `/v1/vllm`

docs/pool-manager.md
```diff
@@ -0,0 +1,74 @@
...
+- Launcher starts the vLLM inference server inside the same pod,
+- Pool Manager patches labels to add to IGW InferencePool.
+
+Queries can now reach the inference server directly.
+
+## Pod States (WIP)
+
+Idle state at startup — the pod should boot with just the launcher (no model loaded). When dual-pod requests for an endpoint and Pool Manager assigns one, it calls POST /vllm inside the pod to start vllm serve. Once active, the pod becomes a valid serving endpoint in the InferencePool.
```
MikeSpreitzer, line 17: > Once active, the pod becomes a valid serving endpoint in the InferencePool.

That skips over some important stuff. The dual-pod controller needs to set the labels on the server-running Pod so that this Pod matches the right InferencePool. The rules for that are described in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/839ede5e77a364fba52a1108dd44ba4de7c16a4a/pkg/api/interface.go#L38-L43

docs/pool-manager.md
```diff
@@ -0,0 +1,74 @@
+# What is the Pool Manager and what does it do?
+
+👉 A component that provisions and manages vLLM pods within the Dual-Pod (DP) Controller architecture. Can be a Kubernetes native service that maintains a pool of idle pods running just the launcher, pre-labeled for IGW’s InferencePool, and transitions them through some different lifecycle states (idle → active → sleeping → deleted) to ensure fast activation of server-running vLLM pods while minimizing cold-start latency.
+
+Flow may look like this for create:
+
+- EPP creates a ISR pod,
```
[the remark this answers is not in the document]
aavarghese: Meant to say the EPP scheduler creates the server requesting ISR pod. I'll be more precise.

docs/pool-manager.md
```diff
@@ -0,0 +1,74 @@
...
+
+Pool manager ensures there are enough idle pods labeled into the pool so EPP always finds capacity.
+
+### Pod spec: (WIP)
+
+Each vLLM pod contains:
+
+ - Launcher (receives POST/DELETE to start/stop inference server).
+
+ - Inference Server (vLLM, activated by Launcher).
```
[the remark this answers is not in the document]
MikeSpreitzer: My point here is that the text is not very clear about the sort of containment. It would be clearer to say that the Pod has one container with one main process (the launcher) and multiple subprocesses (the launched vLLM instances).

docs/pool-manager.md
```diff
@@ -0,0 +1,83 @@
+# What is the Pool Manager and what does it do?
+
+👉 A component that provisions and manages vLLM pods within the Dual-Pod (DP) Controller architecture. Can be a Kubernetes native service that maintains a pool of idle pods running just the launcher, pre-labeled for IGW’s InferencePool, and transitions them through some different lifecycle states (idle → active → sleeping → deleted) to ensure fast activation of server-running vLLM pods while minimizing cold-start latency.
```
MikeSpreitzer, line 3: > pre-labeled for IGW’s InferencePool

A launcher with no awake vLLM instance has no InstancePool that it should be labeled for.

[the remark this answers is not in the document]
aavarghese: Correct, the intention was that the state of the vllm instance running inside the pod will be reflected in the server-running pod's status.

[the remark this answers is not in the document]
aavarghese: Ok, I was not aware that the launcher was only part of the third stage. This might warrant a rethought of the plans here in this PR..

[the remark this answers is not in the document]
aavarghese: Re-worded in latest.

docs/pool-manager.md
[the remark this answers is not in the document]
aavarghese: Done.

docs/pool-manager.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Example:
+
+- A user creates a server-requesting pod,
+- The Dual-Pod Controller watches it and asks Pool Manager for a server-running pod,
+- Pool Manager allocates an idle pod from its pool,
+- Pool Manager issues POST /v1/vllm to the pod's launcher with server-patch options,
+- Launcher starts the vLLM inference server subprocess inside the same pod,
+- Pool Manager patches labels so the pod joins an IGW InferencePool,
+- Requests may be routed directly to the pod bypassing the launcher,
```
diegocastanibm, line 15: I do not get the `may be` in this case. I think requests will always be routed to the inference pod directly

diegocastanibm (approved):
So far, makes sense to me. I just have one tiny comment. 
I approve it, but I suggest to for @lionelvillard or @MikeSpreitzer to approve it too (just to be sure we are all aligned). Also, the building is failing, but I do not think this is your fault

docs/pool-manager.md
MikeSpreitzer: > As we discussed during call today, we won't have sleeping vllms in the initial phase

Unfortunately, I was not on that call. I am a bit confused. The rest of this document pretty clearly seems to contemplate sleeping vLLM instances.

docs/pool-manager.md
```diff
@@ -0,0 +1,72 @@
+# Pool Manager overview
+
+👉 The Pool Manager is a sub-component of the dual-pod controller system. It provisions and manages a pool of idle pods that run only the Launcher process. When requested, it transitions these pods through different lifecycle states (idle → active → sleeping → awake → deleted). This ensures fast activation of server-running vLLM pods while minimizing cold-start latency.
+
```
lionelvillard, line 4: I would remove `→ sleeping → awake →` as it's not in-scope of this PR.

Issue #23, opened 2025-09-12
Design and Implement the readiness relay


lionelvillard:
The Readiness container becomes ready only when
1.  the inference server pod is ready
2. it is bound to the inference server request pod containing the readiness container.

MikeSpreitzer:
See https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/15#discussion_r2350777846

lionelvillard:
@MikeSpreitzer thanks for the proposed design. 

https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/29 takes a slightly different approach:
- The stub in the server-requesting pod can be sent an HTTP **`PUT`** request that conveys the **IP** of the real inference server container. The reason for `PUT` is we cannot assume server-running pods won't restart, as you know. 
- Once given that **IP**, the stub itself responds positively to readiness requests on itself.
- After the dual-pod controller **retrieves**/**configures** (let me know if you find a better word. The idea is the controller does not always create server-running pods) a server-running Pod, the controller waits for the real inference server *pod* to be ready.
- Once the server-running pod is ready, the controller send an HTTP PUT request to the server-requesting pod.

This approach is IMO simpler as the readiness orchestration is encapsulated inside the dual-pod controller only.

Am I missing something?

MikeSpreitzer:
A Pod does not restart. A container can terminate and restart.

lionelvillard:
> What is the point of giving the stub/requester pod an IP address that it does not use? As far as I can tell, the design you stated passes a boolean encoded as (set IP)/(remove IP).

that's a good point. It's a remnant of the older design.

> A Pod does not restart. A container can terminate and restart.

but a Pod can be evicted. 

> (a) this reduces latency (I am not sure what alternative you have in mind, but it will have one or two additional relay steps and possibly slower polling)
> (b) this distributes the polling load rather than centralizes it (if you were thinking of having the dual-pod controller do the polling).

I'm not thinking about polling but instead relying on the existing dual-pod controller pod watch monitoring Pod ready conditions. (b) does not apply.

Note that polling the readiness probe of the inference server container is not enough. For instance the decode pod (for P/D) currently contains 2 containers and both need to the ready. The server-requester pod needs to relay pod readiness, not container readiness. 

As for latency requirements, the server-requester pod readiness requirements come from use cases like rolling updates. Latency is not so much of an issue.

[...]

PR #29, opened 2025-09-15
Add Readiness command

lionelvillard:
Part of https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/23

The readiness command starts 2 HTTP servers, one for k8s probes, the other one for exclusive dual-pod controller to readiness container communication. The second server exposes one endpoint, `/ip`, use by the controller to set/update/unset inference-server pod ip.


dockerfiles/Dockerfile.readiness
```diff
@@ -0,0 +1,37 @@
...
+ARG BUILD_REF
+RUN go build -a -o bin/readiness cmd/readiness/main.go
+
+# Use ubi9 as a minimal base image
+# Refer to https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5 for more details
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+WORKDIR /
+COPY --from=builder /workspace/bin/readiness /app/readiness
+
+USER 65532:65532
```
MikeSpreitzer, line 35: IIRC, this gives OpenShift indigestion (i.e., requires special privileges and handling). IIRC, the easy path with OpenShift is to not specify the userid and groupid in the image, because OpenShift wants to assign those.

pkg/server/requester/probes/server_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	ipCh := make(chan string, 1)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go watchForIPs(ctx, ipCh)
+
+	ipCh <- "192.168.1.100"
+
+	time.Sleep(50 * time.Millisecond) // give goroutine time to process
```
MikeSpreitzer, line 76: I have been surprised by the business of CI servers. 50ms sounds like a dangerously short time to wait.

pkg/server/requester/probes/server_test.go
MikeSpreitzer, line 1: These are checksum tests. They test that the individual functions still do what they were originally intended to do. IMO it is more useful to have behavioral tests, which test that the package accomplishes what its interface says it does.

cmd/readiness/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"k8s.io/klog/v2"
+)
+
+func main() {
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	// set up signals so we handle the shutdown signal gracefully
+	ctx := signals.SetupSignalHandler()
```
MikeSpreitzer, line 36: I do not see a need for a signal handler. There is no cleanup that needs to be done at termination. We can just let the OS terminate the process.

cmd/readiness/main.go
```diff
@@ -0,0 +1,75 @@
...
+	go func() {
+		defer wg.Done()
+
+		err := probes.Start(ctx, probesPort, ipCh)
+		if err != nil {
+			logger.Error(err, "failed to start requester probes server")
+		}
+	}()
+
+	wg.Wait()
```
MikeSpreitzer, line 74: This seems much more complicated than necessary to me. I see no need for two HTTP servers. I see no need for two additional packages. I see no need for the case of taking DOWN readiness. I see only a need for one file, one HTTP server, two entries in the mux (`/readyz` and `/dual-pod/v1/inference-server-ip`). Communication between the two handlers by an atomic that holds the IP (or not).

lionelvillard: 2 servers for security reasons as clients are different. That way it's easy to setup network policies.

Inference server pods can restart which then can receive a new IP.

cmd/readiness/main.go
lionelvillard: it doesn't hurt. It's a common pattern.

pkg/server/requester/probes/server_test.go
lionelvillard: it's a test

MikeSpreitzer: I expect that the unit tests will be required to pass in CI. So we do not want it to be a flaky test.

cmd/readiness/main.go
MikeSpreitzer: A server-running Pod is not "normally" deleted --- except by the dual-pod controller when that deletion is desired. A Node can be deleted, which cascades to deletion of the Pods on it. IIRC, there are also unusual circumstances in which Kubernetes will evict a Pod from a Node.So we _do_ need to think about what happens when something besides the dual-pod controller deletes a server-running Pod, which I hadn't done.

pkg/server/requester/probes/server_test.go
lionelvillard: 50ms is a lot of time. Anyway I can increase.

MikeSpreitzer:
Should this PR be closed now, having been merged into #33?

PR #32, opened 2025-09-15
Add build readiness image gha

lionelvillard:
Add a GitHub action to build the readiness image. The actual readiness Dockerfile will be provided in a follow-up PR

### DIFF
--- a/.github/workflows/build-readiness-image.yml
+++ b/.github/workflows/build-readiness-image.yml
@@ -1,0 +1,62 @@
+name: Readiness Image Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: readiness
+
+jobs:
+  build-llm-d-readiness:
+    strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-readiness-image.yml
```diff
@@ -0,0 +1,62 @@
...
+      fail-fast: false
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Log in to Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
```
MikeSpreitzer, line 31: Given the choices here for username and password, the registry is not a free variable.

.github/workflows/build-readiness-image.yml
MikeSpreitzer, line 1: > Dockerfile will be provided in a follow-up PR

So this workflow will fail until that other PR is merged, right? Do we have to have that grief?

.github/workflows/build-readiness-image.yml
MikeSpreitzer: In other words, the registry has to be GitHub's. Having a variable gives only an illusion of choice.

PR #33, opened 2025-09-16
[rebased onto #29] Implement the requester and the GPU allocator

### LINKED ISSUE #19: Build the inference-server-gpu-allocator
We need to build the 'inference-server-gpu-allocator' as proposed in #15.

The responsibility of this components is to
- get GPUs allocated;
- expose an interface to other components to say what are the allocated GPUs.

The interface is defined as
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/b6aa731247814c38364b2a37363e8ceb22775590/pkg/stub/api/interface.go#L5-L13

waltforme:
Since this PR is based on #29, we can either
- merge #29, then rebasing onto main and merging this PR, or
- merge this PR and close #29.

#29 imlements the readiness/requester.
On top of #29, this PR also implements [the GPU allocator](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/19).

Closes #19


cmd/gpu-allocator/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Check the readiness and exposed GPU.
+```console
+(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$ kubectl get po my-server-request -owide
+NAME                READY   STATUS    RESTARTS   AGE   IP           NODE               NOMINATED NODE   READINESS GATES
+my-server-request   1/1     Running   0          57s   10.0.0.106   ip-172-31-58-228   <none>           <none>
+(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$ thepodip=10.0.0.106
+(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$ curl $thepodip:8080/readyz
+ready(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$
+(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$ curl $thepodip:8080/dual-pod/accelerators
```
MikeSpreitzer, line 46: Maybe put a `/v1` in that path?

cmd/gpu-allocator/README.md
```diff
@@ -0,0 +1,49 @@
...
+["GPU-b88683d2-a275-15af-5633-d569397bd622"]
```
MikeSpreitzer, line 47: We will need the answer in terms that can be put in `CUDA_VISIBLE_DEVICES`. Does vLLM succeed when GPU UUIDs are put there?

cmd/gpu-allocator/main.go
```diff
@@ -0,0 +1,88 @@
...
+// There are at least two types of 'readiness' in the server-requesting pod.
+// 1. The readiness to accept queries asking which GPU(s) are allocated.
+// This readiness is a prerequisite of the corresponding server-running pod
+// because the server-running pod uses the allocated GPU(s).
+// 2. The relayed readiness of the corresponding server-running pod.
+// This readiness is part of the interface for a user to manage his/her
+// inference servers. More broadly, a user uses the server-requseting pod
+// as the interface to a) specify the 'desired state' via the server-requesting
+// pod's annotations, b) monitor the relayed readiness of the corresponding
+// server-running pod as part of the 'reported state'.
```
MikeSpreitzer, line 68: We intend the user to not have to know about server-running pods and relayed readiness. I would simplify part b to speak simply and directly about inference server readiness.

cmd/gpu-allocator/main.go
```diff
@@ -0,0 +1,88 @@
...
+//
+// Here we are dealing with the 1st type of readiness.
+func readyHandler(w http.ResponseWriter, r *http.Request) {
```
MikeSpreitzer, line 71: Wouldn't it be simpler to call `getGPUUUIDs` exactly once in this whole file, in `main`, remember the result, and then start serving HTTP?

Dockerfile.gpu-allocator
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.24 AS builder
```
lionelvillard, line 1: I suggest to follow what's other llm-d projects are doing. See here: https://github.com/llm-d/llm-d-inference-scheduler/blob/main/Dockerfile.

cmd/gpu-allocator/README.md
lionelvillard: it should: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#env-vars

Dockerfile.gpu-allocator
MikeSpreitzer: +1

(A) Pulling from Dockerhub is always something that we live to regret.

(B) Using the latest version of Go is good.

cmd/gpu-allocator/README.md
waltforme: Shouldn't we use `NVIDIA_VISIBLE_DEVICES` here instead of `CUDA_VISIBLE_DEVICES`?

cmd/gpu-allocator/main.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"net/http"
+	"os/exec"
+	"strings"
+
+	stubapi "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/stub/api"
+)
+
+func getGPUUUIDs() ([]string, error) {
+	cmd := exec.Command("nvidia-smi", "--query-gpu=uuid", "--format=csv,noheader")
```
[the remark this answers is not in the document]
waltforme: No. That's not safe. Because as you know, the nvidia device plugin may also be configured to talk about volume mounts when returning allocations to the kubelet.

In that case, the envar reads something like
```
                "NVIDIA_VISIBLE_DEVICES=/var/run/nvidia-container-devices",
```

cmd/gpu-allocator/README.md
waltforme: Added.

cmd/gpu-allocator/main.go
waltforme: Revised to do so.

cmd/gpu-allocator/main.go
MikeSpreitzer: It sounds like we agree on the transparency. My problem is the wording in the current draft of this PR. Let me take it from the top. I currently see the following.

```golang
// ... More broadly, a user uses the server-requseting pod
// as the interface to a) specify the 'desired state' via the server-requesting
// pod's annotations, b) monitor the relayed readiness of the corresponding
// server-running pod as part of the 'reported state'.
```

The sentence starts with "a user uses". That says this is from the user's point of view. I think that we agree that the user looks at the stub container's readiness and takes that to reflect the inference server's readiness. As far as readiness is concerned, the dual-pod design maintains the illusion that the inference server runs in the stub container. I think that this is what you mean by transparency. As far as readiness is concerned, the user and other clients do not need to know or care about the server-running Pod and the fact that readiness is relayed from there. So I suggest rewording along the following lines.

```golang
// ... More broadly, a user uses the server-requseting pod
// as the interface to a) specify the 'desired state' via the server-requesting
// pod's annotations, b) monitor the readiness of the inference server.
```

cmd/gpu-allocator/main.go
```diff
@@ -0,0 +1,105 @@
...
+// This readiness is part of the interface for a user to manage his/her
+// inference servers. More broadly, a user uses the server-requseting pod
+// as the interface to a) specify the 'desired state' via the server-requesting
+// pod's annotations, b) monitor the relayed readiness of the corresponding
+// server-running pod as part of the 'reported state'.
+//
+// Here we are dealing with the 1st type of readiness.
+func readyHandler(w http.ResponseWriter, r *http.Request) {
+	if len(GPUUUIDs) == 0 {
+		http.Error(w, "not ready", http.StatusServiceUnavailable)
```
MikeSpreitzer, line 81: No; even in this case, this server is ready for its gpuHandler method to be invoked.

cmd/gpu-allocator/main.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func main() {
+	err := getGPUUUIDs()
+	if err != nil {
+		fmt.Printf("Error getting GPU UUIDs: %v\n", err)
+	}
+
+	http.HandleFunc("/v1"+stubapi.AcceleratorQueryPath, gpuHandler)
+	http.HandleFunc("/readyz", readyHandler)
```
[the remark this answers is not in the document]
waltforme: I will remove the readiness probe for the gpu-allocator when I rebase this PR onto #33.

cmd/gpu-allocator/main.go
waltforme: I will remove the readiness probe for the gpu-allocator when I rebase this PR onto https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/29.

cmd/gpu-allocator/README.md
MikeSpreitzer: @waltforme: no. You and I have experimentally confirmed that NVIDIA_VISIBLE_DEVICES is not a control knob that a PodSpec can usefully set.

cmd/gpu-allocator/main.go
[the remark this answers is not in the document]
MikeSpreitzer: Oh, ugh.

cmd/gpu-allocator/README.md
```diff
@@ -0,0 +1,49 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: my-server-request
+  labels:
+    app: my-server-request
+spec:
+  containers:
+    - name: gpu-allocator
+      image: quay.io/junatibm/gpu-allocator:latest
```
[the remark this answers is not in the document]
waltforme: Exactly. Addressed this in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/33/commits/1902e1f42836b667c346850592d98aa91b68ac28

cmd/gpu-allocator/main.go
waltforme: Rebased onto #29 and removed.

cmd/gpu-allocator/README.md
waltforme: Our dual-pods technique is special, because
1. By design, the server-runinng pod will bypass the nvidia device plugin, making the user-specified `NVIDIA_VISIBLE_DEVICES` the exactly one instance (instead of two instances) of the envar.
2. Since the user-specified `NVIDIA_VISIBLE_DEVICES` takes precedence over the vLLM image's baked-in `NVIDIA_VISIBLE_DEVICES=all`, the vLLM container will respect the former, and sees precisely what specifed by the user instead of all GPUs.

Therefore, `NVIDIA_VISIBLE_DEVICES` _is_ a precise control knob for the dual-pods use case.

lionelvillard (approved):
LGTM

cmd/requester/README.md
```diff
@@ -0,0 +1,78 @@
+This document shows the steps to exercise the requester in a local k8s environment.
+
+Build the requester container image (use your favorate `REQUESTER_IMG_REG`).
+```shell
+REQUESTER_IMG_REG=quay.io/my-namespace
```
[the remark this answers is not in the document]
MikeSpreitzer: @waltforme: as @lionelvillard's comment illustrates, I think that it would be less confusing if you delete the line `REQUESTER_IMG_REG=quay.io/my-namespace`; the text already tells the user to set it to whatever value works for them, and as @lionelvillard's comment also illustrates, a displayed shell command gets far more attention than text. Without the wrong stuff displayed, the poor reader is forced to actually read.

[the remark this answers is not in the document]
waltforme: Makes sense. Deleted the line.

cmd/gpu-allocator/README.md
MikeSpreitzer: Oh, right, when requesting zero GPUs, the use's specified NVIDIA_VISIBLE_DEVICES is passed through.

Regardless of which variable is used, the more important question here is whether GPU UUIDs are consumable by vLLM. See the earlier comments in this thread.

pkg/server/requester/probes/server.go
```diff
@@ -0,0 +1,104 @@
...
+				ready.Store(false)
+				logger.Info("received empty IP, setting readiness to false")
+
+			} else {
+				ready.Store(true)
+				logger.Info("received IP, setting readiness to true", "ip", ip)
+			}
+		}
+	}
+}
```
MikeSpreitzer, line 66: This function could be deleted if `Start` were simply given the `ready atomic.Bool`. In other words, the interface between the two subpackages were simplified to just be this atomic, rather than a channel.

pkg/server/requester/spi/server.go
```diff
@@ -0,0 +1,157 @@
...
+		logger.Error(err, "failed to get GPU UUIDs")
+	}
+	if len(gpuUUIDs) == 0 {
+		logger.Error(fmt.Errorf("no GPUs found"), "no GPU UUIDs available")
+	} else {
+		logger.Info("Got GPU UUIDs", "uuids", gpuUUIDs)
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc(strings.Join([]string{"GET", "/v1" + stubapi.AcceleratorQueryPath}, " "), gpuHandler(gpuUUIDs))
```
MikeSpreitzer, line 130: Perhaps `stubapi.AcceleratorQueryPath` should include the "/v1".

pkg/server/requester/spi/server.go
```diff
@@ -0,0 +1,157 @@
...
+	mux.HandleFunc("PUT /ip", ipPutHandler(ipCh))
+	mux.HandleFunc("DELETE /ip", ipDeleteHandler(ipCh))
```
MikeSpreitzer, line 132: Shouldn't these also include "v1"?

cmd/requester/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+package main
+
+import (
+	"flag"
+	"os"
+	"sync"
+
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/probes"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/spi"
```
MikeSpreitzer, line 25: Factoring out these small pieces into two separate packages is, in my opinion, unnecessary complexity.

cmd/requester/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+Check the allocated GPU.
+```console
+(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$ kubectl get po my-server-request -owide
+NAME                READY   STATUS    RESTARTS   AGE   IP           NODE               NOMINATED NODE   READINESS GATES
+my-server-request   0/1     Running   0          26s   10.0.0.233   ip-172-31-58-228   <none>           <none>
+(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$ thepodip=10.0.0.233
+(vllm) ubuntu@ip-172-31-58-228:~/llm-d-fast-model-actuation$ curl $thepodip:8081/v1/dual-pod/accelerators
+["GPU-845e0388-9896-3a61-8ac9-3643833770d2"]
```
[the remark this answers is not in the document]
waltforme: Responsed in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/33#discussion_r2368938819

config/examples/requester.yaml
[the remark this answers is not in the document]
waltforme: I need an example with GPUs so I wrote the inline one in `cmd/requester/README.md`.

I don't know what `config/examples/requester.yaml` is used for, so I didn't touch it.

PR #38, opened 2025-09-19
Start latency outline

MikeSpreitzer:
This PR adds a couple of end-to-end outlines, with a focus on latency.

### DIFF
--- a/docs/latency-outline.md
+++ b/docs/latency-outline.md
@@ -1,0 +1,284 @@
+# Latency Outline
+
+Well, really it is another outline of the whole thing, but laser
+focused on latency (since the whole point is good latency).
+
+## Wake_up case
+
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
I still think that we need a document like this.

MikeSpreitzer:
This Issue should, for now, be kept open.

MikeSpreitzer:
The force-push to d680777 is a rebase onto `main`.

MikeSpreitzer:
The force-push to 3d9e0c0 is a rebase onto `main`, so that we can graduate from stage B of #471 .

MikeSpreitzer:
This is still interesting.

MikeSpreitzer:
The force-push to 5090fae is a rebase onto `main`.

Issue #37, opened 2025-09-18
Create ConfigMap for the Launcher


diegocastanibm:
**Description:**
We do not want to create a separate image for the launcher since it depends on vLLM. We want to use the vLLM image and use a ConfigMap to inject the launcher into the vLLM pod.

**Requirements:**

- Create ConfigMap

- Test config map using vLLM image

PR #42, opened 2025-09-22
[DEPLOYMENT]: Launcher as configmap using vLLM image

diegocastanibm:
**Description:**
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/37

**Test:**
- Deployed in `fmaas-platform-eval` cluster (in `caching` project)
- Use `kubectl apply -k .`
- Check launcher with `kubectl port-forward svc/launcher-service 8001:8001`
- Sent `POST`, `GET` and `DELETE` commands

### DIFF
--- a/config/examples/inject-launcher/replicaset.yaml
+++ b/config/examples/inject-launcher/replicaset.yaml
@@ -17,6 +17,8 @@ spec:
         app: inference-server
         version: "1.0"
     spec:
+      nodeSelector:
+        kubernetes.io/hostname: "your-node-name"
       volumes:
         - name: cache-volume
           persistentVolumeClaim:


inference_server/launcher/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+            limits:
+              cpu: "10"
+              memory: 200G
+              nvidia.com/gpu: "1"
+            requests:
+              cpu: "2"
+              memory: 60G
+              nvidia.com/gpu: "1"
+          volumeMounts:
+            - name: code-volume
```
aavarghese, line 50: We'll also need to mount the `cache-volume` volume under `volumeMounts`

aavarghese: Same thing with `shm`

```
    - name: shm
                mountPath: /dev/shm
```

inference_server/launcher/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+            medium: Memory
+            sizeLimit: "2Gi"
+        - name: code-volume
+          configMap:
+            # Kustomize will automatically generate a unique name
+            name: launcher-code
+            defaultMode: 0755
+      containers:
+        - name: launcher
+          image: vllm/vllm-openai:latest
```
aavarghese, line 31: This image has `uvicorn` installed?

inference_server/launcher/deployment.yaml
diegocastanibm: Good catch!

config/examples/inject-launcher/replicaset.yaml
```diff
@@ -0,0 +1,59 @@
...
+  selector:
+    matchLabels:
+      app: inference-server
+      version: "1.0"
+  template:
+    metadata:
+      labels:
+        app: inference-server
+        version: "1.0"
+    spec:
```
MikeSpreitzer, line 19: This needs to insist on running on one particular Node, right?

config/examples/inject-launcher/kustomization.yaml
```diff
@@ -0,0 +1,27 @@
...
+kind: Kustomization
+
+# Generate ConfigMap from files
+configMapGenerator:
+  - files:
+      - __init__.py
+      - launcher.py
+    name: launcher-code
+    options:
+      disableNameSuffixHash: true
```
MikeSpreitzer, line 11: If `kustomize` is invoked per inference server, there will be multiple copies of the same ConfigMap --- i.e., with the same name, which cannot be?

config/examples/inject-launcher/replicaset.yaml
diegocastanibm: I do not understand this. Do you mean that I should add something in the specs to specify the Node?

MikeSpreitzer: Yes

MikeSpreitzer (approved):
I can accept this as an illustration of how to use a ConfigMap to make the server-running Pod. I think that all the one-per-cluster objects should be created differently, and as for one-per-launcher, we only need the Pod object.

PR #45, opened 2025-09-23
[WIP] Pool manager creates set of launcher pods

aavarghese:
Part of issue #18 

Following what is in the current [README](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/main/docs/pool-manager.md) for the component.

Test plan coming up..

### DIFF
--- a/pkg/controller/poolmanager/poolmanager.go
+++ b/pkg/controller/poolmanager/poolmanager.go
@@ -1,0 +1,273 @@
+package poolmanager
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"time"
+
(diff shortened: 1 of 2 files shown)


pkg/controller/poolmanager/poolmanager.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		log.Printf("failed DELETE to launcher: %w", err)
+	}
+	defer resp.Body.Close()
+
+	patch := []byte(`{
+		"metadata": {
+			"labels": {
+				"inference.k8s.io/pool": null,
+				"inference.k8s.io/model": null
```
lionelvillard, line 185: same here, this is the responsibility of the dual-pod controller.

PR #55, opened 2025-09-29
Make dual-pod controller use gpu-map ConfigMap

MikeSpreitzer:
This PR supplies the controller-side functionality to map GPU UUID to index, which is needed in Milestones 1 and 2 but not 3.

This PR also introduces the usual command line flags for kube client configuration.

This PR also introduces focus on one namespace (which I think is right but I am not sure).

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -18,7 +18,9 @@ package dualpods
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"sync/atomic"
 
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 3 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -36,6 +38,7 @@ import (
 )
 
 const ControllerName = "dual-pods-controller"
+const GPUMapName = "gpu-map"
```
jimcadden, line 41: Please add a comment (and/or documentation) that specifies the expected format of gpu-map.

MikeSpreitzer:
The force-push to 0cf8fcc is a rebase onto `main`.

MikeSpreitzer:
That force-push is a rebase.

Issue #50, opened 2025-09-24
Enhancement: Add Multi-vLLM Instances Support to Launcher


diegocastanibm:
## Description

Currently, the launcher only supports managing a single vLLM instance at a time. This limitation prevents efficient resource utilization and scaling scenarios where multiple models or multiple instances of the same model need to run concurrently.

This enhancement request proposes extending the launcher to support managing multiple vLLM instances simultaneously, enabling multi-model serving capabilities.

## Use Cases

Based on the northstar document, multi-vllm launcher will be use in the third stage: model swapping with sleep/wake. 

## Proposed Solution

Extend the launcher architecture to support multiple vLLM instances with individual configuration, monitoring, and lifecycle management.

## Acceptance Criteria

- [x] Launch multiple vLLM instances with different configurations
- [x] Manage instance lifecycle independently
- [x] Handle resource conflicts gracefully
- [ ] Document configuration format and usage examples
- [x] Add new unit test for multi-vLLM instances

cc. @MikeSpreitzer @lionelvillard

diegocastanibm:
Documentation and testing will be accomplish in different issues and PR's for clarity

PR #56, opened 2025-09-29
Multi vllm launcher

diegocastanibm:
**Description:**
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/50

**Testing:**
- Able to create multiple vLLM's (specifying the ID and auto-generated)
- Delete specific and all vLLM instances
- Status of specific and all instances

**Note:**
In order to keep the PR's small and clear enough for review, we will address in different PR's:
- Test for new multi-launcher
- Update documentation


inference_server/launcher/launcher.py
```diff
@@ -133,7 +241,19 @@ async def health():
 async def index():
     """Root URL response"""
     return JSONResponse(
-        content={"name": "REST API Service", "version": "1.0"},
+        content={
+            "name": "Multi-Instance vLLM Management API",
+            "version": "2.0",
```
MikeSpreitzer, line 246: So the endpoints should be /v2, right?

inference_server/launcher/launcher.py
MikeSpreitzer, line 1: Why does an instance have a _name_?

MikeSpreitzer, line 1: So sleeping and waking is Somebody Else's Problem, right? This launcher does not know or care which instances are sleeping.

diegocastanibm: I guess is something we can discuss. Do you want to be handled by the launcher. In that case, I have no problems about it. I will need to know how we set sleep/wake property, but should be feasible

inference_server/launcher/launcher.py
```diff
@@ -142,44 +262,94 @@ async def index():
...
+        raise HTTPException(status_code=500, detail=str(e))
 
-    if vllm_manager.is_running():
-        raise HTTPException(
-            status_code=409, detail="A vLLM instance is already running."
-        )
 
-    result = vllm_manager.start_process(vllm_config)
+@app.delete("/v1/vllm")
+async def delete_all_vllm_instances():
```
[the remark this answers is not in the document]
diegocastanibm: Yes. For me is very useful especially for testing. Also for stress tests at some point. It can be use for emergency shutdown (panic button), to gracefully shut down all instances for maintenance (system updates or GPU updates)...
I would like to keep it.

inference_server/launcher/launcher.py
MikeSpreitzer: But the launcher's management port is different from vLLM's (only) port, right?

In any case, here the code is versioning the launcher's interface. And it is different from the previous launcher interface, which did not support multiple launchees.

inference_server/launcher/launcher.py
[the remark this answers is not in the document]
MikeSpreitzer: "panic ... gracefully" ?

I was wondering why someone would use delete_all in addition to terminating the launcher process. If you have tests that do this, that would be an answer.

inference_server/launcher/launcher.py
diegocastanibm: yes, @lionelvillard had the same comment. I will change the version.

PR #59, opened 2025-09-29
Include linux/arm64 platform in multiarch image

lionelvillard:
To be able to run it on a Mac.


MikeSpreitzer:
It is not enough to work when build ISA = runtime ISA. We also need it to work when build ISA == arm64 and runtime ISA = amd64. I think that #57 handles this.

Testing the current version of this PR (#59) in the scenario above, it fails. The built image is just a single-platform image, for linux/arm64.

MikeSpreitzer (approved):
This looks good to me. https://github.com/lionelvillard/llm-d-fast-model-actuation/actions/runs/18131213304 built a multi-platform manifest for amd64 and arm64, https://github.com/lionelvillard/llm-d-fast-model-actuation/pkgs/container/llm-d-fast-model-actuation-requester/530563799?tag=sha-40b90de .

It would be better squashed; we don't really care about the thrashing along the way.

PR #62, opened 2025-09-30
Remove exclusion of build in shared repo

MikeSpreitzer:
Hoping that UI for explicit allow will appear.


MikeSpreitzer:
Nope. Just got only read permissions for packages. Fail at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/18117255847/job/51555095843#step:1:23

PR #70, opened 2025-10-01
Stop inheriting annotations

MikeSpreitzer:
This PR makes a design change, so that the nominal server-running Pod's annotations are not based on the server-requesting Pod's annotations but only come from the server patch. I make this change because I observed in trials that the server-requesting Pod had several annotations that were (a) added by various platform components and (b) were specific to the network identity of the server-requesting Pod. I doubt that we can write an adequate rule for filtering out all the platform-supplied annotations that are specific to the server-requesting Pod's identity.

This PR also improves the example in cmd/requester/README.md to use a ReplicaSet instead of a Pod.

This PR also removes the userid stipulation in the requester Dockerfile, because this gives OpenShift indigestion (under normal circumstances, which I do not seem to have at the moment in my test cluster).

### DIFF
--- a/cmd/requester/README.md
+++ b/cmd/requester/README.md
@@ -23,56 +23,67 @@ Instantiate the Helm chart for the dual-pods controller. Specify the tag produce
 helm upgrade --install dpctlr charts/dpctlr --set Image="${CONTAINER_IMG_REG}/dual-pods-controller:9010ece"
 ```
 
-Create a server-requesting pod.
+Create a ReplicaSet of 1 server-requesting Pod.
 
 ```shell
 kubectl apply -f - <<EOF
-apiVersion: v1
-kind: Pod
+apiVersion: apps/v1
+kind: ReplicaSet
 metadata:
   name: my-request
-  annotations:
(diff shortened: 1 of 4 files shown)


cmd/requester/README.md
```diff
@@ -23,56 +23,67 @@ Instantiate the Helm chart for the dual-pods controller. Specify the tag produce
...
+        dual-pod.llm-d.ai/server-patch: |
+          spec:
+            containers:
+            - name: inference-server
+              image: docker.io/vllm/vllm-openai:v0.10.2
+              command:
+              - vllm
+              - serve
+              - --port=8000
+              - ibm-granite/granite-3.3-2b-instruct
```
diegocastanibm, line 54: My understanding about this is that if you are not using `--model` , then the model needs to be before everything else. So in this case it would be `vllm server ibm-granite/granite-3.3-2b-instruct --port=8000`. Otherwise it needs to have the `--model` flag. 
I suggest to add `--model` in line 54 for the shake of clarity.

jimcadden (approved):
LGTM

PR #71, opened 2025-10-01
Rename Job to match filename in build-requester-image.yml

MikeSpreitzer:
Experience shows that once you have more than a few workflows, it is very helpful for the `name` in the file to match the file name.


diegocastanibm (approved):
LGTM

PR #72, opened 2025-10-02
Start reporting dual-pod status

MikeSpreitzer:
This PR adds status reporting into the server-requesting Pod from the dual-pods controller. The status goes into a newly defined annotation.

Many permanent error conditions that were previously treated as transient are now reported as errors in the state. Also, in an ambiguous case, it is both reported in the status and retried.

The conversion of the JSON for the nominal server-running Pod into an actual Pod data structure was changed to use real Kubernetes mechanisms, which can detect spurious fields; these are now reported as an error in the state.

This PR also updates the example in cmd/requester/README.md to: (1) illustrate labeling for matching by an InferencePool, choosing to put these labels in the server patch, and (b) illustrate remove of a label by the server patch.

Successfully tested with both (a) spurious fields and (b) invalid strings (detected only by the kube-apiserver).

### DIFF
--- a/pkg/controller/dual-pods/server-requesting.go
+++ b/pkg/controller/dual-pods/server-requesting.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"regexp"
 	"slices"
 	"strconv"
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
BTW, we might consider an alternate way of reporting this information. We might introduce a PodCondition of our own, such as "DualPodExists", and report via that.

For now I will merge this PR, we can change the interface later if we want.

PR #73, opened 2025-10-03
Set nodeSelector in server-running Pod

MikeSpreitzer:
This PR fixes the dual-pod controller, to supply the needed `spec.nodeSelector` in the server-running Pod.

This PR also updates the example that uses caching, to work with vcp-cache-manager. Successfully tested. Loaded model from filesystem, hit the torch.compile cache.


jimcadden (approved):
LGTM

PR #82, opened 2025-10-07
Start documenting end-to-end testing recipe

MikeSpreitzer:
This PR contributes an end-to-end testing recipe in a reasonable place.


docs/e2e-recipe.md
```diff
@@ -0,0 +1,237 @@
...
+
+```shell
+helm upgrade --install dpctlr charts/dpctlr --set Image="${CONTAINER_IMG_REG}/dual-pods-controller:9010ece" --set NodeViewClusterRole=vcp-node-viewer
+```
+
+## Example 1: cycle server-requesting Pod
+
+Create a ReplicaSet of 1 server-requesting Pod. Following are two
+examples. The first is rather minimal. The second uses caching/staging
+from vcp-cache-manager.
```
lionelvillard, line 48: can you provide a link to the vcp-cache-manager since not everyone knows what it is.

MikeSpreitzer: It's not public yet, so not sure whether/what to do about this. I will post a link in Slack for now.

PR #84, opened 2025-10-07
Copy Milestone 1 design from discussion doc into Git

MikeSpreitzer:
This PR replaces the old in-progress dual-pods design document with a copy of the Milestone 1 design from the design discussion document.


docs/dual-pods.md
```diff
@@ -159,10 +216,59 @@ spec:
...
+
+- A set of server-requesting Pods that are bound to Nodes. Each such
+  Pod specifies: model, other command-line parameters, some
+  environment variable settings, and a Node. After its stub has been
+  queried, this Pod is also known to specify a particular set of
+  accelerators on the Node. This server-requesting Pod may be bound
+  (here, in this data structure) to a vLLM instance.
+
+When the server-requesting Pod is bound to a Node that is absent or in the process of being deleted, the dual-pods controller has nothing left to do and the following logic is irrelevant.
+When a server-requesting Pod is bound to a server-running Pod that is in the process of being deleted, the controller (a) ensures that its finalizer is not on the server-running Pod and (b) ensures that the server-requesting Pod is being deleted. (The controller creates server-running Pods with its finalizer on them, so that they cannot evaporate without this interaction with the controller.)
```
aavarghese, line 239: Can we add another line between the different "When" scenarios so they don't all get combined into one big para?

MikeSpreitzer: Sorry, the github diff display for the latest commit does not make it clear, but the commit only adds whitespace.

PR #86, opened 2025-10-09
Fix concurrency bugs in dual-pods controller

MikeSpreitzer:
This PR extends #85 with fixes to some concurrency bugs. In #85, the data about an inference server could be concurrently managed by two goroutines: one processing the server-requesting Pod and one processing the server-running Pod. The main change in this PR (beyond #85) is making the work queue items refer to inference servers rather than Pods.

This PR also corrects the controller's behavior in the following scenario, starting from both Pods existing: the server-running Pod starts being deleted, then the controller removes its finalizer and requests deletion of the server-requesting Pod, then the controller's Pod informer reports deletion of the server-running Pod --- while not yet reporting deletion of the server-requesting Pod (yes, this is possible) --- so the controller sees an unsatisfied request and starts working on satisfying it. (FYI, for the sake of minimizing apiserver traffic, the controller does no additional reads beyond what its informers do.)

Successfully tested using this scenario (because it is relatively challenging):
- setup (as in the E2E doc in PR #82);
- create the ReplicaSet;
- observe the server-running Pod come up;
- take down the dual-pods controller;
- request deletion of the server-running Pod;
- observe that deletion does not complete because of the controller's finalizer;
- bring up the dual-pods controller again;
- observe that the old server-requesting and server-running Pods go away, and new pair comes up;
- examine the logs from both runs of the dual-pods controller, confirm expected content (including no work on zombie server-requesting Pod).

Here are the shell typescript and dual-pods controller logs.

[reorient-trial.log](https://github.com/user-attachments/files/22791156/reorient-trial.log)
[dpctlr.log](https://github.com/user-attachments/files/22791163/dpctlr.log)
[dpctlr2.log](https://github.com/user-attachments/files/22791164/dpctlr2.log)

Slightly less developed versions of this were also successfully tested with gentler scenarios, such as deleting the server-requesting Pod.

If #85 gets reviewed and merged first then I will rebase this on `main`.

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -39,6 +39,22 @@ import (
 	genctlr "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/generic"
 )
 
+// This package implements the dual-pods controller.
+
+// The controller works in the context of one Kubernetes API namespace.
+
+// A Pod is a server-requesting Pod if it has the server patch annotation.
+// A Pod is a server-running Pod if has a controlling OwnerReference to
+// a Pod (the server-requesting Pod).
+
+// There are two types of item in the controller's work queue.
+// One is a reference to the gpu-map ConfigMap.
+
+// The other type of queue item is a reference to an inference server.
+// This reference carries the inference server's UID and the name
+// of the server-requesting Pod.
+// An inference server's UID is the UID of the server-requesting Pod.
+
 const ControllerName = "dual-pods-controller"
 
 // GPUMapName is the name of the ConfigMap(s) parsed to discover the mapping from GPU UUID to location.


pkg/controller/dual-pods/controller.go
```diff
@@ -39,6 +39,22 @@ import (
 	genctlr "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/generic"
 )
 
+// This package implements the dual-pods controller.
+
+// The controller works in the context of one Kubernetes API namespace.
+
+// A Pod is a server-requesting Pod if it has the server patch annotation.
+// A Pod is a server-running Pod if has a controlling OwnerReference to
```
diegocastanibm:
```suggestion
// A Pod is a server-providing Pod if has a controlling OwnerReference to
```

diegocastanibm (approved):

PR #89, opened 2025-10-10
Allow sleep

MikeSpreitzer:
This PR starts the progression to milestone 2. The changes here are as follows.

- Change the way that a server-running Pod references its server-requesting Pod from an entry in `.metadata.ownerReferences` to an attribute; this allows the server-running Pod to continue existing after the server-requesting Pod is deleted.
- Remove the fixed derivation of the server-running Pod's name, make the apiserver generate it instead.
- When releasing a server-running Pod, do not delete it; instead, put it to sleep and remove the binding annotation and the finalizer.
- Make a server-running Pod carry an annotation whose value is a hash of bytes characteristic of the inference server (this anticipates future re-binding (waking up))

### DIFF
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -354,7 +354,8 @@ func (serverDat *serverData) getNominalServerRunningPod(ctx context.Context, req
 		}
 		hasher := sha256.New()
 		hasher.Write(modifiedJSON)
-		hasher.Write([]byte(fmt.Sprintf(";gpus=%s", *serverDat.GPUIndices)))
+		hasher.Write([]byte(";gpus="))
+		hasher.Write([]byte(*serverDat.GPUIndices))
 		var modifiedHash [sha256.Size]byte
 		modifiedHashSl := hasher.Sum(modifiedHash[:0])
 		nominalHash := base64.RawStdEncoding.EncodeToString(modifiedHashSl)


MikeSpreitzer:
The force-push to bb12ccb is a rebase.

MikeSpreitzer:
The force-push to 69f4ece is a rebase onto `main`.

diegocastanibm:
QQ: why do we want to keep the server-running Pod existing after the server-requesting Pod is deleted? Wouldn't be better to have both so, in the future, we can update the requirements of the running pod by modifying the requesting pod?

MikeSpreitzer:
A Pod's Spec can not be modified (except for vertical scaling of CPU and CPU memory).

diegocastanibm (commented):
Why you have the dual pod running:
```
my-request-v66qs                              1/1     Running   0             8m52s
my-request-v66qs-dual-k5btv                   1/1     Running   0             8m42s
```
but then, when you delete the requesting pod with:
```
mspreitz@mjs13 llm-d-fast-model-actuation % kubectl delete pod my-request-v66qs
```
a new requesting pod is generated?
```
my-request-4mp2w                              0/1     Running   0             13s     10.130.7.149   fmaas-vllm-d-wv25b-worker-cl-3-7fk47    <none>           <none>
my-request-v66qs-dual-k5btv                   1/1     Running   0             9m6s    10.130.7.148   fmaas-vllm-d-wv25b-worker-cl-3-7fk47    <none>           <none>
```
What's the point of having a new requesting pod?

diegocastanibm:
I can see the same behavior on my side, but I do not understand why when requesting pod after deleting the controller, a new server requesting pod is created.
[mikeshell.log](https://github.com/user-attachments/files/22957727/mikeshell.log)

pkg/controller/dual-pods/inference-server.go
```diff
@@ -475,3 +577,15 @@ func SliceMap[Domain, Range any](slice []Domain, mapFn func(Domain) (Range, erro
...
+	return func(yield func(Elt) bool) {
+		for elt := range it {
+			if include(elt) {
+				if !yield(elt) {
+					return
+				}
+			}
+		}
+	}
+}
```
diegocastanibm: I find this nested `if` a little awkward. Can we combined the `if` statements?:
```suggestion
func IterFilter[Elt any](it iter.Seq[Elt], include func(Elt) bool) iter.Seq[Elt] {
	return func(yield func(Elt) bool) {
		for elt := range it {
			if include(elt) && !yield(elt) {
				return
			}
		}
	}
}
```

MikeSpreitzer:
@diegocastanibm, regarding your question about why a replacement serve-requesting Pod appears after one is deleted: this is while there is a corresponding ReplicaSet object, right? This is the defined functionality of a ReplicaSet. Its controller maintains the desired number of replicas.

PR #92, opened 2025-10-14
Use wake

MikeSpreitzer:
This PR takes another step towards milestone 2. This step is to wake an existing sleeping inference server, if it fits the bill.

Note: This is difficult to test when many different GPUs are available to allocate. The interesting case is re-use of a GPU, and when there are many available this is not the common case. This PR extends the E2E test recipe doc with an example for testing waking up, with a suggestion for improving the odds while testing.

### DIFF
--- a/docs/e2e-recipe.md
+++ b/docs/e2e-recipe.md
@@ -119,6 +119,8 @@ spec:
                 value: /tmp
               - name: XDG_CONFIG_HOME
                 value: /tmp
+              - name: XDG_CACHE_HOME
+                value: /tmp
               - name: TRITON_HOME
                 value: /tmp
               resources:


MikeSpreitzer:
Successfully tested sleep and wake.  Attached are shell log and dual-pods controller log.

[wake-test-1.log](https://github.com/user-attachments/files/22919518/wake-test-1.log)
[dpctlr.log](https://github.com/user-attachments/files/22919519/dpctlr.log)

pkg/controller/dual-pods/inference-server.go
```diff
@@ -95,7 +95,7 @@ func (item infSvrItem) process(ctx context.Context, ctl *controller) (error, boo
 
 	var shouldAddRequesterFinalizer bool
 	if requestingPod != nil {
-		removed, shouldAdd, err, retry := ctl.removeRequesterFinalizer(ctx, requestingPod, runningPod)
+		removed, shouldAdd, err, retry := ctl.maybeRemoveRequesterFinalizer(ctx, requestingPod, runningPod)
```
MikeSpreitzer: This is just fixing the name of this method.

MikeSpreitzer:
The force-push to f196f0e is a rebase.

docs/e2e-recipe.md
```diff
@@ -322,3 +322,50 @@ dual-pods controller deletes the server-requesting Pod.
 
 Setup and create the dual pods. Then delete the Node that they are
 running on. Observe that eventually both Pods go away.
```
rubambiza: Related to example 3:

The expectations from the first two examples are clear. What is meant by "the right stuff should finish happening"? That new server requesting/running pods will be created from the old ReplicaSet?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -189,9 +189,24 @@ func (item infSvrItem) process(ctx context.Context, ctl *controller) (error, boo
...
-	// relay readiness if needed.
+	// If there is already a server-running Pod then ensure that it is awake,
+	// ensure status reported, and relay readiness if needed.
 	if runningPod != nil {
+		if serverDat.Sleeping == nil || *(serverDat.Sleeping) {
+			_, serverPort, err := getInferenceServerPort(runningPod)
+			if err != nil { // Impossible, because such a runningPod would never be created by this controller
+				return fmt.Errorf("unable to put server to sleep because port not known: %w", err), false
+			}
+			wakeURL := fmt.Sprintf("http://%s:%d/wake_up", runningPod.Status.PodIP, serverPort)
```
rubambiza: Related to this in example 7 of the e2e-recipe, it would be nice to have an example of what an HTTP GET `is_sleeping` for the runningPod looks like in the scenario description. For instance, is it a `kubectl exec <podname> -- curl <something>`? I would presume a direct curl is not possible because the pods do not have an external facing IP.

rubambiza (commented):
Requested trivial changes will be resolved in other PRs as discussed on Slack.

I will approve once other comments are resolved.

docs/e2e-recipe.md
MikeSpreitzer: Since example 3 is really two examples, the right stuff varies.

If you `kubectl delete`d the server-providing Pod while the controller was absent, upon restoration of the controller expect both Pods to go away and be replaced by a new pair.

If you `kubectl delete`d the server-requesting Pod while the controller was absent, expect (in this PR and later ones) that after the controller is restored that server-requesting Pod goes away and is replaced (by the ReplicaSet controller) with a new one. As for the server-running Pod, it should remain in existence; a new one should or should not be added depending on whether the new server-requesting Pod gets assigned the same GPU as the old one.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: There is no example 7 in this PR.

Feel free to do that HTTP GET in any way that you find convenient. If you know of a Pod with a container with `curl` in it, you could indeed `kubectl exec` a call on `curl http://pod:8000/is_sleeping`. I myself do it with a combination of `kubectl port-forward` and `curl http://localhost:8000/is_sleeping`.

PR #93, opened 2025-10-16
Limit unbound server-running Pods

MikeSpreitzer:
This PR takes the last major step towards milestone 2: limiting the collection of server-running Pods that are not bound to a server-requesting Pod, per GPU. It imposes the simplest possible limit: a number of Pods. And it uses a simple policy for picking which to delete: the least recently used. (Testing showed that something deterministic is needed due to the fact that it takes a while to delete a Pod and an informer's cache is only eventually consistent.)

This PR does a bit of re-organization of the controller structure, to provide mutual exclusion per Node rather than per inference server. That is because the culling of excess server-running Pods is not specific to an inference server (multiple concurrent inference server arrivals could trigger this).

The PR also updates the generic controller, to make the `err error` and `retry bool` returned by `process` completely orthogonal.


MikeSpreitzer:
The force-push to 77aaf35 is a rebase.

PR #108, opened 2025-10-24
Start work on end-to-end test

MikeSpreitzer:
This PR starts work on end-to-end testing in a `kind` cluster (using special requester and server that do not acctually use GPUs).

This PR adds the following.

- `cmd/test-requester`, a variant requester that does not involve actual GPUs; instead it keeps its own allocations in a ConfigMap.
- `docs/e2e-recipe.md`. Instructions for very simple manual setup for end-to-end testing.
- `test/e2e/run.sh`. Script to do end-to-end testing. So far, just a few test cases.

Both start with cleanup from previous runs, create a `kind` cluster named "fmatest", build the local version of the repo, load the container images into the `kind` cluster, and deploy the dual-pods controller there.

### DIFF
--- a/cmd/test-requester/gpu-allocation.go
+++ b/cmd/test-requester/gpu-allocation.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"iter"
 	"os"
 	"strings"
 	"time"
@@ -209,15 +208,4 @@ func getPodUIDs(ctx context.Context, podClient corev1client.PodInterface) (sets.
 		return pod.UID, nil
 	})
 	return sets.New(uids...), nil
-}
-
-func MapFilter2to1[Key comparable, Val, Result any](input map[Key]Val, filter func(Key, Val) (Result, bool)) iter.Seq[Result] {
-	return func(yield func(Result) bool) {
-		for key, val := range input {
-			result, include := filter(key, val)
-			if include && !yield(result) {
-				return
-			}
-		}
-	}
 }


cmd/test-requester/gpu-allocation.go
```diff
@@ -0,0 +1,214 @@
...
+// This code maintains a ConfigMap named "gpu-allocs" that holds the current test allocations
+// of GPUs. The data of this ConfigMap is a map from GPU UID to the JSON marshaling of a GPUHOlder.
+
+// gpuMap maps node name to nodeGPUMap
+type gpuMap map[string]nodeGPUMap
+
+// nodeGPUMap maps GPU UID to index
+type nodeGPUMap map[string]int
+
+// GPUHolder identifis a test requester that is currently allocated the use of a GPU
```
rubambiza: Minor: `identifi*e*s`

cmd/test-requester/gpu-allocation.go
```diff
@@ -0,0 +1,214 @@
...
+		ans[gpuUID] = holder
+	}
+	cm = cm.DeepCopy()
+	if cm.Data == nil {
+		cm.Data = map[string]string{}
+	}
+	return ans, cm, nil
+}
+
+func allocateGPUs(ctx context.Context, coreClient corev1client.CoreV1Interface, nodeName, namespace string, podUID apitypes.UID, numGPUs uint) []string {
```
rubambiza: I think there is a lot going on in this function, and it would be nice to have some comments to separate the logical chunks.

For instance, I could not follow why `avail` is being declared way before its first use. Is the gpuMap being changed between the declaration and when the difference with the remaining set is being computed?

At least some comments for what the try and err functions do would be great.

cmd/test-requester/gpu-allocation.go
```diff
@@ -0,0 +1,214 @@
...
+func MapFilter2to1[Key comparable, Val, Result any](input map[Key]Val, filter func(Key, Val) (Result, bool)) iter.Seq[Result] {
+	return func(yield func(Result) bool) {
+		for key, val := range input {
+			result, include := filter(key, val)
+			if include && !yield(result) {
+				return
+			}
+		}
+	}
+}
```
rubambiza: I do not see any usage of this function in this PR. Is it intended to be used somewhere else later? If so, I would suggest removing it. Otherwise, it further muddies somewhat already murky waters.

cmd/test-requester/README.md
```diff
@@ -0,0 +1,47 @@
+# Test requester
+
+This is a variant of the normal requester that does not not actually involve GPUs.
```
rubambiza: Nit: Extra `not`

test/e2e/run.sh
```diff
@@ -0,0 +1,183 @@
...
+kubectl wait --for condition=Ready pod/$prv --timeout=1s
+
+echo Successful re-use
+
+: Test provider deletion
+: expect requester to be deleted and a new pair to appear
+
+kubectl delete pod $prv
+
+# expect "kubectl get pods -o name | grep -c '^pod/$nrq' | grep -w 0"
```
rubambiza: Does this need to be taken out?

cmd/test-requester/gpu-allocation.go
```diff
@@ -0,0 +1,214 @@
...
+type GPUAllocMap map[string]GPUHolder
+
+func getGPUMap(ctx context.Context, cmClient corev1client.ConfigMapInterface) (gpuMap, error) {
+	cm, err := cmClient.Get(ctx, dpctlr.GPUMapName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve gpu-map ConfigMap: %w", err)
+	}
+	ans := gpuMap{}
+	for nodeName, mapStr := range cm.Data {
+		nm := nodeGPUMap{}
```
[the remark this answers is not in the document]
MikeSpreitzer: "ans" is short for "answer"

cmd/test-requester/gpu-allocation.go
MikeSpreitzer: Other PRs move the generic utilities to a separate file and delete this one.

cmd/test-requester/gpu-allocation.go
```diff
@@ -0,0 +1,214 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	dpctlr "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/dual-pods"
+
+	"k8s.io/klog/v2"
+)
+
+// This code maintains a ConfigMap named "gpu-allocs" that holds the current test allocations
+// of GPUs. The data of this ConfigMap is a map from GPU UID to the JSON marshaling of a GPUHOlder.
```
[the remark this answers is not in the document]
MikeSpreitzer: That was a typo. Fixed, thanks.

cmd/test-requester/gpu-allocation.go
MikeSpreitzer: It is not possible to declare everything right before its use.

I have simplified and added comments.

PR #115, opened 2025-10-24
✨ Milestone 2 fixups

MikeSpreitzer:
This PR makes a few improvements in response to reviews and experience. Notable changes are as follows.

- Fix the Helm chart to create a ClusterRoleBinding rather than a RoleBinding to grant Node reading privileges.
- Add one more FYI label maintained by the dual-pods controller, "dual-pods.llm-d.ai/sleeping" showing what the controller thinks that bit is. I have successfully tested this in scenarios of: creation, delete server-requesting Pod, delete server-running Pod, restart controller in the presence of a pair of Pods.

### DIFF
--- a/charts/dpctlr/templates/role-binding.yaml
+++ b/charts/dpctlr/templates/role-binding.yaml
@@ -27,5 +27,6 @@ roleRef:
   name: "{{ .Values.NodeViewClusterRole }}"
 subjects:
 - kind: ServiceAccount
+  namespace: "{{ .Release.Namespace }}"
   name: dual-pods-controller
 {{end}}


waltforme:
@clubanderson 
This PR made a decision to take more responsibility of setting up RBAC and implicitly ask the user to have higher privilege than before.

MikeSpreitzer:
I would say that rather than take more responsibility, this PR fixes the broken implementation of the responsibility that the chart was already alleging that it *can* take. And yes, the fix requires the user to have higher privileges than before. That is unavoidable, it is because of the semantics of RBAC.

The alternative has always been to give the ServiceAccount the needed node-reading privileges in some other way.

PR #124, opened 2025-10-31
E2E fixes

### LINKED ISSUE #122: E2E test passes bogusly
Current runs run out of CPU but pass anyway.

For example, see https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/18966957491/job/54165545647

MikeSpreitzer:
This PR adds checking to catch the fact that the test has been running out of CPU. And reduces the CPU need so that the test passes.

Fixes #122

### DIFF
--- a/test/e2e/mkrs.sh
+++ b/test/e2e/mkrs.sh
@@ -37,7 +37,7 @@ spec:
               - --startup-delay=22
               resources:
                 limits:
-                  cpu: "1"
+                  cpu: "200m"
                   memory: 9Gi
               readinessProbe:
                 httpGet:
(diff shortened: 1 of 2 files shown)


rubambiza (changes requested):
Re-ran the e2e tests successfully ✅ 

On the subject of reducing resources needs: given that the test requester is always building from the current HEAD for each end to end run, may I suggest adding a clean up recipe such that the images are removed if all the tests pass?

This would ensure that the images do not continue to collect over time in the local registry. Though 35-40MB is not much, it could add up with each review and run of the end-to-end tests.

Thoughts?

MikeSpreitzer:
> suggest adding a clean up recipe ... ?

Interesting idea. Non-trivial, actually. I would like to leave that to another PR.

rubambiza:
Makes sense. I figured it might be non-trivial. I will go ahead and approve this PR then.

rubambiza (approved):
LGTM.

PR #133, opened 2025-11-11
DPC Benchmarking Tool

aavarghese:
A benchmarking tool that measures readiness latency and, focused primarily on hit rates (hit == waking up sleeping instance on scale up) in different scenarios...

The benchmarking tool will evolve to meet the needs of the launcher-based method of inference server start from https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/154.

Previous PR https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/176 merged into [milestone-2](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/tree/milestone-2) branch.

### DIFF
--- a/inference_server/benchmark/benchmark.md
+++ b/inference_server/benchmark/benchmark.md
@@ -1,0 +1,113 @@
+# Dual-Pods Benchmarking Tool
+The **Dual-Pods Benchmarking Tool** measures and reports the startup and readiness
+latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
+
+## Purpose
+The goal is to quantify and compare how quickly a model-serving duo (server-requesting
+and server-providing pods) becomes available under different actuation conditions such
+as cold starts, wake-ups from a sleeping state, using prewarmed pods, etc. These metrics
+will guide future optimizations for the **Dual-Pods Controller (DPC)**. Ultimately, the goal
+is *high predictability*, which is defined as achieving close to 100% hit rate of awakening
+available, sleeping pods on cluster GPUs as function of total inference server
+requests for common user scenarios.
+
+## Baseline Startup Latency
+
+**Objective:**
+Measure the time from **deployment (server-request submission)** to **dual-pod readiness**.
+
+### Inputs
+
+| Parameter           | Type   | Required | Default                                 | Description                                                                                            |
+| ------------------ | ------ | -------- | --------------------------------------- | ------------------------------------------------------------------------------------------------------ |
+| `--namespace`      | `str`  | **Yes**  | —                                       | Openshift namespace to run benchmark                                      |
+| `--yaml`           | `str`  | **Yes**  | —                                       | Path to the server-requesting YAML template file              |
+| `--image`          | `str`  | **Yes*** | —                                       | Image repository for the requester pod. Required *only if* `CONTAINER_IMG_REG` env var is **not** set |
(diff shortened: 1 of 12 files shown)


inference_server/benchmark/benchmark_base.py
```diff
@@ -71,26 +75,41 @@ def __init__(
         self.template_files: List[str] = []
         self.provider_pods: List[str] = []
 
-        self.parsed_inputs = self.parse_inputs()
+        # Parse inputs and set as class properties
+        parsed_inputs = self.parse_inputs()
+        self.namespace = parsed_inputs[0]
+        self.yaml_template_file = parsed_inputs[1]
+        self.requester_pod_label = parsed_inputs[2]
```
rubambiza: I think we can remove this as a class property and CLI input since we are now using the `REQUESTER_PATCH_ANNOTATION (dual-pod.llm-d.ai/server-patch)` to filter the requester pods.

inference_server/benchmark/scenarios.py
```diff
@@ -28,22 +43,314 @@
...
+                    "phase": "0_to_1",
+                    "rq_time": rq_ready,
+                    "availability_mode": prv_mode,
+                    "success": rq_ready is not None,
+                }
+            except TimeoutError as e:
+                benchmark.logger.warning(
+                    f"Scaling 0->1 timed out for iteration {iter_num}: {e}"
+                )
+                result = {
```
rubambiza: These lines feel quite repetitive for the scaling operations (i.e., line 173 - 313). The few differences I see are:

- The `phase` assigned to the result varies, which can be passed as param in a sub-method.
- The provider_pods are being created if not already existing, which can be created in a sub-method invocation by passing the benchmark in.
- The benchmark result list is being appended to, which can be returned as part of a sub-method invocation.
- The error that could happen from the wait of each scaling operation can be propagated and return along with the result.

Overall, I think there is an opportunity for more refactoring to make this run_*_scenario a little cleaner. This would also be important once we introduce the number of iterations as an input.

inference_server/benchmark/utils.py
```diff
@@ -82,6 +82,18 @@ def parse_request_args():
         default="fmaas-platform-eval.fmaas.res.ibm.com",
         help="Cluster domain for Prometheus GPU metrics query",
     )
+    parser.add_argument(
+        "--model-path",
+        type=str,
+        help="Path to JSON file containing models for new_variant scenario",
+    )
```
rubambiza: When parsing the arguments on the other side (i.e., DualPodsBenchmark) and the scenario is `new_variant`, we should add a check that triggers an error/early exit if the user does not provide a value for this argument.

By user, I mean this is also me protecting me from myself 😂

inference_server/benchmark/benchmark.md
```diff
@@ -0,0 +1,89 @@
...
+### Inputs
+
+| Parameter                          | Description                                          |
+| ---------------------------------- | -----------------------------------------------------|
+| `--namespace`                      | Kubernetes namespace where benchmarking occurs       |
+| `--yaml`                           | YAML file describing the dual-pod deployment         |
+| `--image`                 | Image repository for the inference server requester container |
+| `--tag`                   | Image tag to use for the inference server requester container |
+| `--num-gpus`           | Number of GPUs to request for the server-providing pod           |
+| `--num-model-variants` | Number of different model variants to deploy during benchmarking |
```
rubambiza: Needs to change to match the example usage below.

PR #141, opened 2025-11-11
Fix chart to adhere to conventions

lionelvillard:
as documented here: https://helm.sh/docs/chart_best_practices/values. Also fix deployment service account.


charts/dual-pod-controller/templates/deployment.yaml
```diff
@@ -19,17 +19,17 @@ spec:
...
-        - "--debug-accelerator-memory={{ .Values.DebugAcceleratorMemory }}"
+        - "--sleeper-limit={{ .Values.sleeperLimit }}"
+        - "--debug-accelerator-memory={{ .Values.debugAcceleratorMemory }}"
         env:
         - name: POD_NAMESPACE
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-      serviceAccount: dual-pods-controller
+        serviceAccountName: dual-pods-controller
```
MikeSpreitzer: +1 on changing from `serviceAccount` to `serviceAccountName` (the former works but is deprecated).

-1 on indenting further. This field is part of the PodSpec, not the Container.

Issue #65, opened 2025-09-30
Update documentation for multi-vllm instances. launcher


diegocastanibm:
Description:

We have implemented a multi-vlllm instances launcher that can be seen in this issue:
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/50

We need to update the documentation for it

PR #142, opened 2025-11-12
[DOCUMENTATION]: Update documentation for launcher

diegocastanibm:
**Description:**
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/65

### DIFF
--- a/docs/launcher.md
+++ b/docs/launcher.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The Multi-Instance vLLM Launcher (a.k.a. the launcher) is a Python program that implements a REST API service that allows clients to dynamically create, manage, and terminate vLLM inference server instances. The goal is to achieve model swapping functionality without changes to vLLM. This enables flexible model serving where clients can spin up different models on demand, and support concurrent inference workloads.
+The Multi-Instance vLLM Launcher (a.k.a. the launcher) is a Python program that implements a REST API service that allows clients to dynamically create, manage, and delete vLLM inference server instances. The goal is to achieve model swapping functionality without changes to vLLM. This enables flexible model serving where clients can spin up different models on demand, and support concurrent inference workloads.
 
 The launcher preloads vLLM’s Python modules to accelerate the initialization of multiple instances. Each vLLM process launched is therefore a subprocess of the launcher.
 
@@ -133,7 +133,6 @@ Response:
 {
   "status": "started",
   "instance_id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
-  "pid": 12345
 }
 ```
 
@@ -242,7 +241,6 @@ Create a new vLLM instance with an auto-generated UUID.
 {
   "status": "started",
   "instance_id": "uuid-string",
-  "pid": 12345
 }
 ```
 
@@ -289,7 +287,6 @@ Stop and delete a specific vLLM instance.
 {
   "status": "terminated",
   "instance_id": "instance-id",
-  "pid": 12345
 }
 ```
 
@@ -311,8 +308,8 @@ Stop and delete all running vLLM instances. This functionality can be specially 
 {
   "status": "all_stopped",
   "stopped_instances": [
-    {"status": "terminated", "instance_id": "id-1", "pid": 12345},
-    {"status": "terminated", "instance_id": "id-2", "pid": 12346}
+    {"status": "terminated", "instance_id": "id-1"},
+    {"status": "terminated", "instance_id": "id-2"}
   ],
   "total_stopped": 2
 }
@@ -353,17 +350,14 @@ Get status information for all instances. `Detail` is `True` by default.
     {
       "status": "running",
       "instance_id": "id-1",
-      "pid": 12345
     },
     {
       "status": "stopped",
       "instance_id": "id-2",
-      "pid": 12346
     },
     {
       "status": "running",
       "instance_id": "id-3",
-      "pid": 12347
     }
   ]
 }
@@ -392,7 +386,6 @@ Get status information for a specific instance.
 {
   "status": "running",
   "instance_id": "instance-id",
-  "pid": 12345
 }
 ```
 
@@ -478,7 +471,7 @@ curl http://localhost:8001/v2/vllm/instances
 
 ### vLLM Options
 
-The `options` field contains stuff added to the command line of the launched `vllm serve`. Common options include:
+The `options` field contains stuff added to the command line of the launched `vllm serve`. Options are listed below.
 
 #### Required Options
 
@@ -609,7 +602,7 @@ Be mindful of system resources:
 - **Memory**: Each instance loads a full model into memory
 - **GPU**: Plan GPU allocation carefully
 - **CPU**: vLLM uses CPU for pre/post-processing
-- **Disk**: Models are cached in `~/.cache/huggingface`
+- **Disk**: Models are cached in the container's filesystem
 
 ### 5. Testing
 


docs/launcher.md
```diff
@@ -1,39 +1,98 @@
+# Multi-Instance vLLM Launcher Documentation
 
-# vLLM Launcher
+## Overview
 
-## Description
+The Multi-Instance vLLM Launcher is a REST API service that allows you to dynamically create, manage, and terminate multiple vLLM inference server instances to achieve model swapping functionality under the Fast Model Actuation (FMA) project. This enables flexible model serving where you can spin up different models on demand, and support multiple concurrent inference workloads.
```
MikeSpreitzer: > under the Fast Model Actuation (FMA) project

The fact that it is in this repo already says that.

The more interesting thing to say was in the original but this PR currently cuts it out: "achieve model swapping without changes to vLLM". Recall that we have also discussed other ways to achieve model swapping but they involve changing vLLM.

docs/launcher.md
```diff
@@ -43,64 +102,552 @@ where the `JSON_EXAMPLE` is:
...
+    "list_instances": "GET /v2/vllm/instances"
+  }
+}
+```
+
+---
+
+#### Create Instance (Auto-Generated ID)
+
+**POST** `/v2/vllm`
```
MikeSpreitzer: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/6d4ac50eae96517f933033af78203ccc7cf1ee4b/inference_server/launcher/launcher.py#L258 says `@app.put("/v2/vllm")`

... and https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/6d4ac50eae96517f933033af78203ccc7cf1ee4b/inference_server/launcher/launcher.py#L242 contradicts that, saying "POST".

docs/launcher.md
```diff
@@ -1,39 +1,98 @@
...
+- **Auto & Custom IDs**: Support for both auto-generated UUIDs and custom instance IDs
+- **Process Isolation**: Each vLLM instance runs in a separate process with isolated configuration
+- **Environment Variable Support**: Set custom environment variables per instance
+- **Graceful Shutdown**: Proper termination with configurable timeout and force-kill fallback
+- **Status Monitoring**: Query status of individual instances or all instances at once
+- **Health Checks**: Built-in health endpoint for monitoring service availability
+
+## Architecture
+
+### Components
```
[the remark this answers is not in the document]
diegocastanibm: I have also removed the PORTS. They can be confusing

docs/launcher.md
diegocastanibm: Yes, you are right. Since the PUT/POST methods has changed in the last launcher version, I have updated all the doc accordingly. I hope I haven't missed any ...

docs/launcher.md
```diff
@@ -1,106 +1,651 @@
...
+    "options": "--model meta-llama/Llama-2-7b-hf --port 8000 --tensor-parallel-size 2",
+    "env_vars": {
+      "CUDA_VISIBLE_DEVICES": "0,1",
+      "VLLM_ATTENTION_BACKEND": "FLASHINFER",
+      "MAX_BATCH_SIZE": "128"
+    }
+  }'
+```
+
+### Example 4: Check All Running Instances
```
MikeSpreitzer: Are there instances that are _not_ running? How would a client get a list of those? I suspect that what you mean to speak of is instances that exist at the moment, rather than instances that are "running".

docs/launcher.md
```diff
@@ -1,106 +1,651 @@
...
+--model gpt2                   # ~500MB
+
+# Then move to production models
+--model meta-llama/Llama-2-7b-hf   # ~14GB
+--model meta-llama/Llama-2-13b-hf  # ~26GB
+```
+
+---
+
+**Version:** 2.0
```
MikeSpreitzer: Is this version number just a version of this document, or of something bigger? If the latter, what keeps it correctly updated here? If the former, when does it change?

docs/launcher.md
```diff
@@ -1,106 +1,651 @@
...
+curl -X DELETE http://localhost:8001/v2/vllm/instances/a1b2c3d4-e5f6-7890-abcd-ef1234567890
+```
+
+## API Reference
+
+### Base URL
+
+```
+http://localhost:8001
+```
```
MikeSpreitzer: This does not look to me like something that belongs in an API reference. The base URL shown is the right one for the quick start above, but not right for all scenarios. For example, the dual-pods controller is going to be creating Pods that run the launcher, and the controller will also be the client of the launcher, and in this scenario the base URL involves the Pod's IP address and the port that the controller chooses to tell the launcher to listen on.

docs/launcher.md
```diff
@@ -1,39 +1,98 @@
...
 
-The client/user will use a launcher-specific command to run the launcher. To swap a model in, the Client will issue a POST request (to the launcher) that includes the model reference and other command line flags. The launcher sends an `OK` response to the client/controller to indicate that queries can be sent to the vLLM instance directly. Upon receiving this response, the vLLM instance has not been fully guaranteed to be in a specific state. It may have executed few or no instructions, might not be listening on any port, and so forth.
+## Features
 
-Setting up process to create a new vLLM inference instance using POST:
+- **Multiple Instance Management**: Run multiple vLLM instances simultaneously with unique identifiers
+- **Dynamic Creation/Deletion**: Start and stop instances on demand via REST API
+- **Auto & Custom IDs**: Support for both auto-generated UUIDs and custom instance IDs
+- **Process Isolation**: Each vLLM instance runs in a separate process with isolated configuration
+- **Environment Variable Support**: Set custom environment variables per instance
```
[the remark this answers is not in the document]
MikeSpreitzer: What do you think of putting in a note that this is something that is planned to be added soon?

docs/launcher.md
diegocastanibm: True, I'll remove it.

docs/launcher.md
```diff
@@ -1,106 +1,635 @@
+# Multi-Instance vLLM Launcher Documentation
 
-# vLLM Launcher
+## Overview
 
-## Description
+The Multi-Instance vLLM Launcher (a.k.a. the launcher) is a Python program that implements a REST API service that allows you to dynamically create, manage, and terminate multiple vLLM inference server instances to achieve model swapping functionality without changes to vLLM. This enables flexible model serving where you can spin up different models on demand, and support multiple concurrent inference workloads.
```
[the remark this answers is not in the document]
diegocastanibm: hahaha, sure. Absolutely right

docs/launcher.md
MikeSpreitzer: > but when we use curl http://localhost:8001/v2/vllm/instances then we just report the running instances

This looks untrue to me. I am looking at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/1ec2d88778d1787620709f6bf677cd30e58af483/inference_server/launcher/launcher.py#L186, which is done for every instance regardless of whether it is "running".

docs/launcher.md
```diff
@@ -1,106 +1,635 @@
...
+  "total_instances": 3,
+  "running_instances": 2,
+  "instances": [
+    {
+      "status": "running",
+      "instance_id": "id-1",
+      "pid": 12345
+    },
+    {
+      "status": "stopped",
```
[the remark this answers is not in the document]
MikeSpreitzer: I think that the above recitation of the lifecycle visible to the REST client is missing at least one thing. It is possible that the vLLM child process gets launched and, after a while, terminates on its own (e.g., due to some misconfiguration in the vLLM command line parameters) --- right? At this time a query would report that instance existing with `"status": "stopped"`, right?

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
+# Multi-Instance vLLM Launcher Documentation
 
-# vLLM Launcher
+## Overview
 
-## Description
+The Multi-Instance vLLM Launcher (a.k.a. the launcher) is a Python program that implements a REST API service that allows you to dynamically create, manage, and terminate multiple vLLM inference server instances to achieve model swapping functionality without changes to vLLM. This enables flexible model serving where clients can spin up different models on demand, and support multiple concurrent inference workloads.
```
rubambiza: This first sentence is a bit overloaded with implementation vs motivation details. I would suggest breaking it up into two, with the second sentence saying "The motivating goal is to achieve model swapping....".

Also, a small nit for succinctness, concurrent implies multiple so we can take that out.

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+- **Dynamic Creation/Deletion**: Create and delete instances on demand via REST API
+- **Auto & Custom IDs**: Support for both auto-generated UUIDs and custom instance IDs
+- **Process Isolation**: Each vLLM instance runs in a separate process with isolated configuration
+- **Environment Variable Support**: Set custom environment variables per instance
+- **Graceful Shutdown**: Proper termination with configurable timeout and force-kill fallback
+- **Status Monitoring**: Query status of individual instances or all instances at once
+- **Health Checks**: Built-in health endpoint for monitoring service availability
+
+> [!NOTE]
+> This is still not implemeted, but the client controls the subset of the node's GPUs that get used by a given vLLM instance.
```
rubambiza: I think understand the `client` here to mean whichever pod is running the launcher, but a naive reader could misinterpret it as a client invoking the REST API. Perhaps rephrase to `the launcher process/pod`?

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+
+**POST** `/v2/vllm/instances`
+
+Create a new vLLM instance with an auto-generated UUID.
+
+**Request Body:**
+
+```json
+{
+  "options": "--model MODEL_NAME --port PORT",
```
rubambiza: Perhaps this is already addressed when I will read further down and our ongoing discussions regarding ports, but I am curious if we have addressed potential port clashes on the host.

For context, I have observed this in the wild where a program in a kind cluster couldn't start properly because its port was clashing with another program on my Mac. It took me days to figure this out and address it.

I imagine this is less likely on a cluster, but worth keeping in mind because, unlike the aforementioned example, we may not have control over the node that the launcher gets scheduled on.

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+- `options` (required): Command-line options for vLLM
+- `env_vars` (optional): Dictionary of environment variables
+
+**Response (201 Created):**
+
+```json
+{
+  "status": "started",
+  "instance_id": "uuid-string",
+  "pid": 12345
```
rubambiza: I do not understand why we are we returning the PID to the client if they will not use for any future calls. I imagine the `instance_id` is sufficient.

rubambiza: Also, this is very unlikely but entertain my dark thoughts. If I am a bad actor who somehow finds out which node the vllm launcher is running on, then I can easily issue a `kill -9` given the PID 🙈

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+- `instance_id`: ID of the instance to delete
+
+**Response (200 OK):**
 
 ```json
 {
   "status": "terminated",
-  "pid": 46910
+  "instance_id": "instance-id",
+  "pid": 12345
```
rubambiza: See my above comment on returning PID and for any subsequent returns below that give the PID.

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+# Delete instance
+curl -X DELETE http://localhost:8001/v2/vllm/instances/abc123...
 ```
 
-As a result, a JSON reply (like in the example below) will be sent to the client:
+### Example 2: Multiple Models on Different Ports
 
-```json
+```bash
+# Start Llama 2 on port 8000
```
rubambiza: I am a bit confusing on the mapping between the comments and the actual ports here. Did you mean 8010, 8011, and 8012 in the comments instead?

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+#   "running_instances": 2,
+#   "instances": [...]
+# }
+```
+
+## Configuration
+
+### vLLM Options
+
+The `options` field accepts any valid vLLM command-line arguments. Common options include:
```
rubambiza: How are we establishing what is valid? If we are checking against a static set, is it forward-compatible in case more options get added to the config?

rubambiza (changes requested):
The requested changes are related to:

- Removing potentially unnecessary returns of PIDs in REST responses
- Adding more details to some of the examples for consistency
- Making some of the sentences a bit more succinct

I also added some clarifying questions to make sure we are all on the same page.

docs/launcher.md
MikeSpreitzer: The launcher will run in a container in a Pod constructed by the dual-pods controller. A Pod intended to match the selector of an InferencePool --- different ones at different times. At the moment when a given launchee is running, the InferencePool for that launchee.

docs/launcher.md
MikeSpreitzer: This section header says "Example 4: Check All Running Instances". But it looks to me like "running" is not accurate, because the response can include instances that are "stopped".

docs/launcher.md
MikeSpreitzer: The launcher does not vet the options. I think a more accurate statement is something like the following.

> The `options` field contains stuff added to the command line of the launched `vllm serve`.

docs/launcher.md
diegocastanibm: I'll create an issue to remove the PID's then. The people have spoken

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+
+**Error Responses:**
 
+- `404 Not Found`: Instance not found
+
+---
+
+#### Delete All Instances
+
+**DELETE** `/v2/vllm/instances`
```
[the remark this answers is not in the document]
diegocastanibm: As you said, I'm constantly using it for testing (instead of killing the instances one-by-one). I still think that it has some value. I think depends on the client or use it or not. That said, I can add a note saying that it can be useful to terminate all the instances at the same time and that this functionality can be specially useful for testing purposes.

docs/launcher.md
diegocastanibm: Nice catch! Looks like the dark arts of copy-paste turned against me this time

docs/launcher.md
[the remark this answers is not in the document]
MikeSpreitzer: I am not sure what sort of note is useful in the documentation. If it is in the API then it needs to be documented. If it is there and documented then it is available for use. Currently the launcher puts no limits on who uses it or how. That is a whole 'nother topic.

docs/launcher.md
diegocastanibm: I have added Mike's sentence

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
-curl -X GET \
+# Create instance
+curl -X POST http://localhost:8001/v2/vllm/instances \
   -H "Content-Type: application/json" \
-  http://localhost:8000/v1/vllm
+  -d '{
+    "options": "--model facebook/opt-125m --port 8000"
+  }'
+
+# Response: {"status": "started", "instance_id": "abc123...", "pid": 12345}
```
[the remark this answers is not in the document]
diegocastanibm: I have deleted the responses from the examples as explained in a comment above. I think the responses are well documented above and prefer not to add in every single example (plus the responses could be different if, for example, there is an error). So for now, examples are just user commands

diegocastanibm:
Great comments and suggestions. Thank you @MikeSpreitzer  and @rubambiza .

docs/launcher.md
```diff
@@ -1,106 +1,633 @@
...
+
+```json
+{
+  "status": "OK"
+}
+```
+
+---
+
+#### Service Information
```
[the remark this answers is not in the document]
diegocastanibm: @rubambiza no rush for the approval. I will need to remove the PID also from the documentation after doing it (and testing that everything works) in the launcher too.

docs/launcher.md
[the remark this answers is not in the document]
MikeSpreitzer: This has been resolved to my satisfaction.

docs/launcher.md
```diff
@@ -1,106 +1,628 @@
+# Multi-Instance vLLM Launcher Documentation
 
-# vLLM Launcher
+## Overview
 
-## Description
+The Multi-Instance vLLM Launcher (a.k.a. the launcher) is a Python program that implements a REST API service that allows clients to dynamically create, manage, and terminate vLLM inference server instances. The goal is to achieve model swapping functionality without changes to vLLM. This enables flexible model serving where clients can spin up different models on demand, and support concurrent inference workloads.
```
MikeSpreitzer: Rather than "terminate", I think that "delete" would be more to the point.

docs/launcher.md
```diff
@@ -1,106 +1,628 @@
...
+```bash
+# Get detailed status
+curl http://localhost:8001/v2/vllm/instances
+```
+
+## Configuration
+
+### vLLM Options
+
+The `options` field contains stuff added to the command line of the launched `vllm serve`. Common options include:
```
MikeSpreitzer: This paragraph ends with an incomplete sentence. (Maybe you think that is not where the paragraph ends. But Markdown is clear on this matter.)

rubambiza (approved):
All comments have been resolved. Looks good for merging.

PR #147, opened 2025-11-14
Reorg so that "spi" declares the _interface_

MikeSpreitzer:
This PR changes the package hierarchy a little, so that the package named "spi" declares the _interface_ between the controller and the requester. This is good because the "i" in "spi" stands for "interface".

Prior to this PR, "spi" holds the _implementation_ of that interface. This PR moves that implementation to pkg/server/coordination.

### DIFF
--- a/cmd/requester/main.go
+++ b/cmd/requester/main.go
@@ -22,8 +22,8 @@ import (
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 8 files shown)


rubambiza (approved):
Conducted a quick review with a run of e2e tests in kind, no errors or weird behavior was observed.

PR #170, opened 2025-11-24
✨ Add log relaying to requester

### LINKED ISSUE #159: Extend requester to accept relay of vLLM instance log
This is part of #157

MikeSpreitzer:
This PR extends the requester container to accept log chunks.

Resolves #159

### DIFF
--- a/pkg/server/requester/coordination/server_test.go
+++ b/pkg/server/requester/coordination/server_test.go
@@ -99,16 +99,19 @@ func TestLogChunking(t *testing.T) {
 		if curLen > fullLogSize-fullLogSize/60 {
 			break
 		}
-		var startPos int
-		if curLen > 0 {
+		var startPos, chunkLen int
+		if rand.IntN(6) == 1 {
+			startPos = max(0, curLen-rand.IntN(4))
+		} else {
 			startPos = rand.IntN((curLen + fullLogSize) / 2)
 		}
-		chunkLen := 100 + rand.IntN(100)
-		if startPos > 400 {
-			chunkLen = rand.IntN(fullLogSize - startPos)
+		if rand.IntN(3) == 1 {
+			chunkLen = rand.IntN(8)
+		} else {
+			chunkLen = rand.IntN(100)
 		}
-		if rand.IntN(7) == 1 {
-			chunkLen = 0
+		if startPos+chunkLen > fullLogSize {
+			chunkLen = rand.IntN(fullLogSize - startPos)
 		}
 		chunkReader := strings.NewReader(string(rightLog[startPos : startPos+chunkLen]))
 		if rand.IntN(10) == 1 {


pkg/server/requester/coordination/server.go
```diff
@@ -143,13 +144,66 @@ func newSetReadyHandler(logger klog.Logger, ready *atomic.Bool, newReady bool) f
...
+		}
+		news, err := io.ReadAll(r.Body)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			_, _ = fmt.Fprintf(w, "Failed to read log chunk from request body: %s", err.Error())
+			return
+		}
+		if startPos+int64(len(news)) <= nextLogPos {
+			w.WriteHeader(http.StatusOK)
+			_, _ = fmt.Fprintf(w, "Accepted startPos=%d, chunkLen=%d, but that has nothing new; next pos=%d\n", startPos, len(news), nextLogPos)
```
diegocastanibm: I prefer to use `contentLen` as you use below. Also, it would be good to be consistent with the terminology in the response messages

pkg/server/requester/coordination/server_test.go
```diff
@@ -74,3 +77,60 @@ func FuzzServer(f *testing.F) {
 	})
 	cancel()
 }
+
+func TestLogChunking(t *testing.T) {
+	var rightLog [600]byte
```
diegocastanibm: can you name `600` as a constant with some explanation?

pkg/server/requester/coordination/server_test.go
```diff
@@ -74,3 +77,60 @@ func FuzzServer(f *testing.F) {
...
+	for idx := range 600 {
+		rightLog[idx] = '0' + byte(rand.IntN(10))
+	}
+	var logBuilder strings.Builder
+	ctx, cancel := context.WithCancel(t.Context())
+	var ready atomic.Bool
+	port := "28084"
```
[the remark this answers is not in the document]
diegocastanibm: OK, but please, could you create an issue and link this comments so we do not forget about it?

pkg/server/requester/coordination/server.go
MikeSpreitzer: I see only "chunkLength" in https://github.com/MikeSpreitzer/llm-d-fast-model-actuation/blob/8e9dea2680d4ee1c047922fbf7b4a7b0af3f4e98/pkg/server/requester/coordination/server.go

MikeSpreitzer:
The latest commits add testing of empty chunks and negative starting positions. I am not clear on what else is needed.

diegocastanibm:
> The latest commits add testing of empty chunks and negative starting positions. I am not clear on what else is needed.

These could be some of the boundary cases to test:
startPos = 0, chunkLen = 0 (empty write at beginning)
startPos = curLen, chunkLen = 1 (exactly one byte forward)
startPos = curLen, chunkLen = 0 (empty write at exact current position)
startPos = curLen - 1, chunkLen = 1 (exactly overwrites last byte)
startPos = curLen - 1, chunkLen = 2 (overwrites last byte + adds one new)
Chunk that exactly fills to fullLogSize (600 bytes)
What do you think?

PR #171, opened 2025-11-25
🐛 Make ensure-nodes-mapped.sh avoid and tolerate problems

MikeSpreitzer:
This PR makes two improvements in this script.

1. If the indexing pod for one node fails, the script continues on to other nodes rather than terminate.
2. The script does not attempt to index a node that is marked as unschedulable.


aavarghese (approved):
Ran script in both clusters for success path..

Issue #154, opened 2025-11-18
Outline of changes from milestone 2 to 3


MikeSpreitzer:
# Software engineering

We have a choice to make about how to develop m3 while maintaining a working m2. I can see three obvious possible approaches.

1. Use different branches. E.g., copy current m2 to branch `milestone-2` and start developing m3 on branch `main`.
2. Use different Go packages. E.g., leave `pkg/controller/dual-pods/` as it is, holding m2, and copy it into new package `pkg/controller/dual-pods-m3` and develop from there. Remember, m3 changes not only the controller but all the recipes and scripts that use it.
3. Introduce an option into the existing controller. E.g., a command line flag `--use-launcher` that takes a Boolean value.

I currently favor approach 3, because I think it will be least disruptive. I expect there may be some on-going maintenance for m2, and I hate maintaining two copies of (more or less) the same code. Also, the change from m2 and m3 can be fairly smooth, in terms of the code in the controller, I think.

# Changes to the dual-pods controller

## Authoritative store of binding state

In m2 this is an annotation on the server-providing Pod. I think that this can remain.

## Index into sleeping vLLM instances

In m2 this is an index maintained by the Pod informer. We can add an additional index into launched vLLM instances, in the controller's data structure. In `nodeData`, have a map from server-providing Pod name to `*launcherData`. Let `launcherData` have a map from nominal hash to last-used time. Use the nominal hash as the instance ID in the launcher. There is no need for a launcher to have multiple instances with the same nominal hash.

## Other data/logic notes

Let a launcher-based server-providing Pod have an annotation that says it is such a thing. Let the controller's handler for Pod notifications queue a reference to the Pod when notified of such a thing.

Let `launcherData` have a boolean indicating whether the set of nominal hash is accurate.

Let the controller use the extension of generic controller that knows whether the initial load of objects has been processed.

When syncing a launcher-based server-providing Pod, create and populate the `launcherData` if it is not already present. If it is present but not accurate, update it to become accurate.

When it is time to create a new vLLM instance, use either the m2 code or the m3 code, as indicated by the option.

Before creating a launcher-based vLLM instance, check that every `lancherData` of the Pod has an accurate set of nominal hash. If not then enqueue a reference to the Pod and consider this a transient failure.

Each launcher has a personality. Create a new instance using only a launcher with the right personality.

Create a new launcher when needed. Delete an old launcher when its set of instances is empty and it has not been used for two minutes.

[...]

PR #177, opened 2025-12-03
Make the controller care about launcher-based pods

waltforme:
Another step implementing the discussions in #154.

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -323,27 +323,27 @@ const (
 	infSvrItemDontCare infSvrItemType = "dont_care"
 )
 
-// careAbout returns an infSvrItem item and a bool owned.
-// Returns owned=true for server-requesting Pods, bound direct server-providing Pods, and launcher-based server-providing Pods;
-// returns owned=false for unbound direct providers and other Pods.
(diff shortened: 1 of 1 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -308,27 +308,42 @@ type infSvrItem struct {
...
+	infSvrItemLauncherBasedProvider infSvrItemType = "launcher_based_provider"
+	// infSvrItemDontCare is not a real infSvrItemType but only a placeholder
+	// saying the corresponding infSvrItem is not relevant to the controller.
+	infSvrItemDontCare infSvrItemType = "dont_care"
+)
+
+// careAbout returns an infSvrItem item and a bool owned.
+// Returns owned=true for server-requesting Pods, bound direct server-providing Pods, and launcher-based server-providing Pods;
+// returns owned=false for unbound direct providers and other Pods.
+func careAbout(pod *corev1.Pod) (item infSvrItem, it infSvrItemType, owned bool) {
```
MikeSpreitzer: This could be simplified by removing `owned bool`, because `owned == (it != infSvrItemDontCare)`.

MikeSpreitzer (approved):
This is good enough, and could be simplified a little.

PR #184, opened 2025-12-08
Create launcher benchmark dockerfile

manoelmarques:
Changes:

-  Add dockerfile for launcher to be used to create images for benchmarks to run at https://github.com/llm-d/llm-d-benchmark
- Change `FlexibleArgumentParser` import path since the previous path is invalid for recent vLLM versions. This path also works for old versions.
- Simplify `requirements.txt` to include only first level dependencies to the launcher.
- Add launcher image build to makefile
- Add launcher image build to github actions


inference_server/launcher/requirements.txt
```diff
@@ -55,7 +55,7 @@ mdurl==0.1.2
 mistral_common==1.8.4
 mlx==0.29.0
 mlx-lm==0.27.1
-mlx-metal==0.29.0
+mlx-metal==0.29.0; sys_platform == 'darwin'
```
rubambiza: I am not knowledgeable on the addition of platforms to Python requirements. Are we intending to run the benchmark only on MacOS?

rubambiza (commented):
Asking a clarifying question regarding the Python requirements. Otherwise, looks good to me.

I got a link to the changed files instead of the PR description. My previous question has been addressed after looking over the description. I will leave it to the originally requested reviewers @diegocastanibm and @lionelvillard to mark it as ready for merge.

PR #187, opened 2025-12-10
Add LauncherPoolPolicy

### LINKED ISSUE #185: FEAT: CRD Design Part 2 - LauncherPoolPolicy
Based on previous discussion, I summarize the following comments as below:

1. From @MikeSpreitzer   This interface needs to explicitly say, in the interface not just in the GitHub PR discussion, what happens if two different counts are specified (via two different CountForLauncher structs, perhaps in different LauncherPoolPolicy objects) for the same (Node, LauncherConfigReference). This interface needs to also explicitly say what happens if no CountForLauncher applies to a given (Node, LauncherConfigReference).
// LauncherCount is the total number of launcher pods to maintain.
	LauncherCount int32 `json:"launcherCount"`

2.  From @MikeSpreitzer  This interface needs to say what happens when multiple ServerProviderTemplateCount --- whether in the same NodePoolSpec or scattered among several --- reference the same LauncherConfig.

// ServerProviderTemplateCount is the total number of launcher or vLLM pods for each ServerProviderConfig
	// to maintain on each matching node.
	ServerProviderTemplateCount []ServerProviderTemplateCount `json:"totalCountPerTemplate"`

osswangxining:
Fix #185 

Introduce an EnhancedNodeSelector that supports combining label-based matching with resource field conditions. Its structure can be defined as follows (YAML example):

```
enhancedNodeSelector:
  # 1. Label selector (compatible with existing metav1.LabelSelector)
  labelSelector:
    matchLabels:
      nvidia.com/gpu.family: ada-lovelace
    matchExpressions:
      - key: node.kubernetes.io/instance-type
        operator: In
        values: ["gx3-48x240x2l40s", "gx3-96x480x4l40s"]

  # 2. Resource condition selector (new capability)
  resourceRequirements:
    allocatable:
      memory: ">=200Gi"
      nvidia.com/gpu: ">=1"
      cpu: ">=32"
    # Optional: also support capacity (if total capacity—not just allocatable—is desired)
    # capacity:
    #   memory: ">=256Gi"

```

This policy will automatically match all nodes that:

Have GPU model NVIDIA-L40S,
Have allocatable memory ≥ 200 GiB,
Have at least 1 GPU,
Have at least 40 CPU cores,

without requiring any manual labeling.


api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+*/
+
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// PoolPolicy defines the proactive provisioning policy for idle launcher pods.
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=launcherpoolpolicies,scope=Cluster
```
aavarghese: Sorry, couldn't attend the joint meeting yday but I believe it was agreed to change this resource to scope=Namespace? Please do inform what the plan is

api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	// ResourceRequirements defines the resource requirements for a node.
+	// +kubebuilder:validation:Optional
+	ResourceRequirements *ResourceRequirementSpec `json:"resourceRequirements,omitempty"`
+}
+
+type CountForLauncher struct {
+	// LauncherConfigName references the name of the LauncherConfig this policy applies to.
+	// +kubebuilder:validation:Required
+	LauncherConfigName string `json:"launcherConfigName"`
```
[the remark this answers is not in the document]
aavarghese: Actually, I take the above comment back. If LauncherPoolPolicy becomes a namespaced object, then we expect LauncherConfig to be located in the same ns. Sorry.

api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// will be ignored for this LauncherConfig.
+	// +kubebuilder:validation:MinItems=1
+	CountForLauncher []CountForLauncher `json:"countForLauncher"`
+}
+
+// ResourceRequirementSpec defines resource requirements for a node.
+type ResourceRequirementSpec struct {
+	// Allocatable defines the allocatable resources for a node.
+	// +kubebuilder:validation:Required
+	Allocatable map[string]string `json:"allocatable,omitempty"`
```
MikeSpreitzer: Using one-sided inequalities --- testing only `<=` --- will cause multiple matches where they are not intended. Consider the case of wanting one policy for nodes will a small amount of main memory and a different policy for nodes with a large amount of main memory. If the selector can only test whether the node's main memory is at least as big as a given value, then the policy intended for small nodes will match big ones too.

I recommend instead using a pair of upper and lower bounds. And express the quantities using https://github.com/kubernetes/apimachinery/blob/v0.34.3/pkg/api/resource/quantity.go#L102 , for consistency with how the quantities appear on Node objects.

api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// LauncherCount is the total number of launcher pods to maintain.
+	// +kubebuilder:validation:Required
+	LauncherCount int32 `json:"launcherCount"`
+}
+
+type LauncherPoolPolicyStatus struct {
+	// `observedGeneration` is the `metadata.generation` last seen by the controller.
+	// +optional
+	ObservedGeneration int32 `json:"observedGeneration,omitempty"`
```
MikeSpreitzer: Oops, I am sorry, I goofed up in an earlier comment recommending this. The type should actually be `int64`. See https://github.com/kubernetes/apimachinery/blob/v0.34.3/pkg/apis/meta/v1/types.go#L177

api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   LauncherPoolPolicySpec   `json:"spec,omitempty"`
+	Status LauncherPoolPolicyStatus `json:"status,omitempty"`
+}
+
+// LauncherPoolPolicySpec defines the node-level idle pool configuration.
+type LauncherPoolPolicySpec struct {
+	// LauncherPoolForNodeType defines pool spec per node type.
+	LauncherPoolForNodeType []LauncherPoolForNodeType `json:"launcherPoolForNodeType"`
```
MikeSpreitzer: The DRA work in Kubernetes contemplates the possibility that one node can have two (or more) types of accelerators. While the nvidia software does not seem to entirely support that today, I would prefer that this work not be equally shortsighted. With this in mind, it seems to me that the proper scoping for a pool is not (Node, LauncherConfig) but rather (Node, type of (perhaps more to the point, amount of memory in) accelerator, LauncherConfig).

api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// PoolPolicy defines the proactive provisioning policy for idle launcher pods.
```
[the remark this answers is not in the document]
aavarghese: @MikeSpreitzer Glad you brought this up here. 

My thoughts are that we create this policy strictly for pre-provisioning a pool of "idle/inactive" launchers to assist DPC with fast launch. Once DPC takes a launcher from the pool, and it officially becomes a "Server providing pod", is there a point for the LaunchPoolPolicy controller to still manage it and keep it as a launcher pool member? I would think not. It does not need to be included in the launcher count (under this policy). Once it becomes a server providing pod (and has a launcher that can start a number of vllm instances under DPC control), what is the advantage of pool controller tracking it?

api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// ResourceRequirementSpec defines resource requirements for a node.
+type ResourceRequirementSpec struct {
+	// Allocatable defines the allocatable resources for a node.
+	// +kubebuilder:validation:Required
+	Allocatable map[string]string `json:"allocatable,omitempty"`
+	// Capacity defines the capacity resources for a node.
+	// +kubebuilder:validation:Optional
+	Capacity map[string]string `json:"capacity,omitempty"`
```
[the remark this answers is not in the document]
osswangxining: remove for now.

api/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,130 @@
...
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// PoolPolicy defines the proactive provisioning policy for idle launcher pods.
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=launcherpoolpolicies,scope=Cluster
+
+type LauncherPoolPolicy struct {
```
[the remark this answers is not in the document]
osswangxining: The LauncherPoolPolicy semantics should be defined as below:

 ## Multiple LauncherPoolPolicy Objects

 1. **Additive Semantics**
   - Multiple LauncherPoolPolicy objects follow additive semantics
   - All policies across all objects are evaluated together
   - Rules from different objects are combined to form the complete policy set

 2. **Zero Objects Behavior**
   - When no LauncherPoolPolicy objects exist, no proactive provisioning occurs
   - System falls back to on-demand launcher pod creation

 ## Multiple LauncherPoolForNodeType Matching

 1. **Multiple Matches**
   - When multiple LauncherPoolForNodeType structs match a single Node (across same or different LauncherPoolPolicy objects):
     - For each unique combination of (Node, Accelerator, LauncherConfig), select the highest LauncherCount value
     - This forms the effective policy for that specific tuple

 2. **Zero Matches**
   - When no LauncherPoolForNodeType matches a Node:
     - No pre-provisioning policy applies to that Node
     - Launcher pods for that Node are created on-demand

 ## Multiple CountForLauncher with Same LauncherConfig

 1. **Duplicate LauncherConfig Names**
   - When multiple CountForLauncher structs specify the same LauncherConfigName for the same (Node, Accelerator) combination:
     - Select the highest LauncherCount value among all matching entries
     - This determines the target pool size for that LauncherConfig on that Node with that Accelerator

 2. **Zero Matching CountForLauncher**
   - When no CountForLauncher matches a specific (Node, Accelerator, LauncherConfig) tuple:
     - No pre-provisioning occurs for that specific combination
     - Launcher pods are created on-demand when needed

api/v1alpha1/launcherpoolpolicy_types.go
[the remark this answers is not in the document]
MikeSpreitzer: So if the policy calls for 8 idle launchers, there is a pool manager that will create 8 launchers, and as soon as one starts being used, the pool manager will create another one, to maintain a count of 8 idle launchers?

I thought we agreed on loose coupling through the kubernetes apiservers. The pool manager will maintain the fact that the desired number of launcher pods exist, regardless of whether any have launched vLLM instances. The dual-pods controller will use an existing launcher if it needs one and an acceptable one exists.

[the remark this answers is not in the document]
osswangxining: > So if the policy calls for 8 idle launchers, there is a pool manager that will create 8 launchers, and as soon as one starts being used, the pool manager will create another one, to maintain a count of 8 idle launchers?

Here I understand that pool manager will trigger one new round creation, if the resourceRequirements meets, it can create new launchers. But not meets, it will not create.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+	// the higher count is used and will be populated into LauncherPoolPolicyStatus.
+	// If no CountForLauncher applies to a given (Node, Accelerator, LauncherConfig), this Node
+	// will be ignored for this LauncherConfig.
+	// +kubebuilder:validation:MinItems=1
+	CountForLauncher []CountForLauncher `json:"countForLauncher"`
+}
+
+// ResourceRequirements defines resource requirements for a node.
+type ResourceRequirements struct {
+	// Allocatable defines the allocatable resources for a node with min/max ranges.	// +kubebuilder:validation:Required
```
MikeSpreitzer: Missing line break here.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+type ResourceRanges struct {
+	// CPU defines the CPU resource range requirement.
+	// +kubebuilder:validation:Optional
+	CPU *ResourceRange `json:"cpu,omitempty"`
+
+	// Memory defines the memory resource range requirement.
+	// +kubebuilder:validation:Optional
+	Memory *ResourceRange `json:"memory,omitempty"`
+
+	// GPUs defines the GPU resource range requirements keyed by GPU type.
```
MikeSpreitzer: This comment should say more about what are valid GPU types.

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: sorry done

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+	// AcceleratorSelector accelerator selector
+	AcceleratorSelector AcceleratorSelector `json:"acceleratorSelector"`
+
+	// CountForLauncher is the total number of launcher for each LauncherConfig
+	// to maintain on each matching node per accelerator.
+	// If two different counts are specified for the same (Node, Accelerator, LauncherConfig),
+	// the higher count is used and will be populated into LauncherPoolPolicyStatus.
+	// If no CountForLauncher applies to a given (Node, Accelerator, LauncherConfig), this Node
+	// will be ignored for this LauncherConfig.
+	// +kubebuilder:validation:MinItems=1
```
MikeSpreitzer: Why must this slice be non-empty, while the other slices have no such requirement?

The common pattern in Kubernetes APIs is to allow an empty slice and include `,omitempty` in the json tag on the field. I think that would be OK here too, although it is also true that an empty slice means the whole struct contributes nothing. On the third hand, forbidding an empty slice is a bit of irregularity: for any length larger than one, removing an entry is allowed, but for length == 1, removing an entry is not allowed.

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: Please @aavarghese @rubambiza  can append this info later?

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+	// +kubebuilder:validation:Optional
+	CPU *ResourceRange `json:"cpu,omitempty"`
+
+	// Memory defines the memory resource range requirement.
+	// +kubebuilder:validation:Optional
+	Memory *ResourceRange `json:"memory,omitempty"`
+
+	// GPUs defines the GPU resource range requirements keyed by GPU type.
+	// +kubebuilder:validation:Optional
+	GPUs map[string]*ResourceRange `json:"gpus,omitempty"`
```
MikeSpreitzer: What about TPUs? Maybe call this field "accelerators"?

osswangxining: fine.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+	Allocatable *ResourceRanges `json:"allocatable,omitempty"`
+}
+
+// ResourceRanges defines min/max ranges for various resources.
+type ResourceRanges struct {
+	// CPU defines the CPU resource range requirement.
+	// +kubebuilder:validation:Optional
+	CPU *ResourceRange `json:"cpu,omitempty"`
+
+	// Memory defines the memory resource range requirement.
```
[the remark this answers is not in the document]
osswangxining: yes, if only Memory wording , it means main memory

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+	// If no CountForLauncher applies to a given (Node, Accelerator, LauncherConfig), this Node
+	// will be ignored for this LauncherConfig.
+	// +kubebuilder:validation:MinItems=1
+	CountForLauncher []CountForLauncher `json:"countForLauncher"`
+}
+
+// ResourceRequirements defines resource requirements for a node.
+type ResourceRequirements struct {
+	// Allocatable defines the allocatable resources for a node with min/max ranges.	// +kubebuilder:validation:Required
+	Allocatable *ResourceRanges `json:"allocatable,omitempty"`
```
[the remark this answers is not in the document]
osswangxining: done

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// Min specifies the minimum quantity required.
+	// +kubebuilder:validation:Optional
+	Min resource.Quantity `json:"min,omitempty"`
+
+	// Max specifies the maximum quantity allowed.
+	// +kubebuilder:validation:Optional
+	Max resource.Quantity `json:"max,omitempty"`
+}
+
+// EnhancedNodeSelector defines node selector with label selector and resource requirements.
```
rubambiza: I would like to suggest moving this to line 120 so that the initial usage of `ResourceRequirements` (and its sub-types) and `AcceleratorSelector` come right before their definition. The `EnhancedNodeSelector` is much easier to parse that way.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+//
+// 2. **Zero Objects Behavior**
+//   - When no LauncherPoolPolicy objects exist, no proactive provisioning occurs
+//   - System falls back to on-demand launcher pod creation
+//
+// ## Multiple LauncherPoolForNodeType Matching
+//
+// 1. **Multiple Matches**
+//   - When multiple LauncherPoolForNodeType structs match a single Node (across same or different LauncherPoolPolicy objects):
+//     - For each unique combination of (Node, Accelerator, LauncherConfig), select the highest LauncherCount value
```
MikeSpreitzer: I think "Accelerator set" is meant here rather than "Accelerator".

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// +kubebuilder:validation:Optional
+	Min resource.Quantity `json:"min,omitempty"`
+
+	// Max specifies the maximum quantity allowed.
+	// +kubebuilder:validation:Optional
+	Max resource.Quantity `json:"max,omitempty"`
+}
+
+// EnhancedNodeSelector defines node selector with label selector and resource requirements.
+type EnhancedNodeSelector struct {
```
MikeSpreitzer: This file is getting large and complicated enough that I think it should have a consistent and readily grasped ordering among its declarations. I suggest that declarations appear after first use, and in order of first use. That would make `EnhancedNodeSelector` come before `PerAcceleratorCount`.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// +required
+	EnhancedNodeSelector EnhancedNodeSelector `json:"enhancedNodeSelector"`
+
+	// PerAcceleratorCount defines pre-configuration quantities for each accelerator type
+	PerAcceleratorCount []PerAcceleratorCount `json:"perAcceleratorCount"`
+}
+
+// PerAcceleratorCount defines configuration for specific accelerators
+type PerAcceleratorCount struct {
+	// AcceleratorSelector accelerator selector
```
rubambiza: I suggest duplicating the comment on [line 178](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/187/changes#diff-7c3f645c43fb2f682a77317ab4b531fbf7d7ceade27b0366dddaaf3bd642ee48R178) here since it's the first definition, regardless of whether the `EnhancedNodeSelector` is moved up or not as suggested in the [below comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/187/changes#r2611994971)

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	AcceleratorSelector AcceleratorSelector `json:"acceleratorSelector"`
```
MikeSpreitzer: In this position we do not need all the content of `AcceleratorSelector`. In particular, its `Count` field is not needed.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	//          min: "2"
+	//	        max: "8"
+	//  acceleratorSelector:
+	//    type: "nvidia.com/gpu"
+	//    memory: "48Gi"
+	//    count: 4
+	// +required
+	EnhancedNodeSelector EnhancedNodeSelector `json:"enhancedNodeSelector"`
+
+	// PerAcceleratorCount defines pre-configuration quantities for each accelerator type
```
MikeSpreitzer: It's actually per _homogeneous set_ of accelerators, right?

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+type ResourceRanges struct {
+	// CPU defines the CPU resource range requirement.
+	// +kubebuilder:validation:Optional
+	CPU ResourceRange `json:"cpu,omitempty"`
+
+	// Memory defines the memory resource range requirement.
+	// +kubebuilder:validation:Optional
+	Memory ResourceRange `json:"memory,omitempty"`
+
+	// Accelerators defines the GPU resource range requirements keyed by GPU type.
```
rubambiza: Is the GPU type here the same as the `Type` in `AcceleratorSelector`? In other words, would the cardinality of `PerAcceleratorCount[]` always match the set of keys in the `Accelerators` map on a given node?

MikeSpreitzer (commented):
I left some comments.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// If no CountForLauncher applies to a given (Node, Accelerator, LauncherConfig), this Node
+	// will be ignored for this LauncherConfig.
+	// +kubebuilder:validation:MinItems=1
+	CountForLauncher []CountForLauncher `json:"countForLauncher"`
+}
+
+// ResourceRequirements defines resource requirements for a node.
+type ResourceRequirements struct {
+	// Allocatable defines the allocatable resources for a node with min/max ranges.
+	// +kubebuilder:validation:Required
```
MikeSpreitzer: It would be better to simply write `// +required`.

rubambiza (changes requested):
- Requesting changes that would make it easier to parse the structs for future reviews and changes.
- Asking clarifying questions to better understand new design decisions.

@MikeSpreitzer, @osswangxining I would like to request that future PRs involving the pool policy/management get an approval from either @aavarghese or myself before merging, especially given that China is 13 hours ahead of us. This ensures we are all on the same page on big changes.

api/v1alpha1/launcherpoolpolicy_types.go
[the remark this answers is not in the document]
osswangxining: Regarding how to manage the Pool, I think we still need to collectively agree on the usage pattern. If we follow Mike's approach, it would be a one-time Pool creation model. However, based on my previous understanding, proactively creating Launcher Pods is meant to be a dynamic, ongoing process.

@MikeSpreitzer @rubambiza @aavarghese

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	Allocatable ResourceRanges `json:"allocatable,omitempty"`
+}
+
+// ResourceRanges defines min/max ranges for various resources of a Node.
+type ResourceRanges struct {
+	// CPU defines the CPU resource range requirement.
+	// +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
osswangxining: done

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: done

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: I think that The GPU type here should be within the Type in AcceleratorSelector.

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: make sense

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: yes

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: done

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: that's fine for me to remove this constraints

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// Accelerators defines the GPU resource range requirements keyed by GPU type.
+	// +optional
+	Accelerators map[string]ResourceRange `json:"accelerators,omitempty"`
+}
+
+// ResourceRange defines a range with minimum and maximum quantity values.
+type ResourceRange struct {
+	// Min specifies the minimum quantity required.
+	// +optional
+	Min resource.Quantity `json:"min,omitempty"`
```
MikeSpreitzer: Since this is optional, I think you want a level of pointer indirection here. The `,omitempty` is meaningless here because the marshaled form of `Resource.Quantity` has no way of saying "empty".

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// `observedGeneration` is the `metadata.generation` last seen by the controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// `errors` reports problems seen in the desired state of this object;
+	// in particular, in the version reported by `observedGeneration`.
+	// +optional
+	Errors []string `json:"errors,omitempty"`
+	// Add status fields if needed (e.g., current idle pod counts)
+}
```
MikeSpreitzer: It would be good for the Status to also have some information about how many such launcher Pods exist, but I am not sure I know a good way to express that.

One idea would be a potentially large slice giving the count for each (Node, accelerator set, Launcher). But that could get big. Another possibility would be to aggregate over the Nodes, collecting only the min and max. That might require a lot of work every time the controller updates the Status. But maybe it is only iterating over an internal data structure, which should be tolerable.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// +required
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	// ResourceRequirements defines the resource requirements for a node.
+	// +optional
+	ResourceRequirements *ResourceRequirements `json:"resourceRequirements,omitempty"`
+	// AcceleratorSelector defines accelerator-specific selection criteria
+	AcceleratorSelector *AcceleratorSelector `json:"acceleratorSelector,omitempty"`
+}
+
+// PerAcceleratorCount defines configuration for specific accelerators
```
MikeSpreitzer: Oh, I am sorry, I did not think through my previous suggestion of ordering very well. The declarations in this file form a tree. For a tree, I think that the best ordering is a pre-order traversal.

MikeSpreitzer:
In service of facilitating simple usage in simple scenarios, I have a suggestion. Currently, the proposed policy data structure is a three level map,

Node -> AcceleratorSet -> LaunchConfig -> count.

I suggest changing the data structure to a two level map, where the inner map has a two-part key combining AcceleratorSet and LaunchConfig:

Node -> (AcceleratorSet, LaunchConfig) -> count.

In particular, we should engineer the expression of "AcceleratorSet" here so that it can be expressed by writing nothing at all in the cases where all the Node's accelerators belong to the set. In practical terms, that would mean using field(s) with default value(s) that get interpreted to refer to all of the Node's accelerators. This allows users in this case to not explicitly mention the AcceleratorSet. Perhaps this means that we should think of it as a _filter_ rather than a _set_; it is common and unsurprising for an empty filter to accept everything (particularly if the filter is defined to be a conjunection of factors).

This also allows a development roadmap in which we start with no accelerator filtering in this position (NOTE: here I am not denying accelerator considerations in the node selector), and add it later.

osswangxining:
> @osswangxining By simple, do you mean assuming all the accelerators on a node are homogeneous?

Yes, and we can refine more on further implementation.

api/fma/v1alpha1/launcherpoolpolicy_types.go
osswangxining: sure, resolved on latest version

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,214 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   LauncherPoolPolicySpec   `json:"spec,omitempty"`
+	Status LauncherPoolPolicyStatus `json:"status,omitempty"`
+}
+
+// LauncherPoolPolicySpec defines the node-level idle pool configuration.
+type LauncherPoolPolicySpec struct {
+	// LauncherPoolForNodeType defines pool spec per node type.
+	LauncherPoolForNodeType []LauncherPoolForNodeType `json:"launcherPoolForNodeType"`
```
MikeSpreitzer: The current design has two levels of grouping, and I do not understand why that is needed. There can be multiple LauncherPoolPolicy objects, and each does nothing but bundle together multiple LauncherPoolForNodeType structs.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,214 @@
...
+	// (convenient for homogeneous clusters).
+	// +optional
+	AcceleratorSelector *AcceleratorSelector `json:"acceleratorSelector,omitempty"`
+}
+
+// ResourceRequirements defines resource requirements for a node.
+type ResourceRequirements struct {
+	// Allocatable defines the allocatable resources for a node with min/max ranges.
+	// +required
+	Allocatable ResourceRanges `json:"allocatable,omitempty"`
```
MikeSpreitzer: `// +required` and `,omitempty` clash.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// EnhancedNodeSelector defines node selector with label selector and resource requirements.
+type EnhancedNodeSelector struct {
+	// LabelSelector defines the label selector for a node.
+	// +required
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	// ResourceRequirements defines the resource requirements for a node.
+	// +optional
+	ResourceRequirements *ResourceRequirements `json:"resourceRequirements,omitempty"`
```
MikeSpreitzer: Can we simplify this by removing one level of structure that adds nothing? That is, replace this field with the following.

```yaml
// +optional
AllocatableResources ResourceRanges `json:"allocatableResources,omitempty"`
```

api/fma/v1alpha1/launcherpoolpolicy_types.go
aavarghese: Resolving this convo based on Mike's comment above

rubambiza (commented):
See this [resolved comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/187#discussion_r2623738437) as a stand-in for my review. Explicit approval will come from Ansu and/or Mike.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// LauncherCount is the total number of launcher pods to maintain.
+	// +required
+	LauncherCount int32 `json:"launcherCount"`
+
+	// Apply this count only to accelerators on the node that
+	// match this selector. When omitted, the count applies to the node regardless
+	// of accelerator sets (convenient for homogeneous clusters).
+	// +optional
+	AcceleratorSelector *AcceleratorSelector `json:"acceleratorSelector,omitempty"`
```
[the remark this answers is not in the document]
delavet: I agree. This field is not required to be added at the current moment, and it can be added when such a scenario is really needed.

aavarghese:
I only have a [follow-up question](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/187#discussion_r2623665180) inside a comment originally opened by Mike due to the recent commit. Nothing more to add other than I'm ok with merging this PR once Mike's comments above to simplify this CRD are addressed minimally in this PR. We can take care of the remaining separately.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+type ResourceRequirements struct {
+	// Allocatable defines the allocatable resources for a node with min/max ranges.	// +kubebuilder:validation:Required
+	Allocatable *ResourceRanges `json:"allocatable,omitempty"`
+}
+
+// ResourceRanges defines min/max ranges for various resources.
+type ResourceRanges struct {
+	// CPU defines the CPU resource range requirement.
+	// +kubebuilder:validation:Optional
+	CPU *ResourceRange `json:"cpu,omitempty"`
```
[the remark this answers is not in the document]
osswangxining: Here I think that using `*ResourceRange` for the CPU & Memory field is more appropriate for the following reasons:


1. **Clear Optional Semantics**: Using pointer types with the [omitempty] tag clearly indicates that the field is optional. When [CPU]  is not set, JSON serialization will omit this field.

2.  **Memory Efficiency**: When CPU resource range specification is not needed, a `nil` pointer avoids allocating memory space for [ResourceRange].

3. **API Design Consistency**: Consistent with Kubernetes API design patterns, optional fields typically use pointer types.

Does this make sense?

osswangxining:
Thanks @MikeSpreitzer @aavarghese @rubambiza  for all the comments.

Basically We've reached an agreement to first remove the Accelerator-related struct definitions and apply some simplifications. Regarding naming, I prefer "LauncherPopulationPolicy"—let's keep that for further discussion.

api/fma/v1alpha1/launcherpoolpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+// ## Multiple LauncherPoolPolicy Objects
+//
+// 1. **Additive Semantics**
+//   - Multiple LauncherPoolPolicy objects follow additive semantics
+//   - All policies across all objects are evaluated together
+//   - Rules from different objects are combined to form the complete policy set
+//
+// 2. **Zero Objects Behavior**
+//   - When no LauncherPoolPolicy objects exist, no proactive provisioning occurs
+//   - System falls back to on-demand launcher pod creation
```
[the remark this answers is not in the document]
MikeSpreitzer: The full answer should be in the interface, not just PR comments.

> In addition, there are also Launcher Pods that are created on-demand in response to inference requests.

That is not clear: are the launcher Pods created in response to server-request Pods strictly _additional_ to the ones called for by policy?

> // 2. **Zero Objects Behavior**
> //   - When no LauncherPoolPolicy objects exist, no proactive provisioning occurs
> //   - System falls back to on-demand launcher pod creation

This does not cover the case of when the server-requesting Pods need more launchers than the policy calls for.

MikeSpreitzer (commented):
I think that this needs more work but this PR already has an unwieldy number of comments. Let's merge this and continue work in follow-on PRs.

PR #195, opened 2025-12-25
Init - launching the pods per the LauncherPopulationPolicy custom resources

osswangxining:
Based on previous discussion , this is the initial version of LauncherPopulationPolicy reconciling logic. Please help review this @MikeSpreitzer @aavarghese @rubambiza

### DIFF
--- a/pkg/controller/launcher-pool/populator.go
+++ b/pkg/controller/launcher-pool/populator.go
@@ -23,9 +23,7 @@ import (
 	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	ctrl "sigs.k8s.io/controller-runtime"
-
+	"k8s.io/utils/ptr"
 	//apierrors "k8s.io/apimachinery/pkg/api/errors"
 	//metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -330,19 +328,18 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
 	if err != nil {
 		return fmt.Errorf("failed to get LauncherConfig %s/%s: %+v", ctl.namespace, launcherConfigName, err)
 	}
-	// Create runtime.Scheme instance
-	scheme := runtime.NewScheme()
-	// Register required types to scheme
-	_ = fmav1alpha1.AddToScheme(scheme)
-	_ = corev1.AddToScheme(scheme)
+
 	// Create the specified number of launcher pods
 	for i := 0; i < count; i++ {
 		pod := ctl.buildPodFromTemplate(launcherConfig.Spec.PodTemplate, key)
 		pod.GenerateName = fmt.Sprintf("launcher-%s-", launcherConfig.Name)
-		// Set owner reference so pods get cleaned up when LauncherConfig is deleted
-		if err := ctrl.SetControllerReference(launcherConfig, pod, scheme); err != nil {
-			return fmt.Errorf("failed to set controller reference: %w", err)
+		// Set owner reference pointing to LauncherConfig
+		ownerRef := *metav1.NewControllerRef(launcherConfig, fmav1alpha1.SchemeGroupVersion.WithKind("LauncherConfig"))
+		ownerRef.BlockOwnerDeletion = ptr.To(false)
+		pod.OwnerReferences = []metav1.OwnerReference{
+			ownerRef,
 		}
+
 		if _, err := ctl.coreclient.Pods(pod.Namespace).Create(ctx, pod, metav1.CreateOptions{}); err != nil {
 			return fmt.Errorf("failed to create launcher pod: %w", err)
 		}


cmd/launcher-pool-populator/main.go
```diff
@@ -0,0 +1,129 @@
...
+	case sig := <-sigChan:
+		logger.Info("Received signal, shutting down...", "signal", sig)
+		cancel() // Cancel the context when receiving a signal
+	}
+
+	// Wait for a period to allow resources to shut down gracefully
+	time.Sleep(5 * time.Second)
+}
+
+func AddFlags(flags pflag.FlagSet, loadingRules *clientcmd.ClientConfigLoadingRules, overrides *clientcmd.ConfigOverrides) {
```
MikeSpreitzer: Why omit the flag for `loadingRules.ExplicitPath` ?

It is OK to move commonly used code into `pkg/<something>` and import it in command packages.

pkg/controller/launcher-pool/node-launcher-key.go
```diff
@@ -0,0 +1,27 @@
+package launcherpool
+
+import (
+	"fmt"
+	"strings"
+)
+
+// NodeLauncherKey defines the unique identifier for a (Node, LauncherConfig) pair
+type NodeLauncherKey struct {
```
MikeSpreitzer: I recommend defining `func (NodeLauncherKey) String() string`.

pkg/controller/launcher-pool/node-matcher.go
```diff
@@ -0,0 +1,31 @@
...
+	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// matchesResourceConditions checks if a node's allocatable resources match the specified ranges
+func matchesResourceConditions(allocatable corev1.ResourceList, ranges fmav1alpha1.ResourceRanges) bool {
+	// If no resource ranges are specified, match all nodes
+	if len(ranges) == 0 {
+		return true
+	}
```
MikeSpreitzer: This is unnecessary, it duplicates the natural behavior of the following code (with no greater efficiency).

cmd/launcher-pool-populator/main.go
```diff
@@ -0,0 +1,129 @@
...
+	)
+	if err != nil {
+		klog.Fatal(err)
+	}
+
+	// Start informers
+	kubePreInformers.Start(ctx.Done())
+	fmaPreInformers.Start(ctx.Done())
+
+	// Start the controller in a goroutine
```
MikeSpreitzer: Why introduce a goroutine for this? Why not just do it inline?

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	genctlr "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/generic"
+)
+
+const ControllerName = "launcher-pool-populator"
+
+const (
+	LauncherComponentAnnotationKey   = "app.kubernetes.io/component"
```
MikeSpreitzer: This key is not specific to launchers, it is defined by the Kubernetes project. So the name should not suggest that it is specific to launchers.

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+	err := ctl.podInformer.AddIndexers(cache.Indexers{
+		launcherPodIndexName: launcherIndexFunc,
+	})
+	if err != nil { //impossible
+		return nil, err
+	}
+
+	ctl.QueueAndWorkers = genctlr.NewQueueAndWorkers(ControllerName, 1, ctl.process)
+	_, err = ctl.podInformer.AddEventHandler(ctl)
```
MikeSpreitzer: The consequent events are ignored by the handler, so why is this here?

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+	scheme := runtime.NewScheme()
+	// Register required types to scheme
+	_ = fmav1alpha1.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+	// Create the specified number of launcher pods
+	for i := 0; i < count; i++ {
+		pod := ctl.buildPodFromTemplate(launcherConfig.Spec.PodTemplate, key)
+		pod.GenerateName = fmt.Sprintf("launcher-%s-%s-", launcherConfig.Namespace, launcherConfig.Name)
+		// Set owner reference so pods get cleaned up when LauncherConfig is deleted
+		if err := ctrl.SetControllerReference(launcherConfig, pod, scheme); err != nil {
```
MikeSpreitzer: This can be done more simply as in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/v0.0.1/pkg/controller/dual-pods/inference-server.go#L333-L335 . That: has fewer module dependencies, no need to constructs a `Scheme` object, and no need to handle errors.

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+		podLister:     corev1PreInformers.Pods().Lister(),
+		nodeInformer:  corev1PreInformers.Nodes().Informer(),
+		nodeLister:    corev1PreInformers.Nodes().Lister(),
+		lppInformer:   fmaInformerFactory.Fma().V1alpha1().LauncherPopulationPolicies().Informer(),
+		lppLister:     fmaInformerFactory.Fma().V1alpha1().LauncherPopulationPolicies().Lister(),
+		lcInformer:    fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Informer(),
+		lcLister:      fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Lister(),
+	}
+	err := ctl.podInformer.AddIndexers(cache.Indexers{
+		launcherPodIndexName: launcherIndexFunc,
```
MikeSpreitzer: I do not see, nor anticipate, usage of this index.

api/fma/v1alpha1/register.go
```diff
@@ -33,6 +33,8 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&InferenceServerConfigList{},
 		&LauncherConfig{},
 		&LauncherConfigList{},
+		&LauncherPopulationPolicy{},
+		&LauncherPopulationPolicyList{},
```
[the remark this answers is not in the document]
osswangxining: yes, we append this here

pkg/controller/launcher-pool/node-launcher-key.go
osswangxining: agree, changed to :

```
func (k NodeLauncherKey) String() string {
	return fmt.Sprintf("%s/%s/%s", k.LauncherConfigNamespace, k.LauncherConfigName, k.NodeName)
}
```

pkg/controller/launcher-pool/node-launcher-key.go
```diff
@@ -0,0 +1,27 @@
...
+	LauncherConfigName      string
+	LauncherConfigNamespace string
+	NodeName                string
+}
+
+// mapToString converts a map of NodeLauncherKey to int32 values into a string representation.
+// This function formats the map as a string with the format "{namespace/name/node:count, ...}"
+// for debugging and logging purposes.
+func mapToString(m map[NodeLauncherKey]int32) string {
```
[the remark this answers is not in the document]
osswangxining: the fmt.Stringer interface cannot be used as a map key type,  map key types must support comparison operations, so I changed  as below:
```
func MapToLoggable(m map[NodeLauncherKey]int32) map[string]int32 {
	result := make(map[string]int32, len(m))
	for k, v := range m {
		result[k.String()] = v
	}
	return result
}
```

cmd/launcher-pool-populator/main.go
osswangxining: The ctlr.Start(ctx) method is blocking - it runs continuously until the context is cancelled
If executed inline, it would prevent the main goroutine from proceeding to signal handling

pkg/controller/launcher-pool/populator.go
osswangxining: yes, changed to dual-pods.llm-d.ai/component

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+	})
+	if err != nil { //impossible
+		return nil, err
+	}
+
+	ctl.QueueAndWorkers = genctlr.NewQueueAndWorkers(ControllerName, 1, ctl.process)
```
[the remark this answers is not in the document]
osswangxining: append the comments,  thanks

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+		return []string{string(pod.UID)}, nil
+	}
+
+	return []string{}, nil
+}
+
+func (ctl *controller) OnAdd(obj any, isInInitialList bool) {
+	switch typed := obj.(type) {
+	case *fmav1alpha1.LauncherPopulationPolicy:
+		if typed == nil {
```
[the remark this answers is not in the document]
osswangxining: done thanks

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+			return
+		}
+		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherPopulationPolicy reference due to notification of add", "name", typed.Name)
+		item := lppItem{cache.ObjectName{Name: typed.Name, Namespace: typed.Namespace}}
```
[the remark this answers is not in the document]
osswangxining: done thanks

api/fma/v1alpha1/register.go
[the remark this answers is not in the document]
MikeSpreitzer: That is OK to do. FYI, you could do that in a separate small PR --- which would be approved without further discussion.

pkg/common/flags.go
```diff
@@ -0,0 +1,14 @@
+package common
+
+import (
+	"github.com/spf13/pflag"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func AddFlags(flags pflag.FlagSet, loadingRules *clientcmd.ClientConfigLoadingRules, overrides *clientcmd.ConfigOverrides) {
```
MikeSpreitzer: Since this has global visibility and the package is not specific to Kubernetes client configuration, I would make the name of this function a little more specific (saying that it is about adding flags for Kubernetes client configuration). Not a blocker.

osswangxining:
> The policy objects are designed to operate independently. See how the code processes them independently? No need to make `process` work on all of them. Better to make `process` work on just the one referenced in the queue item.

will be polished within further PR

pkg/controller/launcher-pool/populator.go
osswangxining: As there are a lot of commits, I keep this for now and refine in another PR, thanks.

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,430 @@
...
+		"app.kubernetes.io/launcher-config-namespace": key.LauncherConfigNamespace,
+		"app.kubernetes.io/launcher-config-name":      key.LauncherConfigName,
+	}
+	// Use podLister's List method with label selector
+	pods, err := ctl.podLister.List(labels.SelectorFromSet(launcherLabels))
+	if err != nil {
+		return nil, fmt.Errorf("failed to list pods with launcher labels: %w", err)
+	}
+
+	// Filter pods that are on the specified node
```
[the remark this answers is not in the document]
MikeSpreitzer: As far as I know, Kubernetes does not maintain such a label. But this package could do that.

MikeSpreitzer:
Ignoring part (b) of https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/122494e0af21e48fa3c288fded52056dc013ecce/api/fma/v1alpha1/launcherpopulationpolicy_types.go#L38-L40 is a functional error.

osswangxining:
> Ignoring part (b) of
> 
> https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/122494e0af21e48fa3c288fded52056dc013ecce/api/fma/v1alpha1/launcherpopulationpolicy_types.go#L38-L40
> 
> is a functional error.

The current (Phase 1) implementation will initially focus only on proactively created Launcher Pods—specifically, calculating the desired number of Launcher Pods per node based on the LauncherPopulationPolicy definition.  

When we move to Phase 2, we’ll need to further clarify how to account for the number of Launcher Pods required to satisfy server-requesting Pods. I’m currently a bit confused about this part and would appreciate the team’s help in working through it together.

For example, the current LauncherPopulationPolicy defines a LauncherCount—the total number of launcher pods to maintain. Once this policy is active, should a server-requesting pod still trigger the creation of a dedicated launcher pod? Or should it instead act as a signal to either create or wake up an inference engine?

In any case, I believe this topic warrants a dedicated discussion thread rather than being addressed within the current PR.

cmd/launcher-pool-populator/main.go
osswangxining: Removing it would cause the program to be unable to properly respond to termination signals, potentially causing resource leaks and state inconsistencies, for example, if may not gracefully handle termination which will product the following result:

Process finished with the exit code 137 (interrupted by signal 9:SIGKILL)

MikeSpreitzer:
> For example, the current LauncherPopulationPolicy defines a LauncherCount—the total number of launcher pods to maintain. Once this policy is active, should a server-requesting pod still trigger the creation of a dedicated launcher pod? Or should it instead act as a signal to either create or wake up an inference engine?

https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/122494e0af21e48fa3c288fded52056dc013ecce/api/fma/v1alpha1/launcherpopulationpolicy_types.go#L38-L40 answers that question. The answer is that both the policy and the server-requesting Pods provide a lower bound on the number of Pods that exist. The reasons to exist are not mutually exclusive. There are not two disjoint populations of Pods. There are no pools. There are only Pods, each of which exists for either or both of two reasons (parts (a) and (b) of the cited statement).

pkg/common/flags.go
```diff
@@ -0,0 +1,32 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func AddKubernetesClientFlags(flags pflag.FlagSet, loadingRules *clientcmd.ClientConfigLoadingRules, overrides *clientcmd.ConfigOverrides) {
+	if loadingRules == nil {
+		return
+	}
+	if overrides == nil {
+		return
+	}
```
rubambiza: Can this be simplified to `if loadingRules == nil || overrides == nil`? Perhaps I am missing something.

MikeSpreitzer:
The dual-pods controller is already maintaining an annotation on bound server-providing Pods, described in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/7a838e25a2005ff031e38a35d5365a9919acee2c/pkg/controller/dual-pods/controller.go#L52-L54 . As noted in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/154#issue-3639885763, section "Authoritative store of binding state", I think that this can remain true in milestone 3. Since we are changing the key of that annotation from a private detail of the dual-pods controller to something shared with the population policy controller, the declaration of that key should move to a more appropriate location. BTW, that would NOT be https://github.com/llm-d-incubation/llm-d-fast-model-actuation/tree/main/pkg/api (which is for the interface with the user of the dual-pods technology).

api/fma/v1alpha1/register.go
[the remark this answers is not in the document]
MikeSpreitzer: I see no rebase. I see a merge of `main` into this PR's branch instead. https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/195#issuecomment-3712279475

pkg/controller/launcher-pool/populator.go
MikeSpreitzer: As discussed elsewhere, a policy object is only one of two possible reasons for a launcher Pod to exist. So it is appropriate for launcher Pod deletion to be handled by FMA controller logic rather than Kubernetes' generic object garbage collection.

The connection from Pod informer events to items in the controller's work queue is done by the controller's event handlers, regardless of what is in the owner reference list.

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,428 @@
...
+	// Create the specified number of launcher pods
+	for i := 0; i < count; i++ {
+		pod := ctl.buildPodFromTemplate(launcherConfig.Spec.PodTemplate, key)
+		pod.GenerateName = fmt.Sprintf("launcher-%s-%s-", launcherConfig.Namespace, launcherConfig.Name)
+		// Set owner reference so pods get cleaned up when LauncherConfig is deleted
+		if err := ctrl.SetControllerReference(launcherConfig, pod, scheme); err != nil {
+			return fmt.Errorf("failed to set controller reference: %w", err)
+		}
+		if _, err := ctl.coreclient.Pods(pod.Namespace).Create(ctx, pod, metav1.CreateOptions{}); err != nil {
+			return fmt.Errorf("failed to create launcher pod: %w", err)
```
MikeSpreitzer: What if the error is because the LauncherConfig's pod template has something bad in it?

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,428 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to delete excess launchers: %w", err)
+		}
+	}
+	return nil
+}
+
+// getCurrentLaunchersOnNode returns launcher pods for a specific config on a specific node
+func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLauncherKey) ([]corev1.Pod, error) {
+	launcherLabels := map[string]string{
```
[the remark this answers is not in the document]
osswangxining: yes, refine and waiting for #199 merged.

MikeSpreitzer:
I merged #199 , let's continue that work here.

osswangxining:
> It's OK with me if you want to start with only implementing part (a) and add part (b) in a later PR.

Sure, thanks @MikeSpreitzer  We can merge this PR firstly with long discussion :-) and then move to new PR for enhancement.

pkg/controller/launcher-pool/populator.go
osswangxining: fine, let me remove the controller-runtime for now and also can keep the flags easy to handle

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,428 @@
...
+		logger.Info("Found matching nodes", "count", len(nodes), "policy", lpp.Name)
+		// For each CountForLauncher rule
+		for _, countRule := range lpp.Spec.CountForLauncher {
+			for _, node := range nodes {
+				key := NodeLauncherKey{
+					NodeName:           node.Name,
+					LauncherConfigName: countRule.LauncherConfigName,
+				}
+				currentCount, exists := populationPolicy[key]
+				logger.Info("Current count for node", "node", node.Name, "countRule.LauncherConfigName",
```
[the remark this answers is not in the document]
osswangxining: updated as below.

```
logger.Info("Current count for node", "node", node.Name, "launcherConfigName",
					countRule.LauncherConfigName, "launcherCount", countRule.LauncherCount, "currentCount", currentCount)
```

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,428 @@
...
+					countRule.LauncherConfigName, "countRule.LauncherCount", countRule.LauncherCount, "currentCount", currentCount)
```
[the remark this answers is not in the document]
osswangxining: ok, alway enable the logging for this

pkg/controller/launcher-pool/populator.go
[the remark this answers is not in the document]
osswangxining: updated and pls help review again

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	pod.Labels[ComponentLabelKey] = LauncherComponentLabelValue
+	pod.Labels[LauncherGeneratedByLabelKey] = LauncherGeneratedByLabelValue
+	pod.Labels[LauncherConfigNameLabelKey] = key.LauncherConfigName
+	// Assign to specific node
+	pod.Spec.NodeName = key.NodeName
+	return pod
+}
+
+// getMatchingNodes returns nodes that match the EnhancedNodeSelector
```
rubambiza: May I suggest moving this function to before `reconcileAllLaunchers`? By the time I got to reviewing it, I had forgotten where it was invoked.

rubambiza (approved):
LGTM.

pkg/common/flags.go
osswangxining: refined https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/204

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,403 @@
...
+import (
+	"context"
+	"fmt"
+
+	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	//apierrors "k8s.io/apimachinery/pkg/api/errors"
+	//metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
osswangxining: refined https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/204

pkg/controller/launcher-pool/populator.go
```diff
@@ -0,0 +1,403 @@
...
+*/
+
+package launcherpool
+
+import (
+	"context"
+	"fmt"
+
+	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
osswangxining: refined https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/204

pkg/controller/launcher-pool/populator.go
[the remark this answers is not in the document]
osswangxining: resolved on https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/205 pls help review again, thanks.

PR #197, opened 2026-01-03
create AddKubernetesClientFlags func within pkg/common

osswangxining:
Split this common codes as one separated PR  which would be approved without further discussion.

### DIFF
--- a/pkg/common/flags.go
+++ b/pkg/common/flags.go
@@ -37,30 +37,9 @@ func AddKubernetesClientFlags(flags pflag.FlagSet, loadingRules *clientcmd.Clien
 		return
 	}
 
-	// Check if the flag already exists to avoid duplicate definition
-	if f := flags.Lookup("kubeconfig"); f != nil {
-		loadingRules.ExplicitPath = f.Value.String()
-	} else {
-		flags.StringVar(&loadingRules.ExplicitPath, "kubeconfig", loadingRules.ExplicitPath, "Path to the kubeconfig file to use")
-	}
-	if f := flags.Lookup("context"); f != nil {
-		overrides.CurrentContext = f.Value.String()
-	} else {
-		flags.StringVar(&overrides.CurrentContext, "context", overrides.CurrentContext, "The name of the kubeconfig context to use")
-	}
-	if f := flags.Lookup("user"); f != nil {
-		overrides.Context.AuthInfo = f.Value.String()
-	} else {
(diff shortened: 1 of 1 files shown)


pkg/common/flags.go
```diff
@@ -0,0 +1,56 @@
...
+func AddKubernetesClientFlags(flags pflag.FlagSet, loadingRules *clientcmd.ClientConfigLoadingRules, overrides *clientcmd.ConfigOverrides) {
+	if loadingRules == nil {
+		return
+	}
+	if overrides == nil {
+		return
+	}
+
+	// Check if the flag already exists to avoid duplicate definition
+	if flags.Lookup("kubeconfig") == nil {
```
MikeSpreitzer: The stated contract of this func does not allow it to avoid adding some flags. Why is this here?

MikeSpreitzer: But if controller-runtime is registering the `--kubeconfig` flag and the FMA code is NOT then the user can supply `--kubeconfig` but the FMA kube client setup code will not be affected by that flag.

We need to either get the FMA code to also access the value that the user provides, or just stop using controller-runtime so that there is not a competing definition of that flag.

osswangxining: @MikeSpreitzer  yes, you are right, I am missing the logic - also access the value if the user provides from controller-runtime

Updated the code.

pkg/common/flags.go
```diff
@@ -0,0 +1,66 @@
...
+		return
+	}
+
+	// Check if the flag already exists to avoid duplicate definition
+	if f := flags.Lookup("kubeconfig"); f != nil {
+		loadingRules.ExplicitPath = f.Value.String()
+	} else {
+		flags.StringVar(&loadingRules.ExplicitPath, "kubeconfig", loadingRules.ExplicitPath, "Path to the kubeconfig file to use")
+	}
+	if f := flags.Lookup("context"); f != nil {
```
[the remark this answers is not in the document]
osswangxining: currently I have problem for kubeconfig only, but I suggest to keep all the same - this also can help avoid if any

pkg/common/flags.go
```diff
@@ -0,0 +1,66 @@
...
+	if loadingRules == nil {
+		return
+	}
+	if overrides == nil {
+		return
+	}
+
+	// Check if the flag already exists to avoid duplicate definition
+	if f := flags.Lookup("kubeconfig"); f != nil {
+		loadingRules.ExplicitPath = f.Value.String()
```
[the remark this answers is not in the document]
osswangxining: fine, let me remove the controller-runtime for now and also can keep the flags easy to handle

PR #207, opened 2026-01-08
Update howto.md for the launcher

waltforme:
This PR tries to improve the launcher's how-to doc a little bit by
- fixing `env_var` in the json payload that differs from its schema;
- writing down the literal commands to make curl users' life easier;
- typo fix.


diegocastanibm (approved):
The full cURL commands are also listed in the launcher documentation in `docs/launcher.md`. That said, I do not think that having them also in the `howto.md` file hurts, so it's OK for me to approve this PR.

PR #208, opened 2026-01-08
✨ Update Go linting, add check of go mod tidy

MikeSpreitzer:
This PR adds explicit checking of whether `go mod tidy` needs to be done. This is good because the existing linting does not always clearly say this.

This PR also upgrades to the latest versions of the actions and linter used.

This PR also switches from identifying GitHub Actions by tag to hash, which is more secure against supply chain attacks.


osswangxining:
LGTM

PR #210, opened 2026-01-09
Add Dockerfile for launcher-populator

osswangxining:
Add Dockerfile for launcher-populator


aavarghese:
LGTM but just a design/consistency question. Since launcher-populator is also a Go controller, should we just use `ko` to build its image like dpctlr does today? Its not a big blocker or change but keep things similar for a FMA deployer person. cc @MikeSpreitzer @osswangxining @rubambiza

rubambiza (approved):
LGTM!

PR #211, opened 2026-01-09
Init the Helmchart for launcher populator

osswangxining:
Init the Helmchart for launcher populator

### DIFF
--- a/charts/launcher-populator/templates/rolebinding.yaml
+++ b/charts/launcher-populator/templates/rolebinding.yaml
@@ -12,7 +12,7 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: launcher-populator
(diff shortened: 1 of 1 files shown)


charts/launcher-populator/templates/rolebinding.yaml
```diff
@@ -0,0 +1,33 @@
...
+    app.kubernetes.io/component: controller
+  name: launcher-populator-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: launcher-populator-role
+subjects:
+  - kind: ServiceAccount
+    name: launcher-populator
+    namespace: {{ .Values.namespace }}
```
aavarghese: I believe this should be `.Release.Namespace` or not specify namespace like DP controller does it..

Issue #201, opened 2026-01-06
Part 2 - LauncherPopulationPolicy & Dual-pods controller logic refinement


osswangxining:
> > Need one sample to explain what the total number defined within policy object means.
> 
> Does the following example help? Suppose that there is one LauncherPopulationPolicy object, containing one CountForLauncher. We restrict our attention to the LauncherConfig and Node mentioned there. Suppose that the CountForLauncher's LauncherCount is 2. Suppose that before any server-requesting Pods are created, the population policy controller creates the 2 launcher Pods. Next, suppose that some user creates a server-requesting Pod that refers to the same LauncherConfig and gets scheduled on the same Node. This does _not_ call for the creation of another launcher Pod; the semantics in
> 
> https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/886b603be9400da62a075cfcf9d091ae16ce9ae5/api/fma/v1alpha1/launcherpopulationpolicy_types.go#L38-L40
> 
> uses a max of two numbers for the count of launchers, not two disjoint pools of launchers. The dual-pods controller will connect that server-requesting Pod with one of the already-existing launcher Pods. After that association is made and the inference server is launched, suppose that some time later the LauncherCount is reduced to 0. That does _not_ call for the deletion of the launcher Pod --- because the server-requesting Pod is still using it; that's part (b) of
> https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/886b603be9400da62a075cfcf9d091ae16ce9ae5/api/fma/v1alpha1/launcherpopulationpolicy_types.go#L38-L40

OK,  lets continue the scenario.  Suppose  all the 2 launcher pods are associated with existing 2 server-requesting pods, and the new server-requesting pod is created,  what it should do?   _not_ call  or call for the creation of another launcher pod?

_Originally posted by @osswangxining in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/195#issuecomment-3713676472_

MikeSpreitzer:
We merged #195 because the review was getting too long, not because it was really "done". It was admittedly just a start on the logic anyway. There are several open discussions in that PR, which should get worked in follow-on PRs.

The main issue that needs further work is correctly implementing the semantics for the population of launcher Pods.

MikeSpreitzer:
[...]

PR #218, opened 2026-01-12
Refine launcher pod deletion logic

osswangxining:
Refine the deletion logic - only for unbound Launcher pods.

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -29,7 +29,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -44,6 +43,7 @@ import (
 	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/api"
 	genctlr "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/generic"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
 	fmainformers "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/generated/informers/externalversions"
 	fmalisters "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/generated/listers/fma/v1alpha1"
 )
--- a/pkg/controller/launcher-populator/interface.go
+++ b/pkg/controller/launcher-populator/interface.go
@@ -28,6 +28,4 @@ const (
 	NodeNameLabelKey = "dual-pods.llm-d.ai/node-name"
 
 	NominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
-
-	RequesterAnnotationKey = "dual-pods.llm-d.ai/requester"
 )
--- a/pkg/controller/launcher-populator/interface.go
+++ b/pkg/controller/launcher-populator/interface.go
@@ -26,4 +26,8 @@ const (
 	LauncherConfigNameLabelKey = "dual-pods.llm-d.ai/launcher-config-name"
 
 	NodeNameLabelKey = "dual-pods.llm-d.ai/node-name"
+
+	NominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
+
+	RequesterAnnotationKey = "dual-pods.llm-d.ai/requester"
 )
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -388,7 +388,7 @@ func (ctl *controller) deleteExcessLaunchers(ctx context.Context, launchers []co
 	for i := 0; i < expectedDeleteCount && i < len(launchers); i++ {
 		pod := launchers[len(launchers)-1-i]
 
-		// 检查 Pod 是否仍处于未绑定状态
+		// Check if the Pod is still in an unbound state
 		refreshedPod, err := ctl.podLister.Pods(pod.Namespace).Get(pod.Name)
 		if err != nil {
 			if apierrors.IsNotFound(err) {
@@ -401,7 +401,7 @@ func (ctl *controller) deleteExcessLaunchers(ctx context.Context, launchers []co
 
 		isBound, requesterPodName := ctl.isLauncherBoundToServerRequestingPod(refreshedPod)
 		if isBound {
-			logger.Info("Skipping deletion of launcher pod as it is bound to a server-requesting pod",
+			logger.V(5).Info("Skipping deletion of launcher pod as it is bound to a server-requesting pod",
 				"pod", pod.Name, "server-requesting pod", requesterPodName)
 			continue
 		}
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -324,7 +324,7 @@ func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLau
 }
 
 // getCurrentLaunchersOnNode returns launcher pods for a specific config on a specific node
-func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLauncherKey) ([]corev1.Pod, error) {
+func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLauncherKey) ([]*corev1.Pod, error) {
 	launcherLabels := map[string]string{
 		ComponentLabelKey:          LauncherComponentLabelValue,
 		LauncherConfigNameLabelKey: key.LauncherConfigName,
@@ -337,9 +337,9 @@ func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLa
 	}
 
 	// Filter pods that are on the specified node
-	var filteredPods []corev1.Pod
+	var filteredPods []*corev1.Pod
 	for _, pod := range pods {
-		filteredPods = append(filteredPods, *pod)
+		filteredPods = append(filteredPods, pod)
 	}
 
 	return filteredPods, nil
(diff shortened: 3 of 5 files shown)


aavarghese (approved):
Makes sense and looks fine!

pkg/controller/dual-pods/controller.go
```diff
@@ -29,6 +29,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
```
rubambiza: I noticed this in #217 , but I didn't raise it in case it was fixed here. Minor nit: Move the import to line 44 so that all FMA imports are together.

pkg/controller/launcher-populator/populator.go
```diff
@@ -18,12 +18,22 @@ package launcherpopulator
...
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"os"
+	"strings"
 
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/api"
```
rubambiza: Minor: see above comment on mixing FMA imports with core Kubernetes imports.

pkg/controller/launcher-populator/populator.go
```diff
@@ -368,22 +378,67 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
...
+	}
+	for i := 0; i < actualDeleteCount && i < len(unboundLaunchers); i++ {
+		pod := unboundLaunchers[len(unboundLaunchers)-1-i]
 		if err := ctl.coreclient.Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{}); err != nil {
 			return fmt.Errorf("failed to delete launcher pod %s: %w", pod.Name, err)
 		}
 		logger.Info("Deleted launcher pod", "pod", pod.Name)
 	}
+
+	if actualDeleteCount < count {
```
rubambiza: Is this accounting for pods that could have gotten bound between the above check and the intermediate deletion of unbound pods?

rubambiza (changes requested):
Minor request for changes to maintain consistency.  Also asking a clarifying question that should not impede the merge, more so a curiosity.

pkg/controller/launcher-populator/populator.go
osswangxining: sure, I will resolve the imports finally.

osswangxining: I updated, and pls help review again, thanks.

MikeSpreitzer:
> BTW, this PR is based on #217 (pls review it firstly).

That remark is outdated now.

image

This PR needs to be rebased onto `main`.

pkg/api/interface.go
```diff
@@ -114,6 +114,10 @@ const DualLabelName string = "dual-pods.llm-d.ai/dual"
 // (it does not rely on this label for anything).
 const SleepingLabelName string = "dual-pods.llm-d.ai/sleeping"
 
+const NominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
+
+const RequesterAnnotationKey = "dual-pods.llm-d.ai/requester"
```
MikeSpreitzer: Neither of these is part of the interface to the dual-pods technique. These are internal to the technique, and should be declared in a different package. Not the generic controller package.

MikeSpreitzer:
See https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/217#discussion_r2692787414

pkg/api/interface.go
osswangxining: Agree, move to internal package, please review again, thanks @MikeSpreitzer

rubambiza (approved):
LGTM.

osswangxining:
> > BTW, this PR is based on #217 (pls review it firstly).
> 
> That remark is still outdated. You can edit the opening comment.

Thanks for reminder. I updated this commet.

pkg/controller/launcher-populator/interface.go
```diff
@@ -26,4 +26,8 @@ const (
 	LauncherConfigNameLabelKey = "dual-pods.llm-d.ai/launcher-config-name"
 
 	NodeNameLabelKey = "dual-pods.llm-d.ai/node-name"
+
+	NominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
+
+	RequesterAnnotationKey = "dual-pods.llm-d.ai/requester"
```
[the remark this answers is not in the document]
osswangxining: @MikeSpreitzer suggested NOT to place into pkg/api/interface.go which is for external instead of internal controller. 

So I think pkg/common/interface.go is better, do you think so? @aavarghese 

Anyway, I keep them here for now in this PR and once we make decision,  we can update 2 controllers together.

osswangxining:
> LGTM.

Thanks @rubambiza .

Do you have any other comments @MikeSpreitzer @aavarghese ? If not, I will merge this PR and keep an eye on this part within further PR.

pkg/controller/launcher-populator/populator.go
```diff
@@ -379,17 +380,89 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
...
-		pod := launchers[len(launchers)-1-i]
+
+	// Filter out pods that are bound to server-requesting pods
+	// Only delete unbound launcher pods
+	var unboundLaunchers []corev1.Pod
+	for _, pod := range launchers {
+		if !ctl.isLauncherBoundToServerRequestingPod(pod) {
+			unboundLaunchers = append(unboundLaunchers, pod)
+		} else {
+			logger.Info("Skipping deletion of launcher pod as it is bound to a server-requesting pod",
```
MikeSpreitzer: (A) Per https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-3717644115, this would be `V(5).Info`, I think.

(B) I would include the name of the server-requesting Pod.

MikeSpreitzer (commented):
Reviewed.

pkg/controller/launcher-populator/populator.go
```diff
@@ -379,17 +380,89 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
...
+				"pod", pod.Name)
+		}
+	}
+
+	// Delete the specified number of unbound launcher pods (starting from the end)
+	expectedDeleteCount := len(unboundLaunchers)
+	if count < expectedDeleteCount {
+		expectedDeleteCount = count
+	}
```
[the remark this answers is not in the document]
osswangxining: 1) switched to use minx(x,y)
2) count value may be less than len(unboundLaunchers)

pkg/controller/launcher-populator/populator.go
```diff
@@ -379,17 +380,89 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
...
+
+	deletedCount := 0
+	for i := 0; i < expectedDeleteCount && i < len(unboundLaunchers); i++ {
+		pod := unboundLaunchers[len(unboundLaunchers)-1-i]
+
+		// Re-check whether the Pod is still unbound
+		refreshedPod, err := ctl.podLister.Pods(pod.Namespace).Get(pod.Name)
```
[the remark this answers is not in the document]
osswangxining: This is based on comments from https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/218#discussion_r2684198307

pkg/controller/launcher-populator/interface.go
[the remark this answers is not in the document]
osswangxining: fine, I move to  pkg/controllers/common. And we can gradually move the common field into this file.

pkg/controller/launcher-populator/populator.go
```diff
@@ -379,17 +381,85 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
...
-	for i := 0; i < count && i < len(launchers); i++ {
+
+	expectedDeleteCount := min(len(launchers), count)
+
+	deletedCount := 0
+	for i := 0; i < expectedDeleteCount && i < len(launchers); i++ {
 		pod := launchers[len(launchers)-1-i]
-		if err := ctl.coreclient.Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{}); err != nil {
+
+		// 检查 Pod 是否仍处于未绑定状态
```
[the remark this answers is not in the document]
osswangxining: sorry for missing this. Updated now.

pkg/controller/launcher-populator/populator.go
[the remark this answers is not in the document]
MikeSpreitzer: BTW, wouldn't it be simpler for `getCurrentLaunchersOnNode` to return a slice of _pointer to_ Pod rather than a slice of Pod?

MikeSpreitzer (commented):
There is further discussion in the thread started at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/218#pullrequestreview-3669426631

pkg/controller/launcher-populator/populator.go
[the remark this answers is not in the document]
osswangxining: The reason why the code doesn't directly use the passed-in launchers parameter is due to data freshness and consistency concern. Do you think that it's unnessary?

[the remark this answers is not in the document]
MikeSpreitzer: The second fetch here is unnecessary. The first fetch in getCurrentLaunchersOnNode happened a very short amount of time ago and the freshness issue is fully handled by the test of the Pod's ResourceVersion in the Delete operation.

pkg/controller/launcher-populator/populator.go
```diff
@@ -377,19 +373,73 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
...
+				continue
+			}
 			return fmt.Errorf("failed to delete launcher pod %s: %w", pod.Name, err)
 		}
 		logger.Info("Deleted launcher pod", "pod", pod.Name)
+		deletedCount++
+	}
+
+	if deletedCount < count {
+		logger.Info("Fewer launcher pods were deleted than requested due to bound pods or concurrent changes",
```
MikeSpreitzer: In this case we want the controller to try again later, after it has reason to suspect that something material has changed. There are two cases here: (1) the concurrent activity that caused the Delete to be refused and (2) usage by a server-requesting Pod. For (1) an immediate retry is warranted. For (2) this needs more logic in the controller's event handlers.

Either or both of these could be addressed either in this PR or later.

MikeSpreitzer (approved):
This is good enough to merge. I left a comment that could be addressed in the PR or later.

PR #53, opened 2025-09-26
GPU Translator

diegocastanibm:
**Description:**
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/51

**Test:**
```
$ python -m pytest tests/test_gputranslator.py 
=========================================================================================================================================== test session starts ============================================================================================================================================
platform darwin -- Python 3.12.10, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/diego/Documents/my_stuff/llm-d-fast-model-actuation/inference_server/launcher
plugins: anyio-4.10.0
collected 2 items                                                                                                                                                                                                                                                                                          

tests/test_gputranslator.py ..                                                                                                                                                                                                                                                                       [100%]

============================================================================================================================================ 2 passed in 0.02s =============================================================================================================================================
```

PR #219, opened 2026-01-12
From GPU UUID into GPU index; launcher integration

diegocastanibm:
**Description:**
In https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/53, we add a capability to translate from UUIDs into index for the GPUs. That capability was not integrated into the launcher.
This PR aims to integrate the GPU translation into the launcher.

We have added a new attribute in the `VllmConfig` dataclass called `gpu_uuids` that receives the UUIDs in a list of strings.

**Tests:**
The test has been modified to add the new attribute.


inference_server/launcher/launcher.py
```diff
@@ -37,18 +38,35 @@
...
-
+        """
+        Initialize vLLMInstance object
         :param instance_id: Instance id (autogenerated or custom)
         :param config: VllmConfig object
         """
+
+        # Check for CUDA device UUIDs and set CUDA_VISIBLE_DEVICES accordingly
+        if config.gpu_uuids:
+            gpu_translator = GpuTranslator()
```
MikeSpreitzer: This will invoke https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/64579825a1538fa2e20757190c4729e81ccb9887/inference_server/launcher/gputranslator.py#L36 every time a vllm instance is created, right? That seems like a lot of redundant work.

Also: to support behavioral testing, it would be good to have a way to create a GpuTranslator with a given list of fake UUIDs rather than reading real ones from real GPUs.

inference_server/launcher/launcher.py
```diff
@@ -37,18 +38,37 @@
...
     """Represents a single vLLM instance"""
 
-    def __init__(self, instance_id: str, config: VllmConfig):
-        """_summary_
-
+    def __init__(
+        self, instance_id: str, config: VllmConfig, gpu_translator: GpuTranslator
+    ):
+        """
+        Initialize vLLMInstance object
```
MikeSpreitzer: Don't you want a lowercase "m" here?

inference_server/launcher/launcher.py
```diff
@@ -37,18 +38,37 @@
...
+        :param gpu_translator: GpuTranslator object
         """
+
+        # Check for CUDA device UUIDs and set CUDA_VISIBLE_DEVICES accordingly
+        if config.gpu_uuids:
+            cuda_indices = []
+            for uuid_str in config.gpu_uuids:
+                index = gpu_translator.uuid_to_index(uuid_str)
+                cuda_indices.append(str(index))
+
```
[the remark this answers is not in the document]
MikeSpreitzer: Or, better yet, change its type from `Optional[Dict[str, Any]]` to `Dict[str, Any]`

PR #220, opened 2026-01-13
Enforce immutability of FMA pods' annotations and labels with CEL ValidatingAdmissionPolicy

aavarghese:
Based on design discussion here: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-3717326061

FYI The exact set of labels and annotation that are checked for may change as milestone M3 and the populator is developed....

### DIFF
--- a/docs/e2e-recipe.md
+++ b/docs/e2e-recipe.md
@@ -64,11 +64,18 @@ to use for Node get/list/watch authorization, or omit if not
 needed. Adjust the SleeperLimit setting to your liking (the default is
 2).
 
+If your cluster does not support `ValidatingAdmissionPolicy` objects with CEL (`admissionregistration.k8s.io/v1`), set EnableValidationPolicy to false (the default is
+true) so that the install does not attempt to create unsupported resources. More about validation admission policies [here](#validating-admission-policies-cel).
+
+```shell
+POLICIES_ENABLED=false # SET TO WHAT FITS YOUR CLUSTER
+```
+
 NOTE: if you have done this before then you will need to delete the
 old Helm chart instance before re-making it.
 
 ```shell
-helm upgrade --install dpctlr charts/dpctlr --set Image="${CONTAINER_IMG_REG}/dual-pods-controller:${CONTROLLER_IMG_TAG}" --set NodeViewClusterRole=vcp-node-viewer --set SleeperLimit=1
+helm upgrade --install dpctlr charts/dpctlr --set Image="${CONTAINER_IMG_REG}/dual-pods-controller:${CONTROLLER_IMG_TAG}" --set NodeViewClusterRole=vcp-node-viewer --set SleeperLimit=1 --set EnableValidationPolicy=${POLICIES_ENABLED}
 ```
 
 Finally, define a shell function that creates a new ReplicaSet whose
@@ -101,20 +108,6 @@ Following are some things to keep in mind about these definitions.
   to fit into requested memory, or `--kv-cache-memory=40974729216` to
   fully utilize gpu memory. Current kv cache memory in use is
   32786058444 bytes".
-
-### Admission policies (CEL)
-
-Kubernetes `ValidatingAdmissionPolicy` (CEL) resources are used to
-protect controller-managed annotations and labels that are
-managed by the controllers (dual-pods controller and
-launcher-populator). The policy manifests and their
-bindings are managed as Helm chart templates under `charts/dpctlr/templates/policies/` and are installed by the chart when `policies.enabled` is `true`.
-
-#### Prerequisites
-- A Kubernetes API server that supports `ValidatingAdmissionPolicy` with CEL (`admissionregistration.k8s.io/v1`).
-
-`./test/e2e/run.sh` invokes the validation checker (`./test/e2e/validate.sh`) as part of the project's e2e suite; GitHub Actions run the suite when the target kind cluster supports the admission APIs.
-
 
 ### Simple ReplicaSet
 
@@ -496,3 +489,14 @@ Or, for more fun, before going past N+1, make a server-requesting Pod
 that causes the oldest runner to be re-used. Then delete that
 requester. Then force a deletion; observe that the deled one is the
 least recently used.
+
+## Validating admission policies (CEL)
+
+Kubernetes `ValidatingAdmissionPolicy` (CEL) resources are used to
+protect a subset of annotations and labels that are critical to FMA
+controller operations (dual-pods controller and
+launcher-populator). The policy manifests and their
+bindings are managed as Helm chart templates under `charts/dpctlr/templates/policies/` and can be installed by the chart when
+`EnableValidationPolicy` is `true`.
+
+For admission policy support, it is required to have a Kubernetes API server that supports `ValidatingAdmissionPolicy` with CEL (`admissionregistration.k8s.io/v1`). If your cluster does not support `ValidatingAdmissionPolicy` objects, set `EnableValidationPolicy=false` (the default is true) when installing the chart so the Helm release does not attempt to create unsupported resources.
--- a/test/e2e/validate.sh
+++ b/test/e2e/validate.sh
@@ -14,19 +14,14 @@ trap 'rc=$?; cleanup; exit $rc' EXIT
 POD_NAME=${POD_NAME:-my-request-test}
 BOUND_POD_NAME=${BOUND_POD_NAME:-my-bound-request-test}
 
-if kubectl api-resources --api-group=admissionregistration.k8s.io -o name | grep -q 'validatingadmissionpolicies'; then
-  echo "Cluster supports ValidatingAdmissionPolicy (CEL)"
-  if command -v helm >/dev/null 2>&1 && [ -d "charts/dpctlr" ]; then
-    helm template dpctlr charts/dpctlr --set policies.enabled=true | kubectl apply -f -
-    # wait for policies to be registered
-    sleep 30
-  else
-    echo "Helm not available or 'charts/dpctlr' missing; cannot install policies automatically."
-    exit 1
-  fi
-else
-  echo "Cluster does not support ValidatingAdmissionPolicy (CEL). Skipping tests."
+if ! kubectl api-resources --api-group=admissionregistration.k8s.io -o name | grep -q 'validatingadmissionpolicies'; then
+  echo "Cluster does not support ValidatingAdmissionPolicy. Skipping."
   exit 0
+fi
+
+if ! kubectl get validatingadmissionpolicy fma-immutable-fields fma-bound-serverreqpod >/dev/null 2>&1; then
+  echo "ERROR: Required validating admission policies not found. Ensure run.sh installed them correctly."
+  exit 1
 fi
 
 cat <<EOF | kubectl apply -f -
@@ -51,7 +46,12 @@ spec:
     command: ["/bin/sh","-c","sleep 3600"]
 EOF
 
-echo "Created launcher pod ${POD_NAME}"
+echo "Created launcher pod ${POD_NAME} and waiting for it to become Ready"
+
+if ! kubectl wait --for=condition=Ready pod/"${POD_NAME}" --timeout=60s >/dev/null 2>&1; then
+  echo "ERROR: pod ${POD_NAME} did not become Ready within 60s"
+  exit 2
+fi
 
 echo "Attempting to change immutable annotation 'dual-pods.llm-d.ai/requester' as a regular user — expect rejection"
 if output=$(kubectl patch pod "${POD_NAME}" -p '{"metadata":{"annotations":{"dual-pods.llm-d.ai/requester":"xyz patched-requester"}}}' --type=merge 2>&1); then
@@ -71,6 +71,15 @@ else
   echo "SUCCESS: label patch was rejected, as expected. kubectl output: ${output}"
 fi
 
+echo "Attempting to delete immutable label 'dual-pods.llm-d.ai/dual' — expect rejection"
+if output=$(kubectl patch pod "${POD_NAME}" -p '{"metadata":{"labels":{"dual-pods.llm-d.ai/dual":null}}}' --type=merge 2>&1); then
+  echo "ERROR: label deletion succeeded but should have been rejected"
+  echo "kubectl output: ${output}"
+  exit 5
+else
+  echo "SUCCESS: label deletion was rejected, as expected. kubectl output: ${output}"
+fi
+
 cat <<EOF | kubectl apply -f -
 apiVersion: v1
 kind: Pod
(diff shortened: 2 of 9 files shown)


pkg/webhook/pod/validate_annotations.go
```diff
@@ -0,0 +1,141 @@
...
+				api.AdminPortAnnotationName,
+			}
+			// Allow the FMA controllers to update these fields.
+			// Usernames for FMA service accounts are of the form
+			// "system:serviceaccount:<ns>:<sa>".
+			userIsController := false
+			if req.UserInfo.Username != "" {
+				ua := req.UserInfo
+				// check for FMA service accounts
+				if strings.HasSuffix(ua.Username, ":launcher-populator") || strings.HasSuffix(ua.Username, ":dual-pods-controller") {
```
rubambiza: To make things a bit cleaner, I'd like to suggest the following:
- Replace `:launcher-populator` with `":" +`[`launcherpopulator.ControllerName`](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/64579825a1538fa2e20757190c4729e81ccb9887/pkg/controller/launcher-populator/populator.go#L41)`
- Replace `:dual-pods-controller` with `":" +` [`dualpods.ControllerName`](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/64579825a1538fa2e20757190c4729e81ccb9887/pkg/controller/dual-pods/controller.go#L91)

This ensures that any future changes to the constants do not have to be updated here, too. It also works with the Copilot suggestion below.

rubambiza (changes requested):
Overall, I think this is a great starting point! I particular like that we can test out the functionality before contemplating its integration into the larger project. See below requested changes and suggestions.

config/policies/validating-admission-policy-binding-fields.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: bind-fma-immutable-fields
+spec:
+  policyName: fma-immutable-fields
+  matchResources:
+    resources:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        kinds: ["Pod"]
+        operations: ["Update"]
```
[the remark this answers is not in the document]
aavarghese: Done

config/policies/validating-admission-policy-bound-serverReqPod.yaml
```diff
@@ -0,0 +1,35 @@
...
+        (
+          (( 'dual-pods.llm-d.ai/inference-server-config' in oldObject.metadata.annotations ? oldObject.metadata.annotations['dual-pods.llm-d.ai/inference-server-config'] : "") != "") &&
+          (( 'dual-pods.llm-d.ai/dual' in oldObject.metadata.labels ? oldObject.metadata.labels['dual-pods.llm-d.ai/dual'] : "") != "") &&
+          (
+              (('dual-pods.llm-d.ai/inference-server-config' in oldObject.metadata.annotations) != ('dual-pods.llm-d.ai/inference-server-config' in object.metadata.annotations)) ||
+              (('dual-pods.llm-d.ai/admin-port' in oldObject.metadata.annotations) != ('dual-pods.llm-d.ai/admin-port' in object.metadata.annotations)) ||
+              (('dual-pods.llm-d.ai/accelerators' in oldObject.metadata.annotations) != ('dual-pods.llm-d.ai/accelerators' in object.metadata.annotations)) ||
+              (('dual-pods.llm-d.ai/dual' in oldObject.metadata.labels) != ('dual-pods.llm-d.ai/dual' in object.metadata.labels))
+          )
+        ) && !request.userInfo.username.matches("system:serviceaccount:[^:]+:(launcher-populator|dual-pods-controller)$")
```
rubambiza: I think the common case here is that the controllers will be the ones updating the pods. Thus, I think it makes sense to move the user check to be the first condition such that the common case short-circuits, and the check for changes is done only when non-controller service accounts are attempting to update the pods. What do you think?

aavarghese: Added in latest commit. Thanks Gloire!

config/policies/validating-admission-policy-bound-serverReqPod.yaml
```diff
@@ -0,0 +1,37 @@
...
+    # once a server-requesting Pod is bound to a server-providing Pod.
+    #
+    # Logic:
+    # 1) If the old object is a server requesting pod (`inference-server-config`),
+    # 2) AND the `dual` label is present and non-empty (bound state),
+    # 3) AND any of the immutable fields change
+    # 4) THEN deny unless the request originates from an FMA controller SA.
+    - expression: |
+        (
+          !request.userInfo.username.matches("system:serviceaccount:[^:]+:(launcher-populator|dual-pods-controller)$") && (
```
MikeSpreitzer: Maybe put a `^` at the start of the regex.

test/e2e/test-validating-admission-policies.sh
```diff
@@ -0,0 +1,129 @@
...
+  kubectl delete -f config/policies/validating-admission-policy-binding-fields.yaml --ignore-not-found --wait=true || true
+  kubectl delete -f config/policies/validating-admission-policy-bindings-serverReqPod.yaml --ignore-not-found --wait=true || true
+  kubectl delete -f config/policies/validating-admission-policy-immutable-fields.yaml --ignore-not-found --wait=true || true
+  kubectl delete -f config/policies/validating-admission-policy-bound-serverReqPod.yaml --ignore-not-found --wait=true || true
+
+  echo "Applying CEL policies and bindings"
+  kubectl apply -f config/policies/validating-admission-policy-immutable-fields.yaml
+  kubectl apply -f config/policies/validating-admission-policy-bound-serverReqPod.yaml
+  kubectl apply -f config/policies/validating-admission-policy-binding-fields.yaml
+  kubectl apply -f config/policies/validating-admission-policy-bindings-serverReqPod.yaml
```
MikeSpreitzer: Installing the policy objects should be part of installing FMA, not part of testing FMA.

test/e2e/test-validating-admission-policies.sh
```diff
@@ -0,0 +1,129 @@
...
+
+echo "Created launcher pod ${POD_NAME} in ${NS}"
+
+echo "Attempting to change immutable annotation as a regular user — expect rejection"
+set +e
+PATCH_CMD=$(kubectl patch pod ${POD_NAME} -n ${NS} -p '{"metadata":{"annotations":{"dual-pods.llm-d.ai/nominal":"test-nominal-patched"}}}' --type=merge 2>&1)
+PATCH_RC=$?
+set -e
+
+if [ ${PATCH_RC} -eq 0 ]; then
```
MikeSpreitzer: This could be done much more directly as follows.

```shell
if output=$(kubectl patch ... 2>&1); then
```

config/policies/validating-admission-policy-bound-serverReqPod.yaml
```diff
@@ -0,0 +1,37 @@
...
+            ((has(oldObject.metadata.annotations) && ('dual-pods.llm-d.ai/inference-server-config' in oldObject.metadata.annotations) ? oldObject.metadata.annotations['dual-pods.llm-d.ai/inference-server-config'] : "") != "") &&
+            ((has(oldObject.metadata.labels) && ('dual-pods.llm-d.ai/dual' in oldObject.metadata.labels) ? oldObject.metadata.labels['dual-pods.llm-d.ai/dual'] : "") != "") &&
```
MikeSpreitzer: This could be written more simply as follows.

```
            (has(oldObject.metadata.labels) && 'dual-pods.llm-d.ai/dual' in oldObject.metadata.labels && oldObject.metadata.labels['dual-pods.llm-d.ai/dual'] != "") &&
```

config/policies/validating-admission-policy-bound-serverReqPod.yaml
```diff
@@ -0,0 +1,37 @@
...
+            (
+                ((has(oldObject.metadata.annotations) && ('dual-pods.llm-d.ai/inference-server-config' in oldObject.metadata.annotations)) != (has(object.metadata.annotations) && ('dual-pods.llm-d.ai/inference-server-config' in object.metadata.annotations))) ||
```
MikeSpreitzer: This is not testing for a change in the value of the "inference-server-config" annotation.
This test only asks whether both the old and new versions of the object have any annotation with that name.

config/policies/validating-admission-policy-immutable-fields.yaml
```diff
@@ -0,0 +1,38 @@
...
+              ((has(oldObject.metadata.annotations) && ('dual-pods.llm-d.ai/admin-port' in oldObject.metadata.annotations)) ? oldObject.metadata.annotations['dual-pods.llm-d.ai/admin-port'] : '') != ((has(object.metadata.annotations) && ('dual-pods.llm-d.ai/admin-port' in object.metadata.annotations)) ? object.metadata.annotations['dual-pods.llm-d.ai/admin-port'] : '')
+            ) || (
+              ((has(oldObject.metadata.labels) && ('dual-pods.llm-d.ai/dual' in oldObject.metadata.labels)) ? oldObject.metadata.labels['dual-pods.llm-d.ai/dual'] : '') != ((has(object.metadata.labels) && ('dual-pods.llm-d.ai/dual' in object.metadata.labels)) ? object.metadata.labels['dual-pods.llm-d.ai/dual'] : '') ||
+              ((has(oldObject.metadata.labels) && ('dual-pods.llm-d.ai/sleeping' in oldObject.metadata.labels)) ? oldObject.metadata.labels['dual-pods.llm-d.ai/sleeping'] : '') != ((has(object.metadata.labels) && ('dual-pods.llm-d.ai/sleeping' in object.metadata.labels)) ? object.metadata.labels['dual-pods.llm-d.ai/sleeping'] : '')
```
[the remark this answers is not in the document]
MikeSpreitzer: Oh, wait. I now see that the other policy is relying on the `dual` annotation on the server-requesting Pod to convey the binding state. In that case it is OK to prevent others from changing it.

Even so, the "sleeping" label is still just FYI and the controllers do not need it protected.

config/policies/validating-admission-policy-bound-serverReqPod.yaml
```diff
@@ -0,0 +1,37 @@
...
+    #
+    # Logic:
+    # 1) If the old object is a server requesting pod (`inference-server-config`),
+    # 2) AND the `dual` label is present and non-empty (bound state),
+    # 3) AND any of the immutable fields change
+    # 4) THEN deny unless the request originates from an FMA controller SA.
+    - expression: |
+        (
+          !request.userInfo.username.matches("system:serviceaccount:[^:]+:(launcher-populator|dual-pods-controller)$") && (
+            ((has(oldObject.metadata.annotations) && ('dual-pods.llm-d.ai/inference-server-config' in oldObject.metadata.annotations) ? oldObject.metadata.annotations['dual-pods.llm-d.ai/inference-server-config'] : "") != "") &&
```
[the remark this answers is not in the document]
MikeSpreitzer: I think it is OK to omit concerns that only apply to milestone 2.

MikeSpreitzer:
I am calling out things that could be removed from the CEL expressions because CEL evaluation is expensive in the apiservers and there are limits on CEL expression complexity.

config/policies/validating-admission-policy-immutable-fields.yaml
```diff
@@ -0,0 +1,38 @@
...
+              ((has(oldObject.metadata.annotations) && ('dual-pods.llm-d.ai/status' in oldObject.metadata.annotations)) ? oldObject.metadata.annotations['dual-pods.llm-d.ai/status'] : '') != ((has(object.metadata.annotations) && ('dual-pods.llm-d.ai/status' in object.metadata.annotations)) ? object.metadata.annotations['dual-pods.llm-d.ai/status'] : '') ||
+              ((has(oldObject.metadata.annotations) && ('dual-pods.llm-d.ai/accelerators' in oldObject.metadata.annotations)) ? oldObject.metadata.annotations['dual-pods.llm-d.ai/accelerators'] : '') != ((has(object.metadata.annotations) && ('dual-pods.llm-d.ai/accelerators' in object.metadata.annotations)) ? object.metadata.annotations['dual-pods.llm-d.ai/accelerators'] : '') ||
+              ((has(oldObject.metadata.annotations) && ('dual-pods.llm-d.ai/launcher-based' in oldObject.metadata.annotations)) ? oldObject.metadata.annotations['dual-pods.llm-d.ai/launcher-based'] : '') != ((has(object.metadata.annotations) && ('dual-pods.llm-d.ai/launcher-based' in object.metadata.annotations)) ? object.metadata.annotations['dual-pods.llm-d.ai/launcher-based'] : '') ||
```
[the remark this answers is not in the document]
aavarghese: Sure. Removing `dual-pods.llm-d.ai/launcher-based`.

test/e2e/validate.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# Test script for ValidatingAdmissionPolicy CEL rules
+POD_NAME=${POD_NAME:-my-request-test}
+BOUND_POD_NAME=${BOUND_POD_NAME:-my-bound-request-test}
+
+if kubectl api-resources --api-group=admissionregistration.k8s.io -o name | grep -q 'validatingadmissionpolicies'; then
+  echo "Cluster supports ValidatingAdmissionPolicy (CEL)"
+  if command -v helm >/dev/null 2>&1 && [ -d "charts/dpctlr" ]; then
+    helm template dpctlr charts/dpctlr --set policies.enabled=true | kubectl apply -f -
+    sleep 5 # wait for policies to be registered
```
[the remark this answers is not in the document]
aavarghese: There is no condition=Ready for validatinadmissionpolicy

test/e2e/validate.sh
```diff
@@ -0,0 +1,102 @@
...
+if kubectl api-resources --api-group=admissionregistration.k8s.io -o name | grep -q 'validatingadmissionpolicies'; then
+  echo "Cluster supports ValidatingAdmissionPolicy (CEL)"
+  if command -v helm >/dev/null 2>&1 && [ -d "charts/dpctlr" ]; then
+    helm template dpctlr charts/dpctlr --set policies.enabled=true | kubectl apply -f -
+    # wait for policies to be registered
+    sleep 30
+  else
+    echo "Helm not available or 'charts/dpctlr' missing; cannot install policies automatically."
+    exit 1
+  fi
```
MikeSpreitzer: The regular installation of FMA at the start of the test suite should be entirely responsible for creating the policy objects if possible.

docs/e2e-recipe.md
```diff
@@ -102,6 +102,20 @@ Following are some things to keep in mind about these definitions.
   fully utilize gpu memory. Current kv cache memory in use is
   32786058444 bytes".
 
+### Admission policies (CEL)
+
+Kubernetes `ValidatingAdmissionPolicy` (CEL) resources are used to
+protect controller-managed annotations and labels that are
+managed by the controllers (dual-pods controller and
+launcher-populator). The policy manifests and their
+bindings are managed as Helm chart templates under `charts/dpctlr/templates/policies/` and are installed by the chart when `policies.enabled` is `true`.
```
MikeSpreitzer: Say that the user is responsible for setting that false if the cluster does not support these kinds of objects.

test/e2e/validate.sh
```diff
@@ -0,0 +1,102 @@
...
+
+if kubectl api-resources --api-group=admissionregistration.k8s.io -o name | grep -q 'validatingadmissionpolicies'; then
+  echo "Cluster supports ValidatingAdmissionPolicy (CEL)"
+  if command -v helm >/dev/null 2>&1 && [ -d "charts/dpctlr" ]; then
+    helm template dpctlr charts/dpctlr --set policies.enabled=true | kubectl apply -f -
+    # wait for policies to be registered
+    sleep 30
+  else
+    echo "Helm not available or 'charts/dpctlr' missing; cannot install policies automatically."
+    exit 1
```
MikeSpreitzer: This is not a test failure, it is just a recognition that the test is inapplicable to this cluster.

test/e2e/validate.sh
```diff
@@ -0,0 +1,102 @@
...
+trap 'rc=$?; cleanup; exit $rc' EXIT
+
+# Test script for ValidatingAdmissionPolicy CEL rules
+POD_NAME=${POD_NAME:-my-request-test}
+BOUND_POD_NAME=${BOUND_POD_NAME:-my-bound-request-test}
+
+if kubectl api-resources --api-group=admissionregistration.k8s.io -o name | grep -q 'validatingadmissionpolicies'; then
+  echo "Cluster supports ValidatingAdmissionPolicy (CEL)"
+  if command -v helm >/dev/null 2>&1 && [ -d "charts/dpctlr" ]; then
+    helm template dpctlr charts/dpctlr --set policies.enabled=true | kubectl apply -f -
```
MikeSpreitzer: Installation is the responsibility of run.sh

test/e2e/validate.sh
aavarghese: Related to [this](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/220#discussion_r2714099166) comment. Will change

docs/e2e-recipe.md
aavarghese: Sure.

PR #223, opened 2026-01-22
Initial integration between the dual-pods controller and the launcher

waltforme:
This PR introduces serveral changes to initialize the integration between the dual-pods controller and the launcher, listed as follows.
- Extend the dual-pods controller's `itemOnNode` interface to allow the codebase to smoothly transform from direct providers to launcher-based providers.
- Outline the basic idea of syncing logic for launcher-based requesters:
   ```text
	// TODO(waltforme): complete the following logic:
	// - if no instances are fulfilling the requester, create an instance to do so;
	// - if an existing instance is already fulfilling the requester, no-op;
	// - if some instances are fulfilling obsolete/non-existing requesters, delete the instances.
	// In other words, empty the symmetric difference between the set of instances and the set of requesters.
   ```
    and begin to implement the logic.
- Let each launcher pod to have an annotation which remembers the hash of the corresponding LauncherConfig object.
- Modifies interal data structure to include the UUIDs of the vLLM instances that assigned by the launcher.
- Formally write down a contract between the launcher implementation to use launcher pod's port number `8001`.

### DIFF
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -306,7 +306,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	if providingPod != nil {
 		var serverPort int16
 		if launcherBased {
-			// TODO(waltforme): currently the port number of an instance is written in the vLLMConfig. Need to parse it out.
+			// TODO(waltforme): use port number in the updated InferenceServerConfig spec
 			serverPort = 8005
 		} else {
 			_, serverPort, err = utils.GetInferenceServerPort(providingPod, false)
@@ -449,19 +449,20 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		return fmt.Errorf("failed to get LauncherConfig %q: %w", lcName, err), false
 	}
 
-	lcTemplateHash, err := ctl.parseLauncherConfig(lc)
+	desiredLauncherPod, err := utils.BuildLauncherPodFromTemplate(lc.Spec.PodTemplate, ctl.namespace, requestingPod.Spec.NodeName, lcName)
 	if err != nil {
-		return fmt.Errorf("parse LauncherConfig %q: %w", lcName, err), true
+		return fmt.Errorf("failed to build launcher Pod from LauncherConfig %q: %w", lcName, err), true
 	}
-	logger.V(5).Info("LauncherConfig's PodTemplate hash", "hash", lcTemplateHash)
-	launcherPodAnys, err := ctl.podInformer.GetIndexer().ByIndex(launcherConfigHashIndexName, lcTemplateHash)
+	lcHash := desiredLauncherPod.Annotations[ctlrcommon.LauncherConfigHashAnnotationKey]
+	logger.V(5).Info("LauncherConfig's hash", "hash", lcHash)
+	launcherPodAnys, err := ctl.podInformer.GetIndexer().ByIndex(launcherConfigHashIndexName, lcHash)
 	if err != nil {
 		return err, false
 	}
 
 	if len(launcherPodAnys) > 0 {
 		// Multiple launcher Pods could exist for one LauncherConfig object on one node.
-		// TODO(waltforme): The controller currently picks the first Pod but should consider all of them, this should be revised.
+		// TODO(waltforme): The logic currently picks the first Pod but should consider all of them
 
 		// Note: Multiple vLLM instances could exist in one launcher Pod, but at most one instance
 		// could be awake at a time.
@@ -507,7 +508,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return fmt.Errorf("wake up vLLM instance: %w", err), true
 			}
 			launcherDat.Instances[iscHash] = time.Now()
-			// TODO(waltforme): the bind method needs to be revised to handle launcher-based server providing Pods
+			// TODO(waltforme): the bind method may need more revision to fully handle launcher-based server providing Pods
 			return ctl.bind(ctx, serverDat, requestingPod, launcherPod, true)
 		} else if insts.TotalInstances == 0 {
 			result, err := lClient.CreateNamedInstance(ctx, iscHash, *cfg)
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
>  // - if some instances are fulfilling obsolete/non-existing requesters, delete the instances.

That sounds like something that could violate the launcher population semantics. Again, they are at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/1d6ca5e7240b71accdb95302d27d73b884f514c7/api/fma/v1alpha1/launcherpopulationpolicy_types.go#L38-L40

pkg/api/interface.go
```diff
@@ -52,6 +52,16 @@ const ServerPatchAnnotationName = "dual-pods.llm-d.ai/server-patch"
...
 
+// LauncherConfigHashAnnotationName is the name of an annotation on the
+// launcher-based server-providing Pod. The value of the annotation is the hash of the
+// LauncherConfig object's PodTemplate that the server-providing Pod uses.
+const LauncherConfigHashAnnotationName = "dual-pods.llm-d.ai/launcher-config-hash"
+
+// LauncherServicePort is the port number on which the launcher exposes its HTTP service
+// for the management of vLLM instances.
+// This is a contract between the dual-pods controller and the launcher implementation.
+const LauncherServicePort = 8001
```
MikeSpreitzer: The launcher population controller also works with this constant. It should not be declared in two places.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -401,6 +409,168 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+	lc, err := ctl.lcLister.LauncherConfigs(ctl.namespace).Get(lcName)
+	if err != nil {
+		// TODO(waltforme): report error in the status annotation
+		// TODO(waltforme): introduce the 'enqueue requesters by launcherconfigs' logic to the controller
+		// It is safe not to retry here because once an event from LauncherConfig occurs, another processing is triggered
+		return fmt.Errorf("failed to get LauncherConfig %q: %w", lcName, err), false
+	}
+
+	// find which launcher Pod is using this launcherConfig, then find its IP
+	lcTemplateHash, err := ctl.parseLauncherConfig(lc)
```
MikeSpreitzer: The Node should contribute to this hash. No point in considering launcher Pods on the wrong Node.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -401,6 +409,168 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+	if err != nil {
+		return err, false
+	}
+	if len(launcherPodAnys) == 0 {
+		// TODO(waltforme): report error in the status annotation
+		// TODO(waltforme): introduce the 'enqueue requesters by launcher Pods' logic to the controller
+		// It will be safe not to retry here because once the launcher Pod exists, another processing is triggered
+		return fmt.Errorf("no launcher Pod found for LauncherConfig %q with PodTemplate hash %q", lcName, lcTemplateHash), false
+	}
+	// Should multiple launcher Pods exist for the same LauncherConfig on one node? The answer is no.
```
MikeSpreitzer: The answer is "yes".
Consider this use case. A node with 8 GPUs. 8 launcher pods, each with one awake child, each using a different GPU.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -401,6 +409,168 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+	// TODO(waltforme): Should we report error if multiple launcher Pods are found? Should we delete the extra ones?
+	launcherPod := launcherPodAnys[0].(*corev1.Pod)
```
MikeSpreitzer: This is the point at which to enforce https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/1d6ca5e7240b71accdb95302d27d73b884f514c7/api/fma/v1alpha1/launcherconfig_types.go#L32

The commentary on that is deficient. But as in milestone 2, the time to enforce is when making a new vllm instance. The reasoning is this: the limit is chosen so that many sleeping instances plus one awake instance fits in memory; thus, that many plus 1 sleeping instances plus zero awake instances will also fit in memory --- and let's not prune sooner than we have to.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -401,6 +409,168 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+		}
+	}
+	launcherDat := nodeDat.Launchers[lcName]
+	for hash, inst := range launcherDat.Instances {
+		if hash == iscHash {
+			InstExists = true
+			inst.LastUsed = time.Now()
+			break
+		}
+	}
```
MikeSpreitzer:
```suggestion
	id, InstExists := launcherDat.Instances[iscHash]
	if InstExists { inst.LastUsed = time.Now() }
```

pkg/controller/dual-pods/inference-server.go
```diff
@@ -401,6 +409,168 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+	// - if some instances are fulfilling obsolete/non-existing requesters, delete the instances.
+	// In other words, empty the symmetric difference between the set of instances and the set of requesters.
+
+	// First, ensure an instance exists for the inferenceserverconfig object.
+	cfg, iscHash, err := ctl.parseInferenceServerConfig(isc)
+	if err != nil {
+		return fmt.Errorf("parse inference server config: %w", err), true
+	}
+	logger.V(5).Info("Nominal hash of InferenceServerConfig", "hash", iscHash)
+	InstExists := false
```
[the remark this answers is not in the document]
waltforme: Corrected.

pkg/api/interface.go
```diff
@@ -52,6 +52,16 @@ const ServerPatchAnnotationName = "dual-pods.llm-d.ai/server-patch"
 // This annotation is mutually exclusive with the 'ServerPatchAnnotationName' annotation.
 const InferenceServerConfigAnnotationName = "dual-pods.llm-d.ai/inference-server-config"
 
+// LauncherConfigHashAnnotationName is the name of an annotation on the
+// launcher-based server-providing Pod. The value of the annotation is the hash of the
+// LauncherConfig object's PodTemplate that the server-providing Pod uses.
+const LauncherConfigHashAnnotationName = "dual-pods.llm-d.ai/launcher-config-hash"
```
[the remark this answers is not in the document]
waltforme: Fixed by #228.

pkg/api/interface.go
waltforme: Moved to pkg/controller/common, but I somehow I don't see the population controller use it so far.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -401,6 +409,168 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+		return err, false
+	}
+	if len(launcherPodAnys) == 0 {
+		// TODO(waltforme): report error in the status annotation
+		// TODO(waltforme): introduce the 'enqueue requesters by launcher Pods' logic to the controller
+		// It will be safe not to retry here because once the launcher Pod exists, another processing is triggered
+		return fmt.Errorf("no launcher Pod found for LauncherConfig %q with PodTemplate hash %q", lcName, lcTemplateHash), false
+	}
+	// Should multiple launcher Pods exist for the same LauncherConfig on one node? The answer is no.
+	// TODO(waltforme): Should we report error if multiple launcher Pods are found? Should we delete the extra ones?
```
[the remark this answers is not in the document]
waltforme: Agreed. Removed this question.

pkg/controller/dual-pods/inference-server.go
waltforme: Fixed.

Issue #230, opened 2026-01-29
Charts and images and testing, oh my


manoelmarques:
In order to proceed work on FMA benchmarks in the `llm-d-benchmark` repository, there is a need for publishing the Dual Pod Controller Chart so it can be accessed there. The benchmark repository only accepts published charts.

Those are two links of chart publishing code that could be looked at as suggestion:

https://github.com/llm-d-incubation/workload-variant-autoscaler/blob/main/.github/workflows/helm-release.yaml
https://github.com/llm-d-incubation/llm-d-modelservice/blob/main/.github/workflows/helm-release.yaml

MikeSpreitzer:
It seems to me that for every use case that we have now, it would be better to have one Helm chart that deploys both controllers (rather than have two single-controller Helm chart).

MikeSpreitzer:
Have a look at the KubeStellar release process, documented at https://main--kubestellar-docs.netlify.app/docs/contributing/release-process and mostly implemented by the workflow at https://github.com/kubestellar/kubestellar/blob/main/.github/workflows/goreleaser.yml . Following are few things that I like about the KubeStellar workflow.

- `helm package` can set the chart's version and appVersion, you do not need hacky text editing to set those.
- Using goreleaser makes it substantially easier.
- There is no need to commit any changes to the files in the repo. There is no need for the files in the repo to contain the latest release string. That is just building in self-references. I hate self-references. KubeStellar needs a few, and each is a pain in the a$$. Luckily, FMA does not need any. Just don't do it at all.

PR #236, opened 2026-02-04
Github actions workflow to publish charts

aavarghese:
Following discussions and work in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/224 and https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/230. 

Implements automated release process that builds and publishes 4 container images and 2 Helm charts to GHCR as OCI artifacts...


.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,141 @@
+name: Publish Helm Charts
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - charts/**
```
MikeSpreitzer: I do not see why we would want this restriction. It means that this workflow is restricted to pushes that only modify chart sources. But the built artifacts depend on a lot more: all the code.

MikeSpreitzer:
The current version of this PR adds a workflow that publishes two Helm charts but DOES NOT build the container images. In comparison, the release processes for WVA and KubeStellar also build and publish the container images and make the published Helm chart default to using the published container images. This is the way.

aavarghese:
> In comparison, https://github.com/llm-d-incubation/workload-variant-autoscaler/blob/main/.github/workflows/helm-release.yaml is a release process that runs when a release is tagged or when specifically requested (which is an exceptional thing, I think) by creating a specially named tag or manually invoking the workflow. That seems like a reasonable practice to me. I do not see why we would want to publish release artifacts every time there is a merge to `main`.

I think this contradicts with what you said in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/236#discussion_r2766635372 :) 

I was also trying to keep in mind the scenarios listed in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/230#issuecomment-3837108661

Either way, we wont be publishing release artifacts every time there is a merge to main. It is ONLY when a new release is published OR the `version` field or something else under `charts/` has been updated. Not every merge to main will create a new artifact. That is too much. So we should keep the `path`.

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,221 @@
+name: Release – Build Images & Publish Helm Charts to GHCR
```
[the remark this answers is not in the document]
aavarghese: Will change to `publish-release.yml`. Thanks!

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,221 @@
...
+      # -----------------------------------------
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # -----------------------------------------
+      # 5. Build and push controller image
```
MikeSpreitzer: Terminology: there are two "controllers", the dual-pods controller and the launcher population controller.

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,221 @@
...
+      # -----------------------------------------
+      - name: Build and push controller image
+        run: |
+          TAG="${{ steps.version.outputs.tag }}"
```
MikeSpreitzer: Why is it better to not have the initial "v" in the image tag? IMO this is an inferior choice.

docs/helm-install.md
```diff
@@ -0,0 +1,57 @@
...
+Or with a values file:
+
+```bash
+helm install dpctlr \
+  oci://ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/charts/dual-pods-controller \
+  --version 0.3.0 \
+  -f custom-values.yaml
+```
+
+## Release Process
```
MikeSpreitzer: The release process definitely should be documented. But not at the end of the file explaining how to use the Helm chart(s). The subject deserves a document of its own. And a complete description, including: applying the Git tag, creating the GitHub release, and what the workflow does.

.github/workflows/helm-release.yaml
```diff
@@ -0,0 +1,220 @@
...
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag (e.g., v0.3.1) - for testing without creating a release'
+        required: false
+        type: string
+  push:
+    tags:
+      - 'v*'
+
```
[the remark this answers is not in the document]
aavarghese: ha - I missed that one sentence in the PR desc. Thanks.

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,215 @@
...
+      # -----------------------------------------
+      # 9. Install Helm and yq
+      # -----------------------------------------
+      - name: Install Helm
+        uses: azure/setup-helm@fe7b79cd5ee1e45176fcad797de68ecaf3ca4814 # v4.2.0
+        with:
+          version: 'v3.16.3'
+
+      - name: Install yq
+        uses: mikefarah/yq@bbdd97482f2d439126582a59689eb1c855944955 # v4.44.3
```
MikeSpreitzer: `yq` is already in the workflow runner. See https://github.com/actions/runner-images/blob/ubuntu24/20260201.15/images/ubuntu/Ubuntu2404-Readme.md

docs/release-process.md
```diff
@@ -0,0 +1,90 @@
...
+4. **Publishes charts to GHCR**:
+   - `oci://ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/charts/dual-pods-controller`
+   - `oci://ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/charts/launcher-populator`
+
+### 4. Alternative Trigger Methods
+
+The workflow can also be triggered by:
+
+- **Tag push**: Simply pushing a tag (e.g., `git push origin v0.3.1`) without creating a GitHub release
+- **Manual dispatch**: Using the GitHub Actions UI to manually trigger the workflow with a specific tag (useful for testing)
```
MikeSpreitzer: Since the workflow checks out the tag, the tag has to exist. So testing is not separate from tag push. Tag push is part of the release process, for both testing releases and regular releases.

rubambiza (changes requested):
Minor requested changes/explanations. Otherwise, LGTM and no need for testing.

docs/helm-install.md
```diff
@@ -0,0 +1,48 @@
+# Helm Chart Installation
+
+This guide covers installing Fast Model Actuation (FMA) Helm charts.
+
+## Available Charts
+
+- **dual-pods-controller**: Main FMA dual pod controller
+- **launcher-populator**: Launcher population controller
+
+## Installation
```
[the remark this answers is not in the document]
aavarghese: Ended up deleting the file because that kind of detail may not be needed for regular developers and isn't part of the other llm-d repos afaik...

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,209 @@
...
+
+      # -----------------------------------------
+      # 5. Build and push dual-pods controller image
+      # -----------------------------------------
+      - name: Build and push dual-pods controller image
+        run: |
+          TAG="${{ steps.version.outputs.tag }}"
+          reg="${{ github.repository }}"
+
+          KO_DOCKER_REPO=ghcr.io/${reg@L} \
```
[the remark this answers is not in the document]
aavarghese: GREAT CATCH!

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,209 @@
...
+
+      # -----------------------------------------
+      # 6. Build and push launcher population controller image
+      # -----------------------------------------
+      - name: Build and push launcher population controller image
+        run: |
+          TAG="${{ steps.version.outputs.tag }}"
+          reg="${{ github.repository }}"
+
+          KO_DOCKER_REPO=ghcr.io/${reg@L} \
```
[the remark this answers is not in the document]
aavarghese: GREAT CATCH!

PR #288, opened 2026-02-26
ci: remove dual-pods finalizers before namespace deletion

### LINKED ISSUE #289: E2E test hangs itself on finalizer
The workflow in `.github/workflows/ci-e2e-openshift.yaml` deletes the dual-pods controller while some relevant Pods exist and then tries to delete the namespace that the test ran in. Some Pods never get deleted because they have a finalizer that the dual-pods controller would normally remove. The namespace never gets deleted because the Pods never get deleted.

MikeSpreitzer:
Before deleting FMA_NAMESPACE, strip any finalizers prefixed with "dual-pods.llm-d.ai/" from all Pods in that namespace. Pods with no matching finalizers are skipped; non-matching finalizers on other pods are preserved. This prevents the namespace from getting stuck in Terminating when the dual-pods controller has set those finalizers and is no longer running to remove them.

The same block is applied in both cleanup sites:
- "Clean up resources for this PR" (pre-run cleanup of leftover state)
- "Cleanup infrastructure" (post-run teardown on success/cancellation)

Resolves #289

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -375,6 +375,21 @@ jobs:
               echo "  Uninstalling helm release: $release"
               helm uninstall "$release" -n "$FMA_NAMESPACE" --ignore-not-found --wait --timeout 60s || true
             done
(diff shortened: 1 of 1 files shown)


aavarghese (approved):
Looks good and is necessary in both places!

PR #313, opened 2026-03-02
Adjust Node viewing ClusterRole in E2E-on-OCP workflow

### LINKED ISSUE #311: Agree and document cluster-sharing approach for Node-reading ClusterRole

MikeSpreitzer:
Modify the GHA workflow that does E2E test in our shared OpenShift cluster so that the ClusterRole for reading Nodes includes the run ID, making it effectively unique.

Resolves #311


waltforme (approved):
lgtm

Issue #337, opened 2026-03-10
[Bug]: Spurious and erroneous special case state dumping


MikeSpreitzer:
### What happened?

In some cases where `.github/workflows/ci-e2e-openshift.yaml` detects an error, it does some special case state dumping.

(A) this is spurious, it should be covered by the unconditional state dumping job steps that come later.

(B) Some of it is erroneous, using the wrong pattern for Pod names. E.g., https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/ea3ba208a5179d22678389f51a2921deed4d8097/.github/workflows/ci-e2e-openshift.yaml#L855 ; see an example failure of that command at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/22889527593/job/66409668351#step:20:113

### Version

main (please specify commit below)

### Commit SHA

ea3ba208a5179d22678389f51a2921deed4d8097

### Relevant log output

```shell
=== Controller logs ===
error: error from server (NotFound): deployments.apps "fma-e2e-22889527593" not found in namespace "fma-e2e-pr-22889527593"
```

MikeSpreitzer:
This has been resolved.

PR #348, opened 2026-03-11
ci: Enable launcher-populator in OpenShift E2E and local tests

aavarghese:
This PR enables the launcher-populator controller in both the OpenShift CI E2E tests and the local run-launcher-based.sh test script.

## Test plan

- [x] CI E2E tests pass on OpenShift with launcher-populator enabled
- [x] Local run-launcher-based.sh passes with launcher-populator

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -760,8 +760,6 @@ jobs:
             fi
             if [ "$ELAPSED" -ge "$LIMIT" ]; then
               echo "::error::Requester pod did not appear within ${LIMIT}s"
-              kubectl get pods -n "$FMA_NAMESPACE" -o wide
-              kubectl get events -n "$FMA_NAMESPACE" --sort-by='.lastTimestamp' | tail -20
               exit 1
             fi
             sleep 5
@@ -786,8 +784,6 @@ jobs:
             fi
             if [ "$ELAPSED" -ge "$LIMIT" ]; then
               echo "::error::Launcher-populator did not create expected $GPU_NODES launcher(s) within ${LIMIT}s (found: $COUNT)"
-              kubectl get pods -n "$FMA_NAMESPACE" -o wide
-              kubectl get events -n "$FMA_NAMESPACE" --sort-by='.lastTimestamp' | tail -20
               exit 1
             fi
             sleep 5
@@ -819,7 +815,6 @@ jobs:
             fi
             if [ "$ELAPSED" -ge "$LIMIT" ]; then
               echo "::error::Launcher-to-requester binding not established within ${LIMIT}s"
-              kubectl get pods -n "$FMA_NAMESPACE" -o wide --show-labels
               exit 1
             fi
             sleep 5
@@ -838,7 +833,6 @@ jobs:
             if [ "$ELAPSED" -ge "$LIMIT" ]; then
               echo "::error::Requester-to-launcher binding not established within ${LIMIT}s"
               echo "  Requester dual label: '$REQUESTER_DUAL' (expected: '$LAUNCHER')"
-              kubectl get pods -n "$FMA_NAMESPACE" -o wide --show-labels
               exit 1
             fi
             sleep 5
@@ -881,7 +875,7 @@ jobs:
 
       - name: Dump vLLM instance logs from launchers
         if: always()
-        run: ./scripts/dump-launcher-vllm-logs.sh "$FMA_NAMESPACE"
+        run: scripts/dump-launcher-vllm-logs.sh "$FMA_NAMESPACE"
 
       - name: Clean up test objects
         if: always()
--- a/scripts/dump-launcher-vllm-logs.sh
+++ b/scripts/dump-launcher-vllm-logs.sh
@@ -2,15 +2,18 @@
 # Dump vLLM instance logs from all launcher pods.
 #
 # Usage: dump-launcher-vllm-logs.sh [namespace]
-#   namespace: Kubernetes namespace (default: default)
+#   namespace: Kubernetes namespace (defaults to kubectl current context)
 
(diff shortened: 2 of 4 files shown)


MikeSpreitzer:
You will also want to rebase this onto `main` to pick up recent improvements there. PR #343, currently in review, also makes related changes.

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -782,17 +805,19 @@ jobs:
...
               echo "Launcher bound to requester: $LAUNCHER -> $REQUESTER"
               break
             fi
             if [ "$ELAPSED" -ge "$LIMIT" ]; then
               echo "::error::Launcher-to-requester binding not established within ${LIMIT}s"
-              echo "  Launcher dual label: '$LAUNCHER_DUAL' (expected: '$REQUESTER')"
               kubectl get pods -n "$FMA_NAMESPACE" -o wide --show-labels
               echo "=== Controller logs ==="
-              kubectl logs deployment/"$FMA_RELEASE_NAME" -n "$FMA_NAMESPACE" --tail=100 || true
+              kubectl logs deployment/"$FMA_RELEASE_NAME-dual-pods-controller" -n "$FMA_NAMESPACE" --tail=100 || true
```
MikeSpreitzer: See #337 . Rather than fixing this command, I think that it should be deleted. There is a later and unconditional state dumping job step that will display this log.

scripts/dump-launcher-vllm-logs.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+# Dump vLLM instance logs from all launcher pods.
+#
+# Usage: dump-launcher-vllm-logs.sh [namespace]
+#   namespace: Kubernetes namespace (default: default)
```
MikeSpreitzer: It would be better if the default were the same as for `kubectl`.
Not a critical problem.

test/e2e/run-launcher-based.sh
```diff
@@ -195,28 +207,37 @@ isc=$(echo $objs | awk '{print $1}')
...
 rslb=$(echo $objs | awk '{print $3}')
 isc2=$(echo $objs | awk '{print $4}')
+isc3=$(echo $objs | awk '{print $5}')
+lpp=$(echo $objs | awk '{print $6}')
 instlb=${rslb#my-request-}
 
+# LauncherPopulationPolicy specifies launcherCount per node with nvidia.com/gpu.present=true
+GPU_NODES=$(kubectl get nodes -l nvidia.com/gpu.present=true --field-selector spec.unschedulable!=true -o name | wc -l | tr -d ' ')
+echo "Expecting launcher-populator to create $GPU_NODES launcher(s) (one per schedulable GPU node)"
+expect "[ \$(kubectl get pods -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | grep -c '^pod/') -ge $GPU_NODES ]"
```
MikeSpreitzer: `grep -c '^pod/'` is a bad way to count lines, it crashes (exits with non-zero code) when there are none. Better to use `wc -l`.

The context matters. In this case, the crash will lead to `[ -ge 1 ]` being executed, which is bad. When `grep -c` is used in a context where the non-zero exit gets the right job done, then it is not a bug.

test/e2e/run-launcher-based.sh
```diff
@@ -251,7 +272,8 @@ expect "kubectl get pods -o name -l app=dp-example,instance=$instlb | grep -c '^
 reqlb2=$(kubectl get pods -o name -l app=dp-example,instance=$instlb | sed s%pod/%%)
 
 # Should still be using the same launcher pod
-launcherlb2=$(kubectl get pods -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | sed s%pod/%%)
+expect "kubectl get pods -o name -l dual-pods.llm-d.ai/dual=$reqlb2 | grep -c '^pod/' | grep -w 1"
```
MikeSpreitzer: `kubectl get pods -o name` guarantees that every output line will start with "pod/". There is no header output. `wc -l` is a better way to count lines, no complexity with different behavior in the zero line case.

scripts/dump-launcher-vllm-logs.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo ""
+  echo "=== vLLM instances status ==="
+  INSTANCES_JSON=$(curl -s "http://localhost:18001/v2/vllm/instances" 2>/dev/null || true)
+  echo "$INSTANCES_JSON" | jq . 2>/dev/null || echo "$INSTANCES_JSON"
+
+  # Get instance IDs
+  INSTANCE_IDS=$(echo "$INSTANCES_JSON" | jq -r '.instances[].instance_id // empty' 2>/dev/null || true)
+
+  if [ -z "$INSTANCE_IDS" ]; then
+    echo "No vLLM instances found on this launcher"
```
rubambiza: Small nit/suggestion:

```suggestion
    echo "No vLLM instances found on launcher: $LAUNCHER_POD"
```

Makes for easier tracing.

test/e2e/run-launcher-based.sh
```diff
@@ -195,28 +207,37 @@ isc=$(echo $objs | awk '{print $1}')
 lc=$(echo $objs | awk '{print $2}')
 rslb=$(echo $objs | awk '{print $3}')
 isc2=$(echo $objs | awk '{print $4}')
+isc3=$(echo $objs | awk '{print $5}')
```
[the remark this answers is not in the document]
MikeSpreitzer: Copilot keeps pointing out that it is unused. I assume somebody put it there anticipating some future usage.

MikeSpreitzer (approved):
IMO this is good enough to merge. Small improvements could be done in this PR or later, and are not critical anyway.

PR #362, opened 2026-03-17
Sync unbound launcher-based server-providing pods

waltforme:
This PR takes unbound launcher pods into the dual-pods controller’s node-local reconciliation flow so launcher readiness, deletion, and restart recovery update internal instance state and re-drive affected requesters.

This PR also extends launcher-based e2e coverage.

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -343,10 +343,10 @@ const (
 	infSvrItemRequester infSvrItemType = "requester"
 	// infSvrItemBoundProvider is for a server-providing Pod that
 	// is bound to a server-requesting Pod.
-	infSvrItemBoundProvider infSvrItemType = "bound_direct_provider"
+	infSvrItemBoundProvider infSvrItemType = "bound_provider"
 	// infSvrItemUnboundLauncherBasedProvider is for a server-providing Pod that
 	// is launcher-based and not bound to any server-requesting Pods.
-	infSvrItemUnboundLauncherBasedProvider infSvrItemType = "launcher_based_provider"
+	infSvrItemUnboundLauncherBasedProvider infSvrItemType = "unbound_launcher_based_provider"
 	// infSvrItemDontCare is not a real infSvrItemType but only a placeholder
 	// saying the corresponding infSvrItem is not relevant to the controller.
 	infSvrItemDontCare infSvrItemType = "dont_care"
@@ -367,12 +367,9 @@ func careAbout(pod *corev1.Pod) (item infSvrItem, it infSvrItemType) {
 	if len(requesterParts) == 2 {
 		return infSvrItem{apitypes.UID(requesterParts[0]), requesterParts[1]}, infSvrItemBoundProvider
 	}
-	// Check for unbound launcher-based server-providing Pod.
-	if pod.Labels != nil {
-		if _, hasLauncherLabel := pod.Labels[ctlrcommon.LauncherConfigNameLabelKey]; hasLauncherLabel {
-			// For an unbound launcher-based server-providing Pod, use the Pod's own UID and name
-			return infSvrItem{pod.UID, pod.Name}, infSvrItemUnboundLauncherBasedProvider
-		}
+	if _, hasLauncherLabel := pod.Labels[ctlrcommon.LauncherConfigNameLabelKey]; hasLauncherLabel {
+		// For an unbound launcher-based server-providing Pod, use the Pod's own UID and name
+		return infSvrItem{pod.UID, pod.Name}, infSvrItemUnboundLauncherBasedProvider
 	}
 	return infSvrItem{}, infSvrItemDontCare
 }
@@ -693,13 +690,13 @@ func (ctl *controller) enqueueRequestersByInferenceServerConfig(isc *fmav1alpha1
 	}
 }
 
-func (ctl *controller) enqueueUnboundInfSvrItemsOnNode(ctx context.Context, nodeName string, why string) {
+func (ctl *controller) enqueueUnboundInfSvrItemsOnNode(ctx context.Context, nodeName string, whyEnqueue string) {
 	logger := klog.FromContext(ctx)
 	nd := ctl.getNodeData(nodeName)
 	itemCount := 0
 	podObjs, err := ctl.podInformer.GetIndexer().ByIndex(nodeNameIndexName, nodeName)
 	if err != nil {
-		logger.Error(err, "Failed to list Pods by nodeName index", "nodeName", nodeName, "why", why)
+		logger.Error(err, "Failed to list Pods by nodeName index", "nodeName", nodeName, "whyEnqueue", whyEnqueue)
 		return
 	}
 	for _, podObj := range podObjs {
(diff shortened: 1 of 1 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -346,18 +351,26 @@ const (
...
 func careAbout(pod *corev1.Pod) (item infSvrItem, it infSvrItemType) {
 	if len(pod.Annotations[api.ServerPatchAnnotationName]) > 0 || len(pod.Annotations[api.InferenceServerConfigAnnotationName]) > 0 {
 		return infSvrItem{pod.UID, pod.Name}, infSvrItemRequester
 	}
 	requesterStr := pod.Annotations[requesterAnnotationKey]
 	requesterParts := strings.Split(requesterStr, " ")
-	if len(requesterParts) != 2 {
-		return infSvrItem{}, infSvrItemDontCare
+	if len(requesterParts) == 2 {
+		return infSvrItem{apitypes.UID(requesterParts[0]), requesterParts[1]}, infSvrItemBoundDirectProvider
```
MikeSpreitzer: Here `infSvrItemBoundDirectProvider` does not exactly match up with `len(requesterParts) == 2`, because the latter will be true of a bound launcher Pod too.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -588,22 +629,29 @@ func (ctl *controller) selectBestLauncherPod(
...
 		insts, err := lClient.ListInstances(ctx)
 		if err != nil {
 			logger.V(5).Info("Failed to list instances from launcher, skipping Pod", "name", launcherPod.Name, "err", err)
 			continue
 		}
 
-		launcherDat := ctl.getLauncherData(nodeDat, launcherPod.Name)
-
 		// Check if this launcher has a sleeping instance matching the iscHash
 		if _, instExists := launcherDat.Instances[iscHash]; instExists {
```
MikeSpreitzer: It seems silly to use this cached (thus possibly outdated) information when fresh information was just fetched 7 lines above. Particularly since this usage is not gated by checking `launcherDat.Accurate`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -62,29 +62,65 @@ func (ni nodeItem) process(ctx context.Context, ctl *controller) (error, bool) {
...
+		if apierrors.IsNotFound(err) {
+			logger.V(2).Info("Launcher pod deleted, cleaning up launcher data")
+			ctl.clearLauncherData(nodeDat, item.LauncherPodName)
+			ctl.enqueueRequestersOnNode(ctx, item.NodeName, "launcher pod deleted")
+			return nil, false
+		}
+		return err, true
+	}
+
+	err, retry := ctl.syncLauncherInstances(ctx, nodeDat, launcherPod)
```
MikeSpreitzer: Why do this here when it is also done synchronously in `infSvrItem.process` ?

pkg/controller/dual-pods/controller.go
```diff
@@ -258,7 +259,7 @@ type nodeData struct {
 	InferenceServers map[apitypes.UID]*serverData
 
 	// Launchers maps name of launcher-based server-providing Pod to launcherData.
-	// Access only while holding controller mutex.
+	// Access only inside the calling hierarchy that `nodeItem.process()` is the root caller.
```
MikeSpreitzer: "root" is not right, since it is not at the coldest end of the stack. Maybe something like the following?

> Access only while `nodeItem.process` is on the call stack.

test/e2e/run-launcher-based.sh
```diff
@@ -370,6 +370,94 @@ kubectl wait --for condition=Ready pod/$launcherlb --timeout=5s
...
+
+# Restart the dual-pods controller to test state recovery
+echo "Restarting dual-pods controller..."
+kubectl rollout restart deployment fma-dual-pods-controller
+kubectl rollout status deployment fma-dual-pods-controller --timeout=60s
+
+# Wait for controller to be ready for ongoing checks
+# In detail: allow some time for the dual-pods controller to do something unexpected in the case that the controller is behaving incorrectly,
+# so that the ongoing checks have some chance to fail thus detect the incorrectness, instead of just quickly and coincidentally passing.
+sleep 10
```
MikeSpreitzer: I would wait a little longer here. Maybe something like 30 seconds?

test/e2e/run-launcher-based.sh
```diff
@@ -370,6 +370,94 @@ kubectl wait --for condition=Ready pod/$launcherlb --timeout=5s
...
+
+# Verify launcher pod set is unchanged and target launcher is still running
```
MikeSpreitzer: The check actually here is that the pod set SIZE eventually is what it was before.

test/e2e/run-launcher-based.sh
```diff
@@ -370,6 +370,94 @@ kubectl wait --for condition=Ready pod/$launcherlb --timeout=5s
...
+reqlb_after_delete=$(kubectl get pods -o name -l app=dp-example,instance=$instlb | sed s%pod/%%)
+expect "kubectl get pods -o name -l dual-pods.llm-d.ai/dual=$reqlb_after_delete | wc -l | grep -w 1"
+launcherlb_after_delete=$(kubectl get pods -o name -l dual-pods.llm-d.ai/dual=$reqlb_after_delete | sed s%pod/%%)
+[ "$launcherlb_after_delete" != "$launcherlb" ]
+expect '[ "$(kubectl get pod $reqlb_after_delete -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$launcherlb_after_delete" ]'
+expect '[ "$(kubectl get pod $launcherlb_after_delete -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$reqlb_after_delete" ]'
+
+date
+kubectl wait --for condition=Ready pod/$reqlb_after_delete --timeout=120s
+kubectl wait --for condition=Ready pod/$launcherlb_after_delete --timeout=120s
```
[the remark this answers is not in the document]
waltforme: Replaced all with a simple test.

pkg/controller/dual-pods/controller.go
```diff
@@ -330,34 +331,50 @@ type infSvrItem struct {
...
 	}
 	requesterStr := pod.Annotations[requesterAnnotationKey]
 	requesterParts := strings.Split(requesterStr, " ")
-	if len(requesterParts) != 2 {
-		return infSvrItem{}, infSvrItemDontCare
+	if len(requesterParts) == 2 {
+		return infSvrItem{apitypes.UID(requesterParts[0]), requesterParts[1]}, infSvrItemBoundProvider
+	}
+	// Check for unbound launcher-based server-providing Pod.
+	if pod.Labels != nil {
```
MikeSpreitzer: This guard is not necessary. The Go language definition says that `pod.Labels[ctlrcommon.LauncherConfigNameLabelKey]` does not crash when `pod.Labels == nil` but rather it reports the absence of that entry.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -588,24 +612,21 @@ func (ctl *controller) selectBestLauncherPod(
...
 			continue
 		}
 
-		launcherDat := ctl.getLauncherData(nodeDat, launcherPod.Name)
-
 		// Check if this launcher has a sleeping instance matching the iscHash
-		if _, instExists := launcherDat.Instances[iscHash]; instExists {
+		hasSleepingInstance := false
+		for _, inst := range insts.Instances {
+			if inst.InstanceID == iscHash {
```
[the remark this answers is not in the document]
MikeSpreitzer: `launcherPodAnys` is the launchers that match the _launcher_ hash produced in https://github.com/waltforme/llm-d-fast-model-actuation/blob/42c4ee8d59cb35f109119e78a5035b41a03cad91/pkg/controller/dual-pods/inference-server.go#L476, extracted at https://github.com/waltforme/llm-d-fast-model-actuation/blob/42c4ee8d59cb35f109119e78a5035b41a03cad91/pkg/controller/dual-pods/inference-server.go#L480, and used to get `launcherPodAnys` at https://github.com/waltforme/llm-d-fast-model-actuation/blob/42c4ee8d59cb35f109119e78a5035b41a03cad91/pkg/controller/dual-pods/inference-server.go#L482 .
(The commit cited is the current rev of this PR, as I write this.)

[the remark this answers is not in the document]
waltforme: Yes. The reason that I mentioned iscHash as well is that the two hashes work here together.

MikeSpreitzer (commented):
LGTM, provided the E2E test on OpenShift succeeds.

PR #363, opened 2026-03-18
Dynamic port allocation with reverse proxy on server-requesting pod

delavet:
This PR makes `InferenceServerConfig.spec.modelServerConfig.Port` optional, implementing the approach discussed in the following Slack conversation to enable integration with InferencePool. Currently, when no port is specified, the dual-pods controller can dynamically allocate a port and record it in the launcher pod’s `inference.networking.k8s.io/active-ports` annotation.

https://llm-d.slack.com/archives/C09TNPEFJUD/p1769183239461429

A current issue indeed exists: users of InferencePool and maintainers of the dual-pods controller must mutually agree upon a predefined list of available ports. Currently, this port range is hardcoded—from `8005` to `8005 + N - 1`—which ironically increases complexity .

I am seriously considering the alternative options proposed earlier: adding TCP proxy functionality to the requester, enabling direct access to the vLLM service via the requester—this should also align all ports. Maybe it will be better.

I would greatly appreciate your comments. @MikeSpreitzer @lionelvillard

### DIFF
--- a/api/fma/v1alpha1/inferenceserverconfig_types.go
+++ b/api/fma/v1alpha1/inferenceserverconfig_types.go
@@ -35,7 +35,10 @@ type InferenceServerConfigSpec struct {
 type ModelServerConfig struct {
 	// Port is the port on which the vLLM server will listen
 	// Particularly, management of vLLM instances' sleep state is done through this port
-	Port int32 `json:"port"`
+	// If not specified, the dual-pods controller will automatically assign an available port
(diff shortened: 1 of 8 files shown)


diegocastanibm (changes requested):
I do not know why the dynamic port controller capability has been documented in the `docs/launcher.md`. I think a better place would be the `docs/dual-pods.md` file. 
Also, please, remember to sign all your commits.

PR #374, opened 2026-03-22
handle LauncherConfig update events

osswangxining:
Previously, when a `LauncherConfig` object was updated:
- The `launcher-populator` controller only processed `LauncherPopulationPolicy` events

## This Solution
Added `LauncherConfig` event handling to the controller to ensure:
1. When a `LauncherConfig` changes, all affected server-requesting pods are not affected
2. New launcher pods are created with the updated configuration
3. Old launcher pods (not bound to any requester) are eventually cleaned up

### DIFF
--- a/pkg/controller/launcher-populator/node-launcher-key.go
+++ b/pkg/controller/launcher-populator/node-launcher-key.go
@@ -19,6 +19,8 @@ package launcherpopulator
 import (
 	"fmt"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
 )
 
@@ -32,18 +34,16 @@ func (k NodeLauncherKey) String() string {
 	return fmt.Sprintf("%s/%s", k.LauncherConfigName, k.NodeName)
 }
 
-// DesiredStateEntry holds the desired count and corresponding LauncherConfig
+// DesiredStateEntry holds the desired count and the LauncherConfig spec
 // for a (Node, LauncherConfig) pair.
 type DesiredStateEntry struct {
-	Count          int32
-	LauncherConfig *fmav1alpha1.LauncherConfig
+	Count              int32
+	LauncherConfigSpec fmav1alpha1.LauncherConfigSpec
+	LauncherConfigOwnerRef metav1.OwnerReference
 }
 
 func (e DesiredStateEntry) String() string {
-	if e.LauncherConfig != nil {
-		return fmt.Sprintf("count=%d,config=%s", e.Count, e.LauncherConfig.Name)
-	}
-	return fmt.Sprintf("count=%d", e.Count)
+	return fmt.Sprintf("count=%d,config=%s", e.Count, e.LauncherConfigOwnerRef.Name)
 }
 
 // MapToLoggable converts a map of NodeLauncherKey to Val values into a string representation.
--- a/pkg/controller/launcher-populator/node-launcher-key.go
+++ b/pkg/controller/launcher-populator/node-launcher-key.go
@@ -37,13 +37,16 @@ func (k NodeLauncherKey) String() string {
 // DesiredStateEntry holds the desired count and the LauncherConfig spec
 // for a (Node, LauncherConfig) pair.
 type DesiredStateEntry struct {
-	Count              int32
-	LauncherConfigSpec fmav1alpha1.LauncherConfigSpec
+	Count                  int32
+	LauncherConfigSpec     *fmav1alpha1.LauncherConfigSpec
 	LauncherConfigOwnerRef metav1.OwnerReference
 }
 
 func (e DesiredStateEntry) String() string {
-	return fmt.Sprintf("count=%d,config=%s", e.Count, e.LauncherConfigOwnerRef.Name)
+	if e.LauncherConfigSpec == nil {
+		return fmt.Sprintf("count=%d,config=%s,spec=<nil>", e.Count, e.LauncherConfigOwnerRef.Name)
+	}
+	return fmt.Sprintf("count=%d,config=%s,spec=%+v", e.Count, e.LauncherConfigOwnerRef.Name, *e.LauncherConfigSpec)
 }
 
 // MapToLoggable converts a map of NodeLauncherKey to Val values into a string representation.
--- a/pkg/controller/launcher-populator/node-launcher-key.go
+++ b/pkg/controller/launcher-populator/node-launcher-key.go
@@ -18,6 +18,8 @@ package launcherpopulator
 
 import (
 	"fmt"
+
+	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
 )
 
 // NodeLauncherKey defines the unique identifier for a (Node, LauncherConfig) pair
@@ -30,8 +32,22 @@ func (k NodeLauncherKey) String() string {
 	return fmt.Sprintf("%s/%s", k.LauncherConfigName, k.NodeName)
 }
 
-// MapToLoggable converts a map of NodeLauncherKey to int32 values into a string representation.
-// This function formats the map as a string with the format "{namespace/name/node:count, ...}"
+// DesiredStateEntry holds the desired count and corresponding LauncherConfig
+// for a (Node, LauncherConfig) pair.
+type DesiredStateEntry struct {
+	Count          int32
+	LauncherConfig *fmav1alpha1.LauncherConfig
+}
+
+func (e DesiredStateEntry) String() string {
+	if e.LauncherConfig != nil {
+		return fmt.Sprintf("count=%d,config=%s", e.Count, e.LauncherConfig.Name)
+	}
+	return fmt.Sprintf("count=%d", e.Count)
+}
+
+// MapToLoggable converts a map of NodeLauncherKey to Val values into a string representation.
+// This function formats the map as a string with the format "{namespace/name/node:val, ...}"
 // for debugging and logging purposes.
 func MapToLoggable[Key interface {
 	comparable
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -121,11 +121,19 @@ type lppItem struct {
 	cache.ObjectName
 }
 
+type lcItem struct {
+	cache.ObjectName
+}
+
 func (ctl *controller) OnAdd(obj any, isInInitialList bool) {
 	switch typed := obj.(type) {
 	case *fmav1alpha1.LauncherPopulationPolicy:
 		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherPopulationPolicy reference due to notification of add", "name", typed.Name)
 		item := lppItem{cache.MetaObjectToName(typed)}
+		ctl.Queue.Add(item)
+	case *fmav1alpha1.LauncherConfig:
+		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherConfig reference due to notification of add", "name", typed.Name)
+		item := lcItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
 	default:
 		ctl.enqueueLogger.V(5).Info("Notified of add of type of ignored object", "type", fmt.Sprintf("%T", obj))
@@ -138,6 +146,10 @@ func (ctl *controller) OnUpdate(prev, obj any) {
 	case *fmav1alpha1.LauncherPopulationPolicy:
 		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherPopulationPolicy reference due to notification of update", "name", typed.Name)
 		item := lppItem{cache.MetaObjectToName(typed)}
+		ctl.Queue.Add(item)
+	case *fmav1alpha1.LauncherConfig:
+		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherConfig reference due to notification of update", "name", typed.Name)
+		item := lcItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
 	default:
 		ctl.enqueueLogger.V(5).Info("Notified of update of type of ignored object", "type", fmt.Sprintf("%T", obj))
@@ -154,7 +166,10 @@ func (ctl *controller) OnDelete(obj any) {
 		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherPopulationPolicy reference due to notification of delete", "name", typed.Name)
 		item := lppItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
-
+	case *fmav1alpha1.LauncherConfig:
+		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherConfig reference due to notification of delete", "name", typed.Name)
+		item := lcItem{cache.MetaObjectToName(typed)}
+		ctl.Queue.Add(item)
 	default:
 		ctl.enqueueLogger.V(5).Info("Notified of delete of type of ignored object", "type", fmt.Sprintf("%T", obj))
 		return
@@ -232,6 +247,63 @@ func (item lppItem) process(ctx context.Context, ctl *controller) (error, bool) 
 		logger.Error(err, "Failed to reconcile launchers")
 		return err, true
 	}
+	return nil, false
+}
+
+func (item lcItem) process(ctx context.Context, ctl *controller) (error, bool) {
+	logger := klog.FromContext(ctx)
+
+	// Get the LauncherConfig
+	lc, err := ctl.lcLister.LauncherConfigs(ctl.namespace).Get(item.Name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("LauncherConfig no longer exists, skipping reconciliation", "name", item.Name)
+			return nil, false
+		}
+		logger.Error(err, "Failed to get LauncherConfig", "name", item.Name)
+		return err, true
+	}
+
+	// Get all LauncherPopulationPolicies that reference this LauncherConfig
+	policies, err := ctl.lppLister.List(labels.Everything())
+	if err != nil {
+		logger.Error(err, "Failed to list LauncherPopulationPolicies")
+		return err, true
+	}
+
+	// Build desired state for this LauncherConfig across all nodes
+	desired := make(map[NodeLauncherKey]int32)
+	for _, lpp := range policies {
+		nodes, err := ctl.getMatchingNodes(ctx, lpp.Spec.EnhancedNodeSelector)
+		if err != nil {
+			logger.Error(err, "Failed to get matching nodes for policy", "policy", lpp.Name)
+			continue
+		}
+
+		for _, countRule := range lpp.Spec.CountForLauncher {
+			if countRule.LauncherConfigName != lc.Name {
+				continue
+			}
+			for _, node := range nodes {
+				key := NodeLauncherKey{
+					NodeName:           node.Name,
+					LauncherConfigName: lc.Name,
+				}
+				// Take the maximum count if multiple rules apply
+				if current, exists := desired[key]; !exists || countRule.LauncherCount > current {
+					desired[key] = countRule.LauncherCount
+				}
+			}
+		}
+	}
+
+	// Reconcile launchers for this LauncherConfig
+	if err := ctl.reconcileAllLaunchers(ctx, desired); err != nil {
+		logger.Error(err, "Failed to reconcile launchers for LauncherConfig", "name", lc.Name)
+		return err, true
+	}
+
+	logger.Info("Successfully reconciled launchers for LauncherConfig", "name", lc.Name)
 	return nil, false
 }
 
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -152,7 +152,7 @@ func (ctl *controller) OnUpdate(prev, obj any) {
 		item := lcItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
 	default:
-		ctl.enqueueLogger.V(5).Info("Notified of update of type of ignored object", "type", fmt.Sprintf("%T", obj))
+		ctl.enqueueLogger.V(5).Info("Notified of update of object of ignored type", "type", fmt.Sprintf("%T", obj))
 		return
 	}
 }
@@ -167,7 +167,7 @@ func (ctl *controller) OnDelete(obj any) {
 		item := lppItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
 	default:
-		ctl.enqueueLogger.V(5).Info("Notified of delete of type of ignored object", "type", fmt.Sprintf("%T", obj))
+		ctl.enqueueLogger.V(5).Info("Notified of delete of object of ignored type", "type", fmt.Sprintf("%T", obj))
 		return
 	}
 }
@@ -261,12 +261,18 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
 					NodeName:           node.Name,
 					LauncherConfigName: countRule.LauncherConfigName,
 				}
-				ownerRef := *metav1.NewControllerRef(lc, fmav1alpha1.SchemeGroupVersion.WithKind("LauncherConfig"))
-				ownerRef.BlockOwnerDeletion = ptr.To(false)
+				ownerRef := metav1.OwnerReference{
+					APIVersion:         fmav1alpha1.SchemeGroupVersion.String(),
+					Kind:               "LauncherConfig",
+					Name:               lc.Name,
+					UID:                lc.UID,
+					Controller:         ptr.To(false),
+					BlockOwnerDeletion: ptr.To(false),
+				}
 				if entry, exists := desired[key]; !exists || countRule.LauncherCount > entry.Count {
 					desired[key] = DesiredStateEntry{
 						Count:                  countRule.LauncherCount,
-						LauncherConfigSpec:     lc.Spec,
+						LauncherConfigSpec:     &lc.Spec,
 						LauncherConfigOwnerRef: ownerRef,
 					}
 				}
@@ -347,10 +353,10 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 	deletionInProgress := false // tracks pods already being deleted (DeletionTimestamp set)
 
 	type creationInfo struct {
-		key    NodeLauncherKey
-		count  int
-		spec   fmav1alpha1.LauncherConfigSpec
-		owner  metav1.OwnerReference
+		key   NodeLauncherKey
+		count int
+		spec  *fmav1alpha1.LauncherConfigSpec
+		owner metav1.OwnerReference
 	}
 	var creations []creationInfo
 
@@ -371,15 +377,17 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 		// BuildLauncherPodFromTemplate computes a hash of the fully built pod spec
 		// and stores it as the LauncherConfigHashAnnotationKey annotation.
 		nominalHash := ""
-		nominalPod, err := utils.BuildLauncherPodFromTemplate(
-			entry.LauncherConfigSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
-		if err != nil {
-			// The only error possible here is that the PodTemplate lacks an inference server container.
-			// In that case we proceed without a nominal hash, so no stale-pod detection occurs for this config.
-			logger.Error(err, "Failed to build nominal pod for hash comparison",
-				"node", nodeName, "config", key.LauncherConfigName)
-		} else if nominalPod.Annotations != nil {
-			nominalHash = nominalPod.Annotations[string(common.LauncherConfigHashAnnotationKey)]
+		if entry.LauncherConfigSpec != nil {
+			nominalPod, err := utils.BuildLauncherPodFromTemplate(
+				entry.LauncherConfigSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
+			if err != nil {
+				// The only error possible here is that the PodTemplate lacks an inference server container.
+				// In that case we proceed without a nominal hash, so no stale-pod detection occurs for this config.
+				logger.Error(err, "Failed to build nominal pod for hash comparison",
+					"node", nodeName, "config", key.LauncherConfigName)
+			} else {
+				nominalHash = nominalPod.Annotations[string(common.LauncherConfigHashAnnotationKey)]
+			}
 		}
 
 		// Categorize current pods: separate live unbound current-spec pods from stale/unbound ones
@@ -432,7 +440,7 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 		effectiveRemaining := liveBoundCount + len(liveUnboundCurrentPods)
 		diff := entry.Count - int32(effectiveRemaining)
 
-		logger.Info("Analyzing config on node",
+		logger.Info("Analyzed config on node",
 			"node", nodeName,
 			"config", key.LauncherConfigName,
 			"current", effectiveRemaining,
@@ -525,7 +533,7 @@ func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLa
 
 // createLaunchers creates the specified number of launcher pods on a node
 // using the given LauncherConfig spec and owner reference directly (no additional lookup needed).
-func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, key NodeLauncherKey, count int, lcSpec fmav1alpha1.LauncherConfigSpec, lcOwnerRef metav1.OwnerReference) error {
+func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, key NodeLauncherKey, count int, lcSpec *fmav1alpha1.LauncherConfigSpec, lcOwnerRef metav1.OwnerReference) error {
 	logger := klog.FromContext(ctx)
 
 	// Create the specified number of launcher pods
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -261,10 +261,13 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
 					NodeName:           node.Name,
 					LauncherConfigName: countRule.LauncherConfigName,
 				}
+				ownerRef := *metav1.NewControllerRef(lc, fmav1alpha1.SchemeGroupVersion.WithKind("LauncherConfig"))
+				ownerRef.BlockOwnerDeletion = ptr.To(false)
 				if entry, exists := desired[key]; !exists || countRule.LauncherCount > entry.Count {
 					desired[key] = DesiredStateEntry{
-						Count:          countRule.LauncherCount,
-						LauncherConfig: lc,
+						Count:                  countRule.LauncherCount,
+						LauncherConfigSpec:     lc.Spec,
+						LauncherConfigOwnerRef: ownerRef,
 					}
 				}
 			}
@@ -344,9 +347,10 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 	deletionInProgress := false // tracks pods already being deleted (DeletionTimestamp set)
 
 	type creationInfo struct {
-		key   NodeLauncherKey
-		count int
-		lc    *fmav1alpha1.LauncherConfig
+		key    NodeLauncherKey
+		count  int
+		spec   fmav1alpha1.LauncherConfigSpec
+		owner  metav1.OwnerReference
 	}
 	var creations []creationInfo
 
@@ -356,6 +360,8 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 
 		currentLaunchers, err := ctl.getCurrentLaunchersOnNode(ctx, key)
 		if err != nil {
+			// The only error possible here is from the lister, which should never fail in practice.
+			// Log and skip this config rather than aborting the entire reconciliation.
 			logger.Error(err, "Failed to get current launchers for config",
 				"node", nodeName, "config", key.LauncherConfigName)
 			continue
@@ -365,10 +371,12 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 		// BuildLauncherPodFromTemplate computes a hash of the fully built pod spec
 		// and stores it as the LauncherConfigHashAnnotationKey annotation.
 		nominalHash := ""
-		nominalPod, hashErr := utils.BuildLauncherPodFromTemplate(
-			entry.LauncherConfig.Spec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
-		if hashErr != nil {
-			logger.Error(hashErr, "Failed to build nominal pod for hash comparison",
+		nominalPod, err := utils.BuildLauncherPodFromTemplate(
+			entry.LauncherConfigSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
+		if err != nil {
+			// The only error possible here is that the PodTemplate lacks an inference server container.
+			// In that case we proceed without a nominal hash, so no stale-pod detection occurs for this config.
+			logger.Error(err, "Failed to build nominal pod for hash comparison",
 				"node", nodeName, "config", key.LauncherConfigName)
 		} else if nominalPod.Annotations != nil {
 			nominalHash = nominalPod.Annotations[string(common.LauncherConfigHashAnnotationKey)]
@@ -457,7 +465,8 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 			creations = append(creations, creationInfo{
 				key:   key,
 				count: int(diff),
-				lc:    entry.LauncherConfig,
+				spec:  entry.LauncherConfigSpec,
+				owner: entry.LauncherConfigOwnerRef,
 			})
 		}
 	}
@@ -476,7 +485,7 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 	// No deletions needed, proceed with planned creations
 	totalCreated := 0
 	for _, creation := range creations {
-		if err := ctl.createLaunchers(ctx, *node, creation.key, creation.count, creation.lc); err != nil {
+		if err := ctl.createLaunchers(ctx, *node, creation.key, creation.count, creation.spec, creation.owner); err != nil {
 			logger.Error(err, "Failed to create launchers for config",
 				"node", nodeName,
 				"config", creation.key.LauncherConfigName,
@@ -490,15 +499,10 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 			"created", creation.count)
 	}
 
-	if totalCreated > 0 {
-		logger.Info("Completed creation of launchers",
-			"node", nodeName,
-			"created", totalCreated)
-	}
-
 	logger.Info("Completed reconciliation for node",
 		"node", nodeName,
-		"configs_processed", len(keys))
+		"configs_processed", len(keys),
+		"created", totalCreated)
 
 	return false, nil
 }
@@ -520,23 +524,18 @@ func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLa
 }
 
 // createLaunchers creates the specified number of launcher pods on a node
-// using the given LauncherConfig directly (no additional lookup needed).
-func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, key NodeLauncherKey, count int, launcherConfig *fmav1alpha1.LauncherConfig) error {
+// using the given LauncherConfig spec and owner reference directly (no additional lookup needed).
+func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, key NodeLauncherKey, count int, lcSpec fmav1alpha1.LauncherConfigSpec, lcOwnerRef metav1.OwnerReference) error {
 	logger := klog.FromContext(ctx)
 
 	// Create the specified number of launcher pods
 	for i := 0; i < count; i++ {
-		pod, err := utils.BuildLauncherPodFromTemplate(launcherConfig.Spec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
+		pod, err := utils.BuildLauncherPodFromTemplate(lcSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
 		if err != nil {
 			return fmt.Errorf("failed to build launcher pod: %w", err)
 		}
-		pod.GenerateName = fmt.Sprintf("launcher-%s-", launcherConfig.Name)
-		// Set owner reference pointing to LauncherConfig
-		ownerRef := *metav1.NewControllerRef(launcherConfig, fmav1alpha1.SchemeGroupVersion.WithKind("LauncherConfig"))
-		ownerRef.BlockOwnerDeletion = ptr.To(false)
-		pod.OwnerReferences = []metav1.OwnerReference{
-			ownerRef,
-		}
+		pod.GenerateName = fmt.Sprintf("launcher-%s-", key.LauncherConfigName)
+		pod.OwnerReferences = []metav1.OwnerReference{lcOwnerRef}
 
 		if _, err := ctl.coreclient.Pods(pod.Namespace).Create(ctx, pod, metav1.CreateOptions{}); err != nil {
 			return fmt.Errorf("failed to create launcher pod: %w", err)
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -193,7 +193,7 @@ func (item lppItem) process(ctx context.Context, ctl *controller) (error, bool) 
 	logger := klog.FromContext(ctx)
 
 	// Build desired state from all policies
-	populationPolicy, err := ctl.buildDesiredStateFromPolicies(ctx, nil)
+	populationPolicy, err := ctl.buildDesiredStateFromPolicies(ctx)
 	if err != nil {
 		logger.Error(err, "Failed to build desired state from policies")
 		return err, true
@@ -202,9 +202,14 @@ func (item lppItem) process(ctx context.Context, ctl *controller) (error, bool) 
 	logger.Info("Final population policy", "policy", MapToLoggable(populationPolicy))
 
 	// Adjust launcher pods according to final requirements
-	if err := ctl.reconcileAllLaunchers(ctx, populationPolicy); err != nil {
+	needsRequeue, err := ctl.reconcileAllLaunchers(ctx, populationPolicy)
+	if err != nil {
 		logger.Error(err, "Failed to reconcile launchers")
 		return err, true
+	}
+
+	if needsRequeue {
+		return nil, true
 	}
 	return nil, false
 }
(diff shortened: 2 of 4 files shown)


waltforme:
I think the concerned handling of LauncherConfig update is described as https://llm-d.slack.com/archives/C09TNPEFJUD/p1774153845072779?thread_ts=1773867574.189679&cid=C09TNPEFJUD, and also in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-4105395399. Am I right?

If so, IMO the handling can be done by any controllers because the necessary info (LauncherConfig name and hash) for the handling is contained within the launcher pods as annotations. Further, I think the launcher popular is the best candidate to do the handling, because "it is currently the only controller deleting launcher Pods and I think the system would be most stable if we keep that responsibility in one controller" as said by Mike in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-4107468213

> BTW, we also need to handle this LauncherConfig update within duals-pod controller, @waltforme do you think that you would update the code or let me try?

rubambiza (commented):
Left a few clarifying questions.

pkg/controller/launcher-populator/populator.go
```diff
@@ -287,32 +338,136 @@ func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLau
...
+			"planned", len(allDeletions),
+			"actual", deletedCount)
+	} else if deletedCount > 0 {
+		logger.Info("Completed Phase 1: Deletions",
+			"node", nodeName,
+			"deleted", deletedCount)
+	}
+
+	// Execute Phase 2: Create all needed pods AFTER deletions complete
+	totalCreated := 0
```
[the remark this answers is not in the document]
osswangxining: Currently I follow the steps below:
// Execute Phase 1: Delete all marked pods FIRST (across all configs on this node): Line 407 ~ 442
// Execute Phase 2: Create all needed pods AFTER deletions complete: Line 445~465

Does this make sense?

pkg/controller/launcher-populator/populator.go
```diff
@@ -287,32 +338,136 @@ func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLau
...
+		}); err != nil {
+			if apierrors.IsNotFound(err) {
+				logger.Info("Launcher pod already deleted", "pod", del.pod.Name)
+				deletedCount++
+				continue
+			}
+			if apierrors.IsConflict(err) {
+				logger.Info("Launcher pod version conflict, skipping deletion",
+					"pod", del.pod.Name, "error", err)
+				continue
```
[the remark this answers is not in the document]
osswangxining: I got your idea, now fixed, please help review again, thanks.

osswangxining:
> > When handling deletions, would it be sufficient to just compare the name? It might not be necessary to compare the spec content.
> 
> Agreed regarding LauncherConfig deletions. But that is beside the point that I raised in [#374 (comment)](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#issuecomment-4156312219) , which is about what happens when the Spec of a LauncherConfig object _changes_.

When the Spec of a LauncherConfig object changes, existing Launcher Pods should remain unaffected; only newly created Launcher Pods will utilize the updated LauncherConfig spec.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,29 +300,36 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
-	// Reconcile for each (Node, LauncherConfig) pair
-	for key, desiredCount := range desired {
-		if err := ctl.reconcileLaunchersOnNode(ctx, key, desiredCount); err != nil {
-			logger.Error(err, "Failed to reconcile launchers on node",
-				"node", key.NodeName,
-				"config", key.LauncherConfigName)
-			// Continue processing other combinations
+
+	// Group by node to optimize resource usage across all LauncherConfigs on each node
+	nodeGroups := make(map[string][]NodeLauncherKey)
```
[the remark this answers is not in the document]
osswangxining: As variable "desired" is map[NodeLauncherKey]int32, then it's simpler to use the same type.

pkg/controller/launcher-populator/populator.go
```diff
@@ -180,59 +191,92 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
...
+		return err, true
+	}
+
+	logger.Info("Successfully reconciled launchers for LauncherConfig", "name", lc.Name)
+	return nil, false
+}
+
+// buildDesiredStateFromPolicies builds the desired state map from all policies.
+// If filterByConfig is provided, only policies referencing that config are considered.
+func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context, filterByConfig *string) (map[NodeLauncherKey]int32, error) {
```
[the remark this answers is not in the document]
osswangxining: Sorry, I'm not sure I follow. Could you elaborate a bit more?"

pkg/controller/launcher-populator/populator.go
```diff
@@ -287,32 +338,136 @@ func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLau
...
+
+		if diff < 0 {
+			// Need to delete some pods for this config
+			numToDelete := int(-diff)
+			for i := 0; i < numToDelete && i < len(currentLaunchers); i++ {
+				pod := currentLaunchers[len(currentLaunchers)-1-i]
+				isBound, requesterPodName := ctl.isLauncherBoundToServerRequestingPod(pod)
+				if isBound {
+					logger.V(5).Info("Skipping deletion of bound pod",
+						"pod", pod.Name, "server-requesting pod", requesterPodName)
```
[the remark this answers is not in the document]
aavarghese: Right, since the for loop has now changed, numToDelete needs no adjustment anymore...

MikeSpreitzer:
@osswangxining:

> When the Spec of a LauncherConfig object changes, existing Launcher Pods should remain unaffected; only newly created Launcher Pods will utilize the updated LauncherConfig spec.

No. We agreed on A3 to Q3 in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-3717326061 . This means that when the Spec of a LauncherConfig object changes, this should lead to deletion of each existing Launcher Pod that (a) was created from the wrong LauncherConfigSpec and (b) is _not_ bound to a server-requesting Pod. (Also considering sleep/wake is NOT necessary, since considering binding covers that.)

pkg/controller/launcher-populator/populator.go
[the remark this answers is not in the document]
MikeSpreitzer: This method does not need the `filterByConfig *string` parameter, because no LauncherConfig needs special treatment. _Every_ LauncherConfig that does not exist should get no processing. Because we are using A3 to Q3 in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-3717326061 , this method needs to read (from the informer's local cache is preferred) _every_ LauncherConfig object to get its current LauncherConfigSpec; these reads will reveal _all_ of the non-existent LauncherConfig objects.

pkg/controller/launcher-populator/populator.go
```diff
@@ -287,32 +338,145 @@ func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLau
...
+					continue
+				}
+				allDeletions = append(allDeletions, deletionInfo{
+					key:    key,
+					pod:    pod,
+					reason: "excess launcher",
+				})
+				collectedForDeletion++
```
MikeSpreitzer: It would be simpler to do the deletion here. No need to queue for later.

pkg/controller/launcher-populator/populator.go
```diff
@@ -287,32 +338,145 @@ func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLau
...
+		logger.Info("Deleted launcher pod (Phase 1)",
+			"pod", del.pod.Name,
+			"node", nodeName,
+			"config", del.key.LauncherConfigName,
+			"reason", del.reason)
+		deletedCount++
+	}
+
+	if len(conflictErrors) > 0 {
+		return fmt.Errorf("encountered %d version conflicts during deletion (pods: %v), will retry",
```
MikeSpreitzer: Not a big deal, but I would use https://pkg.go.dev/errors@go1.26.2#Join to build the returned error.

pkg/controller/launcher-populator/populator.go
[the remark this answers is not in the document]
MikeSpreitzer: That is not what I see now. I still see building up both a list of deletions and a list of creations, then doing the deletions, then (without waiting) doing the creations.

pkg/controller/launcher-populator/populator.go
```diff
@@ -287,32 +338,136 @@ func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLau
...
+		logger.Info("Completed Phase 1: Deletions",
+			"node", nodeName,
+			"deleted", deletedCount)
+	}
+
+	// Execute Phase 2: Create all needed pods AFTER deletions complete
+	totalCreated := 0
+	for _, creation := range allCreations {
+		if err := ctl.createLaunchers(ctx, *node, creation.key, creation.count); err != nil {
+			logger.Error(err, "Failed to create launchers for config",
```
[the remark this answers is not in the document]
MikeSpreitzer: > why we're not catching partial success like we are doing for the deletion (i.e., checking for isConflict to skip adding deletedCount)

I do not understand the suggestion. The launcher creation code does not stipulate the Pod's name, so there is not concept of that particular Pod having already been created.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/populator.go
```diff
@@ -121,12 +121,20 @@ type lppItem struct {
...
 	case *fmav1alpha1.LauncherPopulationPolicy:
 		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherPopulationPolicy reference due to notification of add", "name", typed.Name)
 		item := lppItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
+	case *fmav1alpha1.LauncherConfig:
+		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherConfig reference due to notification of add", "name", typed.Name)
+		item := lcItem{cache.MetaObjectToName(typed)}
+		ctl.Queue.Add(item)
 	default:
 		ctl.enqueueLogger.V(5).Info("Notified of add of type of ignored object", "type", fmt.Sprintf("%T", obj))
```
rubambiza: **Nit**: For code readability

```suggestion
		ctl.enqueueLogger.V(5).Info("Notified of add of ignored object", "type", fmt.Sprintf("%T", obj))
```

pkg/controller/launcher-populator/populator.go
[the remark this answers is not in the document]
rubambiza: @osswangxining 's comment clarified things for me. @MikeSpreitzer , I was suggesting flagging/retrying in case we do not get the full `creation.count` of launchers created, like the case where some failed deletions are flagged [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374/changes/be2599664292142a1a01bcef69e474d6506a7a6b..b082cf9339e3b0617057a884fba0bb5c91e5e7eb#diff-4fa6d5c4dc86eff0054d35b46748081d11cd647e4463fcedff1bbb30a890ea74R436).

rubambiza (commented):
Did another round of review for the new commits since my last review two weeks ago. Added some clarification on a previous suggestion (which has been addressed) and a minor readability. The concern about the [deletion/creation steps](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374/changes/be2599664292142a1a01bcef69e474d6506a7a6b..b082cf9339e3b0617057a884fba0bb5c91e5e7eb#r2988275778) still lingers.

pkg/controller/launcher-populator/populator.go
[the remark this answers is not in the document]
MikeSpreitzer: @rubambiza: thanks for the clarification. I agree that there should be a retry if the desired number of launchers does not get created.

pkg/controller/launcher-populator/populator.go
osswangxining: resolved within latest commit

osswangxining:
@MikeSpreitzer @rubambiza @waltforme @aavarghese  I rebase together with the fix for all the comments. Looking forward to your reviews again, thanks a lot.

pkg/controller/launcher-populator/populator.go
```diff
@@ -121,14 +121,22 @@ type lppItem struct {
...
 		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherPopulationPolicy reference due to notification of add", "name", typed.Name)
 		item := lppItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
+	case *fmav1alpha1.LauncherConfig:
+		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherConfig reference due to notification of add", "name", typed.Name)
+		item := lcItem{cache.MetaObjectToName(typed)}
+		ctl.Queue.Add(item)
 	default:
-		ctl.enqueueLogger.V(5).Info("Notified of add of type of ignored object", "type", fmt.Sprintf("%T", obj))
+		ctl.enqueueLogger.V(5).Info("Notified of add of ignored object", "type", fmt.Sprintf("%T", obj))
```
MikeSpreitzer: The "type" is important to include in the log message, since the object's type is the critical factor here and appears in the key/value pairs. I think that the original is a commonly used English way of stating the intended idea, but perhaps is not strictly correct grammar. Perhaps it is better to be more expansive. Perhaps a better message would be "Notified of add of object of ignored type".

The same wording issue applies in the update and delete handlers.

pkg/controller/launcher-populator/populator.go
```diff
@@ -180,59 +191,104 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
...
+	needsRequeue, err := ctl.reconcileAllLaunchers(ctx, populationPolicy)
+	if err != nil {
+		logger.Error(err, "Failed to reconcile launchers")
+		return err, true
+	}
+
+	if needsRequeue {
+		return nil, true
+	}
+	return nil, false
```
MikeSpreitzer: These can be simplified to just `return nil, needsRequeue`.

pkg/controller/launcher-populator/populator.go
```diff
@@ -180,59 +191,104 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
...
+}
+
+func (item lcItem) process(ctx context.Context, ctl *controller) (error, bool) {
```
MikeSpreitzer, line 196: The body of this method is now exactly the same as the body of `lppItem.process`. Rather than maintain two copies, I think that it would be better to have both simply call one function that has the common code.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer, line 291: This error needs to get to the `Errors` of the `LauncherPopulationPolicyStatus`. That can be done in a later PR.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
 	return matchedNodes, nil
 }
 
-// reconcileAllLaunchers adjusts all launcher pods according to final requirements
-func (ctl *controller) reconcileAllLaunchers(ctx context.Context, desired map[NodeLauncherKey]int32) error {
+// reconcileAllLaunchers adjusts all launcher pods according to final requirements.
+// It returns true if a requeue is needed (deletions were performed or are in progress),
+// so that creations happen only after deletions have taken effect.
+func (ctl *controller) reconcileAllLaunchers(ctx context.Context, desired map[NodeLauncherKey]DesiredStateEntry) (bool, error) {
```
MikeSpreitzer, line 310: FYI, everywhere else that is analogous returns the two values in the opposite order (`(error, bool)`). This is not a big deal.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+	anyRequeueNeeded := false
+	for nodeName, keys := range nodeGroups {
+		needsRequeue, err := ctl.reconcileLaunchersOnSingleNode(ctx, nodeName, keys, desired)
+		if err != nil {
+			logger.Error(err, "Failed to reconcile launchers on node", "node", nodeName)
+			continue
+		}
+		if needsRequeue {
+			anyRequeueNeeded = true
 		}
```
MikeSpreitzer:
```suggestion
			anyRequeueNeeded = anyRequeueNeeded || needsRequeue
```

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+
+	return anyRequeueNeeded, nil
 }
 
-// reconcileLaunchersOnNode ensures the number of launchers with a specific launcher config on a node matches the requirement
-func (ctl *controller) reconcileLaunchersOnNode(ctx context.Context, key NodeLauncherKey, desiredCount int32) error {
+// reconcileLaunchersOnSingleNode handles all LauncherConfigs for a single node.
+// For each LauncherConfig, it does deletions immediately as they are identified
+// and remembers creations called for. If any deletions were performed (or are in
+// progress from a previous cycle), it returns true to request a requeue so that
```
MikeSpreitzer, line 335: This controller should be reacting to Pod deletions and creations (#448). If it did that then it would not need to requeue due to a deletion is in progress, because it will get notified when the deletion completes. Can be addressed in a later PR.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+	// Process each LauncherConfig on this node
+	for _, key := range keys {
+		entry := desired[key]
+
+		currentLaunchers, err := ctl.getCurrentLaunchersOnNode(ctx, key)
 		if err != nil {
-			return fmt.Errorf("failed to create launchers: %w", err)
+			logger.Error(err, "Failed to get current launchers for config",
+				"node", nodeName, "config", key.LauncherConfigName)
+			continue
```
MikeSpreitzer, line 373: It might be worth a comment here documenting why it is OK to not do anything about this error. It is because the only error that can get here is from a failure of a lister, which I expect will never actually happen.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
-		// Need to delete excess launchers
-		err := ctl.deleteExcessLaunchers(ctx, currentLaunchers, int(-diff))
-		if err != nil {
-			return fmt.Errorf("failed to delete excess launchers: %w", err)
+
+		// Compute the nominal hash for spec-change detection.
+		// BuildLauncherPodFromTemplate computes a hash of the fully built pod spec
+		// and stores it as the LauncherConfigHashAnnotationKey annotation.
+		nominalHash := ""
+		nominalPod, hashErr := utils.BuildLauncherPodFromTemplate(
```
MikeSpreitzer: `hashErr` is a strange name for this error, coming from a func named "BuildLauncherPodFromTemplate". The problem is not necessarily in the hashing. Looking in that func, I see that the only error that it will return is a complaint about not finding the inference server container in PodSpec.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+			entry.LauncherConfig.Spec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
+		if hashErr != nil {
+			logger.Error(hashErr, "Failed to build nominal pod for hash comparison",
```
MikeSpreitzer: Since the only error that actually reaches here is a user error (not putting an inference server container in the launcher Pod template), the proper handling of this error is to, in addition to ignoring the LauncherConfig here, expose the error to users in the `Status` section of the `LauncherConfig` (which, I see, does not yet have any place to report problems). Can be handled in other PR(s).

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+		logger.Info("Created launchers for config",
+			"node", nodeName,
+			"config", creation.key.LauncherConfigName,
+			"created", creation.count)
+	}
+
+	if totalCreated > 0 {
+		logger.Info("Completed creation of launchers",
+			"node", nodeName,
+			"created", totalCreated)
```
MikeSpreitzer: I would put this count in the unconditional log message below, rather than in a separate and conditional log message.

pkg/controller/launcher-populator/populator.go
```diff
@@ -365,54 +565,6 @@ func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, ke
...
-		if isBound {
-			logger.V(5).Info("Skipping deletion of launcher pod as it is bound to a server-requesting pod",
-				"pod", pod.Name, "server-requesting pod", requesterPodName)
-			continue
-		}
-
-		if err := ctl.coreclient.Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{
-			Preconditions: &metav1.Preconditions{
-				ResourceVersion: &pod.ResourceVersion,
-			},
```
MikeSpreitzer, line 385: I think that this belongs in the Delete calls that happen only when the Pod is unbound, because that condition can change.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +312,214 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+		nominalPod, hashErr := utils.BuildLauncherPodFromTemplate(
+			entry.LauncherConfig.Spec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
+		if hashErr != nil {
+			logger.Error(hashErr, "Failed to build nominal pod for hash comparison",
+				"node", nodeName, "config", key.LauncherConfigName)
+		} else if nominalPod.Annotations != nil {
+			nominalHash = nominalPod.Annotations[string(common.LauncherConfigHashAnnotationKey)]
+		}
+
+		// Categorize current pods: separate live unbound current-spec pods from stale/unbound ones
```
MikeSpreitzer, line 393: I think that the coding here for deletions is more complex than necessary. Rather than compute counts and collections up front, simply compute the target number to delete then iterate through the launchers, deleting launchers to which either criterion applies (unbound and wrong config, unbound and target not yet reached).

MikeSpreitzer (commented):
I finished another round of review.

pkg/controller/launcher-populator/populator.go
```diff
@@ -261,16 +261,28 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context, filter
...
+			// Read the LauncherConfig from informer's local cache to verify existence
+			// and get the current spec (needed for A3: spec-change detection)
+			lc, err := ctl.lcLister.LauncherConfigs(ctl.namespace).Get(countRule.LauncherConfigName)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					logger.Info("LauncherConfig referenced in policy does not exist, skipping",
+						"config", countRule.LauncherConfigName, "policy", lpp.Name)
+					continue
+				}
+				return nil, fmt.Errorf("failed to get LauncherConfig %s: %w", countRule.LauncherConfigName, err)
```
rubambiza, line 256: I am confused why were returning here without having processed:
- The count for each launcher
- The remaining policies

pkg/controller/launcher-populator/populator.go
```diff
@@ -301,62 +313,65 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+// It returns true if a requeue is needed (deletions were performed or are in progress),
+// so that creations happen only after deletions have taken effect.
+func (ctl *controller) reconcileAllLaunchers(ctx context.Context, desired map[NodeLauncherKey]DesiredStateEntry) (bool, error) {
 	logger := klog.FromContext(ctx)
 
-	// Group by node to optimize resource usage across all LauncherConfigs on each node
+	// Group by node to process each node separately
 	nodeGroups := make(map[string][]NodeLauncherKey)
 	for key := range desired {
 		nodeGroups[key.NodeName] = append(nodeGroups[key.NodeName], key)
```
rubambiza, line 316: What if `nodeGroups[key.NodeName]` does not exist yet?

pkg/controller/launcher-populator/populator.go
```diff
@@ -365,119 +380,146 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
...
+		}
+
+		// Delete stale pods immediately (spec changed → delete and replace)
+		for _, pod := range staleUnboundPods {
+			if err := ctl.coreclient.Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{}); err != nil {
+				if apierrors.IsNotFound(err) {
+					logger.Info("Stale launcher pod already deleted", "pod", pod.Name)
+					continue
+				}
+				return false, fmt.Errorf("failed to delete stale launcher pod %s: %w", pod.Name, err)
```
rubambiza, line 430: I'm curious, is the design intent here to return as soon as we encounter the first failed deletion with the hopes that the requeueing will capture the other pods in the next iteration?

In other words, why not just:
- Flag that one of the deletions failed
- Delete the rest that _can_ be deleted
- Retry the ones that _do_ exist (but deletion failed in the current iteration) later

Same comment applies for [line 465](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374/changes/b082cf9339e3b0617057a884fba0bb5c91e5e7eb..c6a515fb575c7f6b290b8fced42018a65c5a7f62#diff-4fa6d5c4dc86eff0054d35b46748081d11cd647e4463fcedff1bbb30a890ea74R465)

pkg/controller/launcher-populator/populator.go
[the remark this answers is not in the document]
rubambiza: Resolving this as it appears the creations now only proceed if there were no deletions.

pkg/controller/launcher-populator/populator.go
```diff
@@ -301,62 +313,65 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+		lc    *fmav1alpha1.LauncherConfig
 	}
-	var allCreations []creationInfo
+	var creations []creationInfo
 
-	// Analyze each LauncherConfig on this node
+	// Process each LauncherConfig on this node
 	for _, key := range keys {
-		desiredCount := desired[key]
+		entry := desired[key]
```
rubambiza, line 365: I can't quite put my finger on it, but it seems like there's duplication of information being passed through `keys` and `desired`, especially given the definition of [`NodeLauncherKey`](https://github.com/osswangxining/llm-d-fast-model-actuation/blob/f87ffe00e49d611b3cdb44a37bb788bfe329063f/pkg/controller/launcher-populator/node-launcher-key.go#L24). The only reasonable assumption I could make is if one data structure is more stable than the other between function calls. Otherwise, I think passing in the `nodeName` and `keys` seem sufficient here.

MikeSpreitzer:
@osswangxining:

>  I rebase

I see no rebase here. The first commit contributed here is from March 22.

image

pkg/controller/launcher-populator/node-launcher-key.go
```diff
@@ -30,8 +32,22 @@ func (k NodeLauncherKey) String() string {
 	return fmt.Sprintf("%s/%s", k.LauncherConfigName, k.NodeName)
 }
 
-// MapToLoggable converts a map of NodeLauncherKey to int32 values into a string representation.
-// This function formats the map as a string with the format "{namespace/name/node:count, ...}"
+// DesiredStateEntry holds the desired count and corresponding LauncherConfig
+// for a (Node, LauncherConfig) pair.
+type DesiredStateEntry struct {
+	Count          int32
+	LauncherConfig *fmav1alpha1.LauncherConfig
```
MikeSpreitzer: We do not need the whole LauncherConfig here. All that is really needed is the Spec.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: A failure to read a LauncherConfig object for any reason other than its absence is a transient failure of some basic infrastructure --- e.g., inability to even make requests to the apiserver at all. The right reaction to an infrastructure failure when trying to read one particular LauncherConfig is not to ignore that one but rather to try the whole ensemble all over again later.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: In that case the read of `nodeGroups[key.NodeName]` will evaluate to the value `nil` of the right slice type, and `append([]elttype(nil), something)` returns `[]elttype{something}`.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: I can put my finger on two bits of duplication here.
1. Every key in `keys` contains the node name already given separately in `nodeName`. I pointed out this duplication earlier, and @osswangxining said that he finds that it makes the code simpler overall.
2. Every `entry` contains a whole `LauncherConfig` object, including its `.Name` --- which is also in the `key`. I pointed out elsewhere that the `entry` only needs to hold the `LauncherConfigSpec`.

If this func is not given the `LauncherConfigSpec`s then it will have to read the `LauncherConfig` objects again, which would be bad because (a) it is unnecessary extra work, having just been done by the caller, and (b) requires adding complexity for dealing with the possibility that the second read returns a different result than the first (e.g., was present at first and is gone now).

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Calling BuildLauncherPodFromTemplate here is a lot of wasted work. I see that it is necessary because the only hash available takes the Node name into account. If BuildLauncherPodFromTemplate also returned an additional hash that does NOT take the Node name into account then, for purposes of this consistency checking, it would only be necessary to call BuildLauncherPodFromTemplate once per ResourceVersion of the LauncherConfig. This improvement could be made in a follow-on PR, but should be made.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Besides "not found", the errors that can arise here are concurrency conflicts and infrastructure failures. For both it is better to start over again later.

pkg/controller/launcher-populator/populator.go
osswangxining: Fine, I will replace with "Notified of add of object of ignored type". @rubambiza  FYI.

pkg/controller/launcher-populator/populator.go
osswangxining: sure

pkg/controller/launcher-populator/populator.go
osswangxining: make sense

pkg/controller/launcher-populator/populator.go
osswangxining: done  thanks

pkg/controller/launcher-populator/populator.go
osswangxining: append the comments.

pkg/controller/launcher-populator/populator.go
osswangxining: fine, changed the name.

pkg/controller/launcher-populator/populator.go
osswangxining: agree, updated.

pkg/controller/launcher-populator/populator.go
osswangxining: make sense

pkg/controller/launcher-populator/populator.go
osswangxining: agree

pkg/controller/launcher-populator/populator.go
osswangxining: make sense

pkg/controller/launcher-populator/populator.go
osswangxining: will be handled in #450

osswangxining:
> @osswangxining:
> 
> > I rebase
> 
> I see no rebase here. The first commit contributed here is from March 22.
> 
>  

I thought there were some conflicts before, while no any conflict so I continue to refine this file.

pkg/controller/launcher-populator/populator.go
osswangxining: Will be enhanced in  #451

pkg/controller/launcher-populator/populator.go
osswangxining: fine

pkg/controller/launcher-populator/populator.go
osswangxining: refine in the new commit.

osswangxining:
I've made improvements based on all the current comments. Please give it another review, thanks. @MikeSpreitzer @aavarghese @waltforme @rubambiza

pkg/controller/launcher-populator/populator.go
rubambiza: Thanks for the clarification Mike.

pkg/controller/launcher-populator/populator.go
```diff
@@ -539,23 +524,18 @@ func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLa
...
 		}
-		pod.GenerateName = fmt.Sprintf("launcher-%s-", launcherConfig.Name)
-		// Set owner reference pointing to LauncherConfig
-		ownerRef := *metav1.NewControllerRef(launcherConfig, fmav1alpha1.SchemeGroupVersion.WithKind("LauncherConfig"))
-		ownerRef.BlockOwnerDeletion = ptr.To(false)
-		pod.OwnerReferences = []metav1.OwnerReference{
-			ownerRef,
-		}
+		pod.GenerateName = fmt.Sprintf("launcher-%s-", key.LauncherConfigName)
+		pod.OwnerReferences = []metav1.OwnerReference{lcOwnerRef}
```
rubambiza, line 545: I see the introduction of `OwnerReferences` throughout the most recent changes. Is this to ensure the new launchers are owned by the same Object(?) as the old ones?

This is not a blocker, mostly for my own better understanding of k8s.

rubambiza (commented):
Leaving a clarifying comment that is not a blocker. Otherwise looks good to me. I'll leave approval to Mike once his comments are addressed. I, too, am interested in his [comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374/changes/c6a515fb575c7f6b290b8fced42018a65c5a7f62..276a8da97a6e3074c9e30b1f5914be625b8f4da8#r3124918889) on simplifying the process of deletion resolved.

pkg/controller/launcher-populator/node-launcher-key.go
```diff
@@ -30,8 +34,20 @@ func (k NodeLauncherKey) String() string {
...
+// DesiredStateEntry holds the desired count and the LauncherConfig spec
+// for a (Node, LauncherConfig) pair.
+type DesiredStateEntry struct {
+	Count              int32
+	LauncherConfigSpec fmav1alpha1.LauncherConfigSpec
+	LauncherConfigOwnerRef metav1.OwnerReference
+}
+
+func (e DesiredStateEntry) String() string {
+	return fmt.Sprintf("count=%d,config=%s", e.Count, e.LauncherConfigOwnerRef.Name)
```
MikeSpreitzer: This does not change when the LauncherConfigSpec changes. That's less than ideal.

pkg/controller/launcher-populator/populator.go
```diff
@@ -139,6 +147,10 @@ func (ctl *controller) OnUpdate(prev, obj any) {
 		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherPopulationPolicy reference due to notification of update", "name", typed.Name)
 		item := lppItem{cache.MetaObjectToName(typed)}
 		ctl.Queue.Add(item)
+	case *fmav1alpha1.LauncherConfig:
+		ctl.enqueueLogger.V(5).Info("Enqueuing LauncherConfig reference due to notification of update", "name", typed.Name)
+		item := lcItem{cache.MetaObjectToName(typed)}
+		ctl.Queue.Add(item)
 	default:
 		ctl.enqueueLogger.V(5).Info("Notified of update of type of ignored object", "type", fmt.Sprintf("%T", obj))
```
MikeSpreitzer: Did you want to update this log message too? And the one for delete?

pkg/controller/launcher-populator/populator.go
```diff
@@ -179,60 +190,91 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
...
-	// Build the PopulationPolicy map, storing the maximum count for each (Node, LauncherConfig) pair
-	populationPolicy := make(map[NodeLauncherKey]int32)
+	desired := make(map[NodeLauncherKey]DesiredStateEntry)
 	for _, lpp := range policies {
-		// Get matching nodes
 		nodes, err := ctl.getMatchingNodes(ctx, lpp.Spec.EnhancedNodeSelector)
 		if err != nil {
 			logger.Error(err, "Failed to get matching nodes for policy", "policy", lpp.Name)
-			return err, true
+			continue
```
MikeSpreitzer, line 243: If the error is from `metav1.LabelSelectorAsSelector(&selector.LabelSelector)` then this is a user error that should be reported in the `Status` of the `LauncherPopulationPolicy`. Can be done in a follow-on PR.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Oops, sorry, I goofed. This is not reading from the apiservers. It is reading from the informer's local cache. That should never suffer an error other than "not found", so it doesn't much matter how the impossible errors are handled.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +298,213 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
-	// Reconcile for each (Node, LauncherConfig) pair
-	for key, desiredCount := range desired {
-		if err := ctl.reconcileLaunchersOnNode(ctx, key, desiredCount); err != nil {
-			logger.Error(err, "Failed to reconcile launchers on node",
-				"node", key.NodeName,
-				"config", key.LauncherConfigName)
-			// Continue processing other combinations
+
+	// Group by node to process each node separately
+	nodeGroups := make(map[string][]NodeLauncherKey)
```
MikeSpreitzer, line 314: BTW, it is wasteful to have buildDesiredStateFromPolicies create a data structure that is then immediately reshaped. It would be better to have buildDesiredStateFromPolicies directly create the useful shape of data structure. Can be dealt with in a later PR.

pkg/controller/launcher-populator/node-launcher-key.go
```diff
@@ -30,8 +34,20 @@ func (k NodeLauncherKey) String() string {
 	return fmt.Sprintf("%s/%s", k.LauncherConfigName, k.NodeName)
 }
 
-// MapToLoggable converts a map of NodeLauncherKey to int32 values into a string representation.
-// This function formats the map as a string with the format "{namespace/name/node:count, ...}"
+// DesiredStateEntry holds the desired count and the LauncherConfig spec
+// for a (Node, LauncherConfig) pair.
+type DesiredStateEntry struct {
+	Count              int32
+	LauncherConfigSpec fmav1alpha1.LauncherConfigSpec
```
MikeSpreitzer: A LauncherConfigSpec is quite large (it includes inline a PodSpec). I would put a level of pointer indirection here.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +298,213 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
-	if err != nil {
-		return fmt.Errorf("failed to get current launchers: %w", err)
+
+	didDelete := false
+	deletionInProgress := false // tracks pods already being deleted (DeletionTimestamp set)
+
+	type creationInfo struct {
+		key    NodeLauncherKey
+		count  int
+		spec   fmav1alpha1.LauncherConfigSpec
```
MikeSpreitzer: A LauncherConfigSpec is quite large (it includes inline a PodSpec). I would put a level of pointer indirection here.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +298,213 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+		nominalHash := ""
+		nominalPod, err := utils.BuildLauncherPodFromTemplate(
+			entry.LauncherConfigSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
 		if err != nil {
-			return fmt.Errorf("failed to delete excess launchers: %w", err)
+			// The only error possible here is that the PodTemplate lacks an inference server container.
+			// In that case we proceed without a nominal hash, so no stale-pod detection occurs for this config.
+			logger.Error(err, "Failed to build nominal pod for hash comparison",
+				"node", nodeName, "config", key.LauncherConfigName)
+		} else if nominalPod.Annotations != nil {
```
MikeSpreitzer: You can delete the `if nominalPod.Annotations != nil` leaving just `} else {`.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +298,213 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+				"node", nodeName,
+				"config", key.LauncherConfigName)
+			didDelete = true
+		}
+
+		// Calculate diff based on effective remaining pods after stale deletion
+		effectiveRemaining := liveBoundCount + len(liveUnboundCurrentPods)
+		diff := entry.Count - int32(effectiveRemaining)
+
+		logger.Info("Analyzing config on node",
```
MikeSpreitzer: Did you mean "Analyzed"? Or "Adjusting"?

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +298,213 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+			"config", key.LauncherConfigName,
+			"current", effectiveRemaining,
+			"stale", len(staleUnboundPods),
+			"desired", entry.Count,
+			"diff", diff)
+
+		if diff < 0 {
+			// Need to delete excess pods from live unbound current pods
+			numToDelete := int(-diff)
+			for i := len(liveUnboundCurrentPods) - 1; i >= 0 && numToDelete > 0; i-- {
```
MikeSpreitzer, line 454: Why not the simpler enumeration from 0 upward?

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Oh, I see. The stale unbound ones should be deleted first and unconditionally.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +298,213 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+			logger.Error(err, "Failed to build nominal pod for hash comparison",
+				"node", nodeName, "config", key.LauncherConfigName)
+		} else if nominalPod.Annotations != nil {
+			nominalHash = nominalPod.Annotations[string(common.LauncherConfigHashAnnotationKey)]
+		}
+
+		// Categorize current pods: separate live unbound current-spec pods from stale/unbound ones
+		var liveBoundCount int
+		var liveUnboundCurrentPods []*corev1.Pod // live, unbound, spec matches current LauncherConfig
+		var staleUnboundPods []*corev1.Pod       // live, unbound, spec is stale
```
MikeSpreitzer, line 396: It would be simpler to delete these as encountered in the `for _, pod := range currentLaunchers` loop rather than collecting them and then immediately doing another loop to delete them. Can be addressed in follow-on.

pkg/controller/launcher-populator/populator.go
```diff
@@ -333,29 +524,18 @@ func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLa
...
 			return fmt.Errorf("failed to build launcher pod: %w", err)
 		}
-		pod.GenerateName = fmt.Sprintf("launcher-%s-", launcherConfig.Name)
-		// Set owner reference pointing to LauncherConfig
-		ownerRef := *metav1.NewControllerRef(launcherConfig, fmav1alpha1.SchemeGroupVersion.WithKind("LauncherConfig"))
-		ownerRef.BlockOwnerDeletion = ptr.To(false)
-		pod.OwnerReferences = []metav1.OwnerReference{
-			ownerRef,
-		}
+		pod.GenerateName = fmt.Sprintf("launcher-%s-", key.LauncherConfigName)
```
MikeSpreitzer: BuildLauncherPodFromTemplate already has responsibility for setting this field. There should not be two divergent settings for this field. Can be addressed in follow-on.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: It is positively wrong to create that OwnerReference by https://github.com/osswangxining/llm-d-fast-model-actuation/blob/276a8da97a6e3074c9e30b1f5914be625b8f4da8/pkg/controller/launcher-populator/populator.go#L264-L265 --- which sets the reference's `Controller` field to point to `true` --- and use it here. We do NOT want to delete all the launchers unconditionally if the LauncherConfig object is deleted.

The ownership of launcher Pods is complex. It is the result of the interplay among LauncherConfig, LauncherPopulationPolicy, and server-requesting Pod objects. I think that it is confusing to list just the LauncherConfig here. The launchers can only be correctly deleted by FMA controllers.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: I see that my 2nd bit was addressed.
The first was replied to earlier.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/node-launcher-key.go
osswangxining: OK, add the LauncherConfigSpec here: 	return fmt.Sprintf("count=%d,config=%s,spec=%+v", e.Count, e.LauncherConfigOwnerRef.Name, e.LauncherConfigSpec)

pkg/controller/launcher-populator/populator.go
osswangxining: sure, keep all using the same log format now

pkg/controller/launcher-populator/populator.go
osswangxining: OK  create new issue #456

pkg/controller/launcher-populator/node-launcher-key.go
osswangxining: fine, change to * pointer

pkg/controller/launcher-populator/populator.go
osswangxining: fine, change to * pointer

pkg/controller/launcher-populator/populator.go
osswangxining: Since BuildLauncherPodFromTemplate guarantees that Annotations are initialized upon success, the nil check is redundant.

pkg/controller/launcher-populator/populator.go
osswangxining: oh,  Analyzed

pkg/controller/launcher-populator/populator.go
osswangxining: The reverse iteration is intentional: it deletes the oldest unbound pods (which tend to appear earlier in the list returned by the lister), preserving the most recently created ones.

pkg/controller/launcher-populator/populator.go
osswangxining: I noticed that BuildLauncherPodFromTemplate  will create the pod.GenerateName = "launcher-" instead of missing LauncherConfigName, I think that LauncherConfigName is better if have. Do you think so?

pkg/controller/launcher-populator/populator.go
osswangxining: Agreed, the two loops could be merged. However, the current structure intentionally separates classification from action for readability, and len(staleUnboundPods) is reused in the log below. Happy to consolidate in a follow-on as suggested.

pkg/controller/launcher-populator/populator.go
osswangxining: Agreed, the reshaping is redundant. The current separation keeps buildDesiredStateFromPolicies focused on policy evaluation with a clean composite key, while the grouping is a reconciliation concern. The reshaping cost is negligible, but consolidating would reduce indirection. Happy to address in a follow-on PR as suggested.

pkg/controller/launcher-populator/populator.go
osswangxining: Make sense, this is one bug, I fixed this.

pkg/controller/launcher-populator/populator.go
osswangxining: Record in #457

osswangxining:
Thanks for your further review ! I've made improvements based on the comments above.  @MikeSpreitzer  @aavarghese @waltforme @rubambiza

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Yes, I think that the setting here for `pod.GenerateName` is better. However, you do need to pay attention to one thing: in Kubernetes there is a limit on how long a name can be. I know that the generation process will make some adaptation if it would normally produce a name that is too long, but I am not clear on whether it is an error to submit a `GenerateName` that is too long.

The right way to make this improvement is to change the assignment in BuildLauncherPodFromTemplate.

BTW, @rubambiza @dumb0002, our tests tend to put "launcher config" in the start of the name of the LauncherConfig objects, which (a) is redundant, because an API object's `kind` already says its kind and there is no name confusion between objects of different kinds, and (b) make the computation here (`fmt.Sprintf("launcher-%s-", key.LauncherConfigName)`) produce a name that stutters.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: I think this answer is worth documenting, with a comment here or in a document. The question of which launchers to delete is an interesting one. Can be addressed in a later PR.

pkg/controller/launcher-populator/populator.go
aavarghese: Will you be changing this @osswangxining ?

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: We want FMA to be usable in a large cluster, so the reshaping could be costly.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Yes, and it is also redundant because of Go language semantics: `aMap[something]`, when `aMap == nil`, will not err and will evaluate to the same thing it would if `aMap != nil && len(aMap) == 0`.

pkg/controller/launcher-populator/populator.go
osswangxining: @aavarghese  Yes, in fact that I already updated the codes, but forgot to submit,  and sorry for this and submitted the latsest codes

pkg/controller/launcher-populator/populator.go
osswangxining: @MikeSpreitzer   agree, I updated the assignment in BuildLauncherPodFromTemplate. Please review this change @waltforme @MikeSpreitzer @aavarghese @rubambiza

pkg/controller/launcher-populator/populator.go
osswangxining: sure. I mark it #459

pkg/controller/launcher-populator/populator.go
osswangxining: OK. will be resolved in #460

osswangxining:
@MikeSpreitzer @rubambiza @aavarghese @waltforme  Thanks for all your review.  I updated the codes almost except for the following issues to be resolved separately:
#450 
#451  
#456 
#457 
#459 
#460

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +304,215 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+					continue
+				}
+			}
+
+			liveUnboundCurrentPods = append(liveUnboundCurrentPods, pod)
+		}
+
+		// Delete stale pods immediately (spec changed → delete and replace)
+		for _, pod := range staleUnboundPods {
+			if err := ctl.coreclient.Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{}); err != nil {
```
MikeSpreitzer, line 425: The `metav1.DeleteOptions{}` here should include a test on `ResourceVersion`, to avoid deleting a Pod that has been bound since the read that cause it to go into `staleUnboundPods`.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +304,215 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+				if apierrors.IsNotFound(err) {
+					logger.Info("Stale launcher pod already deleted", "pod", pod.Name)
+					continue
+				}
+				return false, fmt.Errorf("failed to delete stale launcher pod %s: %w", pod.Name, err)
```
MikeSpreitzer, line 430: This `return false, <some error>` causes the caller to skip processing this Node and move on to the next. But a retry is the right response.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,64 +304,215 @@ func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha
...
+			numToDelete := int(-diff)
+			for i := len(liveUnboundCurrentPods) - 1; i >= 0 && numToDelete > 0; i-- {
+				pod := liveUnboundCurrentPods[i]
+				if err := ctl.coreclient.Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{}); err != nil {
+					if apierrors.IsNotFound(err) {
+						logger.Info("Launcher pod already deleted", "pod", pod.Name)
+						numToDelete--
+						continue
+					}
+					return false, fmt.Errorf("failed to delete launcher pod %s: %w", pod.Name, err)
```
MikeSpreitzer, line 462: As noted above, returning `false, <some error>` here fails to cause a retry. This is risky. One failure mode in a distributed system is: client sends a request, server receives it and handles it, but connection is broken before reply is returned. That can happen here. So the launcher might actually get deleted.

pkg/controller/utils/pod-helper.go
```diff
@@ -147,7 +147,7 @@ func BuildLauncherPodFromTemplate(template corev1.PodTemplateSpec, ns, nodeName,
 		Spec:       *DeIndividualize(template.Spec.DeepCopy()),
 	}
 	pod.Namespace = ns
-	pod.GenerateName = "launcher-"
+	pod.GenerateName = fmt.Sprintf("launcher-%s-", launcherConfigName)
```
MikeSpreitzer, line 150: Is there a length limit on `pod.GenerateName` ?

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: See #456

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: I see that this PR now proposes to add an OwnerReference whose `.Controller` points to `false`. The clear functional bug has been removed. However, I still think that it is confusing to mark the LauncherConfig as an owner but not also the relevant LauncherPopulationPolicy objects and server-requesting Pod objects. All three kinds of objects are relevant to the determination that this launcher should exist.

I also think that it would be unnecessary latency and code complexity to do that, since the collection of those other objects can change over time. So, in my opinion, it is best to simply not list any owners.

MikeSpreitzer (approved):
I think that there are still real bugs that need to be fixed,
but I think that we can do that in follow-on PRs (some of which are urgent,
the real bugs are sources of flakiness).

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: See #461

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: See #462

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: See #462

pkg/controller/utils/pod-helper.go
MikeSpreitzer: I asked: https://kubernetes.slack.com/archives/C0EG7JC6T/p1777400655256509

pkg/controller/launcher-populator/populator.go
osswangxining: Thanks  #461  merged

PR #381, opened 2026-03-25
docs: Update benchmark scenarios matrix and integration strategy

rubambiza:
## Summary
- Add layered measurement model (L1 actuation, L2 inference readiness, L3 steady-state) with metric definitions using precise FMA terminology (requester, launcher, DPC)
- Populate 4x3 benchmarking matrix: 4 user-story scenarios (Fast Replica Scale Up, Introducing New Variant, Resource Request Justification, Maintenance Planning) x 3 actuation paths (Cold Start, Warm Start, Hot Start)
- Adopt the team's hot/warm/cold start taxonomy for actuation paths
- Add scenario rationale and actuation path rationale tables for team discussion
- Note caching and simulation as orthogonal dimensions rather than matrix columns
- Document llm-d-benchmark integration strategy via the [fma branch](https://github.com/manoelmarques/llm-d-benchmark/tree/fma) and reference [WVA PR #900](https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/900)
- Mention WVA integration in the Purpose section
- Extract legacy `benchmark_base.py` docs to `benchmark_legacy.md`

## Test plan
- [x] Verify markdown renders correctly on GitHub (tables, links, collapsible section)
- [x] Cross-check llm-d-benchmark references against the `fma` branch
- [x] Review matrix cell annotations (L1, L1+L2, P, --) for accuracy


inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
+| **L2: Inference Readiness** | First inference response | T_first_token (requester ready to first inference response), T_e2e (requester creation to first inference response) | llm-d-benchmark nop/inference-perf harness |
+| **L3: Steady-State** | Throughput/latency | T_actuation (requester creation to readiness), TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
+
+**Metric definitions:**
+
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving.
+- **T_wake**: Time from the DPC sending `/wake_up` to a sleeping vLLM instance on the server-providing pod to that instance reporting ready to serve. A part of T_actuation when a GPU hit occurs.
+- **Hit_rate**: Fraction of requesters that get bound to an existing sleeping pod on the correct GPU (hit) vs. requiring a cold start (miss).
+- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading.
```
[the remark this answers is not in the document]
aavarghese: Ideally we should not have a case where "DPC requests launcher creation". The launcher populator should have done its job beforehand. Hence why "Time from the launcher receiving a create request". 
We can change "to the new vLLM instance reporting healthy" to "to the launcher being Ready"

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
+| **L2: Inference Readiness** | First inference response | T_first_token (requester ready to first inference response), T_e2e (requester creation to first inference response) | llm-d-benchmark nop/inference-perf harness |
+| **L3: Steady-State** | Throughput/latency | T_actuation (requester creation to readiness), TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
+
+**Metric definitions:**
+
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving.
+- **T_wake**: Time from the DPC sending `/wake_up` to a sleeping vLLM instance on the server-providing pod to that instance reporting ready to serve. A part of T_actuation when a GPU hit occurs.
+- **Hit_rate**: Fraction of requesters that get bound to an existing sleeping pod on the correct GPU (hit) vs. requiring a cold start (miss).
```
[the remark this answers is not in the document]
aavarghese: A cold start (miss) could be both 2. and 3. where 3. is less likely when launcher-populator has done its job and will be part of initial bootstrap.

inference_server/benchmark/benchmark.md
MikeSpreitzer: No "etc" needed, there are three actuation paths.

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
 
-### Next steps
+| Actuation Path                          | What It Measures | llm-d-benchmark Config |
+| --------------------------------------- | ---------------- | ---------------------- |
+| **Cold Start (Standalone)**             | No launcher, no sleeping pods. Raw Kubernetes deploy-to-ready latency. Baseline for all other paths. | `-t standalone` comparison baseline |
+| **Cold Start (w/ Launcher)**            | Launcher is present and pre-loads vLLM Python modules. Measures the launcher's contribution to reducing cold start latency vs. standalone. | `-t fma` with default `LLMDBENCH_FMA_LAUNCHER_*` env vars |
+| **Wake Sleeping Instance**              | A sleeping vLLM instance exists on the correct GPU. DPC sends `/wake_up`. Measures the best-case actuation path. | `-t fma` with `LLMDBENCH_VLLM_COMMON_ENABLE_SLEEP_MODE=true`, `LLMDBENCH_FMA_DUAL_POD_SLEEPER_LIMIT>=1` |
+| **Create Instance (Existing Launcher)** | No sleeping instance available on the assigned GPU, but a launcher pod exists. Launcher creates a new vLLM instance. | `-t fma`, no sleeping instances pre-provisioned |
```
MikeSpreitzer: To me this seems the same as "Cold Start (w/ Launcher)".

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
-| **Resource Request Justification** |                      |                           |                       |                       |                     |                                 |
-| **Maintenance Planning**      |                           |                           |                       |                       |                     |                                 |
+### Actuation Paths (columns)
 
+The columns represent the different paths FMA can take to satisfy a new server-requesting pod:
 
-### Next steps
+| Actuation Path                          | What It Measures | llm-d-benchmark Config |
+| --------------------------------------- | ---------------- | ---------------------- |
+| **Cold Start (Standalone)**             | No launcher, no sleeping pods. Raw Kubernetes deploy-to-ready latency. Baseline for all other paths. | `-t standalone` comparison baseline |
```
MikeSpreitzer: Is this

(A) the non-FMA actuation where the model deployer simply deploys a Pod that runs vllm

or

(B) the slowest FMA actuation, which involves both launcher Pod creation and then vllm instance creation under that launcher?

(I think we want to measure both.)

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
+| **Wake Sleeping Instance**              | A sleeping vLLM instance exists on the correct GPU. DPC sends `/wake_up`. Measures the best-case actuation path. | `-t fma` with `LLMDBENCH_VLLM_COMMON_ENABLE_SLEEP_MODE=true`, `LLMDBENCH_FMA_DUAL_POD_SLEEPER_LIMIT>=1` |
+| **Create Instance (Existing Launcher)** | No sleeping instance available on the assigned GPU, but a launcher pod exists. Launcher creates a new vLLM instance. | `-t fma`, no sleeping instances pre-provisioned |
+| **Model Swap (Launcher)**               | Launcher swaps the model on an existing vLLM instance without restarting the process. Tests the launcher's dynamic model management capability. | `-t fma`, sequential model deployment via `07_deploy_fma_models.py` |
+| **Cached Model (PVC)**                  | Model weights pre-cached on a PersistentVolumeClaim, eliminating download time. Isolates the non-download portion of actuation latency. | `-t fma` with `LLMDBENCH_VLLM_COMMON_EXTRA_PVC_NAME` + `LLMDBENCH_VLLM_COMMON_VLLM_CACHE_ROOT` (configured in `fma.sh`) |
```
MikeSpreitzer: There are two things that might be cached, independently: the model tensors, and the CUDA graph compilation.

Caching is not one actuation path, it is a potential aspect of every actuation path besides "Wake Sleeping Instance".

inference_server/benchmark/benchmark.md
rubambiza: Fixed. The Purpose now explicitly lists the three actuation paths in decreasing latency order and mentions WVA integration.

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,139 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
 
 ## Purpose
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under different actuation conditions such
-as cold starts, wake-ups from a sleeping state, using prewarmed pods, etc. These metrics
-will guide future optimizations for the **Dual-Pods Controller (DPC)**. Ultimately, the goal
+and server-providing pods), when integrated with the Workload Variant Autoscaler (WVA),
```
MikeSpreitzer: I do not think that this is where WVA comes in. The L1, L2, and L3 metrics do not involve WVA. I think that the reasons to talk about WVA are:
- some Benchmarking Scenarios(s) might involve WVA (I am not actually sure about this),
- we want to demonstrate FMA working with WVA, and
- we view WVA as pioneering the 2nd generation of benchmarking and think that is where FMA benchmarking should appear (because the 1st generation benchmark framework is unable to conceive of FMA).

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
+| **L2: Inference Readiness** | First inference response | T_e2e (requester creation to first inference response), T_first_token (requester ready to first inference response) | llm-d-benchmark nop/inference-perf harness |
+| **L3: Steady-State** | Throughput/latency | T_actuation (requester creation to readiness), TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
+
+**Metric definitions:**
+
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving.
+- **T_wake**: Time from the DPC sending `/wake_up` to a sleeping vLLM instance on the server-providing pod to that instance reporting ready to serve. A part of T_actuation when a GPU hit occurs.
+- **Hit_rate**: Fraction of requesters that get bound to an existing sleeping pod on the correct GPU (hit) vs. requiring a cold start (i.e., new vLLM instance in existing launcher pod or new launcher pod + new vLLM instance).
```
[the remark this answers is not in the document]
rubambiza: Thanks for catching it. Oversight between editing in vim and with Claude. Fixed -- Hit_rate now reads: "Fraction of server-requesting Pods that get satisfied by waking a sleeping vLLM instance."

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
-- Validate improvements across llm-d releases and DPC changes.
+Cell annotations indicate which measurement layers apply:
+- **L1** -- Layer 1 actuation metrics (T_actuation, T_wake, Hit_rate, T_launcher)
+- **L1+L2** -- Actuation metrics plus inference readiness (T_first_token, T_e2e)
+- **P** -- Planned but not yet implemented
+- **--** -- Not applicable to this combination
+
+| Scenario                           | Cold Start (Standalone) | Cold Start (Launcher) | Wake Sleeping Instance | Create Instance (Existing Launcher) | Model Swap (Launcher) | Cached Model (PVC) |
+| ---------------------------------- | :---------------------: | :-------------------: | :--------------------: | :---------------------------------: | :-------------------: | :-----------------: |
+| **Fast Replica Scale Up**          | L1                      | L1+L2                 | L1+L2                  | L1+L2                               | --                    | L1+L2               |
```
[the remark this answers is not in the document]
rubambiza: Good call out. I've added L3 to the last two scenarios: Resource Scaling and Stress Test (L1+L3, since it involves sustained load for performance/cost characterization) and Maintenance Planning (L1+L2+L3, since it compares steady-state performance before/after maintenance).

For the other two: Fast Replica Scale Up could include L3 if we run sustained inference after actuation to measure steady-state under the new replica count, and Introducing New Variant is similar. But that starts to blur the line between "FMA benchmark" and "regular inference benchmark that happens to start with FMA actuation," so I'm keeping those focused on L1/L2 for now.

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
+| **Introducing New Variant**        | L1                      | L1+L2                 | --                     | L1+L2                               | L1+L2                 | L1+L2               |
+| **Free Up Cluster Resources**      | --                      | P                     | P                      | --                                  | P                     | --                  |
+| **Resource Request Justification** | L1                      | L1                    | L1                     | L1                                  | P                     | L1                  |
```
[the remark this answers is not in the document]
rubambiza: Fair enough. Renamed the scenario to "Resource Scaling and Stress Test" with the description: "As a Workload Owner, I can stress-test my namespace's resource capacity at scale (many models, many requesters) for numerous purposes (e.g., management reports, load characterization, future resource requests)."

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,109 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
+
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving.
+- **T_wake**: Time from the DPC instructing the launcher to wake a sleeping vLLM instance on the server-providing pod to that instance reporting ready to serve. A subset of T_actuation when a GPU hit occurs.
+- **Hit_rate**: Percentage of scale-up events where the DPC binds a requester to an existing sleeping pod on the correct GPU (hit) vs. requiring a cold start (miss).
+- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading.
+- **T_first_token**: Time from requester pod readiness to first successful inference response received through the server-providing pod's vLLM instance (time-to-first-token, post-actuation).
+- **T_e2e**: Total time from requester pod creation to first successful inference response. Spans the full path: requester scheduling, DPC binding, launcher wake/create, vLLM ready, first inference (T_actuation + T_first_token).
```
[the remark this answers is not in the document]
rubambiza: Just to clarify, you are effectively arguing for a fourth column (let's call it "luke warm start" for now), right? I want to confirm because it would require more rethinking for the matrix.

Re: instance deletion, thanks for the clarification.

inference_server/benchmark/benchmark.md
```diff
@@ -11,103 +11,127 @@ is *high predictability*, which is defined as achieving close to 100% hit rate o
...
+
+FMA benchmarking uses a layered measurement model. Layer 1 is what FMA benchmarks own
+directly. Layer 2 bridges actuation to inference readiness. Layer 3 is out of FMA's
+direct scope but is referenced for completeness and handoff to other frameworks.
+
+| Layer | Focus | Metrics | Measured By |
+| ----- | ----- | ------- | ----------- |
+| **L1: Actuation** | Requester pod readiness | T_actuation (requester creation to readiness), T_wake (DPC wakes sleeping vLLM instance), Hit_rate (GPU hits), T_launcher (launcher creates new vLLM instance) | llm-d-benchmark new harness |
+| **L2: Inference Readiness** | First inference response | T_first_token (requester ready to first inference response), T_e2e (requester creation to first inference response) | llm-d-benchmark nop/inference-perf harness |
+| **L3: Steady-State** | Throughput/latency | T_actuation (requester creation to readiness), TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
```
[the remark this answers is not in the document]
rubambiza: As stated in my previous comment, T_actuation is the bridge that we care most about. It'\''s the one metric that FMA uniquely contributes to the L3 picture, letting WVA and llm-d-benchmark consumers see how actuation latency affects their steady-state metrics. The other L1/L2 metrics are either FMA-internal (T_wake, Hit_rate, T_launcher) or already captured by L3 tools in their own terms (TTFT maps to T_first_token).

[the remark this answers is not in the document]
MikeSpreitzer: I am not clear on what sort of "bridge" you are talking about. The table already shows T_activation being applicable to every scenario due to L1 being applicable. If L1 and L3 both apply to a scenario, then why duplicate T_activation in L3? If L3 does not apply to a scenario but we care about throughput metrics for that scenario, those words seem mutually inconsistent to me.

inference_server/benchmark/benchmark.md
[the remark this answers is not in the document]
MikeSpreitzer: Yes, I am arguing for measuring a fourth activation path. It is very similar to the one currently called "cold start", but involves FMA. So maybe replace the one "cold start" path with two: "cold start without FMA" and "cold start through FMA".

inference_server/benchmark/benchmark.md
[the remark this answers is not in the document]
rubambiza: Agreed that it's confusing. I have [removed](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/381/changes/a4dcd00282fd8f2421a91e5c6bb4a6a3ce48746e) actuation from L3

inference_server/benchmark/benchmark.md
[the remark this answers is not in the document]
rubambiza: I considered "cold start without/through FMA". However, it becomes a bit confusing because "cold start through FMA" could also cover the cover the case of M2 FMA. Therefore, I stuck with lukewarm for the path we are describing. Updated throughout the doc in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/381/changes/a2754d4712e63dbe9e5ced52a4e0544cad6c2902.

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,208 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
 ## Purpose
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under different actuation conditions such
-as cold starts, wake-ups from a sleeping state, using prewarmed pods, etc. These metrics
-will guide future optimizations for the **Dual-Pods Controller (DPC)**. Ultimately, the goal
+and server-providing pods) becomes available under four different actuation conditions
+in order of decreasing latency:
+
+- **Cold start**: creating a new vLLM instance without using a launcher
+- **Luke warm start**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
```
MikeSpreitzer, line 11: "lukewarm" is one word.

MikeSpreitzer, line 11: This actuation path is not warmed at all. It is a longer cold start path. For my taste, it would make more sense to define "cold start without FMA" and "cold start through FMA" than to define "cold start" (to mean without FMA) and "lukewarm start" (to be just as cold but also involving a launcher).

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,208 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
+
+**Metric definitions:**
+
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or luke warm start (T_luke_warm).
+- **T_wake**: Request-response time for the DPC's `/wake_up` call to a sleeping vLLM instance on the server-providing pod. A part of T_actuation when a hot start occurs.
+- **Hit_rate**: Fraction of server-requesting Pods that get satisfied by waking a sleeping vLLM instance.
+- **T_luke_warm**: Time from the DPC requesting launcher pod creation to the new vLLM instance reporting healthy. Covers the full luke warm start span: launcher pod scheduling, launcher readiness, DPC reconciliation, and vLLM instance creation. Measured end-to-end because the boundary between launcher readiness and instance creation is not directly observable from outside the DPC.
```
MikeSpreitzer, line 43: It is a little odd to call out "DPC reconciliation" in this list of things that are mostly one-shot actions. This path goes through the DPC multiple times.

MikeSpreitzer, line 43: This list ("launcher pod scheduling, launcher readiness, DPC reconciliation, and vLLM instance creation") consists of things that are grammatical actions except for one: "launcher readiness". I suppose that an action is meant there, something like "launcher startup".

MikeSpreitzer, line 43: > Measured end-to-end because the boundary between launcher readiness and instance creation is not directly observable from outside the DPC.

I am not sure what this sentence is trying to say.
I think maybe this is trying to say that this metric is calculated as the difference between the initial and final timestamp rather than as a sum of constituent durations because one of them can't be observed? That sounds like a problem to solved, because the constituents are each important to measure. Can you say more about exactly what the problem is? I am not clear on what sort of observations you are wiling to consider here (log message understanding, Kubernetes object creation timestamps, Kubernetes watch notifications, Prometheus metrics, ...).

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,208 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
+- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to the warm start path, where a launcher pod already exists.
```
MikeSpreitzer, line 44: "Time from the launcher receiving a create request to the new vLLM instance reporting healthy" is relevant to lukewarm start too.

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,208 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
+## Benchmarking Matrix
+
+### Actuation Paths (columns)
+
+The columns represent the different paths FMA can take to satisfy a new server-requesting pod,
+using the team's established terminology:
+
+| Actuation Path            | What It Measures | Why Included | llm-d-benchmark Config |
+| ------------------------- | ---------------- | ------------ | ---------------------- |
+| **Cold Start**            | No launcher, no sleeping pods. Raw Kubernetes deploy-to-ready latency (non-FMA baseline, or FMA milestone 2 without launcher). | Establishes the baseline that all FMA paths should improve upon. | `-t standalone` comparison baseline |
```
MikeSpreitzer, line 67: The no-FMA path is different from the FMA-milestone-2-cold path. It is wrong to treat them as a single path; they are two different paths. If you want to skip benchmarking the milestone 2 cold path, that would be fine by me (because we are planning to drop milestone 2 eventually and only advocate milestone 3).

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,208 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
+| **Luke Warm Start**       | No launcher pod on the assigned GPU. The DPC creates a new launcher pod, then the launcher creates a new vLLM instance. | Worst-case FMA path, relevant for dynamic situations such as LauncherConfig rollouts or newly added nodes where the LPC has not yet populated launchers. | `-t fma` with LauncherPopulationPolicy that does not cover the target GPU node |
```
MikeSpreitzer, line 68: A launcher is not "on" a GPU. A launcher is on a Node, and has constraints on what more it can do. I think that rather than "No launcher pod on the assigned GPU", it would be better to simply say "no suitable launcher Pod exists at first".

MikeSpreitzer, line 68: "then the launcher creates a new vLLM instance" is not wrong, but perhaps "then DPC commands the launcher to create a new vLLM instance" would shed more light.

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,208 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
+| **Warm Start**            | A launcher pod exists (pre-created by the LPC) but no sleeping instance is available on the assigned GPU. Launcher creates a new vLLM instance with module preloading. | Measures the launcher's contribution when no sleeping instance is available. | `-t fma` with default `LLMDBENCH_FMA_LAUNCHER_*` env vars |
```
MikeSpreitzer, line 69: Re "(pre-created by the LPC)": the DPC code does not care whether that pre-existing launcher Pod was created by the LPC or was created earlier by the DPC (and was used for a while and then the awake vllm instance on it was put to sleep). If you intend specifically to benchmark only the first of those two then it is right to be that specific, but you might want to be clear that you are deliberately not trying the other path. Or, if you do not intend to be that specific, then do not specify one but not the other.

inference_server/benchmark/benchmark.md
```diff
@@ -4,110 +4,208 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
+- **L1+L2** -- Actuation metrics plus inference readiness (T_first_token, T_e2e)
+- **L1+L3** -- Actuation metrics plus steady-state performance (TPOT, throughput, queue depth, KV cache, replica stability)
+- **L1+L2+L3** -- All three layers
+- **--** -- Not applicable to this combination
+
+| Scenario                           | Cold Start | Luke Warm Start | Warm Start | Hot Start |
+| ---------------------------------- | :--------: | :-------------: | :--------: | :-------: |
+| **Fast Replica Scale Up**          | L1+L2      | L1+L2           | L1+L2      | L1+L2     |
+| **Introducing New Variant**        | L1+L2      | L1+L2           | L1+L2      | --        |
+| **Resource Scaling and Stress Test** | L1+L3    | L1+L3           | L1+L3      | L1+L3     |
```
MikeSpreitzer, line 99: I do not understand why L1 is relevant in this row but L2 is not. Both L1 and L2 are about latency at the start of an instance's life.

inference_server/benchmark/benchmark.md
MikeSpreitzer: If you are concerned about the milestone 2 vs. milestone 3 ambiguity from saying just "FMA" then say "launcher" instead. Perhaps define the two paths as "cold start without FMA" and "cold start launcher + vllm".

aavarghese:
As I was writing up [this comment](https://github.com/llm-d/llm-d-benchmark/pull/900#issuecomment-4239575612), for the llm-d benchmarking work to get the different metrics to merge, I noticed it was hard to follow the L1 metrics we have. Does it make sense to change:
T_wake ----> T_hot
T_launcher ---> T_warm?

We already have T_luke_warm.

MikeSpreitzer:
> Does it make sense to change:
> T_wake ----> T_hot
> T_launcher ---> T_warm?

Those are not T_actuation for the different actuation paths. Those are timings for smaller, constituent steps. See their definitions at https://github.com/rubambiza/llm-d-fast-model-actuation/blob/a4dcd00282fd8f2421a91e5c6bb4a6a3ce48746e/inference_server/benchmark/benchmark.md?plain=1#L41 and https://github.com/rubambiza/llm-d-fast-model-actuation/blob/a4dcd00282fd8f2421a91e5c6bb4a6a3ce48746e/inference_server/benchmark/benchmark.md?plain=1#L44

T_luke_warm is also not an end-to-end metric, although it does roll up a few constituent pieces. See its definition at https://github.com/rubambiza/llm-d-fast-model-actuation/blob/a4dcd00282fd8f2421a91e5c6bb4a6a3ce48746e/inference_server/benchmark/benchmark.md?plain=1#L43

(The commit cited above is the current revision of this PR.)

A little more broadly, I see two sorts of L1 thing to do that are more specific than T_actuation: (1) break it down into constituent pieces, each covering part of the end-to-end path, and (2) for each actuation path, a version of T_actuation that applies only to that path (but still the whole end-to-end path).

MikeSpreitzer (approved):
This needs more work, and it can be done in follow-on PRs.

inference_server/benchmark/benchmark.md
rubambiza: Fixed! All `lukewarm` instances have been replaced with `cold start (with launcher)`

rubambiza: Addressed! The two cold start paths are now: (`cold start without FMA` and `cold start launcher + vllm)` -> As of 4/23

As of 4/24, we now have three cold start paths. https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454/changes#diff-26a1ffde90bb76e53782af51f5443fe2c2b314250c5d19532a72d0d9136caf90R86

inference_server/benchmark/benchmark.md
rubambiza: Fixed.

- Fixed typo for lukewarm -> `T_lukewarm`
- Rephrased `full lukewarm start` to `full cold start launcher + vllm span`
- Removed `DPC reconciliation` from list

rubambiza: Fixed.`launcher readiness` -> [`launcher startup`](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454/changes#diff-26a1ffde90bb76e53782af51f5443fe2c2b314250c5d19532a72d0d9136caf90R44)

rubambiza: What I was trying to convey is that some of the consistent durations cannot be observed from watching Kube objects/watch notifications only. You're right that it is a problem to be solved, and a combination of (log message understanding, Kubernetes object creation timestamps, Kubernetes watch notifications, Prometheus metrics) can get us there. I also agree that it's important to measure the constituents durations.

I'll need to think through the fidelity of some of them (e.g., delay in retrieving/parsing DPC logs for some of the computations) and where the benchmarking code is today. For now, I have added the [constituent durations as new metrics](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454/changes#diff-26a1ffde90bb76e53782af51f5443fe2c2b314250c5d19532a72d0d9136caf90R49) planned, and I have removed the sentence stating that they are not observable.

inference_server/benchmark/benchmark.md
rubambiza: Fixed. Rephrased `Applies to the warm start path, where a launcher pod already exists` -> `Applies to both cold start launcher + vllm and warm start paths`

inference_server/benchmark/benchmark.md
rubambiza: I have added the FMA milestone 2 benchmarking as a [new `planned` actuation path](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454/changes#diff-26a1ffde90bb76e53782af51f5443fe2c2b314250c5d19532a72d0d9136caf90R87) for now. Given the divergent nature of components:
- Current llm-d benchmark is entirely M3-oriented (deploys ISC, LauncherConfig, LPP, and requesting pods with the inference-server-config annotation)
- Some complexity in constructing correct server-patches + GPU mapping circa M2

It would take a bit of work to add M2 as part of the harness. Claude estimates it to be 2-3 days worth of work. Still, I want to pursue it because I am convinced that it solves one of our core challenges, and it could be an interesting ablation study for a paper.

inference_server/benchmark/benchmark.md
rubambiza: Thanks for the suggestion. Fixed!

rubambiza: Thanks for the suggestion. Fixed!

inference_server/benchmark/benchmark.md
rubambiza: I think the `no sleeping instance is available on the assigned GPU` part of the description rules out the ambiguity as you stated `was created earlier by the DPC (and was used for a while and then the awake vllm instance on it was put to sleep)`. In other words, if the launcher was created by the DPC and put to sleep, then it is not the kind pre-created by the LPC. Still, I have taken your suggestion and removed the `pre-created by the LPC` part for conciseness.

inference_server/benchmark/benchmark.md
rubambiza: This was a judgement call because the primary concern for that scenario is how fast all requesters come up at scale and how the system performs under stress, not the latency of the inference response for each individual requester. However, you do make a good point that L2 is also about latency. More importantly, the TTFT request is low compared to the total of L1 metrics. I have added L2 in. Thanks!

inference_server/benchmark/benchmark.md
rubambiza: On second thought, I have decided to remove the milestone 2 cold path altogether.

PR #413, opened 2026-04-06
Add watcher as sentinel in launcher

### LINKED ISSUE #378: Add watch endpoint for launcher instance lifecycle events
## Summary

We want the launcher to proactively detect when vLLM instances terminate and expose a streaming API for clients to observe these events in real time (Kubernetes watch-style).

## Scope

This issue covers only **instance termination detection** — when a vLLM child process exits for any reason (clean exit, crash, SIGKILL, etc.).

## Detection Mechanism

Python's `multiprocessing.Process.sentinel` provides a file descriptor that the OS kernel makes readable when the child process terminates. We register it on the asyncio event loop for immediate, zero-polling detection. No cooperation from the child process is needed — the kernel handles it.

## Watch Endpoint

`GET /v2/vllm/instances/watch` returns a chunked NDJSON stream. An internal `EventBroadcaster` (using `asyncio.Condition`) fans out lifecycle events to all connected watchers.

Each event is a JSON line:
```json
{"type": "CREATED", "object": {"instance_id": "abc", "status": "started"}}
{"type": "STOPPED", "object": {"instance_id": "abc", "status": "stopped", "exit_code": -9}}
{"type": "DELETED", "object": {"instance_id": "abc", "status": "terminated"}}
```

### Event Types

- `CREATED` — new instance spawned
- `STOPPED` — instance process terminated (detected via sentinel)
- `DELETED` — instance explicitly removed via DELETE API

## Implementation Components

1. **Sentinel watcher** — an `asyncio.Task` per instance that registers `Process.sentinel` on the event loop. When the fd becomes ready, it publishes a `STOPPED` event to the broadcaster.

2. **`EventBroadcaster`** — internal pub/sub using `asyncio.Condition` that fans out events to all connected watchers. Each watcher maintains its own cursor.

3. **Watch endpoint** — `GET /v2/vllm/instances/watch` returns a `StreamingResponse` with NDJSON. Client disconnection is handled automatically by FastAPI cancelling the async generator.

cc. @MikeSpreitzer @waltforme

diegocastanibm:
**Description:**
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/378

**Details:**

Previously, the launcher had no proactive mechanism to detect when vLLM instances terminated unexpectedly. Clients had to poll `GET /v2/vllm/instances/{id}` to discover a crash.

This PR adds:

  - `WatchEvent` — Pydantic model for lifecycle events with type and object fields
  - `EventBroadcaster` — pub/sub mechanism using `asyncio.Condition` with a bounded event buffer. Late-joining watchers catch up from the buffer.
  - `Sentinel watcher` — Each instance registers its `Process.sentinel fd` on the asyncio event loop. When the kernel signals process exit, a `STOPPED` event is published immediately.
  - `GET /v2/vllm/instances/watch` — Returns a StreamingResponse with application/x-ndjson content type. Client disconnection is handled automatically by FastAPI cancelling the async generator.

**Tests:**
Unit tests passed:
```
============================== test session starts ===============================
platform darwin -- Python 3.12.11, pytest-9.0.2, pluggy-1.6.0 -- /Users/diego/Documents/my_stuff/llm-d-fast-model-actuation/.venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/diego/Documents/my_stuff/llm-d-fast-model-actuation/inference_server/launcher
plugins: anyio-4.12.1
collected 79 items                                                               

tests/test_launcher.py::TestVllmConfig::test_vllm_config_with_env_vars PASSED [  1%]
tests/test_launcher.py::TestVllmConfig::test_vllm_config_without_env_vars PASSED [  2%]
tests/test_launcher.py::TestVllmInstance::test_instance_creation PASSED    [  3%]
tests/test_launcher.py::TestVllmInstance::test_instance_start PASSED       [  5%]
tests/test_launcher.py::TestVllmInstance::test_instance_start_already_running PASSED [  6%]
tests/test_launcher.py::TestVllmInstance::test_instance_stop PASSED        [  7%]
tests/test_launcher.py::TestVllmInstance::test_instance_stop_not_running PASSED [  8%]
tests/test_launcher.py::TestVllmInstance::test_instance_force_kill PASSED  [ 10%]
tests/test_launcher.py::TestVllmInstance::test_instance_get_status PASSED  [ 11%]
tests/test_launcher.py::TestVllmInstance::test_instance_uuid_to_index_translation PASSED [ 12%]
tests/test_launcher.py::TestVllmInstance::test_instance_uuid_translation_creates_env_vars_if_none PASSED [ 13%]
tests/test_launcher.py::TestVllmInstance::test_instance_uuid_translation_preserves_existing_env_vars PASSED [ 15%]
tests/test_launcher.py::TestVllmInstance::test_instance_no_uuid_translation_when_gpu_uuids_none PASSED [ 16%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_create_instance_auto_id PASSED [ 17%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_create_instance_custom_id PASSED [ 18%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_create_instance_duplicate_id PASSED [ 20%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_stop_instance PASSED [ 21%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_stop_nonexistent_instance PASSED [ 22%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_stop_all_instances PASSED [ 24%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_get_instance_status PASSED [ 25%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_get_instance_status_nonexistent PASSED [ 26%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_get_all_instances_status PASSED [ 27%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_list_instances PASSED [ 29%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_get_instance_log_bytes PASSED [ 30%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_get_instance_log_bytes_nonexistent PASSED [ 31%]
tests/test_launcher.py::TestVllmMultiProcessManager::test_get_instance_log_bytes_with_end PASSED [ 32%]
tests/test_launcher.py::TestAPIEndpoints::test_health_endpoint PASSED      [ 34%]
tests/test_launcher.py::TestAPIEndpoints::test_index_endpoint PASSED       [ 35%]
tests/test_launcher.py::TestAPIEndpoints::test_create_vllm_instance PASSED [ 36%]
tests/test_launcher.py::TestAPIEndpoints::test_create_id_vllm_instance PASSED [ 37%]
tests/test_launcher.py::TestAPIEndpoints::test_create_duplicate_instance PASSED [ 39%]
tests/test_launcher.py::TestAPIEndpoints::test_delete_vllm_instance PASSED [ 40%]
tests/test_launcher.py::TestAPIEndpoints::test_delete_nonexistent_instance PASSED [ 41%]
tests/test_launcher.py::TestAPIEndpoints::test_delete_all_instances PASSED [ 43%]
tests/test_launcher.py::TestAPIEndpoints::test_list_instances PASSED       [ 44%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_status PASSED  [ 45%]
tests/test_launcher.py::TestAPIEndpoints::test_get_nonexistent_instance_status PASSED [ 46%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_endpoint PASSED [ 48%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_with_range_header PASSED [ 49%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_nonexistent_endpoint PASSED [ 50%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_range_not_available PASSED [ 51%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_partial_content_206 PASSED [ 53%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_open_ended_range PASSED [ 54%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_invalid_range PASSED [ 55%]
tests/test_launcher.py::TestAPIEndpoints::test_get_instance_logs_suffix_range_rejected PASSED [ 56%]
tests/test_launcher.py::TestVllmInstanceLogs::test_get_log_bytes_no_file PASSED [ 58%]
tests/test_launcher.py::TestVllmInstanceLogs::test_get_log_bytes_empty_file PASSED [ 59%]
tests/test_launcher.py::TestVllmInstanceLogs::test_get_log_bytes_with_content PASSED [ 60%]
tests/test_launcher.py::TestVllmInstanceLogs::test_get_log_bytes_with_end PASSED [ 62%]
tests/test_launcher.py::TestVllmInstanceLogs::test_get_log_bytes_persists_across_calls PASSED [ 63%]
tests/test_launcher.py::TestHelperFunctions::test_set_env_vars PASSED      [ 64%]
tests/test_launcher.py::TestHelperFunctions::test_max_log_response_bytes_constant PASSED [ 65%]
tests/test_launcher.py::TestLogFile::test_basic_retrieval PASSED           [ 67%]
tests/test_launcher.py::TestLogFile::test_empty_file PASSED                [ 68%]
tests/test_launcher.py::TestLogFile::test_end_limit_enforcement PASSED     [ 69%]
tests/test_launcher.py::TestLogFile::test_end_larger_than_content PASSED   [ 70%]
tests/test_launcher.py::TestLogFile::test_unicode_bytes PASSED             [ 72%]
tests/test_launcher.py::TestLogFile::test_partial_retrieval_with_end PASSED [ 73%]
tests/test_launcher.py::TestEventBroadcaster::test_publish_and_watch PASSED [ 74%]
tests/test_launcher.py::TestEventBroadcaster::test_multiple_watchers PASSED [ 75%]
tests/test_launcher.py::TestEventBroadcaster::test_watch_since_revision PASSED [ 77%]
tests/test_launcher.py::TestSentinelWatcher::test_sentinel_publishes_stopped_event PASSED [ 78%]
tests/test_launcher.py::TestSentinelWatcher::test_sentinel_cancelled_on_explicit_stop PASSED [ 79%]
tests/test_launcher.py::TestWatchAPIEndpoint::test_watch_endpoint_content_type PASSED [ 81%]
tests/test_launcher.py::TestWatchAPIEndpoint::test_watch_endpoint_streams_events PASSED [ 82%]
tests/test_launcher.py::TestLogStartOffset::test_start_offset PASSED       [ 83%]
tests/test_launcher.py::TestLogStartOffset::test_start_at_message_boundary PASSED [ 84%]
tests/test_launcher.py::TestLogStartOffset::test_start_beyond_available_raises_error PASSED [ 86%]
tests/test_launcher.py::TestLogStartOffset::test_start_equal_to_length_raises_error PASSED [ 87%]
tests/test_launcher.py::TestLogStartOffset::test_no_file_with_start_raises_error PASSED [ 88%]
tests/test_launcher.py::TestLogFileCleanup::test_stop_terminated_cleans_up_log_file PASSED [ 89%]
tests/test_launcher.py::TestLogFileCleanup::test_cleanup_missing_file_no_error PASSED [ 91%]
tests/test_launcher.py::TestParseRangeHeader::test_full_range PASSED       [ 92%]
tests/test_launcher.py::TestParseRangeHeader::test_open_ended_range PASSED [ 93%]
tests/test_launcher.py::TestParseRangeHeader::test_large_values PASSED     [ 94%]
tests/test_launcher.py::TestParseRangeHeader::test_suffix_range_rejected PASSED [ 96%]
tests/test_launcher.py::TestParseRangeHeader::test_invalid_unit PASSED     [ 97%]
tests/test_launcher.py::TestParseRangeHeader::test_garbage_input PASSED    [ 98%]
tests/test_launcher.py::TestParseRangeHeader::test_end_less_than_start PASSED [100%]

=============================== 79 passed in 0.75s ===============================
```

End-to-end test passed (with new revision field as suggested in @MikeSpreitzer comments):
```
MacBook-Pro ~ $ curl -N http://localhost:8001/v2/vllm/instances/watch
{"type":"CREATED","object":{"status":"running","instance_id":"my_diego","options":"--model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --port 8006","env_vars":{"VLLM_USE_V1":"1","VLLM_LOGGING_LEVEL":"DEBUG"}},"revision":1}
{"type":"STOPPED","object":{"instance_id":"my_diego","status":"stopped","exit_code":null},"revision":2}
```


docs/launcher.md
```diff
@@ -367,6 +369,40 @@ Content-Type: application/octet-stream
...
+
+- `CREATED` — A new instance was spawned
+- `STOPPED` — An instance process terminated (detected automatically via kernel-level process monitoring, zero-polling)
+- `DELETED` — An instance was explicitly removed via the DELETE API
+
+**Response Headers:**
+
+- `Content-Type: application/x-ndjson`
+- `X-Content-Type-Options: nosniff`
+
```
MikeSpreitzer, line 399: Doesn't the chunked transfer encoding also involve response headers?
(I do not recommend copying HTTP documentation into here; referencing it is enough.)

docs/launcher.md
```diff
@@ -367,6 +369,40 @@ Content-Type: application/octet-stream
 
 ---
 
+#### Watch Instance Lifecycle Events
+
+**GET** `/v2/vllm/instances/watch`
+
+Stream real-time instance lifecycle events as newline-delimited JSON (NDJSON), similar to a Kubernetes watch. The response uses chunked transfer encoding and remains open until the client disconnects.
```
MikeSpreitzer: The documentation should also say what (if anything) the client gets told about events that happened before the `watch` began.

docs/launcher.md
```diff
@@ -367,6 +369,40 @@ Content-Type: application/octet-stream
...
+
```
MikeSpreitzer, line 380: I recommend following a few key aspects of the design of the watch operation in Kubernetes apiservers.

1. A server progresses through a totally ordered sequence of states, where each state is about all of the server's accessible objects. The state of one object includes the sequence number at which that object's state last changed (in Kubernetes this is called "ResourceVersion"). Each watch event includes all of the state of the relevant object. For a delete event, the ResourceVersion is the state number in which the object became deleted. It is not critical that the sequence numbers are assigned densely: it would be OK, for example, to go from sequence number 42 to 1701.
2. When a client opens a watch, the client is first given CREATE events for every object that existed when the watch began, each conveying the state of the object at the beginning of the watch. Objects deleted before the start of the watch are not mentioned.
3. A watch request can optionally specify the state number at which to start the watch. If the requested start is too long ago, the request fails with an indicative error.

inference_server/launcher/launcher.py
```diff
@@ -411,12 +514,31 @@ async def index():
...
                 "get_instance_logs": "GET /v2/vllm/instances/{instance_id}/log",
+                "watch_instances": "GET /v2/vllm/instances/watch",
             },
         },
         status_code=HTTPStatus.OK,
     )
 
 
+######################################################################
+# WATCH ENDPOINT
```
MikeSpreitzer: Isn't this one of the "vLLM MANAGEMENT ENDPOINTS"?

inference_server/launcher/launcher.py
```diff
@@ -411,12 +514,31 @@ async def index():
...
+######################################################################
+@app.get("/v2/vllm/instances/watch")
+async def watch_instances():
+    """Stream instance lifecycle events as NDJSON (Kubernetes watch-style)."""
+
+    async def event_stream():
+        async for event in vllm_manager.broadcaster.watch():
+            yield event.model_dump_json() + "\n"
```
MikeSpreitzer: Is this going to correctly (i.e., flatly) include the VllmConfig?

docs/launcher.md
diegocastanibm: Yes. I added something like:
```
The response uses [chunked transfer encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding) with `Content-Type: application/x-ndjson
```
As you can see, I have linked the documentation

docs/launcher.md
diegocastanibm: At this moment, the response sends previous events in the buffer. Since you are asking a different design in the comment below, I will modify this part of the documentation accordingly.

inference_server/launcher/launcher.py
diegocastanibm: I'll move it

inference_server/launcher/launcher.py
diegocastanibm: Yes, `_make_state()` already unpacks the `VllmConfig` flat `**self.config.model_dump(exclude_none=True)`, so the object dict contains the config fields directly (not nested under a config key). No changes needed here.

docs/launcher.md
diegocastanibm: OK, the watch endpoint now follows the three Kubernetes watch conventions that you described:
1. Revision number: Every event carries a monotonically increasing revision field
2. Initial state on connect: When a client opens a watch without ?since, it first receives a synthetic CREATED event for every instance that exists at that point. Instances deleted before the watch began are not mentioned.
3. Resumption with ?since=N: A client can reconnect with ?since=N to resume from a specific revision. Error if the revision falls out of the event buffer.

Also, documentation has been updated

docs/launcher.md
```diff
@@ -30,6 +30,7 @@ The launcher preloads vLLM’s Python modules to accelerate the initialization o
 - **Status Monitoring**: Query status of individual instances or all instances at once
 - **Log Capture**: Retrieve stdout/stderr logs from running instances via REST API
 - **Health Checks**: Built-in health endpoint for monitoring service availability
+- **Instance Lifecycle Watch**: Stream real-time instance lifecycle events (created, stopped, deleted) via a Kubernetes watch-style NDJSON endpoint
```
MikeSpreitzer: I wouldn't say "real-time" here. To me, that suggests something done with tight timing constraints, and does not suggest reflection of history.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +65,75 @@ class VllmConfig(BaseModel):
...
+        oldest = self.oldest_revision
+        if since_revision < oldest:
+            raise RevisionTooOld(since_revision, oldest)
+        pos = since_revision
+        while True:
+            async with self._condition:
+                while pos >= self._revision:
+                    await self._condition.wait()
+                offset = pos - self.oldest_revision
+                new_events = self._events[offset:]
```
MikeSpreitzer, line 139: What if the HTTP client is VERY slow (or grinds to a halt without closing the connection) consuming the response chunks? At some point that backs up into something in the launcher getting clogged up, right? Could it cause offset >= _MAX_BROADCASTER_EVENTS here?

The launcher should not silently omit notifications required by its API. If it cannot continue with more response chunks then it has to close the connection. A Kubernetes informer responds to such situations correctly.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +65,75 @@ class VllmConfig(BaseModel):
...
+        return self._revision
+
+    @property
+    def oldest_revision(self) -> int:
+        return self._revision - len(self._events)
+
+    async def publish(self, event: WatchEvent):
+        async with self._condition:
+            self._revision += 1
+            event.revision = self._revision
```
MikeSpreitzer: In the Kubernetes API object model, the revision is part of the object itself, and appears in every dump of the object's state, not just in watch responses.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +65,75 @@ class VllmConfig(BaseModel):
     annotations: Optional[Dict[str, str]] = None
 
 
+class WatchEvent(BaseModel):
+    """Represents an instance lifecycle event for the watch stream."""
+
+    type: str  # "CREATED", "STOPPED", "DELETED"
+    object: dict
+    revision: int = 0
```
MikeSpreitzer: I suspect that there are off-by-1 errors, due to lack of clarity about whether this field holds the _next_ number or the _most recent_ number. A simple comment here would help.

inference_server/launcher/launcher.py
MikeSpreitzer: Also, the doc that you cited has some more relevant info.

https://kubernetes.io/docs/reference/using-api/api-concepts/#resourceversion-in-metadata mentions that a LIST response includes the ResourceVersion that characterizes the list. Following is a demonstration of that.

```console
mspreitz@mjs13 llm-d-fast-model-actuation % kubectl get --raw https://127.0.0.1:56988/api/v1/namespaces/default/configmaps | jq . | head -15     
{
  "kind": "ConfigMapList",
  "apiVersion": "v1",
  "metadata": {
    "resourceVersion": "200988"
  },
  "items": [
    {
      "metadata": {
        "name": "gpu-allocs",
        "namespace": "default",
        "uid": "bfce6c27-4d0a-4af3-8ba1-3ffb7d683c45",
        "resourceVersion": "1766",
        "creationTimestamp": "2026-04-03T07:10:21Z"
      },
```

The section https://kubernetes.io/docs/reference/using-api/api-concepts/#the-resourceversion-parameter is also relevant.

docs/launcher.md
diegocastanibm: removed real-time

inference_server/launcher/launcher.py
diegocastanibm: This one was a nice catch! I did not think about it and had an interesting conversation with Claude to solve it. Apparently it is true that can silently misbehave if the offset is negative (taking wrong events from the tail of the buffer). Fixed: if offset < 0 -> raise RevisionTooOld

inference_server/launcher/launcher.py
diegocastanibm: Every instance state dict now includes a revision field. I have re-run and posted the example in the description

inference_server/launcher/launcher.py
diegocastanibm: Added a comment to clarify it in the init: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/413/changes/10343203b002517f6d3eccced0316fef900f101e#diff-6aa4261ef88911a6c58527353454b522cda583dfd2762bf66167df2f9aa20834R95

MikeSpreitzer: GitHub seems to have lost track of the line that this thread is commenting on.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +65,79 @@ class VllmConfig(BaseModel):
...
+    async def watch(self, since_revision: int = 0):
+        """Async generator that yields WatchEvents as they arrive.
+
+        :param since_revision: Resume from this revision.  Events with
+            revision > since_revision are yielded.
+        :raises RevisionTooOld: If since_revision is older than the
+            oldest event still in the buffer.
+        """
+        oldest = self.oldest_revision
+        if since_revision < oldest:
```
MikeSpreitzer, line 129: I think there is an off-by-1 error in this func.

If `oldest <= since_revision < self.revision` then the event with revision == since_revision will be yielded (even though the function comment says "Events with revision > since_revision are yielded").

inference_server/launcher/launcher.py
```diff
@@ -255,19 +373,39 @@ def create_instance(
...
+        result = instance.start()
+        try:
+            loop = asyncio.get_running_loop()
+            instance.start_sentinel_watcher(self.broadcaster)
+            loop.create_task(
+                self.broadcaster.publish(WatchEvent(type="CREATED", object=result))
+            )
+        except RuntimeError:
+            pass  # No running event loop (e.g. in sync tests)
+        result["revision"] = self.broadcaster.revision
```
MikeSpreitzer: This is a bit fragile (as are the other statements like it). It does not assign increasing revisions in the absence of a running loop (even though a running loop is not logically necessary for assigning increasing revisions). It assumes that nothing else increments the self.broadcaster.revision between (a) the assignment of a revision number to this create operation and (b) this statement being executed.

Could this be simplified and made more robust by defining an ordinary method of broadcaster that assigns a revision number and puts it in the object and then, if there is a running loop, creates a task on it to publish the event?

My comments in get_instance_status and get_all_instances_status are about a functional problem with the current approach, also motivating the approach that I suggested.

inference_server/launcher/launcher.py
```diff
@@ -293,20 +431,25 @@ def get_instance_status(self, instance_id: str) -> dict:
         if instance_id not in self.instances:
             raise KeyError(f"Instance {instance_id} not found")
 
-        return self.instances[instance_id].get_status()
+        result = self.instances[instance_id].get_status()
+        result["revision"] = self.broadcaster.revision
```
MikeSpreitzer: No. The revision number in this result should be the revision number at which there was last a change (i.e., create or stop) in this instance.

inference_server/launcher/launcher.py
```diff
@@ -293,20 +431,25 @@ def get_instance_status(self, instance_id: str) -> dict:
...
 
     def get_all_instances_status(self) -> dict:
         """Get status of all instances"""
         instances_status = []
         running_count = 0
+        revision = self.broadcaster.revision
 
         for instance in self.instances.values():
             status = instance.get_status()
+            status["revision"] = revision
```
MikeSpreitzer: No. The revision number in this result should be the revision number at which there was last a change (i.e., create or stop) in this instance.

inference_server/launcher/launcher.py
MikeSpreitzer: Or maybe have the VllmMultiProcessManager in directly charge of assigning revision numbers, and leave the broadcaster to just the publishing/watching?

inference_server/launcher/launcher.py
```diff
@@ -420,6 +564,53 @@ async def index():
...
+        if since is None:
+            # Initial state: emit a CREATED event for every existing instance
+            # and start streaming from the current revision.
+            start_revision = vllm_manager.broadcaster.revision
+            for instance in vllm_manager.instances.values():
+                state = instance.get_status()
+                initial = WatchEvent(
+                    type="CREATED",
+                    object=state,
+                    revision=start_revision,
```
MikeSpreitzer: (A) The revision should be in the object, not adjacent to it.
(B) The object's revision should be the last revision in which the object was written.

inference_server/launcher/launcher.py
```diff
@@ -420,6 +564,53 @@ async def index():
...
+        try:
+            async for event in vllm_manager.broadcaster.watch(start_revision):
+                yield event.model_dump_json() + "\n"
+        except RevisionTooOld as exc:
+            error = {
+                "error": "revision_too_old",
+                "message": f"Requested revision {exc.requested} is no longer "
+                f"available. Oldest available: {exc.oldest}.",
+            }
+            yield json.dumps(error) + "\n"
```
MikeSpreitzer: The HTTP stream of response chunks needs to be closed; the client is not expecting to receive a JSON object describing an error.

inference_server/launcher/launcher.py
```diff
@@ -487,7 +678,11 @@ async def get_all_vllm_instances(detail: bool = True):
         result = vllm_manager.get_all_instances_status()
     else:
         instances = vllm_manager.list_instances()
-        result = {"instance_ids": instances, "count": len(instances)}
+        result = {
+            "revision": vllm_manager.broadcaster.revision,
+            "instance_ids": instances,
+            "count": len(instances),
```
MikeSpreitzer, line 714: IMO, having this count in the response schema just makes life hard for a carefully written client. The client has to handle the case where the count is different from the length of the array of instances.

This problem was not introduced in this PR and does not have to be addressed in this PR.

MikeSpreitzer (commented):

inference_server/launcher/launcher.py
```diff
@@ -255,19 +373,39 @@ def create_instance(
...
         # the HalfMade exception.
-        return instance.start()
+        result = instance.start()
+        try:
+            loop = asyncio.get_running_loop()
+            instance.start_sentinel_watcher(self.broadcaster)
+            loop.create_task(
+                self.broadcaster.publish(WatchEvent(type="CREATED", object=result))
+            )
+        except RuntimeError:
```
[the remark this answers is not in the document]
MikeSpreitzer: IMO simplest would be to make the VllmMultiProcessManager be the thing that assigns revision numbers.

inference_server/launcher/launcher.py
```diff
@@ -255,19 +373,39 @@ def create_instance(
         # Because start() is always the first method called on a VllmInstance,
         # the other methods of VllmMultiProcessManager do not need to handle
         # the HalfMade exception.
-        return instance.start()
+        result = instance.start()
+        try:
```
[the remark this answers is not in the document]
diegocastanibm: Wrapped `instance.start()` in `try/except` as suggested

inference_server/launcher/launcher.py
```diff
@@ -255,19 +373,39 @@ def create_instance(
...
             del self.instances[instance_id]
 
+        try:
+            loop = asyncio.get_running_loop()
+            loop.create_task(
+                self.broadcaster.publish(WatchEvent(type="DELETED", object=result))
+            )
+        except RuntimeError:
+            pass  # No running event loop (e.g. in sync tests)
+        result["revision"] = self.broadcaster.revision
```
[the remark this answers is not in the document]
diegocastanibm: Revision assignment has been moved out of the broadcaster and into `VllmMultiProcessManager`, which now owns a `_revision counter` and a `_next_revision()` method. The revision is assigned synchronously before `create_task()` is called, so the result dict always carries the correct value.

docs/launcher.md
```diff
@@ -367,6 +372,46 @@ Content-Type: application/octet-stream
...
+
+Stream instance lifecycle events as newline-delimited JSON ([NDJSON](https://github.com/ndjson/ndjson-spec)), following a design similar to a [Kubernetes watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes). The response uses [chunked transfer encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding) with `Content-Type: application/x-ndjson` and remains open until the client disconnects.
+
+Each event carries the full state of the relevant instance and a monotonically increasing `revision` number:
+
+```json
+{"type": "CREATED", "object": {"status": "started", "instance_id": "abc123", "options": "--model test"}, "revision": 1}
+{"type": "STOPPED", "object": {"instance_id": "abc123", "status": "stopped", "exit_code": -9}, "revision": 2}
+{"type": "DELETED", "object": {"status": "terminated", "instance_id": "abc123", "options": "--model test"}, "revision": 3}
+```
```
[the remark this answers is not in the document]
diegocastanibm: Updated the docs

inference_server/launcher/launcher.py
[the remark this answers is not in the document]
diegocastanibm: Same fix as above: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/413#discussion_r3111920391

inference_server/launcher/launcher.py
```diff
@@ -255,19 +373,39 @@ def create_instance(
...
+        try:
+            loop = asyncio.get_running_loop()
+            instance.start_sentinel_watcher(self.broadcaster)
+            loop.create_task(
+                self.broadcaster.publish(WatchEvent(type="CREATED", object=result))
+            )
+        except RuntimeError:
+            pass  # No running event loop (e.g. in sync tests)
+        result["revision"] = self.broadcaster.revision
+        return result
```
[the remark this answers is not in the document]
diegocastanibm: Same

inference_server/launcher/launcher.py
diegocastanibm: The comment on `_revision` is now at the `EventBroadcaster.__init__` level, clarifying that revision numbers are assigned by the manager and the broadcaster only records them

inference_server/launcher/launcher.py
diegocastanibm: I think the code is correct. `oldest_revision` is an exclusive lower bound (not the revision of the oldest event in the buffer). I also checked it with Claude and gave me this example:
```
For example, with events [rev1, rev2, rev3]: oldest_revision = 0, and watch(since_revision=2) gives offset = 2, yielding events[2:] = [rev3] — only revision 3, not 2. The invariant is:
events[i] has revision oldest_revision + 1 + i, so events[offset:] starts at revision since_revision + 1, which matches the documented contract "> since_revision".
```

inference_server/launcher/launcher.py
diegocastanibm: Refactored so that `VllmMultiProcessManager` owns the revision counter directly as you suggested

inference_server/launcher/launcher.py
diegocastanibm: Each `VllmInstance` now has a `last_revision` field set when the instance is created, stopped, or exits via sentinel

inference_server/launcher/launcher.py
diegocastanibm: same fix as above

inference_server/launcher/launcher.py
diegocastanibm: Done. The revision is now inside the object itself (set from `instance.last_revision`).

inference_server/launcher/launcher.py
diegocastanibm: I'll leave it for now and address it separately in a different PR

inference_server/launcher/launcher.py
diegocastanibm: OK, changed to just close the stream when the revision falls out of the buffer.

diegocastanibm: Also, When `?since=N` references a revision that has already fallen out of the buffer, the server now responds with 410 Gone (HTTP error) before opening the stream. If the revision falls out of the buffer while the stream is already open (slow client), the server silently closes the connection.

docs/launcher.md
```diff
@@ -367,6 +372,43 @@ Content-Type: application/octet-stream
...
+#### Watch Instance Lifecycle Events
+
+**GET** `/v2/vllm/instances/watch[?since=N]`
+
+Stream instance lifecycle events as newline-delimited JSON ([NDJSON](https://github.com/ndjson/ndjson-spec)), following a design similar to a [Kubernetes watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes). The response uses [chunked transfer encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding) with `Content-Type: application/x-ndjson` and remains open until the client disconnects.
+
+Each event carries the full state of the relevant instance and a monotonically increasing `revision` number:
+
+```json
+{"type": "CREATED", "object": {"status": "started", "instance_id": "abc123", "options": "--model test", "revision": 1}, "revision": 1}
```
MikeSpreitzer: We do not need `revision` to appear at the top level of each row, appearing inside the object is enough.

docs/launcher.md
```diff
@@ -423,21 +467,24 @@ Get status information for all instances. `Detail` is `True` by default.
...
+      "revision": 5
     },
     {
       "status": "stopped",
       "instance_id": "id-2",
       "options": <options 2>,
       "gpu_uuids": <gpus 2>,
-      "env_vars": <envars 2>
+      "env_vars": <envars 2>,
+      "revision": 5
```
MikeSpreitzer: The instances will each have a different `revision`, right?

inference_server/launcher/launcher.py
MikeSpreitzer: Now there are two revision numbers in each event: one inside the object and one outside the object. We only need the one inside the object.

inference_server/launcher/launcher.py
```diff
@@ -293,7 +463,10 @@ def get_instance_status(self, instance_id: str) -> dict:
         if instance_id not in self.instances:
             raise KeyError(f"Instance {instance_id} not found")
 
-        return self.instances[instance_id].get_status()
+        instance = self.instances[instance_id]
+        result = instance.get_status()
+        result["revision"] = instance.last_revision
```
MikeSpreitzer: I notice that this statement follows every call to `instance.get_status()`. Why isn't this functionality part of `instance.get_status()` ? Why is it not part of `instance._make_state()` ?

inference_server/launcher/launcher.py
```diff
@@ -62,6 +64,81 @@ class VllmConfig(BaseModel):
...
+    @property
+    def oldest_revision(self) -> int:
+        """Exclusive lower bound: events in the buffer have revision
+        strictly greater than this value."""
+        return self._revision - len(self._events)
+
+    async def publish(self, event: WatchEvent):
+        async with self._condition:
+            self._revision = event.revision
+            self._events.append(event)
```
MikeSpreitzer: These need to be done in the same sequence as incrementing the global revision counter and assigning that to a VllmInstance, not fobbed off in two levels of `async`.

MikeSpreitzer (commented):

docs/launcher.md
diegocastanibm: Now, just inside the object:
```
{"type": "CREATED", "object": {"status": "started", ..., "revision": 1}}
```

inference_server/launcher/launcher.py
diegocastanibm: Done. `_make_state()` now includes `self.last_revision` directly

inference_server/launcher/launcher.py
diegocastanibm: OK. Refactored `EventBroadcaster.publish()` into sync `_append()` + async `_notify()`. The manager now calls `_append()` synchronously right after `_next_revision()`, ensuring the event is in the buffer before the method returns. Only the condition notification is deferred via `create_task`.

docs/launcher.md
diegocastanibm: Right. Instances now show their individual last_revision instead of all having the same value in the documentation

docs/launcher.md
```diff
@@ -415,6 +458,7 @@ Get status information for all instances. `Detail` is `True` by default.
 
 ```json
 {
+  "revision": 5,
```
MikeSpreitzer: In general this can be higher than the highest revision of an existing instance. It would be good if this example illustrated this possibility.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +64,90 @@ class VllmConfig(BaseModel):
     annotations: Optional[Dict[str, str]] = None
 
 
+class WatchEvent(BaseModel):
+    """Represents an instance lifecycle event for the watch stream."""
+
+    type: str  # "CREATED", "STOPPED", "DELETED"
+    object: dict
+    revision: int = Field(default=0, exclude=True)
```
MikeSpreitzer: Having this field seems like unnecessary complexity to me, since the revision is also in the `object: dict`.

inference_server/launcher/launcher.py
```diff
@@ -110,6 +196,8 @@ def __init__(
         self.instance_id = instance_id
         self.config = config
         self.process: Optional[multiprocessing.Process] = None
+        self.last_revision: int = 0
```
MikeSpreitzer: Wouldn't `self.last_revision: Optional[int] = None` be more accurate?

inference_server/launcher/launcher.py
```diff
@@ -237,6 +367,33 @@ def __init__(
...
+            self.instances[instance_id].last_revision = revision
+        event = WatchEvent(
+            type="STOPPED",
+            object={
+                "instance_id": instance_id,
+                "status": "stopped",
+                "exit_code": exitcode,
+                "revision": revision,
+            },
+            revision=revision,
```
MikeSpreitzer: Two revisions here.
Object state is missing stuff.
Better to use object.get_status here.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +64,90 @@ class VllmConfig(BaseModel):
...
+
+    async def _notify(self):
+        """Acquire condition lock and wake all watchers."""
+        async with self._condition:
+            self._condition.notify_all()
+
+    async def publish(self, event: WatchEvent):
+        self._append(event)
+        async with self._condition:
+            self._condition.notify_all()
```
MikeSpreitzer: This method is obsolete and should be deleted (unit tests need to be updated).

inference_server/launcher/launcher.py
```diff
@@ -252,22 +409,49 @@ def create_instance(
...
-        return instance.start()
+        try:
+            start_result = instance.start()
+        except Exception:
+            self.instances.pop(instance_id, None)
+            raise
+
+        revision = self._next_revision()
+        instance.last_revision = revision
+        result = instance._make_state(start_result["status"])
```
MikeSpreitzer: Now there are two calls to `_make_state()`: the one in `instance.start()` and the one right here. The result from the first of those two is almost entirely unused, and could be completely unused if `get_status()` were called here instead of `_make_state()`.

Similar things happen in similar methods of this class.

MikeSpreitzer (commented):

docs/launcher.md
diegocastanibm: Changed the list-level revision to 7

inference_server/launcher/launcher.py
diegocastanibm: Done. Removed the revision field from WatchEvent entirely. The broadcaster now reads the revision from `event.object["revision"]`.

inference_server/launcher/launcher.py
diegocastanibm: Changed to `Optional[int] = None`

inference_server/launcher/launcher.py
diegocastanibm: Fallback to a minimal dict only. `_on_instance_stopped` now uses `instance.get_status()` to build the object state

inference_server/launcher/launcher.py
diegocastanibm: Done. Deleted the `publish()` method. All callers now use `_append()` plus` _notify()` directly.

inference_server/launcher/launcher.py
diegocastanibm: Done. Both create_instance and stop_instance now use `get_status()` instead of calling `_make_state()` a second time. This also normalizes the status values: create returns "running" (the actual process state) and delete returns "stopped". Docs updated accordingly.

docs/launcher.md
```diff
@@ -367,6 +372,43 @@ Content-Type: application/octet-stream
...
+
+Each event carries the full state of the relevant instance and a monotonically increasing `revision` number:
+
+```json
+{"type": "CREATED", "object": {"status": "running", "instance_id": "abc123", "options": "--model test", "revision": 1}}
+{"type": "STOPPED", "object": {"instance_id": "abc123", "status": "stopped", "exit_code": -9, "revision": 2}}
+{"type": "DELETED", "object": {"status": "stopped", "instance_id": "abc123", "options": "--model test", "revision": 3}}
+```
+
+**Event Types:**
```
MikeSpreitzer, line 389: This looks like a subsection header, but it is not made using a Markdown construct for a subsection header. I think that it would be better to use a Markdown construct.

Same comment for the other ersatz subsection headers.

docs/launcher.md
```diff
@@ -367,6 +372,43 @@ Content-Type: application/octet-stream
...
+
+- `CREATED` — A new instance was spawned
+- `STOPPED` — An instance process terminated (detected automatically via kernel-level process monitoring, zero-polling)
+- `DELETED` — An instance was explicitly removed via the DELETE API.
+
+**Initial state and resumption:**
+
+- **Without `?since`**: The stream begins with a synthetic `CREATED` event for every instance that exists at the time the watch starts, followed by live events. Instances deleted before the watch began are not mentioned.
+- **With `?since=N`**: The stream resumes from revision `N`, yielding only events with `revision > N`. This allows a client to reconnect without missing events. If the requested revision is older than the server's event buffer, the server responds with **410 Gone** and closes the connection. The client should start a fresh watch without `?since` to re-sync. If the revision falls out of the buffer while the stream is already open (slow client), the server silently closes the connection.
```
MikeSpreitzer: > The client should start a fresh watch without `?since` to re-sync.

No, that does not work; this is why the LIST operation returns the revision number representing the latest revision at the server.

Consider a client trying to sync using only a watch without `?since`. How does this client identify instances that were deleting while the client was not watching? These will not be mentioned in the response stream of the new watch. How does the client know when it has gotten CREATE events for everything that existed at the server at the start of the watch? There is no way.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +64,84 @@ class VllmConfig(BaseModel):
...
+        """Acquire condition lock and wake all watchers."""
+        async with self._condition:
+            self._condition.notify_all()
+
+    async def watch(self, since_revision: int = 0):
+        """Async generator that yields WatchEvents as they arrive.
+
+        :param since_revision: Resume from this revision.  Events with
+            revision > since_revision are yielded.
+        :raises RevisionTooOld: If since_revision is older than the
```
MikeSpreitzer: I would change that "If" to "If/when", because it can arise at any time during the generating.

inference_server/launcher/launcher.py
```diff
@@ -62,6 +64,84 @@ class VllmConfig(BaseModel):
...
+
+    @property
+    def oldest_revision(self) -> int:
+        """Exclusive lower bound: events in the buffer have revision
+        strictly greater than this value."""
+        return self._revision - len(self._events)
+
+    def _append(self, event: WatchEvent):
+        """Synchronously buffer event and advance the revision cursor."""
+        self._revision = event.object["revision"]
```
MikeSpreitzer, line 110: The typing here looks potentially problematic. Here, `event.object` is a general `dict`, not guaranteed to have an entry named `revision`, not guranteeing the type of the value if it exists. Yet line 106 does arithmetic on `self._revision`.

To be fair, the type checker in my IDE (VS Code) is not complaining about this (even though it has a few other type complaints). But I do not understand why.

docs/launcher.md
```diff
@@ -367,6 +372,43 @@ Content-Type: application/octet-stream
...
+
+**GET** `/v2/vllm/instances/watch[?since=N]`
+
+Stream instance lifecycle events as newline-delimited JSON ([NDJSON](https://github.com/ndjson/ndjson-spec)), following a design similar to a [Kubernetes watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes). The response uses [chunked transfer encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding) with `Content-Type: application/x-ndjson` and remains open until the client disconnects.
+
+Each event carries the full state of the relevant instance and a monotonically increasing `revision` number:
+
+```json
+{"type": "CREATED", "object": {"status": "running", "instance_id": "abc123", "options": "--model test", "revision": 1}}
+{"type": "STOPPED", "object": {"instance_id": "abc123", "status": "stopped", "exit_code": -9, "revision": 2}}
```
MikeSpreitzer, line 385: `exit_code` is not defined (in this doc) to be a part of the instance state.

docs/launcher.md
```diff
@@ -367,6 +372,43 @@ Content-Type: application/octet-stream
 
 ---
 
+#### Watch Instance Lifecycle Events
+
+**GET** `/v2/vllm/instances/watch[?since=N]`
+
+Stream instance lifecycle events as newline-delimited JSON ([NDJSON](https://github.com/ndjson/ndjson-spec)), following a design similar to a [Kubernetes watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes). The response uses [chunked transfer encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding) with `Content-Type: application/x-ndjson` and remains open until the client disconnects.
+
+Each event carries the full state of the relevant instance and a monotonically increasing `revision` number:
```
MikeSpreitzer: "state and revision" suggests that this is the structure of an event. But no, the structure of an event is actually "type and state".

inference_server/launcher/launcher.py
```diff
@@ -237,6 +361,34 @@ def __init__(
...
+            instance.last_revision = revision
+            obj = instance.get_status()
+            obj["exit_code"] = exitcode
+        else:
+            obj = {
+                "instance_id": instance_id,
+                "status": "stopped",
+                "exit_code": exitcode,
+                "revision": revision,
+            }
```
MikeSpreitzer: This is does not follow the documented schema for objects in watch events.
If I follow correctly, this case is only necessary because sometimes the manager removes `instance_id` from `self.instances` too soon. IMO it would be better for the manager to not do that, and remove this divergent case from here.

MikeSpreitzer (commented):
I finished another round of review.

docs/launcher.md
diegocastanibm: I prefer to keep the bold style here for aesthetic reasons. A level-5 heading would feel heavy in this context.

inference_server/launcher/launcher.py
diegocastanibm: Done

inference_server/launcher/launcher.py
diegocastanibm: `event.object["revision"]` is guaranteed to be present at every call site that creates a `WatchEvent`. There are four such sites in this file, and all of them set `revision` in the `object` dict. The type checker doesn't complain because dict is typed as `Dict[str, Any]` so key access is valid syntax.

docs/launcher.md
diegocastanibm: Right. I have added a note to the `STOPPED` event description documenting the `exit_code` field

docs/launcher.md
diegocastanibm: OK. Added that each event has a `type` field and an `object` field carrying the full state of the relevant
  instance, including a monotonically increasing `revision` number.

inference_server/launcher/launcher.py
diegocastanibm: The else branch is in fact unreachable. The sentinel fd reader is registered in `create_instance` and is **always** cancelled in `cancel_sentinel_watcher()` before the instance is removed from `self.instances`. Since `stop_instance` calls `cancel_sentinel_watcher()` as its first step and since the asyncio event loop is single-threaded there is no window in which `_on_instance_stopped` can fire for an instance that has already been removed from the dict. Removed.

docs/launcher.md
diegocastanibm: I see. OK, changed the documentation to to prescribe the correct re-sync procedure. Now it says that it needs to call LIST to obtain the current state and its revision, then open a new watch with `?since=<list-revision>`.
Also, according to Claude:
```
This mirrors the Kubernetes informer pattern and ensures no events are missed while correctly reflecting deletions that happened during the gap.
```

docs/launcher.md
MikeSpreitzer: Do you mean the source feels heavy or the rendering? I find that the rendering of level 5 headings is pretty light.

docs/launcher.md
```diff
@@ -367,6 +372,43 @@ Content-Type: application/octet-stream
 
 ---
 
+#### Watch Instance Lifecycle Events
+
+**GET** `/v2/vllm/instances/watch[?since=N]`
+
+Stream instance lifecycle events as newline-delimited JSON ([NDJSON](https://github.com/ndjson/ndjson-spec)), following a design similar to a [Kubernetes watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes). The response uses [chunked transfer encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding) with `Content-Type: application/x-ndjson` and remains open until the client disconnects.
```
MikeSpreitzer, line 379: The Kubernetes document set changes shape occasionally (on minor release boundaries). It would be better to have version in the URL, such as https://v1-34.docs.kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes

docs/launcher.md
MikeSpreitzer: My point here is that if an object can have an `exit_code` then that needs to be a documented part of the schema for an object. Compare, for example, the way in Kuberntes a Container in a Pod can report that it ran and terminated at a certain time with a certain result.

inference_server/launcher/launcher.py
MikeSpreitzer: To be clear, this is not a critical problem, since neither of us seems to be getting any resistance from our IDE.

In my IDE, `event.object` is typed `dict[Unknown, Unknown]`, and `event.object["revision"]` is typed as `Unknown`.

You report `event.object` being typed `dict[str, Any]` --- i.e., no guarantee about what keys are in the dict. So `event.object["revision"]` must be a complete unknown.

The thing that troubles me is assigning this unknown to an integer variable (`self._revision`). In my IDE, `self._revision` is typed `int` everywhere but here; here it is typed `Unknown | int`. I do not know how/why my IDE is sanguine about not propagating the unknown-ness of `event.object["revision"]` through `self._revision`.

inference_server/launcher/launcher.py
```diff
@@ -252,22 +396,47 @@ def create_instance(
         )
         self.instances[instance_id] = instance
 
-        # Because start() is always the first method called on a VllmInstance,
-        # the other methods of VllmMultiProcessManager do not need to handle
-        # the HalfMade exception.
```
MikeSpreitzer, line 257: This comment is still true and pertinent, but needs a little expansion to mention the new possibility that `start()` fails AND the instance is removed from `self.instances`.

inference_server/launcher/launcher.py
```diff
@@ -420,6 +591,56 @@ async def index():
...
+                detail=f"Requested revision {since} is no longer available. "
+                f"Oldest available: {oldest}.",
+            )
+
+    async def event_stream():
+        if since is None:
+            # Initial state: emit a CREATED event for every existing instance
+            # and start streaming from the current revision.
+            start_revision = vllm_manager.revision
+            for instance in vllm_manager.instances.values():
```
MikeSpreitzer, line 621: Is this safe to do with a `yield` in the loop body? https://docs.python.org/3.13/library/stdtypes.html#dict-views says, under `iter(dictview)`,

> Iterating views while adding or deleting entries in the dictionary may raise a [RuntimeError](https://docs.python.org/3.13/library/exceptions.html#RuntimeError) or fail to iterate over all entries.

MikeSpreitzer (dismissed):
This could use more work, but I think that will be safe to do in follow-on PRs;
this one is about new functionality that no client uses yet.

MikeSpreitzer:
I looked into the test failure of the CI test of launcher, E2E, kind.

The run of the test cases shows the date "Wed May 13 16:56:07 UTC 2026" just before it does `kubectl wait --for condition=Ready pod/$reqlb_after_delete -n "$NS" --timeout=120s` and that times out. 119 seconds later the dual-pods controller reports having set the readiness of the requester. I do not see evidence of when the kubelet last probed readiness before the timeout.

Here are some excerpts from the dual-pods controller log.

> I0513 16:56:21.002293       1 inference-server.go:464] "Created vLLM instance" worker=1 node="fmatest-worker" serverUID="b159923b-45b2-4119-bee3-d580d4097079" requesterName="my-request-13-16-47-54-9fcqk" instanceID="IiG2tMamvGMRzpMiB07ZpPkkd6PMtKYgw1gIgSojFu70i" providerName="launcher-13-16-47-54-5675s" instance_id="IiG2tMamvGMRzpMiB07ZpPkkd6PMtKYgw1gIgSojFu70i" status="running"

I include the following exceptional one because it puzzles me. The launcher---once---fails to list instances.

> E0513 16:57:06.388302       1 inference-server.go:1751] "Failed to list instances from launcher" err="Get \"[http://10.244.1.16:8001/v2/vllm/instances\](http://10.244.1.16:8001/v2/vllm/instances/)": context deadline exceeded (Client.Timeout exceeded while awaiting headers)" worker=1 node="fmatest-worker" serverUID="b159923b-45b2-4119-bee3-d580d4097079" requesterName="my-request-13-16-47-54-9fcqk" instanceID="IiG2tMamvGMRzpMiB07ZpPkkd6PMtKYgw1gIgSojFu70i" providerName="launcher-13-16-47-54-5675s"

> I0513 16:58:06.421099       1 inference-server.go:1691] "Set status/finalizers" worker=1 node="fmatest-worker" serverUID="b159923b-45b2-4119-bee3-d580d4097079" requesterName="my-request-13-16-47-54-9fcqk" providerName="launcher-13-16-47-54-5675s" serverRequestingPod="my-request-13-16-47-54-9fcqk" status={"Errors":null} accelerators="GPU-2" boundName="launcher-13-16-47-54-5675s" instanceID="IiG2tMamvGMRzpMiB07ZpPkkd6PMtKYgw1gIgSojFu70i" finalizers=["dual-pods.llm-d.ai/requester"] newResourceVersion="2069"

> I0513 16:58:06.421914       1 inference-server.go:513] "Successfully relayed the readiness" worker=1 node="fmatest-worker" serverUID="b159923b-45b2-4119-bee3-d580d4097079" requesterName="my-request-13-16-47-54-9fcqk" instanceID="IiG2tMamvGMRzpMiB07ZpPkkd6PMtKYgw1gIgSojFu70i" providerName="launcher-13-16-47-54-5675s" name="launcher-13-16-47-54-5675s" readiness="ready"

MikeSpreitzer (approved):
As noted before: this could use more work, but can be safely merged because its changes involve stuff not used yet.

Issue #375, opened 2026-03-23
[Feature]: Make the dual-pods controller better informed about launcher hosted vLLM instances


waltforme:
### Feature Area

Other

### Problem Statement

By using launchers, individual vLLM instances are packed into a single Pod instead of each running in its own Pod, which changes the way the dual-pods controller is informed about status changes of individual vLLM instances.

Previously, these changes were exposed through the status of each Pod. The dual-pods controller passively received the events delivered by the Kubernetes API machinery in a timely and reliable manner.

Today, the dual-pods controller actively queries the launcher's API. These queries are initiated by the dual-pods controller, sometimes periodically and sometimes not. As a result, the awareness of status changes may be delayed, or the changes may even be missed.

It would be better to restore timely and reliable event-driven informing of status changes.

### Proposed Solution

TBD

### Willingness to Contribute

Yes, I can submit a PR

MikeSpreitzer:
Stage 1 was completed by #414. That is sufficient to get decent milestone 3 functionality.

What remains now is Stage 2, which is an efficiency improvement.

MikeSpreitzer:
Closing this Issue, because stage 1 has been implemented. We can open a more focused Issue for stage 2 if/when we want.

PR #414, opened 2026-04-06
React on externalized instance state change

waltforme:
This PR finishes 'Stage 1' of the roadmap proposed in #375.

### DIFF
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -85,7 +85,7 @@ func (ni nodeItem) process(ctx context.Context, ctl *controller) (error, bool) {
 	return nil, retries > 0
 }
 
-func (item launcherPodItem) process(ctx context.Context, ctl *controller, nodeDat *nodeData) (error, bool) {
+func (item unboundLauncherPodItem) process(ctx context.Context, ctl *controller, nodeDat *nodeData) (error, bool) {
 	logger := klog.FromContext(ctx).WithValues("launcherPod", item.LauncherPodName, "node", item.NodeName)
 	ctx = klog.NewContext(ctx, logger)
 
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -108,15 +108,12 @@ func (item unboundLauncherPodItem) process(ctx context.Context, ctl *controller,
 
 	// Sync launcher instances to keep internal state fresh and clean up stopped instances.
 	_, syncErr, syncRetry := ctl.syncLauncherInstances(ctx, nodeDat, launcherPod)
-	if syncErr != nil {
-		if syncRetry {
-			logger.V(4).Info("Failed to sync launcher instances, will retry", "err", syncErr)
-		} else {
-			logger.Error(syncErr, "Failed to sync launcher instances")
-		}
-	}
 
 	ctl.enqueueUnboundInfSvrItemsOnNode(ctx, item.NodeName, fmt.Sprintf("launcher pod %s changed", item.LauncherPodName))
+
+	if syncErr != nil {
+		return fmt.Errorf("failed to sync launcher instances: %w", syncErr), syncRetry
+	}
 	return nil, syncRetry
 }
 
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -370,7 +370,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 			syncResult, err, retry := ctl.syncLauncherInstances(ctx, nodeDat, providingPod)
 			if err != nil || retry {
 				if err != nil {
-					return fmt.Errorf("failed to sync launcher instances for bound launcher Pod: %w", err), true
+					return fmt.Errorf("failed to sync launcher instances for bound launcher Pod: %w", err), retry
 				}
 				return nil, true
 			}
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -1412,7 +1412,7 @@ func (ctl *controller) syncLauncherInstances(ctx context.Context, nodeDat *nodeD
 		if inst.Status == InstanceStatusStopped {
 			// Clean up stopped instances from the launcher.
 			_, delErr := lClient.DeleteInstance(ctx, inst.InstanceID)
-			if delErr != nil && !IsLauncherNotFoundError(delErr) {
+			if delErr != nil && !IsInstanceNotFoundError(delErr) {
 				logger.V(3).Info("Failed to delete stopped instance from launcher during sync",
 					"instanceID", inst.InstanceID, "err", delErr)
 				// Deletion failed — the instance still occupies a slot in the launcher.
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -389,6 +389,8 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				_, delErr := lClient.DeleteInstance(ctx, serverDat.InstanceID)
 				if delErr != nil && !IsLauncherNotFoundError(delErr) {
 					logger.V(3).Info("Failed to delete stopped instance from launcher", "instanceID", serverDat.InstanceID, "err", delErr)
+					return ctl.ensureReqStatus(ctx, requestingPod, serverDat,
+						fmt.Sprintf("failed to delete stopped instance %q from launcher: %s", serverDat.InstanceID, delErr.Error()))
 				}
 				// Mark as sleeping so that ensureUnbound (called during requester deletion)
 				// does not attempt to POST /sleep on the dead instance.
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -89,7 +89,7 @@ func (item launcherPodItem) process(ctx context.Context, ctl *controller, nodeDa
 	logger := klog.FromContext(ctx).WithValues("launcherPod", item.LauncherPodName, "node", item.NodeName)
 	ctx = klog.NewContext(ctx, logger)
 
-	_, err := ctl.podLister.Pods(ctl.namespace).Get(item.LauncherPodName)
+	launcherPod, err := ctl.podLister.Pods(ctl.namespace).Get(item.LauncherPodName)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			logger.V(2).Info("Launcher pod deleted, cleaning up launcher data")
@@ -100,8 +100,19 @@ func (item launcherPodItem) process(ctx context.Context, ctl *controller, nodeDa
 		return err, true
 	}
 
+	// Sync launcher instances to keep internal state fresh and clean up stopped instances.
+	// This is triggered by annotation changes from the sidecar notifier.
+	_, syncErr, syncRetry := ctl.syncLauncherInstances(ctx, nodeDat, launcherPod)
+	if syncErr != nil {
+		if syncRetry {
+			logger.V(4).Info("Failed to sync launcher instances, will retry", "err", syncErr)
+		} else {
+			logger.Error(syncErr, "Failed to sync launcher instances")
+		}
+	}
+
 	ctl.enqueueUnboundInfSvrItemsOnNode(ctx, item.NodeName, fmt.Sprintf("launcher pod %s changed", item.LauncherPodName))
-	return nil, false
+	return nil, syncRetry
 }
 
 func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *nodeData) (error, bool) {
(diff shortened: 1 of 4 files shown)


pkg/controller/dual-pods/inference-server.go
```diff
@@ -350,6 +361,58 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+					return fmt.Errorf("failed to get instance state from launcher: %w", err), true
+				}
+			} else if instState.Status == InstanceStatusStopped {
+				logger.V(2).Info("Bound instance is stopped", "instanceID", serverDat.InstanceID)
+				instanceStopped = true
+			}
+			if instanceStopped {
+				// Clean up the stopped instance from the launcher.
+				_, delErr := lClient.DeleteInstance(ctx, serverDat.InstanceID)
+				if delErr != nil && !IsLauncherNotFoundError(delErr) {
```
MikeSpreitzer: I am getting more unhappy with the name "IsLauncherNotFoundError". The meaning here is _instance_ not found.

Maybe just make it "IsNotFoundError", and really just check for HTTP 404?

Can be addressed in a later PR.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -350,6 +361,58 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+					Preconditions:     &metav1.Preconditions{UID: &item.UID, ResourceVersion: &requestingPod.ResourceVersion}})
+				if err == nil {
+					logger.V(2).Info("Requested deletion of server-requesting Pod because bound instance stopped")
+				} else if apierrors.IsGone(err) || apierrors.IsNotFound(err) {
+					logger.V(5).Info("The server-requesting Pod is already gone")
+				} else {
+					return fmt.Errorf("failed to delete server-requesting Pod for stopped instance: %w", err), true
+				}
+				serverDat.RequesterDeleteRequested = true
+				return nil, false
```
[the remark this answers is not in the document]
MikeSpreitzer: I think that @copilot has a valid point here.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1337,14 +1400,43 @@ func (ctl *controller) syncLauncherInstances(ctx context.Context, nodeDat *nodeD
 
 	launcherDat := ctl.getLauncherData(nodeDat, launcherPod.Name)
 	newInstances := make(map[string]time.Time)
+	remainingInstances := make([]InstanceState, 0, len(insts.Instances))
 	for _, inst := range insts.Instances {
+		if inst.Status == InstanceStatusStopped {
+			// Clean up stopped instances from the launcher.
+			_, delErr := lClient.DeleteInstance(ctx, inst.InstanceID)
```
MikeSpreitzer, line 1413: This is repeating some of the logic in `infSvrItem.process`, and I do not understand why it is correct to omit the other associated logic (e.g., deleting the server-requesting Pod) (when syncLauncherInstances is called from `launcherPodItem.process`).

I think that it would be simpler and easier to understand if there is just one copy of the handling logic, and it is in `infSvrItem.process`. Let `launcherPodItem.process` enqueue `infSvrItem`s that need such processing. Yes, that means that `launcherData` has to grow to hold the current state of each instance. Yes, that means that `launcherClient.CreateNamedInstance` has to return the full instance state (it will be easy, the launcher's creation operations now return the full instance state).

I do not think that `selectBestLauncherPod` needs the stopped instances to be filtered out; their stopped status should clearly mark them for correct handling by `selectBestLauncherPod`.

pkg/controller/dual-pods/inference-server.go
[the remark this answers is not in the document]
waltforme: Addressed. Now the controller deletes a requester only if it's confirmed that the corresponding instance does not exist.

pkg/controller/dual-pods/inference-server.go
waltforme: Now there is no repeating logic.

Now the only logic is inside `syncLauncherInstances`. The logic can't be put in `infSvrItem.process` because an unbound launcher pod does not contribute to any `infSvrItem`.

Deleting a requester is only necessary for a bound instance. When an unbound launcher pod is cleaned up, there is no requester to delete.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -184,6 +189,12 @@ func (c *LauncherClient) create(
 	return &out, nil
 }
 
+// IsLauncherNotFoundError returns true if the error from the launcher client
+// indicates a 404 Not Found response (e.g., instance does not exist).
+func IsLauncherNotFoundError(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "launcher error 404:")
+}
```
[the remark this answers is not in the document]
MikeSpreitzer: I agree on a less brittle detection, along with renaming this func (whose meaning is NOT "launcher not found").

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1337,14 +1407,48 @@ func (ctl *controller) syncLauncherInstances(ctx context.Context, nodeDat *nodeD
...
+					"instanceID", inst.InstanceID, "err", delErr)
+				// Deletion failed — the instance still occupies a slot in the launcher.
+				// Include it in remainingInstances so capacity accounting stays accurate.
+				failedStoppedInstanceErrs[inst.InstanceID] = delErr
+			} else {
+				logger.V(2).Info("Deleted stopped instance from launcher during sync",
+					"instanceID", inst.InstanceID)
+				deletedStoppedInstanceIDs.Insert(inst.InstanceID)
+			}
+			if delErr == nil || IsLauncherNotFoundError(delErr) {
```
MikeSpreitzer: Wouldn't it be simpler if these two lines weren't here?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1337,14 +1407,48 @@ func (ctl *controller) syncLauncherInstances(ctx context.Context, nodeDat *nodeD
...
 
+	// Replace the returned instance list and counts with the filtered view
+	// so that callers (e.g. selectBestLauncherPod) see accurate capacity.
+	insts.Instances = remainingInstances
+	insts.TotalInstances = len(remainingInstances)
+	runningCount := 0
+	for _, inst := range remainingInstances {
+		if inst.Status == "running" {
+			runningCount++
+		}
```
MikeSpreitzer: FYI, this could be folded into the earlier loop over `insts.Instances`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -100,8 +106,18 @@ func (item launcherPodItem) process(ctx context.Context, ctl *controller, nodeDa
 		return err, true
 	}
 
+	// Sync launcher instances to keep internal state fresh and clean up stopped instances.
+	_, syncErr, syncRetry := ctl.syncLauncherInstances(ctx, nodeDat, launcherPod)
+	if syncErr != nil {
+		if syncRetry {
+			logger.V(4).Info("Failed to sync launcher instances, will retry", "err", syncErr)
```
MikeSpreitzer: This is redundant with https://github.com/waltforme/llm-d-fast-model-actuation/blob/03254d8938f25ea2a51f9ab9b989a809b428e040/pkg/controller/dual-pods/inference-server.go#L78, right? Would wrapping the error be better?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -561,7 +622,6 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	}
 	// Sleeper budget is met. Make a new launcher Pod.
 
-	// TODO(waltforme): introduce the 'enqueue requesters by launcher pods' logic to the controller.
 	echo, err := podOps.Create(ctx, desiredLauncherPod, metav1.CreateOptions{})
```
MikeSpreitzer, line 621: This problem existed before this PR, but it becomes more important with this PR: I think that the launcher should be created in a bound state. Otherwise the launcher population controller might delete the launcher before it gets bound.

MikeSpreitzer:
Commit 48664fefec7b4fd306aea997c3da9e0d8b87522d ("Don't delete requester if deletion of a stopped instance fails") does not have a DCO sign-off.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1337,14 +1407,48 @@ func (ctl *controller) syncLauncherInstances(ctx context.Context, nodeDat *nodeD
...
 	for _, inst := range insts.Instances {
+		if inst.Status == InstanceStatusStopped {
+			// Clean up stopped instances from the launcher.
+			_, delErr := lClient.DeleteInstance(ctx, inst.InstanceID)
+			if delErr != nil && !IsLauncherNotFoundError(delErr) {
+				logger.V(3).Info("Failed to delete stopped instance from launcher during sync",
+					"instanceID", inst.InstanceID, "err", delErr)
+				// Deletion failed — the instance still occupies a slot in the launcher.
+				// Include it in remainingInstances so capacity accounting stays accurate.
+				failedStoppedInstanceErrs[inst.InstanceID] = delErr
```
MikeSpreitzer, line 1418: Copilot's objection is still valid. By not putting this instance in `remainingInstances` in this case, this causes `launcherDat.Instances` to omit the instance --- which is inaccurate. Currently nothing _consumes_ `launcherDat.Instances`, but clearly we hope to change that in the future. We should not put a logic time-bomb in here now.

waltforme: When this comment was made, the issue raised by Copilot had already been addressed. Would you like to double-check?

pkg/controller/dual-pods/inference-server.go
waltforme: Fixed.

pkg/controller/dual-pods/inference-server.go
waltforme: Folded.

pkg/controller/dual-pods/launcherclient.go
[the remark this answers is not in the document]
waltforme: Addressed.

pkg/controller/dual-pods/inference-server.go

pkg/controller/dual-pods/inference-server.go
waltforme: Wrapped.

pkg/controller/dual-pods/inference-server.go
waltforme: Why does the separate issue become more important with this PR?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -350,6 +363,52 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+			if err != nil || retry {
+				if err != nil {
+					return fmt.Errorf("failed to sync launcher instances for bound launcher Pod: %w", err), true
+				}
+				return nil, true
+			}
+
+			_, instancePresent := findInstanceState(syncResult.instances.Instances, serverDat.InstanceID)
+			if delErr, failedCleanup := syncResult.failedStoppedInstanceErrs[serverDat.InstanceID]; failedCleanup {
+				return ctl.ensureReqStatus(ctx, requestingPod, serverDat,
```
MikeSpreitzer: The case of a failed instance deletion calls for retry, not note-and-give-up.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -350,6 +363,52 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+					fmt.Sprintf("failed to delete stopped instance %q from launcher: %s", serverDat.InstanceID, delErr.Error()))
+			}
+			if _, deletedStopped := syncResult.deletedStoppedInstanceIDs[serverDat.InstanceID]; deletedStopped || !instancePresent {
+				if deletedStopped {
+					logger.V(2).Info("Deleted stopped bound instance from launcher during sync", "instanceID", serverDat.InstanceID)
```
MikeSpreitzer: The "instanceID" was added to the logger way up near the start of `infSvrItem.process`, no need to repeat it at this point (nor in the `else` case) (`serverDat.InstanceID` will not have changed yet).

pkg/controller/dual-pods/inference-server.go
```diff
@@ -350,6 +363,52 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+				// Mark as sleeping so that ensureUnbound (called during requester deletion)
+				// does not attempt to POST /sleep on the dead instance.
+				// The instance process is dead — this is not a real sleeping state,
+				// but it prevents ensureUnbound from hitting a dead endpoint and retrying forever.
+				serverDat.Sleeping = ptr.To(true)
+				// Delete the server-requesting Pod.
+				// This is analogous to the direct-provider case where a providing Pod's
+				// deletion is reflected to deletion of the server-requesting Pod.
+				// The ReplicaSet will recreate the requesting Pod, triggering a fresh bind.
+				err = podOps.Delete(ctx, requestingPod.Name, metav1.DeleteOptions{
```
MikeSpreitzer, line 397: This deletion will not complete until this very controller removes its finalizer, which it does earlier in this func --- thus, the current code structure requires a round trip to get the finalizer removed. Wouldn't it be better to move the detection of instance stop/absence up so that it can contribute to the test currently in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/1811815361d97472a3da67cbde001448fdfdf0ac/pkg/controller/dual-pods/inference-server.go#L184 ?

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: This PR made me think more about that issue. But I do not think that this PR makes that case more likely to happen. I think that it is fine to address this issue in another PR.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: @waltforme: you're right.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -34,9 +35,22 @@ type LauncherClient struct {
 	httpClient *http.Client
 }
 
+type launcherError struct {
+	StatusCode int
+	Body       string
+}
+
+func (e *launcherError) Error() string {
```
MikeSpreitzer, line 43: FYI, this method could have been defined as `func (e launcherError) ...`; the level of pointer indirection is not necessary or particularly helpful. These values are going to get boxed and put on the heap anyway, so an additional level of heaping is more expensive at runtime than it is worth. Not a big deal.

pkg/controller/dual-pods/inference-server.go
waltforme: Done.

pkg/controller/dual-pods/inference-server.go
waltforme: Removed.

pkg/controller/dual-pods/inference-server.go
waltforme: Expanding the test (thus moving the logic into the proposed `if` block) wouldn't be better because the logic of that `if` block itself also needs a round trip to remove the finalizer.

MikeSpreitzer: What I am suggesting is to move the logic that fetches the instance state above the test for deleted provider and change that test to be deleted provider Pod OR instance used to be running but is now stopped/deleted.

MikeSpreitzer: The finalizer removal is an unconditionally added round trip.

waltforme: > The finalizer removal is an unconditionally added round trip.

This is exactly what confused me. Moving doesn’t affect the finalizer or the round trip — so why does the comment start by talking about them?

waltforme: That wouldn’t be cleaner, because the logic for the “direct” provider and the launcher don’t overlap well.

> What I am suggesting is to move the logic that fetches the instance state above the test for deleted provider and change that test to be deleted provider Pod OR instance used to be running but is now stopped/deleted.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -350,6 +363,51 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 			}
 		}
+		// For launcher-based providers, check whether the bound instance is still alive.
+		// The sidecar notifier updates the Pod annotation when instance status changes,
+		// which triggers reconciliation through the informer.
+		if launcherBased && serverDat.InstanceID != "" && providingPod.Status.PodIP != "" {
+			syncResult, err, retry := ctl.syncLauncherInstances(ctx, nodeDat, providingPod)
+			if err != nil || retry {
+				if err != nil {
+					return fmt.Errorf("failed to sync launcher instances for bound launcher Pod: %w", err), true
```
MikeSpreitzer: I would return `retry` rather than `true` here.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: The `if` statement at https://github.com/waltforme/llm-d-fast-model-actuation/blob/17c9377fdc6eaf20fe574d9bde061ca427e111f2/pkg/controller/dual-pods/inference-server.go#L197-L228 applies to both a launcher-based provider and a direct provider.

The general flow of `infSvrItem.process` has been handling the case of reflecting deletion of vllm instance into deletion of requester Pod in that `if` statement. My suggestion was to keep that there. That would require expanding the test in that `if` to also catch the case of a goner (`deletedStopped || !instancePresent`) vllm instance in a launcher. It would also require conditioning https://github.com/waltforme/llm-d-fast-model-actuation/blob/17c9377fdc6eaf20fe574d9bde061ca427e111f2/pkg/controller/dual-pods/inference-server.go#L201-L206 to not happen in this new case, since that possible finalizer _addtion_ is for the milestone 2 pathway of keeping the requester Pod in existence while the provider Pod is being deleted.

I think that it is a reasonable alternative to do as the current version of this PR does, reflect stop/removal of a vllm instance of a launcher into a request for deletion of the server-providing Pod and let the existing logic respond as it does to a server-requesting Pod going through deletion.

MikeSpreitzer (commented):
The changes here are good enough to commit, but one of the commits (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/414/changes/48664fefec7b4fd306aea997c3da9e0d8b87522d) lacks a DCO sign-off.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: I opened #421 for this.

waltforme:
The force-push to https://github.com/llm-d-incubation/llm-d-fast-model-actuation/commit/eca4a3a3a4ae3d379a6ccb2d4ca08f62df0819a2 adds the missing DCO sign-off. There is no code change.

pkg/controller/dual-pods/inference-server.go
waltforme: The 'existing logic' and the logic introduced by this PR don't overlap well. The statement '...applies to both a launcher-based provider and a direct provider' is incorrect: An immediate example is that lines 219-227 don't apply to the launcher-based cases. Also, the loggings will need to be conditional. There is only a few lines of shared code.

MikeSpreitzer (approved):
LGTM

PR #417, opened 2026-04-10
Add FYI instance label on server-requesting Pods

MikeSpreitzer:
## Summary

- Add a purely-FYI label `dual-pods.llm-d.ai/instance` on server-requesting Pods carrying the launcher's vLLM instance ID, to aid debugging and observability
- The label is present while bound to a launcher-based server-providing Pod and absent otherwise
- Wrap the ISC hash with "I" prefix and "i" suffix so it is always a valid Kubernetes label value (must start/end with alphanumeric)

Resolves #398

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Existing unit tests pass
- [ ] Verify in a launcher-based e2e test that the requesting pod gets the `dual-pods.llm-d.ai/instance` label after binding

### DIFF
--- a/.github/workflows/launcher-based-e2e-test.yml
+++ b/.github/workflows/launcher-based-e2e-test.yml
@@ -62,7 +62,7 @@ jobs:
 
       - name: show test pods with labels
         if: always()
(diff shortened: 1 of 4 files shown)


MikeSpreitzer:
Opened from wrong repo; need to use shared repo so that the updated workflows can be tested.

PR #418, opened 2026-04-10
Add FYI instance label on server-requesting Pods

MikeSpreitzer:
- Add a purely-FYI label `dual-pods.llm-d.ai/instance` on server-requesting Pods carrying the launcher's vLLM instance ID, to aid debugging and observability
- The label is present while bound to a launcher-based server-providing Pod and absent otherwise
- Wrap the ISC hash with "I" prefix and "i" suffix so it is always a valid Kubernetes label value (must start/end with alphanumeric)

Resolves #398

### DIFF
--- a/.github/workflows/launcher-based-e2e-test.yml
+++ b/.github/workflows/launcher-based-e2e-test.yml
@@ -62,7 +62,7 @@ jobs:
 
       - name: show test pods with labels
         if: always()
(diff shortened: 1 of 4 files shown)


aavarghese (approved):
Code updates lgtm.  This label is useful for the FMA benchmark instead of steps to query the launcher API...
For nominalHash, were the I and i randomly chosen?

MikeSpreitzer:
The hash had to be wrapped with _some_ alphanumeric to make sure that it is a valid label value. I chose "I" to stand for instance. Capital at the start, because that's English, lowercase at the end for contrast and complementarity.

PR #423, opened 2026-04-14
Enable reverse proxy in server-requesting pod

delavet:
Based on our discussion in #363, I have split PR #363 and am resubmitting this PR to introduce the first part of the changes: adding a reverse proxy for server-requesting pods.
I have re-conducted the reverse proxy overhead experiments, primarily updating the experimental parameter configurations and adding monitoring and analysis of pod resource usage. For details, please refer to the documentation below.

https://docs.google.com/document/d/1qX4KtcTJEfdatgVmtsveOMfjm883MG-HW-e7ddmlqkA/edit?usp=sharing

The part on dynamic port allocation in PR #363 can be considered for addition after Milestone 3.


pkg/spi/interface.go
```diff
@@ -59,3 +59,15 @@ const SetLogPath = "/v1/set-log"
 // LogStartPosParam is the name of the query parameter that
 // holds that starting position of a log chunk.
 const LogStartPosParam = "startPos"
+
+// InitProxy is the path for initializing the HTTP reverse proxy。
```
aavarghese: There's a strange period at the end here...

cmd/requester/main.go
aavarghese: Must change to wg.Add(3)?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -574,6 +574,15 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return fmt.Errorf("launcher Pod %q has no IP assigned yet", launcherPod.Name), true
 			}
 
+			// Initialize the reverse proxy between the launcher Pod and the requester Pod.
+			// Requests can be proxied to the launcher Pod from the requester Pod.
+			url := fmt.Sprintf("http://%s:%s%s", requestingPod.Status.PodIP, adminPort, stubapi.InitProxy)
+			if err := doPostWithData(url, bytes.NewReader([]byte(fmt.Sprintf("{\"address\":\"%s\",\"port\":%d}",
+				launcherIP, desiredPort)))); err != nil {
+				logger.Error(err, "Failed to initialize requester proxy")
```
rubambiza: Perhaps we should include the requesting pod `ip:port` pair in the error for troubleshooting an environment with multiple requesters?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,187 @@
...
+
+// ConfigRequest is the request body to configure the proxy target
+type ConfigRequest struct {
+	Address string `json:"address"`
+	Port    int    `json:"port"`
+}
+
+// proxy is a lazy HTTP reverse proxy that only starts after receiving
+// the first configuration request
+type proxy struct {
```
rubambiza: The type being `proxy` and having a field named `proxy` is a bit confusing. I'd like to suggest making the type more specific to what it does in FMA, though I don't have an appropriate name off the top of my head.

rubambiza (commented):
A few suggested changes for debugging and code readability. Not major blockers.

pkg/spi/interface.go
```diff
@@ -59,3 +59,15 @@ const SetLogPath = "/v1/set-log"
...
+// The proxy is used to forward requests from the server-requesting
+// pod to the server-providing pod.
+// Supports two HTTP methods:
+//   - GET: retrieves the initialization status of the proxy.
+//     Returns  status info.
+//   - POST: initializes the proxy with a target address and port.
+//     The request body should contain a JSON object with "address"
+//     and "port" fields. After successful initialization,
+//     the proxy will forward requests to the configured target server.
+const InitProxy = "/v1/proxy/init"
```
aavarghese: Dumb question but is it possible for this single proxy server to route requests to the launcher mgmt API (/v2/vllm/instances....*) running on port 8001 to get status of the vllm instances, get logs etc in addition to the actual vllm's inference endpoints?

cmd/requester/main.go
```diff
@@ -72,5 +78,14 @@ func main() {
 		}
 	}()
 
+	// Start the reverse proxy server
+	go func() {
+		defer wg.Done()
+		err := proxy.Run(ctx, proxyPort)
+		if err != nil {
+			logger.Error(err, "failed to start requester proxy server")
```
MikeSpreitzer: "start" or "run"?

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: The error message should include _both_ host:port pairs.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,187 @@
+/*
+Copyright 2025 The llm-d Authors.
```
MikeSpreitzer: New files should be born with the current year here.

pkg/spi/interface.go
MikeSpreitzer: The requester should not expose information about more than itself and its corresponding vllm instance.

pkg/spi/interface.go
MikeSpreitzer: Yes. Please change it to a regular period (character code 0x2E).

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,187 @@
...
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+// ConfigRequest is the request body to configure the proxy target
+type ConfigRequest struct {
+	Address string `json:"address"`
+	Port    int    `json:"port"`
+}
```
MikeSpreitzer: This should be declared in a common place where the dual-pods controller can use it too.

pkg/server/requester/proxy/server.go
MikeSpreitzer: I agree. But I think that this is moot because it should be a TCP proxy rather than an HTTP proxy.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,187 @@
...
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", serveProxy)
+
+	server := &http.Server{
+		Addr:         fmt.Sprintf(":%s", port),
+		Handler:      mux,
+		ReadTimeout:  30 * time.Second,
+		WriteTimeout: 5 * time.Minute, // Long timeout for inference requests
+		IdleTimeout:  120 * time.Second,
```
MikeSpreitzer: Maybe these two timeouts should be configurable?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,187 @@
...
+// the first configuration request
+type proxy struct {
+	mu          sync.RWMutex
+	targetURL   *url.URL
+	proxy       *httputil.ReverseProxy
+	initialized atomic.Bool
+}
+
+// singleton instance initialized once at startup
+var instance = &proxy{}
```
MikeSpreitzer: The level of pointer indirection here is just unhelpful complexity. It could be just

```go
var instance proxy
```

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,187 @@
...
+	if r.Method != http.MethodPost {
+		http.Error(w, "invalid method", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Try initialize server
+	if instance.initialized.Load() {
+		http.Error(w, "proxy already initialized", http.StatusConflict)
+		return
+	}
```
MikeSpreitzer: This code is unnecessary complexity. Just delete it.

docs/dual-pods.md
```diff
@@ -125,6 +125,39 @@ assigned to the server-requesting Pod) for running `vllm serve`. To
...
+1. **Initialization**: When the dual-pods controller binds a
+   server-requesting Pod to a server-providing Pod, it sends an HTTP
+   POST request to the requester's proxy initialization endpoint
+   (`/v1/proxy/init`). The request body contains the target address
+   (launcher Pod IP) and the allocated port:
+
+   ```json
+   {"address": "10.244.1.5", "port": 8005}
+   ```
+
```
MikeSpreitzer, line 153: The proxy also implements `GET /v1/proxy/init`, and that should be documented too.

docs/dual-pods.md
```diff
@@ -125,6 +125,39 @@ assigned to the server-requesting Pod) for running `vllm serve`. To
...
+The requester container includes a reverse proxy server that forwards
+inference requests to the actual vLLM instance running in the
+server-providing Pod (typically managed by the launcher). This
+abstraction allows clients to send requests to the server-requesting
+Pod without needing to know the actual port vLLM is listening on.
+
+The reverse proxy operates as follows:
+
+1. **Initialization**: When the dual-pods controller binds a
```
MikeSpreitzer: Since there is also a `GET` on the same path, the design should be modified to conform to REST. Define a schema for the resource at `/v1/proxy/init` (or maybe just `/v1/proxy` or `/v1/proxy/config`?), and define PUT and GET to write and read this resource. GET when uninitialized returns HTTP status 404.

docs/dual-pods.md
```diff
@@ -125,6 +125,39 @@ assigned to the server-requesting Pod) for running `vllm serve`. To
 swap a model out, the controller issues a request that does not
 include those details.
 
+#### Requester Reverse Proxy
+
```
MikeSpreitzer, line 129: HTTP is too specific. It forecloses things like HTTPS and HTTP 2 or 3. The proxy should be simply a TCP proxy.

MikeSpreitzer (commented):
I have some comments on the study in https://docs.google.com/document/d/1qX4KtcTJEfdatgVmtsveOMfjm883MG-HW-e7ddmlqkA

"20 req/sec" is cited as a concurrency level, but that is actually a rate. Please be accurate and give a short sharp statement of the behavior of the benchmarking client(s).

Does every request go on a new TCP connection? If not then there is an additional thing to measure, the added latency in TCP connection setup. The measurement would be something like time from (a) client sending request to open connection to (b) client receiving first token from first request.

When doing performance studies we usually pay attention to the distribution of the result. Stuff like average, median, and high percentiles (90, 95, 99, 99.9). Of course, for high percentiles you need enough cycles to get statistically significant results.

pkg/spi/interface.go
aavarghese: Fair point!

cmd/requester/main.go
delavet: IMO, “start” is used in error messages (describing the action of failing to start the server). “Run” as a function name indicates a blocking call—the server has started and continues running.
Both terms are semantically correct: the Run() function is invoked to start the server.

delavet:
> I have some comments on the study in https://docs.google.com/document/d/1qX4KtcTJEfdatgVmtsveOMfjm883MG-HW-e7ddmlqkA
> 
> "20 req/sec" is cited as a concurrency level, but that is actually a rate. Please be accurate and give a short sharp statement of the behavior of the benchmarking client(s).
> 
> Does every request go on a new TCP connection? If not then there is an additional thing to measure, the added latency in TCP connection setup. The measurement would be something like time from (a) client sending request to open connection to (b) client receiving first token from first request.
> 
> When doing performance studies we usually pay attention to the distribution of the result. Stuff like average, median, and high percentiles (90, 95, 99, 99.9). Of course, for high percentiles you need enough cycles to get statistically significant results.

Thanks! I’ve just refactored the proxy into a pure TCP proxy. I still need a bit more time to set up a new experiment and investigate these issues.

aavarghese:
Can we add a new test case to our e2e suite [here ](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/main/test/e2e/test-cases.sh#L439) right after `Multiple Instances Share One Launcher`?

Claude generated test below:
```
# ---------------------------------------------------------------------------
# Reverse Proxy Initialization and Forwarding
# ---------------------------------------------------------------------------

intro_case Reverse Proxy Initialization and Forwarding

# Verify the proxy is initialized and pointing at launcher1's IP
kubectl port-forward pod/"$req3" 28091:8081 -n "$NS" &
PF_SPI_PID=$!
sleep 2

proxy_resp=$(curl -sf "http://localhost:28091/v1/proxy/init" 2>/dev/null || echo "")
kill "$PF_SPI_PID" 2>/dev/null || true
wait "$PF_SPI_PID" 2>/dev/null || true

if ! echo "$proxy_resp" | grep -q "proxying to"; then
    echo "ERROR: expected proxy to be initialized, got: '$proxy_resp'" >&2
    exit 1
fi
echo "Proxy is initialized: $proxy_resp"

launcher1_ip=$(kubectl get pod "$launcher1" -n "$NS" -o jsonpath='{.status.podIP}')
if ! echo "$proxy_resp" | grep -qF "$launcher1_ip"; then
    echo "ERROR: proxy target does not contain launcher IP $launcher1_ip: '$proxy_resp'" >&2
    exit 1
fi
echo "Proxy target matches launcher IP: $launcher1_ip"

# On OpenShift the launcher runs a real vLLM — verify traffic actually flows
# through the TCP proxy port (8082) to the launcher's vLLM /health endpoint.
if [ "$E2E_PLATFORM" = "openshift" ]; then
    kubectl port-forward pod/"$req3" 28092:8082 -n "$NS" &
    PF_PROXY_PID=$!
    sleep 2

    health_status=$(curl -s -o /dev/null -w "%{http_code}" \
        "http://localhost:28092/health" 2>/dev/null || echo "000")
    kill "$PF_PROXY_PID" 2>/dev/null || true
    wait "$PF_PROXY_PID" 2>/dev/null || true

    if [ "$health_status" != "200" ]; then
        echo "ERROR: vLLM /health via proxy port returned $health_status (expected 200)" >&2
        exit 1
    fi
    echo "Proxy forwarding verified: /health via proxy port → 200"
fi

cheer Successful reverse proxy initialization and forwarding
```

cmd/requester/main.go
MikeSpreitzer: If the `Run` function really is blocking (and I mention this because Kubernetes code has a bad habit of using the wrong name) then saying "start" in the error message is, at best, risky code: it depends on `Run` only erring during the start phase --- and violations of this expectation are not checked (except by you, too late, when you are debugging). It would be correct to say "failed to run" in response to an error from `Run`.

delavet:
> Can we add a new test case to our e2e suite [here ](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/main/test/e2e/test-cases.sh#L439) right after `Multiple Instances Share One Launcher`?
> 
> Claude generated test below:
> 
> ```
> # ---------------------------------------------------------------------------
> # Reverse Proxy Initialization and Forwarding
> # ---------------------------------------------------------------------------
> 
> intro_case Reverse Proxy Initialization and Forwarding
> 
> # Verify the proxy is initialized and pointing at launcher1's IP
> kubectl port-forward pod/"$req3" 28091:8081 -n "$NS" &
> PF_SPI_PID=$!
> sleep 2
> 
> proxy_resp=$(curl -sf "http://localhost:28091/v1/proxy/init" 2>/dev/null || echo "")
> kill "$PF_SPI_PID" 2>/dev/null || true
> wait "$PF_SPI_PID" 2>/dev/null || true
> 
> if ! echo "$proxy_resp" | grep -q "proxying to"; then
>     echo "ERROR: expected proxy to be initialized, got: '$proxy_resp'" >&2
>     exit 1
> fi
> echo "Proxy is initialized: $proxy_resp"
> 
> launcher1_ip=$(kubectl get pod "$launcher1" -n "$NS" -o jsonpath='{.status.podIP}')
> if ! echo "$proxy_resp" | grep -qF "$launcher1_ip"; then
>     echo "ERROR: proxy target does not contain launcher IP $launcher1_ip: '$proxy_resp'" >&2
>     exit 1
> fi
> echo "Proxy target matches launcher IP: $launcher1_ip"
> 
> # On OpenShift the launcher runs a real vLLM — verify traffic actually flows
> # through the TCP proxy port (8082) to the launcher's vLLM /health endpoint.
> if [ "$E2E_PLATFORM" = "openshift" ]; then
>     kubectl port-forward pod/"$req3" 28092:8082 -n "$NS" &
>     PF_PROXY_PID=$!
>     sleep 2
> 
>     health_status=$(curl -s -o /dev/null -w "%{http_code}" \
>         "http://localhost:28092/health" 2>/dev/null || echo "000")
>     kill "$PF_PROXY_PID" 2>/dev/null || true
>     wait "$PF_PROXY_PID" 2>/dev/null || true
> 
>     if [ "$health_status" != "200" ]; then
>         echo "ERROR: vLLM /health via proxy port returned $health_status (expected 200)" >&2
>         exit 1
>     fi
>     echo "Proxy forwarding verified: /health via proxy port → 200"
> fi
> 
> cheer Successful reverse proxy initialization and forwarding
> ```

I’ve implemented a connection pooling and reuse mechanism for the TCP proxy, eliminating the need for frequent TCP connection establishment. Additionally, I’ve added e2e test as per your suggestion. Could you please review the changes again? Thanks a lot!

delavet:
I am still optimizing and re-running the experiments, and I expect to share the results of this new implementation within this week.

cmd/requester/main.go
```diff
@@ -72,5 +88,14 @@ func main() {
 		}
 	}()
 
+	// Start the reverse proxy server
+	go func() {
+		defer wg.Done()
+		err := proxy.RunWithConfig(ctx, proxyPort, proxyConfig)
+		if err != nil {
+			logger.Error(err, "failed to start requester proxy server")
```
MikeSpreitzer: I sill see "start" here (and in the test-requester) rather than "run".

docs/dual-pods.md
```diff
@@ -125,6 +125,45 @@ assigned to the server-requesting Pod) for running `vllm serve`. To
...
+   ```json
+   {"address": "10.244.1.5", "port": 8005}
+   ```
+
+   A successful response returns HTTP 200. If the proxy is already
+   configured, the request returns HTTP 409 (Conflict).
+
+2. **Status checking**: The proxy's configuration status can be
+   queried via an HTTP GET request to `/v1/proxy/init`. If the proxy
+   has been configured, it returns HTTP 200 with the target address.
```
MikeSpreitzer: Only address, not port too?
In REST, PUT and GET manipulate the same resource, same schema.

docs/dual-pods.md
```diff
@@ -125,6 +125,45 @@ assigned to the server-requesting Pod) for running `vllm serve`. To
...
+   (`/v1/proxy/init`). The request body contains the target address
+   (launcher Pod IP) and the allocated port:
+
+   ```json
+   {"address": "10.244.1.5", "port": 8005}
+   ```
+
+   A successful response returns HTTP 200. If the proxy is already
+   configured, the request returns HTTP 409 (Conflict).
+
```
MikeSpreitzer, line 157: We might also want a way to unbind. Can be addressed in a follow-on PR.

pkg/spi/interface.go
```diff
@@ -59,3 +59,21 @@ const SetLogPath = "/v1/set-log"
...
+// Supports two HTTP methods:
+//   - GET: retrieves the initialization status of the proxy.
+//     Returns 200 if initialized, 404 if not.
+//   - PUT: configures the proxy with a target address and port.
+//     The request body should contain a JSON object with "address"
+//     and "port" fields. After successful configuration,
+//     the proxy will forward TCP connections to the configured target.
+const InitProxy = "/v1/proxy/init"
+
+// ProxyConfigRequest is the request body to configure the proxy target
```
MikeSpreitzer: I hope that this type describes responses as well as requests. I would call it something simpler, like `ProxyConfig`.

pkg/spi/interface.go
```diff
@@ -59,3 +59,21 @@ const SetLogPath = "/v1/set-log"
...
+type ProxyConfigRequest struct {
+	Address string `json:"address"`
+	Port    int    `json:"port"`
```
MikeSpreitzer: I would type this `int16`, because port numbers are 16 bits long.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1005,6 +1005,27 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
 			return fmt.Errorf("unable to wake up server because port not known: %w", err), true
 		}
 	}
+	// Initialize a reverse proxy between the launcher Pod and the requester Pod.
+	// Requests can be proxied to the launcher Pod from the requester Pod.
+	// This could be done during binding.
```
MikeSpreitzer, line 1010: In my opinion, doing it in `bind` is too early; the vllm instance is not ready yet. Either the inference client is waiting for the server-requesting Pod to be ready or the client will try inference requests too soon and get errors. The right time to do the proxy binding is just before the relay of positive readiness, which starts at line 509 in the current revision of this PR.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1615,21 +1636,35 @@ func (ctl *controller) ensureReqState(ctx context.Context, requestingPod *corev1
...
+}
+
+func doPut(url string, data io.Reader) error {
+	return doHTTPRequest(http.MethodPut, url, data)
+}
+
+func doHTTPRequest(method, url string, data io.Reader) error {
+	req, err := http.NewRequest(method, url, data)
+	if err != nil {
+		return fmt.Errorf("http %s %q: %w", strings.ToUpper(method), url, err)
```
MikeSpreitzer: This func is never called with lowercase method names, no need to use `strings.ToUpper` here and later.

pkg/server/requester/proxy/pool.go
MikeSpreitzer, line 1: So by "connection pooling" you mean re-use of connections from the proxy to vllm. When one connection to the proxy ends, a following connection to the proxy might re-use the same connection from proxy to vllm.

Is this really something that we want? Is it safe? Will it confuse anything?

pkg/server/requester/proxy/pool.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// idle connections protected by mu
+	mu   sync.Mutex
+	idle []*idleConn
+
+	// configuration
+	maxIdle  int
+	idleTTL  time.Duration
+	dialer   *net.Dialer
+	onDial   func(ctx context.Context, network, addr string) (net.Conn, error)
```
MikeSpreitzer: Why is this here? I see no writes to this field.

pkg/server/requester/proxy/pool.go
```diff
@@ -0,0 +1,333 @@
...
+		_ = conn.Close()
+		p.totalDiscards.Add(1)
+		return
+	}
+
+	p.idle = append(p.idle, &idleConn{conn: conn, idleSince: time.Now()})
+}
+
+// drain closes all idle connections and resets the pool.
+// Must be called with the target lock held (which SetTarget does).
```
MikeSpreitzer: No, SetTarget does _not_ call this with a mutex held.

pkg/server/requester/proxy/pool.go
```diff
@@ -0,0 +1,333 @@
...
+// PoolConfig holds configuration for the connection pool
+type PoolConfig struct {
+	// MaxIdleConnections is the maximum number of idle connections to keep
+	MaxIdleConnections int
+	// IdleTimeout is how long a connection can sit idle before being closed
+	IdleTimeout time.Duration
+	// DialTimeout is the timeout for creating new connections
+	DialTimeout time.Duration
+	// CleanInterval is how often the background cleaner runs
+	CleanInterval time.Duration
```
MikeSpreitzer: Nothing reads this.

pkg/server/requester/proxy/pool.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	p.target.Store("")
+	return p
+}
+
+// SetTarget updates the target address and drains existing connections.
+// Returns the old target for logging purposes.
+func (p *connPool) SetTarget(target string) string {
+	old := p.target.Swap(target).(string)
+	p.drain()
```
MikeSpreitzer: I think that the mutually exclusive code block here includes the target change as well as the drain. Otherwise, you do not have the invariant that the target describes what the pooled connections are too.

With the current code, for example, the following is possible.

Goroutine A enters SetTarget and does the target swap. Then comes the call to p.drain. The first thing in there is a mutex entry. Goroutine A blocks for a while waiting on this mutex.

After goroutine A starts waiting but before it is resumed, other goroutines do Get and, a little later, do Put on those just-gotten connections --- causing the pool to hold some connections to the new target.

pkg/server/requester/proxy/pool.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	var buf [1]byte
+	_, err := tcpConn.Read(buf[:])
+	// Reset deadline
+	_ = tcpConn.SetReadDeadline(time.Time{})
+
+	if err == nil {
+		// Got data — connection is alive, but we consumed a byte.
+		// This is unlikely during idle health check but handle gracefully.
```
MikeSpreitzer: If this idle connection is returning any bytes at all, it will be a nasty surprise to the next client.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,284 @@
...
+
+// Run starts the TCP proxy server on the given port
+func Run(ctx context.Context, port string) error {
+	return RunWithConfig(ctx, port, DefaultProxyConfig)
+}
+
+// RunWithConfig starts the TCP proxy server with custom configuration
+func RunWithConfig(ctx context.Context, port string, cfg ProxyConfig) error {
+	logger := klog.FromContext(ctx).WithName("proxy-server")
+	logger.Info("starting TCP proxy server")
```
MikeSpreitzer: A log message should be a sentence, which starts with a capital letter.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// Create or update the connection pool
+	instance.mu.Lock()
+	instance.cfg = cfg.PoolConfig
+	if instance.pool == nil {
+		instance.pool = newConnPool(cfg.PoolConfig)
```
MikeSpreitzer: If `instance.pool != nil` this potentially makes `instance.cfg` inconsistent with `instance.pool`.

And what if the pool is not empty?

pkg/server/requester/proxy/pool.go
MikeSpreitzer: It seems to me that this definitely has functionality problems. It assumes that (a) there are no unsolicited bytes sent from the server and (b) the client and server have no state associated with the connection. I think that these are not valid assumptions.

Even if we know that the protocol is HTTP (without Upgrade), consider this possibility: the client sends a request to the server, then closes the connection. After the proxy unbinds and puts the connection in the idle pool, the server sends the HTTP response message.

Consider HTTP Upgrade. This is an overall protocol that starts out in HTTP then switches to some other protocol that is NOT based on HTTP messaging. Suppose that switch has been done, then the client closes the connection and the proxy saves the idle connection. The next client gets that connection that is now in non-HTTP mode but sends an HTTP message.

The protocol might not be HTTP.

delavet: This sounds reasonable. Indeed, introducing “connection pooling” at this stage may constitute over-engineering—it stemmed from my concern about the frequent establishment of TCP connections mentioned in the earlier comments. Before proceeding with this change, we should first conduct experiments to quantify the overhead of establishing TCP connections. I will therefore revert these changes and perform further experiments.

docs/dual-pods.md
delavet: I'll create a follow-on PR for this

pkg/controller/dual-pods/inference-server.go
delavet: I indeed intend to initialize the TCP proxy during the bind phase, as I believe we can subsequently modify the requester’s readiness probe to point to the TCP proxy’s endpoint. This way, health and readiness check requests would be directly forwarded to the backend vLLM endpoint, thereby reflecting the vLLM instance’s health and readiness status directly in the requester pod’s health and readiness status—without requiring any intervention from a controller. I’d appreciate your thoughts on this idea.

pkg/server/requester/proxy/server.go
delavet: Note these comments down first, in case we really need “connection pooling.”

pkg/spi/interface.go
delavet: Would using `uint16` be more appropriate here?

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: A PodSpec is almost entirely immutable (the only exceptions are for "vertical scaling").

pkg/spi/interface.go
MikeSpreitzer: I think so.

cmd/requester/main.go
```diff
@@ -48,10 +50,21 @@ func main() {
 		spiPort = "8081"
 	}
 
+	proxyPort := os.Getenv("PROXY_PORT")
+	if proxyPort == "" {
+		proxyPort = "8082"
+	}
```
MikeSpreitzer: And if not the empty string, maybe check that it correctly parses as a 16-bit integer? And print an error message and fail if not.

cmd/requester/main.go
```diff
@@ -48,10 +50,21 @@ func main() {
...
+
+	// Configure proxy timeouts from environment variables (with defaults)
+	proxyConfig := proxy.DefaultProxyConfig
+	if dur, err := time.ParseDuration(os.Getenv("PROXY_DIAL_TIMEOUT")); err == nil && dur != 0 {
+		proxyConfig.DialTimeout = dur
+	}
```
MikeSpreitzer: Also, I would print an error message and fail if `err != nil`.

We commonly use `github.com/spf13/pflag` to bundle up parsing and failing. And implementing `--help` (which gives users a reliable way to discover all the parameters, with their defaults and explanations).

cmd/requester/main.go
```diff
@@ -48,10 +50,21 @@ func main() {
...
+
 	var ready atomic.Bool
 
 	var wg sync.WaitGroup
-	wg.Add(2)
+	wg.Add(3)
```
MikeSpreitzer: It is not wrong to have a single `wg.Add` up front. However, the code is a little more localized if there is a `wg.Add(1)` right in front of each `go ...`. Not a big deal.

pkg/spi/interface.go
```diff
@@ -59,3 +59,20 @@ const SetLogPath = "/v1/set-log"
...
+// Supports two HTTP methods:
+//   - GET: retrieves the configuration status of the proxy.
+//     Returns 200 if configured, 404 if not.
+//   - PUT: configures the proxy with a target address and port.
+//     After successful configuration, the proxy will forward TCP
+//     connections to the configured target.
+const ProxyConfigPath = "/v1/proxy/config"
+
+// ProxyConfig is the request and response body for configuring the proxy.
+type ProxyConfig struct {
```
MikeSpreitzer: Oops! I am sorry, when I suggested calling this type "ProxyConfig" I was not taking into account the fact that this PR also introduces a different type that is also called "ProxyConfig". Both types are about configuring the proxy, so "ProxyConfig" is appropriate for each --- but not specific enough. One is configuration that is given once at startup, the other is configuration that is given once by the dual-pods controller. I recommend giving each a more specific name, perhaps "ProxyInitialConfig" and "ProxyDynamicConfig".

cmd/requester/main.go
```diff
@@ -72,5 +85,14 @@ func main() {
 		}
 	}()
 
+	// Start the reverse proxy server
+	go func() {
+		defer wg.Done()
+		err := proxy.RunWithConfig(ctx, proxyPort, proxyConfig)
```
MikeSpreitzer: Isn't the port number part of the proxy's configuration? Why not include the port number in the configuration struct?

Also the parsing and erring and parameter discovery gets simpler and more uniform if all the parameters are in one struct and that struct has a method to add command line parameters for all of its fields. Then both requesters can just invoke that one method.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,221 @@
...
+	dialer: &net.Dialer{},
+}
+
+// Run starts the TCP proxy server on the given port
+func Run(ctx context.Context, port string) error {
+	return RunWithConfig(ctx, port, DefaultProxyConfig)
+}
+
+// RunWithConfig starts the TCP proxy server with custom configuration
+func RunWithConfig(ctx context.Context, port string, cfg ProxyConfig) error {
```
MikeSpreitzer: In my opinion it is not worth the complexity to have two different "run"-ish methods. I recommend having just one, named "Run", which takes a config struct as a parameter. Callers will start with the Default...() method to construct a config object and then expose all its fields on the command line in a uniform way.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (f *forwarder) getTarget() string {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.target
+}
+
+// singleton instance initialized once at startup
+var instance = forwarder{
```
MikeSpreitzer: At the current level of ambition it is not wrong to have a global variable holding the proxy's state. But as soon as we want concurrent testing, that will be a problem. I suggest instead defining `New(config) *forwarder {...}` and making `Run` and `Configure` be methods of `*forwarder`. Not a big deal.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,221 @@
...
+			case <-ctx.Done():
+				logger.Info("TCP proxy server stopped")
+				return nil
+			default:
+				logger.Error(err, "Failed to accept connection")
+				continue
+			}
+		}
+
+		go handleConnection(ctx, conn, &instance, logger)
```
MikeSpreitzer: FYI, it is considered bad form to pass both logger and ctx. The usual form is to

```go
    ctx = klog.NewContext(ctx, logger)
```

and then pass just `ctx`. Not a big deal.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	// Start shutdown goroutine
+	go func() {
+		<-ctx.Done()
+		logger.Info("Shutting down TCP proxy server")
+		_ = listener.Close()
+	}()
+
+	logger.Info("TCP proxy server started", "port", port)
```
MikeSpreitzer: I myself would log the whole proxy config.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,221 @@
...
+		if err != nil {
+			logger.V(4).Info("Target-to-client copy finished", "err", err)
+		}
+		done <- struct{}{}
+	}()
+
+	// Wait for either direction to finish or context to be cancelled
+	select {
+	case <-done:
+		// One side finished, close the other side
```
MikeSpreitzer: That's not exactly how TCP works. TCP is two independent unidirectional streams. One stream can continue after the other one finishes. See https://pkg.go.dev/net@go1.26.2#TCPConn.CloseRead and CloseWrite.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,221 @@
...
+		logger.Error(err, "Failed to dial backend", "target", target)
+		return
+	}
+	defer func() { _ = targetConn.Close() }()
+
+	// Bidirectional forwarding using two goroutines
+	done := make(chan struct{}, 2)
+
+	go func() {
+		_, err := io.Copy(targetConn, clientConn)
```
MikeSpreitzer: Note that this is not connected to `ctx.Done()`.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	go func() {
+		err := RunWithConfig(ctx, fmt.Sprintf("%d", proxyPort), cfg)
+		if err != nil {
+			t.Logf("proxy Run error: %v", err)
+		}
+	}()
+
+	// Wait for listener to be ready
+	time.Sleep(50 * time.Millisecond)
```
MikeSpreitzer: Picking a time constant here is problematic: too short causes false failures (BTW, we have been surprised more than once by how overloaded the GitHub runners get), too long makes the test take longer than necessary. There should be no need to wait here. Nothing tries to connect to that port until after `Configure` succeeds. In my opinion the proper solution is to make `Configure` not return until the proxy is listening for connections from clients.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		t.Fatalf("failed to write to proxy: %v", err)
+	}
+
+	reader := bufio.NewReader(conn)
+	resp, err := reader.ReadString('\n')
+	if err != nil {
+		t.Fatalf("failed to read echo: %v", err)
+	}
+	if !strings.Contains(resp, "hello proxy") {
```
MikeSpreitzer: Why not check for exact equality with `testMsg`?

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if w1.Code != http.StatusNotFound {
+		t.Errorf("GET before configure should return 404, got %d", w1.Code)
+	}
+
+	// Now configure and check status again
+	backendAddr, backendCloser := startTestEchoServer(t)
+	defer backendCloser()
+
+	backendParts := strings.Split(backendAddr, ":")
+	backendPort := backendParts[1]
```
MikeSpreitzer: Suggestion for code simplification: merge this, and the integer parse, into `startTestEchoServer`.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if config.Address != "127.0.0.1" {
+		t.Errorf("expected address 127.0.0.1, got %q", config.Address)
+	}
+	expectedPort := mustParsePort(backendPort, t)
+	if config.Port != expectedPort {
+		t.Errorf("expected port %d, got %d", expectedPort, config.Port)
+	}
+}
+
+func TestConfigure_Reconfigure(t *testing.T) {
```
MikeSpreitzer: FYI, FMA does not need the functionality tested here. The server-requesting Pod only ever gets bound to at most one inference server.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	jsonBody1, _ := json.Marshal(body1)
+	req1 := httptest.NewRequest(http.MethodPut, stubapi.ProxyConfigPath, strings.NewReader(string(jsonBody1)))
+	w1 := httptest.NewRecorder()
+	Configure(w1, req1)
+
+	if w1.Code != http.StatusOK {
+		t.Fatalf("first Configure failed: %d — %s", w1.Code, w1.Body.String())
+	}
+	if !strings.Contains(w1.Body.String(), "configured proxy to") {
```
MikeSpreitzer: This is testing an unimportant detail. What is important is that a subsequent GET returns the right result.

Also, while we are at it, shouldn't we test actually using the first config (i.e., have a client make and use a connection)?

And shouldn't we also test that the reconfig is effective (i.e., have a client exercise it)?

If this test tests this much then we do not need a separate test that only tests a prefix of this.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,381 @@
...
+	if _, err := fmt.Sscanf(s, "%d", &port); err != nil {
+		t.Fatalf("failed to parse port %q: %v", s, err)
+	}
+	if port < 0 || port > 32767 {
+		t.Fatalf("port %d out of int16 range", port)
+	}
+	return int16(port)
+}
+
+// Suppress klog output during tests
```
MikeSpreitzer: In my opinion it is helpful to get debug logging from tests that fail, and it is very rare for anybody to care about the output from tests that pass.

test/e2e/test-cases.sh
```diff
@@ -490,6 +490,64 @@ if [ "$E2E_PLATFORM" = "openshift" ]; then check_gpu_pin $req4; fi
...
+# Starting state: $req4 is bound to $launcher1, both Ready.
+
+echo "Checking proxy configuration on requester pod $req4"
+
+# Port-forward to the requester's SPI port to query proxy config
+kubectl port-forward pod/"$req4" 28091:8081 -n "$NS" &
+PF_SPI_PID=$!
+sleep 2
+
+proxy_resp=$(curl -sf "http://localhost:28091/v1/proxy/config" 2>/dev/null || echo "")
```
MikeSpreitzer: Better: using `-w` and `-o` to capture and test both status code and response body.

test/e2e/test-cases.sh
```diff
@@ -490,6 +490,64 @@ if [ "$E2E_PLATFORM" = "openshift" ]; then check_gpu_pin $req4; fi
...
+# GET /v1/proxy/config returns JSON {"address":"...","port":...} or 404 if not configured
+if ! echo "$proxy_resp" | grep -q '"address"'; then
+    echo "ERROR: expected proxy to be configured, got: '$proxy_resp'" >&2
+    exit 1
+fi
+echo "Proxy is configured: $proxy_resp"
+
+# Verify proxy target contains the launcher pod IP
+launcher1_ip=$(kubectl get pod "$launcher1" -n "$NS" -o jsonpath='{.status.podIP}')
+if ! echo "$proxy_resp" | grep -qF "$launcher1_ip"; then
```
MikeSpreitzer: We can use `jq` here. In my opinion, a better test would construct the correct response JSON and use `jq` to compare for JSON object equality.

test/e2e/test-cases.sh
```diff
@@ -490,6 +490,64 @@ if [ "$E2E_PLATFORM" = "openshift" ]; then check_gpu_pin $req4; fi
...
+    echo "ERROR: proxy config does not contain launcher IP $launcher1_ip: '$proxy_resp'" >&2
+    exit 1
+fi
+echo "Proxy config contains launcher IP: $launcher1_ip"
+
+# Verify traffic forwarding through the TCP proxy port (8082)
+# The launcher has a /health endpoint that returns 200
```
MikeSpreitzer: The proxy forwards to the vllm inference port, not the launcher's management port.

MikeSpreitzer:
Did you consider whether there is a pre-existing TCP proxy that we can import?

docs/dual-pods.md
MikeSpreitzer: I really did mean that as a "maybe". The dual-pod lifecycle involves only one binding. The only possible use of unbinding would be in the teardown path, and I am not sure that would be worth the latency.

delavet: Yes, same opinion. In fact, I think this TCP Proxy may not even be worth cleaning up, as it will naturally terminate when the requester pod is terminated.

pkg/controller/dual-pods/inference-server.go
delavet: Sorry, there might be ambiguity here. What I meant is that I’m considering submitting a follow-up PR, which would modify the requester’s current readiness probe to instead probe the TCP proxy’s port—ensuring that the requester pod consistently reflects the actual health status of its bound vLLM instance.

delavet:
> Did you consider whether there is a pre-existing TCP proxy that we can import?

I’m currently considering and investigating this, as I’ve found that maintaining a TCP proxy so far has indeed introduced excessive additional complexity.

pkg/spi/interface.go
delavet: Yes, I’ve also identified this issue—I’ll fix the code.

MikeSpreitzer:
> I’ve found that maintaining a TCP proxy so far has indeed introduced excessive additional complexity.

I am not sure about that. The proxy itself is about 300 lines of code in this PR now, and I think that it is not far from being done.

But it is always worth explicitly considering whether you can import an adequate solution.

pkg/spi/interface.go
delavet: Current revision plan:
1. `proxy.ProxyConfig`: Retain this name, as it defines configuration parameters for the proxy itself.
2. `stubapi.ProxyConfig` → `stubapi.ProxyTargetConfig`: Renamed because it actually describes the target endpoint to which the proxy forwards requests.

How about this?

pkg/server/requester/proxy/server.go
delavet: Converting `Configure` into a method of `forwarder` may introduce additional complexity, as it affects the logic for registering with the coordination server. I’ll note this for now and revisit the change logic later if needed.

MikeSpreitzer:
It might have been good to first present a list of candidate proxies to import and the reasoning behind choosing a particular one.

MikeSpreitzer:
My own survey and evaluation leads to the same choice, inetaf/tcpproxy.

The only other prominent alternative that showed up in my search is https://github.com/jpillora/go-tcp-proxy, and that was last updated 6 years ago.

inetaf/tcpproxy was updated last year. Still not great, but perhaps good enough for something simple. This module looks like it has more functionality than we need; I hope that our limited usage will limit exposure to the unnecessary functionality.

delavet:
# Go TCP Proxy Library Research Report

## Current Implementation Analysis

Your current TCP proxy implementation in `pkg/server/requester/proxy/server.go` follows a classic pattern:

```go
// Bidirectional forwarding
go func() {
    _, err := io.Copy(targetConn, clientConn)
    done <- struct{}{}
}()
go func() {
    _, err := io.Copy(clientConn, targetConn)
    done <- struct{}{}
}()
```

This is the standard approach for TCP proxying in Go — using two goroutines with `io.Copy` for bidirectional data forwarding. The implementation also includes:

- Context-based graceful shutdown
- Dynamic target configuration via HTTP API (GET/PUT)
- Thread-safe target reconfiguration with `sync.Mutex`
- Singleton pattern for the forwarder
- klog integration for logging

## Survey of Mature Go TCP Proxy Libraries

### 1. inetaf/tcpproxy (Recommended)

**Repository**: https://github.com/inetaf/tcpproxy  
**GitHub (legacy redirect)**: https://github.com/google/tcpproxy → 301 → inetaf/tcpproxy  
**Stars**: ~1.3k  
**License**: Apache-2.0 (compatible with your project)  
**Maintained by**: Google's inetaf (Internet Architecture Foundation)

This is the most mature and production-ready TCP proxy library in the Go ecosystem. It's used by major projects including Caddy and various infrastructure tools.

**Key Features**:
- Route-based architecture with support for HTTP Host header and TLS SNI routing
- `DialProxy` type implements `Target` interface — exactly what you need for simple TCP forwarding
- Configurable `KeepAlivePeriod`, `DialTimeout`, and custom dial functions
- PROXY protocol support (v1 and v2) via `ProxyProtocolVersion`
- Clean connection lifecycle management

**How it compares to our implementation**:

| Feature | Our Implementation | inetaf/tcpproxy |
|---------|-------------------|-----------------|
| Bidirectional forwarding | Manual `io.Copy` x2 | Built-in `DialProxy.HandleConn` |
| Graceful shutdown | Context cancellation | `Close()` method |
| Configurable dial timeout | Yes (manual) | Yes (`DialTimeout` field) |
| PROXY protocol | No | Yes (v1/v2 support) |
| Keep-alive | No | Yes (`KeepAlivePeriod`) |
| Dynamic reconfiguration | Custom singleton | Route-based (AddRoute/TargetListener) |
| Production readiness | Self-maintained | Used by Caddy, etc. |


### 2. jpillora/go-tcp-proxy

**Repository**: https://github.com/jpillora/go-tcp-proxy  
**Stars**: ~796  
**License**: MIT

A simpler TCP proxy focused on text protocol debugging. Less suitable for production infrastructure use.

**Limitations for your use case**:
- Designed primarily as a CLI tool/debugger
- Regex-based pattern matching on text packets (not binary-safe)
- No programmatic API for dynamic reconfiguration
- Lacks graceful shutdown support

### 3. Other Notable Mentions

| Library | Stars | Notes |
|---------|-------|-------|
| `snail007/goproxy` | ~20k+ | Full-featured proxy suite (HTTP/SOCKS), but heavyweight for simple TCP forwarding |
| `armon/go-proxyproto` | ~1k+ | PROXY protocol parser only, not a full TCP proxy |
| `caddyserver/forwardproxy` | - | HTTP forward proxy, not TCP-level |

## Recommendation

### Option A: Use inetaf/tcpproxy (Recommended)

The `inetaf/tcpproxy` library is the best fit because:

1. **Same license**: Apache-2.0, no compatibility issues
2. **Production-proven**: Used by Caddy and other major projects
3. **Feature-rich out of the box**: Keep-alive, PROXY protocol, configurable timeouts
4. **Maintained by Google-backed organization**: Long-term sustainability
5. **Simple API**: `DialProxy.HandleConn()` replaces your `handleConnection()` function entirely

The main adaptation needed is changing your singleton pattern to work with `tcpproxy`'s route-based architecture. You could use `TargetListener` for dynamic target switching.

### Option B: Keep current implementation with improvements

If you prefer to avoid an external dependency, your current implementation is actually quite clean. The `io.Copy` bidirectional pattern is the standard approach and is used by many production systems. Consider these improvements:

1. Add `KeepAlivePeriod` configuration for TCP connections
2. Use `CloseRead()`/`CloseWrite()` for half-close support (if on Go 1.16+)
3. Add connection counting/metrics

### Option C: Hybrid approach

Use `inetaf/tcpproxy`'s `DialProxy` for the forwarding logic while keeping your own listener and configuration management. This gives you the battle-tested forwarding code without changing your architecture significantly.

## Migration Effort Estimate

| Option | Effort | Risk |
|--------|--------|------|
| A. Full adoption | Medium (refactor singleton + config) | Low (well-tested library) |
| B. Keep + improve | Low | Medium (self-maintained) |
| C. Hybrid (use DialProxy only) | Low | Low |

## Conclusion

For reducing maintenance burden, I recommend **Option C** as the most pragmatic approach: import `inetaf/tcpproxy` and use its `DialProxy` type to replace the `io.Copy` bidirectional forwarding in `handleConnection()`. This keeps your existing listener, configuration API, and singleton pattern while leveraging a production-tested forwarding implementation.

pkg/spi/interface.go
MikeSpreitzer: Good enough.
A clarifying comment on the bare "ProxyConfig" would make it better.

cmd/requester/main.go
```diff
@@ -18,57 +18,123 @@ package main
...
+		Proxy:      proxy.DefaultProxyConfig,
+	}
+
 	klog.InitFlags(nil)
-	flag.Parse()
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	addFlags(*pflag.CommandLine, &cfg)
+	pflag.Parse()
+
+	// Override from environment variables if not explicitly set via flags
```
MikeSpreitzer: This is defaulting behavior. The better way to do this is to put it into the logic that initializes `cfg` before calling `addFlags`, so that the user sees the actual defaults that really apply.

MikeSpreitzer: More importantly, I do not think that we need the complexity of supporting environment variables AS WELL AS command line flags. Just the flags is sufficient.

cmd/requester/README.md
```diff
@@ -2,6 +2,39 @@ This document shows the steps to exercise the requester and dual-pods controller
...
 
+## Requester Overview
+
+The server-requesting Pod runs three servers:
+
+| Server | Default Port | Purpose |
+|--------|-------------|---------|
+| Probes | 8080 | Readiness relay (`/ready` endpoint) |
+| SPI | 8081 | Dual-pods controller interface (`/v1/dual-pods/accelerators`) |
+| Proxy | 8082 | TCP reverse proxy to the inference server |
```
MikeSpreitzer: I recommend being clear about whether  this the configuration port for the proxy, or the port where it actually listens for connections to forward.

cmd/requester/README.md
```diff
@@ -2,6 +2,39 @@ This document shows the steps to exercise the requester and dual-pods controller
...
+
+### TCP Proxy
+
+The requester includes a **TCP proxy** that forwards traffic to the bound inference server.
+After the dual-pods controller binds the requester to a server-providing Pod, it configures
+the proxy target via a `PUT` to the SPI endpoint at `/v1/proxy/config`. You can query the
+current proxy configuration with `GET /v1/proxy/config`.
+
+**Environment variables:**
```
MikeSpreitzer: Use real Markdown subsection headers, not visual formatting.
This comment applies to all of these.

cmd/requester/README.md
```diff
@@ -2,6 +2,39 @@ This document shows the steps to exercise the requester and dual-pods controller
 in a local k8s environment with model `ibm-granite/granite-3.3-2b-instruct`
 cached on local PV in the cluster.
 
+## Requester Overview
```
MikeSpreitzer, line 5: Hmm, I see that there was no overview before this PR. This PR is not adding a full overview. Since it is adding a partial overview, it should include a placeholder for the missing rest of the overview.

Also, after the `## Requester Overview` section, the transition to non-overview material should also have a `##` section header.

cmd/requester/README.md
```diff
@@ -2,6 +2,39 @@ This document shows the steps to exercise the requester and dual-pods controller
...
+**Environment variables:**
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `PROBES_PORT` | `8080` | Probes server port |
+| `SPI_PORT` | `8081` | SPI server port |
+| `PROXY_PORT` | `8082` | TCP proxy port (must be 1-65535) |
+| `PROXY_DIAL_TIMEOUT` | `10s` | Timeout for dialing the inference server |
+
+**Command-line flags:**
```
MikeSpreitzer: In my opinion it is better to list all of the command line flags. I typically start with the output from `--help` and then organize that into topical groups.

cmd/requester/main.go
```diff
@@ -18,57 +18,123 @@ package main
...
+	pflag.CommandLine.VisitAll(func(f *pflag.Flag) {
+		logger.V(1).Info("Config", "flag", f.Name, "value", f.Value.String())
+	})
 
 	var ready atomic.Bool
 
 	var wg sync.WaitGroup
-	wg.Add(2)
 
+	wg.Add(1)
```
MikeSpreitzer, line 57: Did you intend to use this style in the test-requester too?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,231 @@
...
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/inetaf/tcpproxy"
+	"github.com/spf13/pflag"
+	"k8s.io/klog/v2"
```
MikeSpreitzer, line 31: See #447

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,231 @@
...
+// Run starts the TCP proxy server with the given configuration.
+// It blocks until the context is cancelled or a fatal error occurs.
+func Run(ctx context.Context, cfg ProxyConfig) error {
+	logger := klog.FromContext(ctx).WithName("proxy-server")
+	logger.Info("Starting TCP proxy server", "config", cfg)
+
+	// Set dial timeout for later use in Configure()
+	targetMu.Lock()
+	dialTimeout = cfg.DialTimeout
+	targetMu.Unlock()
```
MikeSpreitzer: Does any use of `dialTimeout` run concurrently with this code? If not, why is the locking here?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,231 @@
...
+		if dp == nil {
+			w.WriteHeader(http.StatusNotFound)
+			_, _ = w.Write([]byte("proxy not configured"))
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		addr, portStr, _ := net.SplitHostPort(dp.Addr)
+		var portVal uint16
+		if portNum, err := net.LookupPort("tcp", portStr); err == nil {
```
MikeSpreitzer: By the construction below, `portStr` necessarily holds an integer literal, right? If so then it would be simpler to use an integer parser here.

cmd/requester/main.go
```diff
@@ -18,57 +18,123 @@ package main
...
+	ProbesPort int16
+	SPIPort    int16
+	Proxy      proxy.ProxyConfig
+}
+
+func addFlags(fs pflag.FlagSet, cfg *requesterConfig) {
+	fs.Int16Var(&cfg.ProbesPort, "probes-port", cfg.ProbesPort, "port number for readiness/liveness probes")
+	fs.Int16Var(&cfg.SPIPort, "spi-port", cfg.SPIPort, "port for dual-pods SPI requests")
+	cfg.Proxy.AddFlags(fs)
+}
```
MikeSpreitzer: Doesn't the test requester need the same stuff (and the default constructor too)? Why isn't this in some common package?

pkg/server/requester/proxy/server.go
MikeSpreitzer, line 1: This package should not be making its own listener for the proxy service port. The imported proxy does that. This package, a client of tcpproxy, should be similar to the example in https://pkg.go.dev/github.com/inetaf/tcpproxy:

- This package's init should construct a tcpproxy.Proxy.
- This package's Run gets the startup config parameters.
- This package's Configure gets the remaining config from the config client (i.e., dual-pods controller).
- Run and Configure are called by distinct goroutines, which could start those calls in either order.
- Regardless of the order of those two calls:
- Run needs to wait until both kinds of configuration are received, then call Proxy.AddRoute with the DialProxy, then call Proxy.Run.
- Configure needs to make its received configuration available to Run.

cmd/requester/main.go
delavet: Sure, will handle that

pkg/server/requester/proxy/server.go
delavet: Agree with this -- I will refactor a new version of code

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1005,6 +1005,27 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
...
+	launcherAddr := fmt.Sprintf("%s:%d", providingPod.Status.PodIP, serverPort)
+	url := fmt.Sprintf("http://%s%s", requesterAddr, stubapi.ProxyConfigPath)
+	proxyConfig, _ := json.Marshal(stubapi.ProxyTargetConfig{
+		Address: providingPod.Status.PodIP,
+		Port:    uint16(serverPort),
+	})
+
+	if err := doPut(url, bytes.NewReader(proxyConfig)); err != nil {
+		logger.Error(err, "Failed to initialize requester proxy",
+			"requesterAddr", requesterAddr, "launcherAddr", launcherAddr)
```
MikeSpreitzer: Rather than `"launcherAddr", launcherAddr`, I think what would be more helpful here would be `"proxyConfig", string(proxyConfig)`.

pkg/server/requester/proxy/server.go
delavet: I've refactored the proxy package to use tcpproxy.Proxy as the framework.

The key tension is: tcpproxy.Proxy.AddRoute must be called before Proxy.Run() starts, but the backend target address isn't known at startup — it arrives later via Configure() (HTTP PUT from the dual-pods controller), at an unpredictable time.   

To address this,  a placeholder target `lazyTarget` is introduced. Let me know if you see a cleaner way to handle this dynamic-bind requirement.

cmd/requester/main.go
```diff
@@ -19,56 +19,70 @@ package main
...
 
+	"github.com/spf13/pflag"
+
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/config"
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/coordination"
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/probes"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/proxy"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
 	"k8s.io/klog/v2"
```
MikeSpreitzer, line 34: See #447 .
A little more generally: we typically organize imports into groups that proceed from more basic to more dependent. The imports from the same repo (or Go module, if there is a difference) go in the last group. The imports from Kubernetes would come in an earlier group.

pkg/server/requester/config/config.go
```diff
@@ -0,0 +1,46 @@
...
+func NewDefault() Config {
+	return Config{
+		ProbesPort: 8080,
+		SPIPort:    8081,
+		Proxy:      proxy.DefaultProxyConfig,
+	}
+}
+
+// AddFlags registers command-line flags for all Config fields.
+func (cfg *Config) AddFlags(fs pflag.FlagSet) {
```
MikeSpreitzer, line 42: The formal parameter here needs to be a _pointer to_ the `pflag.FlagSet`, because this func's job is to modify that.

cmd/requester/README.md
```diff
@@ -2,6 +2,63 @@ This document shows the steps to exercise the requester and dual-pods controller
 in a local k8s environment with model `ibm-granite/granite-3.3-2b-instruct`
 cached on local PV in the cluster.
 
+## Requester Overview
+
+The requester binary runs inside the server-requesting Pod of the dual-pods
+architecture. It acts as a lightweight coordinator between the dual-pods
+controller and the actual inference server running in the server-providing Pod.
```
MikeSpreitzer, line 9: I would summarize it a little differently. The requester executable is what runs in the requester container, which is intended to 

(a) look to the rest of llm-d like an inference server, and
(b) relay the information needed to implement this appearance to and from the dual-pods controller.

Adding the reverse TCP proxy makes the requester look _more_ like an inference server than the requester otherwise would.

config/examples/requester.yaml
```diff
@@ -6,14 +6,13 @@ spec:
   containers:
     - name: requester-container
       image: ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/requester:latest
+      command:
+        - /app/requester
+        - --probes-port=8080
+        - --spi-port=8081
       ports:
         - containerPort: 8080
         - containerPort: 8081
```
MikeSpreitzer, line 15: Don't you want to list the proxy port too?

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: That is, you intend a follow-on PR to add another level of proxying, for the readiness probe. Do we even need to work that hard? Why not just have the PodSpec direct the readiness probe to the proxy service port?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1615,21 +1636,35 @@ func (ctl *controller) ensureReqState(ctx context.Context, requestingPod *corev1
...
-// doPost does the HTTP POST request/response to the given URL.
 func doPost(url string) error {
+	return doHTTPRequest(http.MethodPost, url, nil)
+}
+
+func doPut(url string, data io.Reader) error {
+	return doHTTPRequest(http.MethodPut, url, data)
+}
+
+func doHTTPRequest(method, url string, data io.Reader) error {
```
MikeSpreitzer, line 1647: FYI, you could factor this change out into a small PR that would merge quickly.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/inetaf/tcpproxy"
+	"github.com/spf13/pflag"
+	"k8s.io/klog/v2"
+
+	stubapi "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/spi"
+)
+
+// ProxyConfig holds the proxy server's own runtime parameters (listening port,
+// dial timeout, etc.). This is distinct from spi.ProxyTargetConfig, which
+// describes the backend target the proxy forwards traffic to.
```
MikeSpreitzer, line 38: I still think that both of config structs hold some of "the proxy server's own runtime parameters". The difference is that one holds parameters provided on the requester command line and the other holds parameters that are provided later by the dual-pods controller (or, less specifically, the SPI client).

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// DefaultProxyConfig provides sensible defaults
+var DefaultProxyConfig = ProxyConfig{
+	Port:        8082,
+	DialTimeout: 10 * time.Second,
+}
+
+// AddFlags registers command-line flags for all proxy configuration fields.
+func (cfg *ProxyConfig) AddFlags(fs pflag.FlagSet) {
```
MikeSpreitzer, line 53: The formal parameter here needs to be a _pointer to_ the FlagSet.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+type lazyTarget struct {
+	mu          sync.Mutex
+	dialProxy   *tcpproxy.DialProxy    // nil until Configure() delivers
+	proxyTarget *stubapi.ProxyTargetConfig
+	configured  bool                   // true once setConfig has been called
+}
+
+// HandleConn checks if the backend is configured. If not, the connection is
+// rejected immediately. Otherwise it delegates to tcpproxy.DialProxy.HandleConn.
+func (l *lazyTarget) HandleConn(conn net.Conn) {
```
MikeSpreitzer, line 70: I would put debug logging in this method.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+	lt := proxyTarget
+	proxyMu.Unlock()
+
+	if ps != nil {
+		<-ps
+	}
+
+	if lt == nil {
+		http.Error(w, "proxy server not started", http.StatusServiceUnavailable)
+		return
```
MikeSpreitzer, line 164: This is not an error condition. It is valid for Configure to be called before Run. The callers are two different goroutines, with no relevant synchronization.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+	proxyStarted = make(chan struct{})
+	proxyMu.Unlock()
+
+	// Start the proxy in a goroutine (it creates its own listener and blocks).
+	// Signal started immediately after so Configure() can proceed.
+	proxyErr := make(chan error, 1)
+	go func() {
+		proxyErr <- proxy.Run()
+	}()
+	close(proxyStarted)
```
MikeSpreitzer, line 138: You realize that this channel close is not synchronized with `proxy.Run()` doing anything, right? You could just as well move this channel close up before the `go func...` statement above. You could just as well move it to right after `proxyStarted = make(chan struct{})`. This channel closure conveys no useful timing information to a receiver that already knows that `proxyStarted != nil`.

pkg/server/requester/proxy/server.go
MikeSpreitzer: Yes, I tried to speak to that tension in my previous comment.

I think that the current draft is more complicated than necessary. There is no need to call `Proxy.Run` before `Configure` has done its work. I suspect that all you need for coordination between this package's `Run` and `Configure` is the following.

- A `sync.Condition` and associated `Mutex`.
- They guard access to a variable holding a `*ProxyTargetConfig`.
- `Run` uses that Condition to wait for the `*ProxyTargetConfig` to be non-nil and react as soon as that condition is met and the mutex is not held.
- At that point, `Run` has everything it needs to construct, configure, and run a `tcpproxy.Proxy`; there is no need to do any of that any earlier, no need for a fake or lazy `tcpproxy.Target`.
- `Configure`, in the write case, checks whether the write has already been done and, if not, sets that `*ProxyTargetConfig` to the right value and signals the `Condition`.

pkg/server/requester/proxy/server.go
MikeSpreitzer: I do not see a real problem there.

pkg/server/requester/proxy/server.go
MikeSpreitzer: As I mentioned in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#pullrequestreview-4216876731 , I think that it would be beneficial if `Configure` did not return until after the proxy was listening. You could accomplish that by the following additional steps.

- Initialize a variable to hold a channel that `Run` closes once the proxy starts listening.
- Make `Configure` wait for that channel to be closed before returning.
- Make `Run` use `tcpproxy.Proxy.Start` and `Wait` instead of `tcpproxy.Proxy.Run` (look at the implementation of `Run`). Between the `Start` and the `Wait`, close that channel.

MikeSpreitzer:
This PR has over 100 comments, which is awkward in GitHub. I suggest closing this one and opening a new one to make progress on.

pkg/server/requester/proxy/server.go
MikeSpreitzer: Actually, perhaps it can be even simpler. Just use two channel objects. One to get the `ProxyTargetConfig` from `Configure` to `Run`, and one to get the started signal from `Run` to `Configure`.

pkg/controller/dual-pods/inference-server.go
delavet: Yes, I think so too—I’ll just implement it directly in the new PR.

delavet:
Close this since #498 is opened

PR #424, opened 2026-04-14
✨ Initial commit of AI agent instruction files

MikeSpreitzer:
Added two skills and `AGENTS.md` saying when to use those skills.

Skills:
- our-git-commits
- pr-security-review

These are my attempt to extract into AAIF files what I have accumulated in my personal Claude instructions. The explicit direction to use the skills was added to my Claude instructions because I found Claude Code not using the skills some times and asked Claude how to make it not make that mistake.

### DIFF
--- a/.agents/skills/our-git-commits/SKILL.md
+++ b/.agents/skills/our-git-commits/SKILL.md
@@ -1,0 +1,9 @@
+name: our-git-commits
+description: Things to do when making a Git commit
+trigger: When the user asks to commit changes, create a commit, make a git commit, amend a commit, revise a commit, reword a commit message, or any other action that results in running a `git commit` command
+---
+
+When creating a Git commit:
+
+1. Before committing, run `typos` on the changed files to find and fix any typographical errors. Ask the user about any findings you are unsure about.
+2. Include a DCO sign-off and a cryptographic signature.
--- a/.agents/skills/pr-security-review/SKILL.md
+++ b/.agents/skills/pr-security-review/SKILL.md
@@ -1,0 +1,11 @@
+name: pr-security-review
+description: When reviewing PRs that bump dependencies or GitHub Actions, actively search for known vulnerabilities rather than reasoning abstractly
+---
+
+When reviewing dependency bump PRs, verify security by actually checking external sources — don't just reason about it in the abstract.
+
+Checklist:
+1. Verify the pinned SHA matches the upstream tag
+2. Search github.com/advisories for the dependency
+3. Web search for CVEs/vulnerabilities in the specific new version being introduced
+4. Check the release notes for any security-relevant changes
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,5 @@
+# Instructions for coding agents
+
+## Workflow Preferences
+- Always use the `our-git-commits` skill when making git commits.
+- Always use the `pr-security-review` skill when reviewing PRs that bump dependencies or GitHub Actions.


.agents/skills/our-git-commits/SKILL.md
```diff
@@ -0,0 +1,9 @@
+name: our-git-commits
+description: Things to do when making a Git commit
+trigger: When the user asks to commit changes, create a commit, make a git commit, amend a commit, revise a commit, reword a commit message, or any other action that results in running a `git commit` command
```
MikeSpreitzer, line 3: I know that `trigger` is not specified in AAIF. It _is_ understood by Claude Code. Maybe that is good enough?

AGENTS.md
```diff
@@ -0,0 +1,5 @@
+# Instructions for coding agents
+
+## Workflow Preferences
+- Always use the `my-commits` skill when making git commits.
```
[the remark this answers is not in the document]
MikeSpreitzer: Fixed.

.agents/skills/pr-security-review/SKILL.md
```diff
@@ -0,0 +1,11 @@
+name: PR security review
```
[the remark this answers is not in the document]
MikeSpreitzer: Fixed.

.agents/skills/our-git-commits/SKILL.md
```diff
@@ -0,0 +1,9 @@
...
+---
+
+When creating a Git commit:
+
+1. Before committing, run `typos` on the changed files to find and fix any typographical errors. Ask the user about any findings you are unsure about.
+2. Include a DCO sign-off and a GPG signature.
```
diegocastanibm: This line means that even when the user is not adding `git commit -s -S` to the command, the agent will apply automatically those flags in this repo?

.agents/skills/our-git-commits/SKILL.md
diegocastanibm: Maybe the trigger can be part of the description as you are doing below. However, if claude gets it, looks good to me

AGENTS.md
```diff
@@ -0,0 +1,5 @@
+# Instructions for coding agents
```
diegocastanibm, line 1: I suppose that `AGENTS.md` is not creating any conflict with `CLAUDE.md`, am I right?

.agents/skills/our-git-commits/SKILL.md
MikeSpreitzer: Claude always reads the trigger, but does not always read the body. The whole point of the trigger is getting Claude to use the body under the right circumstances. Claude has been, to my mind, surprisingly inattentive otherwise. I am not sure whether for other agents this file will be relevant and also suffering from lack of agent attention. Let us cross that bridge if and when we come to it.

.agents/skills/our-git-commits/SKILL.md
MikeSpreitzer: Yes, and it is deeper than that. The user is not necessarily issuing `git` commands at all, but possibly talking at a more or less higher level that implies `git` commands. I have done this, with some success.

AGENTS.md
MikeSpreitzer: (A) This project has no `CLAUDE.md` to conflict with.

(B) If both files existed then Claude Code would read both.

.agents/skills/our-git-commits/SKILL.md
```diff
@@ -0,0 +1,9 @@
+name: our-git-commits
+description: Things to do when making a Git commit
```
waltforme, line 2: I remember there were some very interesting general discussions about “adverb”-style skills vs. “verb”-style skills. This one would probably fall into the “adverb” category.
As I recall, the main issue with “adverb”-style skills is that they don’t always get triggered. I’m curious whether that problem has been resolved, and if so, how.

MikeSpreitzer: I solved it in my personal Claude instruction files, using Claude-specific techniques. One is the trigger header item. Another is feedback pointing Claude at the skill. I kept the header item, and tried to map the feedback item into AAIF-speak.

.agents/skills/our-git-commits/SKILL.md
rubambiza: It is probably unlikely, but I wondered if there might be any ambiguity here if the user opts for SSH-key-based signing vs GPG-based.

I am looking here to a future where we can't make assumptions about the contributors being mostly us, who I presume prefer GPG.

This is also coming from a recent experience on the kagenti project where I was reviewing someone's PR.
1. They [forgot DCO](https://github.com/kagenti/.github/pull/57#issuecomment-4207192796), which I asked them to fix
2. They fixed the DCO [without cryptographic signature](https://github.com/kagenti/.github/pull/57#pullrequestreview-4100660655), which I asked them again to fix

The easiest fix suggested by Claude, which I passed to them, was to use ssh-key based.

In terms of action here, would `cryptographic signature` be more appropriate (as opposed to GPG specifically)?

rubambiza (commented):
Minor experiential comment with suggested change.

.agents/skills/our-git-commits/SKILL.md
MikeSpreitzer: Interestingly, `git` itself seems to use the term "GPG signature" loosely, to include the possibility of signing with an SSH key. Look at the commands in that other comment that you cited:

```shell
git config --global user.signingkey ~/.ssh/id_ed25519.pub
git config --global commit.gpgsign true
```

I have a setting for `user.signingkey` in my own git configuration; in my case, the value is a GPG key ID.

That said, it would not be bad for our own language to be properly inclusive. "cryptographic signature" sounds good to me.

MikeSpreitzer:
The force-push to f899f67 makes the change that @rubambiza suggested.

.agents/skills/our-git-commits/SKILL.md
MikeSpreitzer: I have made the suggested change.

rubambiza (approved):
lgtm

waltforme (approved):

AGENTS.md
diegocastanibm: I have a CLAUDE.md in my local copy (that I excluded). I do not want Claude to have any conflict between those 2. I just hope you are right with (B).

diegocastanibm (approved):

PR #428, opened 2026-04-15
Garbage collect obsolete vLLM instances

waltforme:
This PR carries out 'A2 to Q4' in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-3717326061.

This PR does more than that. The PR also covers corresponding instances that are _not_ sleeping when an ISC changes --- they are garbage collected when unbounding.

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -349,7 +349,7 @@ type unboundLauncherPodItem struct {
 	NodeName        string
 }
 
-type iscItem struct {
+type instanceGCItem struct {
 	ISCName string
 }
 
@@ -527,7 +527,7 @@ func (ctl *controller) OnUpdate(prev, obj any) {
 	case *fmav1alpha1.InferenceServerConfig:
 		prevTyped, ok := prev.(*fmav1alpha1.InferenceServerConfig)
 		if ok && !reflect.DeepEqual(prevTyped.Spec, typed.Spec) {
-			ctl.Queue.Add(iscItem{ISCName: typed.Name})
+			ctl.enqueueInstanceGCItemsForISC(typed.Name)
 		}
 		ctl.enqueueInfSvrItemsByISC(typed, false)
 	case *corev1.ConfigMap:
@@ -664,16 +664,13 @@ func (item cmItem) process(ctx context.Context, ctl *controller) (error, bool) {
 	return nil, false
 }
 
-func (item iscItem) process(ctx context.Context, ctl *controller) (error, bool) {
-	isc, err := ctl.iscLister.InferenceServerConfigs(ctl.namespace).Get(item.ISCName)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil, false
-		}
-		return err, true
+func (ctl *controller) enqueueInstanceGCItemsForISC(iscName string) {
+	ctl.mutex.Lock()
+	defer ctl.mutex.Unlock()
+	for nodeName, nodeDat := range ctl.nodeNameToData {
+		nodeDat.add(instanceGCItem{ISCName: iscName})
+		ctl.Queue.Add(nodeItem{nodeName})
 	}
-	ctl.cleanupObsoleteSleepingInstances(ctx, isc)
-	return nil, false
 }
 
 func (ctl *controller) enqueueRequesters(ctx context.Context) {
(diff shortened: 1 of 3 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -509,7 +510,15 @@ func (ctl *controller) OnUpdate(prev, obj any) {
...
 		}
 	case *fmav1alpha1.InferenceServerConfig:
-		ctl.enqueueRequestersByInferenceServerConfig(typed, false)
+		prevTyped, ok := prev.(*fmav1alpha1.InferenceServerConfig)
+		if ok && !reflect.DeepEqual(prevTyped.Spec, typed.Spec) {
+			go func() {
+				ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+				defer cancel()
+				ctl.cleanupObsoleteSleepingInstances(ctx, prevTyped, typed)
+			}()
```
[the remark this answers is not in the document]
waltforme: This is already fixed.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1108,6 +1115,63 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
...
+	}
+	// Instance is obsolete — delete from launcher instead of sleeping.
+	if _, err := lClient.DeleteInstance(ctx, serverDat.InstanceID); err != nil {
+		if !IsInstanceNotFoundError(err) {
+			logger.Error(err, "Failed to delete obsolete instance during unbinding",
+				"instanceID", serverDat.InstanceID)
+			return false
+		}
+	}
+	nodeName, _ := getProviderNodeName(providingPod)
```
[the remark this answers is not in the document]
waltforme: The error should not happen.

MikeSpreitzer:
> This PR carries out 'A2 to Q4' in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/201#issuecomment-3717326061.
> 
> This PR does more than that. The PR also covers corresponding instances that are not sleeping when an ISC changes --- they are garbage collected when unbounding.

Good. Q4 and A2 are stated colloquially in terms of one relevant edge, but the proper framing of the question and answer are state based. What, if anything, should be done about a vllm instance that was created by a non-current revision of the relevant InferenceServerConfig? Answer: deletion if the vllm instance is sleeping.

pkg/controller/dual-pods/controller.go
```diff
@@ -75,14 +76,18 @@ import (
 // deletion of any server-providing Pod. Nor does the controller ever try to bind
 // one that is unbound; they are only created in the bound state.
 
-// There are two types of item in the controller's work queue.
+// There are three types of item in the controller's work queue.
```
MikeSpreitzer, line 79: This commentary about queuing was already outdated before this PR, not mentioning `cmItem` nor `unboundLauncherItem` nor `nodeItem` and its nested structure.

pkg/controller/dual-pods/controller.go
```diff
@@ -788,3 +813,107 @@ func (ctl *controller) clearServerData(nodeDat *nodeData, uid apitypes.UID) {
...
+	logger := ctl.enqueueLogger.WithValues("inferenceServerConfig", isc.Name)
+	type launcherRef struct {
+		nodeName        string
+		launcherPodName string
+	}
+
+	var launcherRefs []launcherRef
+	ctl.mutex.Lock()
+	for nodeName, nodeDat := range ctl.nodeNameToData {
+		for launcherPodName := range nodeDat.Launchers {
```
MikeSpreitzer: This also violates https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/750d1b2262643c2474609a299b11101178b6ffe8/pkg/controller/dual-pods/controller.go#L262

Copilot's complaint about locking is correct.

I recommend the following approach: the controller's handler for a change in an InferenceServeConfig should iterate over `ctl.nodeNameToData` and for each one enqueue in the `nodeData` an item for ISC tracking on the launchers on that node.

pkg/controller/dual-pods/controller.go
```diff
@@ -788,3 +813,107 @@ func (ctl *controller) clearServerData(nodeDat *nodeData, uid apitypes.UID) {
...
+			iscName := inst.Annotations[VllmConfigISCNameAnnotationKey]
+			if iscName != isc.Name {
+				continue
+			}
+			if len(inst.GpuUUIDs) == 0 {
+				logger.V(4).Info("Skipping instance cleanup because GPU UUIDs are unknown",
+					"launcherPod", lr.launcherPodName, "instanceID", inst.InstanceID)
+				continue
+			}
+			_, currentHash, err := ctl.configInferenceServer(isc, inst.GpuUUIDs)
```
MikeSpreitzer: Ick. This is an expensive method to call. I recommend a lighter approach. Change this method to return two hashes, one that does NOT take the GPU UUIDs into account and one that DOES. Put the GPU-insensitive hash in the instance's annotations and use that hash for currency checking. Also, extend `launcherData` to hold that hash (or the whole `VllmConfig`) for each instance, so this currency checking does not add more network traffic (when `launcherData.Accurate`).

waltforme:
Force-push to https://github.com/llm-d-incubation/llm-d-fast-model-actuation/commit/236789486fa5e771f2fb037f5999747b2d41a597 includes rebase onto `main` and changes in response to review comments.

pkg/controller/dual-pods/controller.go
waltforme: Addressed.

pkg/controller/dual-pods/controller.go
waltforme: Improved.

pkg/controller/dual-pods/controller.go
```diff
@@ -788,3 +797,112 @@ func (ctl *controller) clearServerData(nodeDat *nodeData, uid apitypes.UID) {
...
+	defer ctl.mutex.Unlock()
+	nodeDat := ctl.nodeNameToData[nodeName]
+	if nodeDat == nil {
+		return
+	}
+	launcherDat := nodeDat.Launchers[launcherPodName]
+	if launcherDat == nil {
+		return
+	}
+	delete(launcherDat.Instances, instanceID)
```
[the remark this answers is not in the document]
waltforme: Addressed.

pkg/controller/dual-pods/controller.go
waltforme: This reads more like a scalability optimization than a correctness guarantee. Should we prioritize finishing M3 for now and defer this after M3?

pkg/controller/dual-pods/controller.go
```diff
@@ -513,7 +525,11 @@ func (ctl *controller) OnUpdate(prev, obj any) {
 			ctl.Queue.Add(nodeItem{nodeName})
 		}
 	case *fmav1alpha1.InferenceServerConfig:
-		ctl.enqueueRequestersByInferenceServerConfig(typed, false)
+		prevTyped, ok := prev.(*fmav1alpha1.InferenceServerConfig)
+		if ok && !reflect.DeepEqual(prevTyped.Spec, typed.Spec) {
+			ctl.enqueueInstanceGCItemsForISC(typed.Name)
```
MikeSpreitzer, line 530: For Kubernetes API objects we usually use https://github.com/kubernetes/apimachinery/blob/v0.34.7/pkg/api/equality/semantic.go#L29

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1118,36 +1118,115 @@ func (ctl *controller) removeProviderFinalizer(ctx context.Context, providingPod
...
+		launcherBaseURL := fmt.Sprintf("http://%s:%d", launcherPod.Status.PodIP, ctlrcommon.LauncherServicePort)
+		lClient, err := NewLauncherClient(launcherBaseURL)
+		if err != nil {
+			logger.Error(err, "Failed to create launcher client during instance GC", "launcherPod", launcherPodName)
+			continue
+		}
+		allInsts, err := lClient.ListInstances(ctx)
+		if err != nil {
+			logger.Error(err, "Failed to list instances during instance GC", "launcherPod", launcherPodName)
+			continue
```
MikeSpreitzer, line 1153: Isn't this worth retrying?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1227,6 +1306,64 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
...
+	}
+	iscName := instState.Annotations[VllmConfigISCNameAnnotationKey]
+	if iscName == "" {
+		logger.V(4).Info("Cannot check instance obsolescence: no ISC name annotation on instance", "instanceID", serverDat.InstanceID)
+		return false
+	}
+	currentISC, err := ctl.iscLister.InferenceServerConfigs(ctl.namespace).Get(iscName)
+	if err != nil {
+		logger.V(4).Info("Cannot check instance obsolescence: ISC not found", "iscName", iscName, "err", err)
+		return false
```
MikeSpreitzer, line 1337: If `IsNotFoundError(err)` then the desired behavior is no check for obsolescence, right?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1227,6 +1306,64 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
...
+	}
+	if len(instState.GpuUUIDs) == 0 {
```
MikeSpreitzer, line 1339: Under what circumstances would `len(instState.GpuUUIDs) == 0` ?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1108,6 +1115,63 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
...
+		return false
+	}
+	launcherBaseURL := fmt.Sprintf("http://%s:%d", providingPod.Status.PodIP, ctlrcommon.LauncherServicePort)
+	lClient, err := NewLauncherClient(launcherBaseURL)
+	if err != nil {
+		logger.V(4).Info("Cannot check instance obsolescence: failed to create launcher client", "err", err)
+		return false
+	}
+	instState, err := lClient.GetInstanceState(ctx, serverDat.InstanceID)
+	if err != nil {
```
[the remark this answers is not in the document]
MikeSpreitzer: There is normally no concurrent cleanup.

This noise is tolerable, it will rarely happen.

test/e2e/test-cases.sh
```diff
@@ -543,6 +548,58 @@ if [ "$E2E_PLATFORM" = "openshift" ]; then check_gpu_pin $req_post_restart; fi
 
 cheer Successful controller restart state recovery
 
+# ---------------------------------------------------------------------------
+# Delete Obsolete Sleeping Instances After ISC Update
+# ---------------------------------------------------------------------------
+
+intro_case Delete Obsolete Sleeping Instances After ISC Update
+
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$req_post_restart" ]'
```
MikeSpreitzer, line 557: This is redundant, it was established above (on line 540 in commit 236789486fa5e771f2fb037f5999747b2d41a597).

If this is just documentation then make it a comment.

If this is intended as a prereq check to make this test case fail in case of incorrect script editing then:
- a one-shot test would suffice, no need to wait, and
- this test case has a lot of other prereqs that are not being tested.

test/e2e/test-cases.sh
```diff
@@ -543,6 +548,58 @@ if [ "$E2E_PLATFORM" = "openshift" ]; then check_gpu_pin $req_post_restart; fi
...
+# Delete Obsolete Awake Instance on Unbinding
+# ---------------------------------------------------------------------------
+
+intro_case Delete Obsolete Awake Instance on Unbinding
+
+# This test verifies that when an awake instance's ISC is updated (making it
+# obsolete) and then the requester scales down, the controller deletes the
+# instance during unbinding rather than putting it to sleep.
+
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$req_post_restart" ]'
```
MikeSpreitzer, line 579: Also redundant.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1050,30 +1050,37 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
...
+						return fmt.Errorf("unable to put server to sleep because port not known: %w", err)
+					}
 				}
 			}
+			endpoint := fmt.Sprintf("%s:%d", providingPod.Status.PodIP, serverPort)
+			sleepURL := "http://" + endpoint + "/sleep"
+			resp, err := http.Post(sleepURL, "", nil)
+			if err != nil {
+				return fmt.Errorf("failed to put provider %q to sleep, POST %s got error: %w", serverDat.ProvidingPodName, sleepURL, err)
+			}
```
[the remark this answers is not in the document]
MikeSpreitzer: I think this is a reasonable suggestion.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: It is an efficiency matter. It can be addressed in a later PR. There are other efficiency matters to address too.

MikeSpreitzer (approved):
This needs more work, but is good enough to merge as progress.

PR #436, opened 2026-04-17
Propagate ISC labels and annotations to launcher Pod on bind/unbind

### LINKED ISSUE #432: Propagate ISC labels to launcher Pod on bind/unbind for InferencePool discovery
## Problem

When integrating FMA with the Inference Gateway (IGW), the InferencePool's EPP discovers backend pods via label selectors (e.g., `llm-d.ai/inference-serving: "true"`, `llm-d.ai/guide: "inference-scheduling"`). Launcher Pods created by FMA need these labels to be visible to the EPP.

The `InferenceServerConfig` (ISC) already has a `spec.modelServerConfig.labels` field designed for this purpose, and there is an existing TODO in the codebase:

```go
// TODO: sync desired and actual providingPod wrt labels (spec is mostly immutable, possible mutations are allowed)
```
— `pkg/controller/dual-pods/inference-server.go:458`

However, the dual-pods controller does **not** currently propagate ISC labels to the launcher Pod during binding.

## Expected behavior

When the dual-pods controller **binds** a server-requesting Pod to a launcher Pod:
- The labels from `isc.Spec.ModelServerConfig.Labels` should be applied to the launcher Pod

When the controller **unbinds** (requester deleted, scale-down):
- Those labels should be removed from the launcher Pod

This way, the EPP only discovers launcher Pods that have an active vLLM instance bound to a requester — sleeping or unbound launchers would not be visible to the InferencePool.

## Current workaround

For the PoC, we use two workarounds:

1. **CRD patch**: The `LauncherConfig` CRD's `spec.podTemplate.metadata` does not accept `labels` due to missing `x-kubernetes-preserve-unknown-fields: true`. We patch the CRD at runtime:
   ```bash
   kubectl patch crd launcherconfigs.fma.llm-d.ai --type=json -p='[
     {"op": "add",
      "path": "/spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/podTemplate/properties/metadata/x-kubernetes-preserve-unknown-fields",
      "value": true}
   ]'
   ```

2. **Static labels in LauncherConfig**: We set the InferencePool labels in `LauncherConfig.spec.podTemplate.metadata.labels`, so they are always present on the launcher Pod regardless of binding state.

This works but means the EPP sees all launcher Pods (including unbound/sleeping ones) as valid backends, which is not ideal.

## Proposed implementation

In the `bind()` function (`inference-server.go:~911`), after setting `DualLabelName`:
- Read `isc.Spec.ModelServerConfig.Labels`
- Apply them to the launcher Pod via `utils.MapSet`

In the `ensureUnbound()` function (`inference-server.go:~1081`), when putting the instance to sleep:
- Remove the ISC labels from the launcher Pod

## Related

- The CRD for `LauncherConfig` should also be updated to add `x-kubernetes-preserve-unknown-fields: true` on `spec.podTemplate.metadata` so that custom labels/annotations can be set in the pod template. This is a separate but related fix.
- This is needed for FMA + IGW/HPA integration as described in the `workload-autoscaling` well-lit path.

cc. @MikeSpreitzer

MikeSpreitzer:
## Summary

- Propagates `InferenceServerConfig.spec.modelServerConfig.labels` and `.annotations` to the launcher Pod on bind, removes them on unbind
- Tracks applied keys via `dual-pods.llm-d.ai/isc-label-keys` and `dual-pods.llm-d.ai/isc-annotation-keys` annotations on the Pod, enabling correct removal after controller restart
- Protects the tracking annotations from non-controller mutation via the `fma-immutable-fields` ValidatingAdmissionPolicy

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [ ] E2E: verify ISC label/annotation appear on launcher Pod after bind
- [ ] E2E: verify ISC label/annotation removed from launcher Pod after unbind
- [ ] E2E: verify ISC label/annotation re-appear after re-bind
- [ ] E2E: verify ValidatingAdmissionPolicy rejects non-controller changes to tracking annotations

Resolves #432

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -368,6 +368,13 @@ jobs:
           rm -f openshift-client-linux.tar.gz kubectl README.md
           # Install helm
           curl -fsSL --retry 3 --retry-delay 5 https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+          # Install yq if not already present (for YAML-to-JSON conversion in deploy_fma.sh)
+          if ! command -v yq &>/dev/null; then
+            YQ_VERSION="v4.53.2"
+            curl -fsSL --retry 3 --retry-delay 5 -o yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
+            chmod +x yq
+            sudo mv yq /usr/local/bin/
+          fi
 
       - name: Verify cluster access
         run: |
--- a/api/fma/v1alpha1/inferenceserverconfig_types.go
+++ b/api/fma/v1alpha1/inferenceserverconfig_types.go
@@ -44,7 +44,16 @@ type ModelServerConfig struct {
 	// EnvVars are the environment variables for the vLLM instance
 	// +optional
 	EnvVars     map[string]string `json:"env_vars,omitempty"`
-	Labels      map[string]string `json:"labels,omitempty"`
+	// Labels are applied to the server-providing Pod while bound.
+	// Keys must not start with "dual-pods.llm-d.ai/".
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.all(k, !k.startsWith('dual-pods.llm-d.ai/'))",message="label keys must not start with 'dual-pods.llm-d.ai/'"
+	Labels map[string]string `json:"labels,omitempty"`
+
+	// Annotations are applied to the server-providing Pod while bound.
+	// Keys must not start with "dual-pods.llm-d.ai/".
+	// +optional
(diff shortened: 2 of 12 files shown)


MikeSpreitzer:
The force-push above adds the following.

  - CRD validation: ISC rejects label/annotation keys starting with dual-pods.llm-d.ai/ via CEL XValidation rules
  - Collision check in bind(): refuses to overwrite existing pod labels/annotations with ISC values, reports collisions as status errors on the requesting pod
  - ensureUnbound() recovery: populates ISC key cache from tracking annotations when reached before the normal reconciliation path
  - E2E test: new "ISC Label and Annotation Key Validation" test case verifying rejected ISC creation, placed before the GPU probe

MikeSpreitzer:
Closing to re-open from upstream repo branch.

MikeSpreitzer:
See #437, since this now involves a change in the E2E-on-OpenShift workflow.

PR #443, opened 2026-04-21
Create launcher pod in bound state

### LINKED ISSUE #421: [Bug]: dual-pods controller should create launcher in bound state
### Contact Details

_No response_

### What happened?

This is from code inspection. I noticed that the dual-pods controller creates launcher Pods that are not bound to a requester. This is problematic because it means that there is a window of time during which the launcher population controller might delete that new launcher Pod.

### Version

### Branch name

_No response_

### Commit SHA

64262a54a652f2bc9204cf2fb6e687347ec62f76

### Relevant log output

```shell

```

waltforme:
Fixes #421

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -304,9 +304,9 @@ type serverData struct {
 	GPUIndices    []string
 	GPUIndicesStr *string
 
-	ProvidingPodName  string
-	InstanceID        string // ISC hash; set when computed, independent of instance existence
-	InstanceExists    *bool  // nil=unknown, *true=confirmed present, *false=confirmed absent
+	ProvidingPodName     string
+	InstanceID           string // ISC hash; set when computed, independent of instance existence
+	InstanceKnownToExist bool   // meaningful only for launcher-based providers
 	ISCLabelKeys      []string // keys of ISC labels applied to providingPod
 	ISCAnnotationKeys []string // keys of ISC annotations applied to providingPod
 
@@ -818,4 +818,3 @@ func (ctl *controller) clearServerData(nodeDat *nodeData, uid apitypes.UID) {
 	defer ctl.mutex.Unlock()
 	delete(nodeDat.InferenceServers, uid)
 }
-
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -305,7 +305,8 @@ type serverData struct {
 	GPUIndicesStr *string
 
 	ProvidingPodName  string
-	InstanceID        string   // if provider launcher-based
+	InstanceID        string // ISC hash; set when computed, independent of instance existence
+	InstanceExists    *bool  // nil=unknown, *true=confirmed present, *false=confirmed absent
 	ISCLabelKeys      []string // keys of ISC labels applied to providingPod
 	ISCAnnotationKeys []string // keys of ISC annotations applied to providingPod
 
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -982,7 +982,6 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
 		providingPod.Annotations[iscAnnotationKeysAnnotationKey] = strings.Join(annotationKeys, " ")
 	}
 	serverDat.Sleeping = nil
-	serverDat.InstanceExists = nil
 	echo, err := ctl.coreclient.Pods(ctl.namespace).Update(ctx, providingPod, metav1.UpdateOptions{FieldManager: ControllerName})
 	if err != nil {
 		return fmt.Errorf("failed to bind server-providing Pod %s: %w", providingPod.Name, err), true
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -425,10 +425,10 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 			_, instanceStopped := syncResult.stoppedInstanceIDs[serverDat.InstanceID]
 
 			if instanceStopped || !instancePresent {
-				if instanceStopped || (serverDat.InstanceExists != nil && *serverDat.InstanceExists) {
+				if instanceStopped || serverDat.InstanceKnownToExist {
 					// instanceStopped is an objective signal that the instance existed
-					// and died — no dependency on in-memory InstanceExists state.
-					// When !instancePresent && InstanceExists==true the instance vanished
+					// and died — no dependency on in-memory InstanceKnownToExist state.
+					// When !instancePresent && InstanceKnownToExist==true the instance vanished
 					// (e.g. launcher restart) — same treatment.
 					// Delete the requesting Pod first so the intent is durable in the
 					// Kubernetes API; the stopped vLLM instance is cleaned up by the
@@ -454,7 +454,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 					serverDat.RequesterDeleteRequested = true
 					return nil, false
 				}
-				// InstanceExists is nil (unknown) and instance is absent (not stopped) —
+				// InstanceKnownToExist is false and instance is absent (not stopped) —
 				// not yet created (bind-first path) or controller restarted and lost tracking.
 				// We just synced, so we know the instance is not on the launcher — create directly.
 				launcherBaseURL := fmt.Sprintf("http://%s:%d", providingPod.Status.PodIP, ctlrcommon.LauncherServicePort)
@@ -466,6 +466,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				if err != nil {
 					return fmt.Errorf("failed to create vLLM instance %q: %w", serverDat.InstanceID, err), true
 				}
+				serverDat.InstanceKnownToExist = true
 				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
 				launcherDat.Instances[serverDat.InstanceID] = time.Now()
 				logger.V(5).Info("Created vLLM instance", "instance_id", result.InstanceID, "status", result.Status)
@@ -475,7 +476,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 						isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations, true)
 				}
 			}
-			serverDat.InstanceExists = ptr.To(true)
+			serverDat.InstanceKnownToExist = true
 		}
 		if serverDat.Sleeping == nil {
 			sleeping, err := ctl.querySleeping(ctx, providingPod, serverPort)
@@ -1303,7 +1304,7 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
 	serverDat.ProvidingPodName = ""
 	serverDat.ServerPort = -1
 	serverDat.InstanceID = ""
-	serverDat.InstanceExists = nil
+	serverDat.InstanceKnownToExist = false
 	return nil
 }
 
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -469,8 +469,8 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
 				launcherDat.Instances[serverDat.InstanceID] = time.Now()
 				logger.V(5).Info("Created vLLM instance", "instance_id", result.InstanceID, "status", result.Status)
-				// If ISC tracking annotations are missing (pre-bound pod), complete the bind metadata.
-				if _, bound := providingPod.Annotations[iscLabelKeysAnnotationKey]; !bound {
+				// If ISC tracking annotations are missing (pre-bound pod), propagate the ISC metadata.
+				if _, propagated := providingPod.Annotations[iscLabelKeysAnnotationKey]; !propagated {
 					return ctl.bind(ctx, serverDat, requestingPod, providingPod, &serverDat.InstanceID, int16(isc.Spec.ModelServerConfig.Port),
 						isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations, true)
 				}
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -93,7 +93,15 @@ func ensureNamedLauncherInstance(
 
 	result, err := lClient.CreateNamedInstance(ctx, instanceID, cfg)
 	if err != nil {
-		return nil, fmt.Errorf("create vLLM instance: %w", err)
+		if IsInstanceAlreadyExistsError(err) {
+			inst, retryErr := lClient.GetInstanceState(ctx, instanceID)
+			if retryErr != nil {
+				return nil, fmt.Errorf("get existing vLLM instance %q after conflict: %w", instanceID, retryErr)
+			}
+			launcherDat.Instances[instanceID] = time.Now()
+			return inst, nil
+		}
+		return nil, fmt.Errorf("create vLLM instance %q: %w", instanceID, err)
 	}
 	launcherDat.Instances[instanceID] = time.Now()
 	return &InstanceState{
@@ -422,33 +430,34 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return fmt.Errorf("unable to wake up server because port not known: %w", err), true
 			}
 		}
-		// A launcher pod may be created in a bound state (requester annotation set at
-		// creation) before its vLLM instance is set up. Handle that here.
-		// The absence of DualLabelName distinguishes this from a controller restart
-		// where InstanceID is unknown but the instance already exists on the launcher.
+		// A launcher pod may be created in a pre-bound state, or the controller
+		// may have restarted and lost the in-memory InstanceID. In both cases,
+		// ensureNamedLauncherInstance recovers or creates the instance.
 		if launcherBased && serverDat.InstanceID == "" {
-			if _, hasDual := providingPod.Labels[api.DualLabelName]; !hasDual {
-				if providingPod.Status.PodIP == "" || !utils.IsPodReady(providingPod) {
-					return nil, true
-				}
-				cfg, iscHash, err := ctl.configInferenceServer(isc, serverDat.GPUIDs)
-				if err != nil {
-					return fmt.Errorf("failed to configure inference server config: %w", err), true
-				}
-				launcherBaseURL := fmt.Sprintf("http://%s:%d", providingPod.Status.PodIP, ctlrcommon.LauncherServicePort)
-				lClient, err := NewLauncherClient(launcherBaseURL)
-				if err != nil {
-					return err, true
-				}
-				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
-				result, err := ensureNamedLauncherInstance(ctx, lClient, launcherDat, iscHash, *cfg)
-				if err != nil {
-					return err, true
-				}
-				logger.V(5).Info("Ensured vLLM instance for pre-bound launcher Pod", "instance_id", result.InstanceID, "status", result.Status)
-				return ctl.bind(ctx, serverDat, requestingPod, providingPod, &iscHash, int16(isc.Spec.ModelServerConfig.Port),
-					isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations)
+			if providingPod.Status.PodIP == "" || !utils.IsPodReady(providingPod) {
+				return nil, true
 			}
+			cfg, iscHash, err := ctl.configInferenceServer(isc, serverDat.GPUIDs)
+			if err != nil {
+				return fmt.Errorf("failed to configure inference server config: %w", err), true
+			}
+			launcherBaseURL := fmt.Sprintf("http://%s:%d", providingPod.Status.PodIP, ctlrcommon.LauncherServicePort)
+			lClient, err := NewLauncherClient(launcherBaseURL)
+			if err != nil {
+				return err, true
+			}
+			launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
+			result, err := ensureNamedLauncherInstance(ctx, lClient, launcherDat, iscHash, *cfg)
+			if err != nil {
+				return err, true
+			}
+			logger.V(5).Info("Ensured vLLM instance", "instance_id", result.InstanceID, "status", result.Status)
+			if _, bound := providingPod.Annotations[iscLabelKeysAnnotationKey]; !bound {
+				return ctl.bind(ctx, serverDat, requestingPod, providingPod, &iscHash, int16(isc.Spec.ModelServerConfig.Port),
+					isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations, true)
+			}
+			serverDat.InstanceID = iscHash
+			serverDat.ServerPort = int16(isc.Spec.ModelServerConfig.Port)
 		}
 		// For launcher-based providers, check whether the bound instance is still alive.
 		// The sidecar notifier updates the Pod annotation when instance status changes,
@@ -580,7 +589,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				logger.V(2).Info("Unexpected: multiple sleeping Pods match; using the first", "requesterName", requestingPod.Name)
 			}
 			providingPod = sleepingAnys[0].(*corev1.Pod)
-			return ctl.bind(ctx, serverDat, requestingPod, providingPod, nil, -1, nil, nil)
+			return ctl.bind(ctx, serverDat, requestingPod, providingPod, nil, -1, nil, nil, false)
 		}
 		// What remains is to make a new server-providing Pod --- if the sleeper budget allows.
 
@@ -677,7 +686,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 					return fmt.Errorf("wake up vLLM instance: %w", err), true
 				}
 				launcherDat.Instances[iscHash] = time.Now()
-				return ctl.bind(ctx, serverDat, requestingPod, launcherPod, &iscHash, int16(isc.Spec.ModelServerConfig.Port), isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations)
+				return ctl.bind(ctx, serverDat, requestingPod, launcherPod, &iscHash, int16(isc.Spec.ModelServerConfig.Port), isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations, false)
 			} else {
 				// Slower path: create new instance in launcher with capacity
 				logger.V(5).Info("Creating new vLLM instance", "iscHash", iscHash)
@@ -690,7 +699,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 					"status", result.Status,
 				)
 				launcherDat.Instances[iscHash] = time.Now()
-				return ctl.bind(ctx, serverDat, requestingPod, launcherPod, &iscHash, int16(isc.Spec.ModelServerConfig.Port), isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations)
+				return ctl.bind(ctx, serverDat, requestingPod, launcherPod, &iscHash, int16(isc.Spec.ModelServerConfig.Port), isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations, false)
 			}
 		}
 	}
@@ -704,9 +713,9 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	// Sleeper budget is met. Make a new launcher Pod.
 
 	// Bind at creation time so the launcher-populator cannot delete this pod
-	// while the vLLM instance is being set up. The dual label is added
-	// later by bind() once the instance is ready.
+	// while the vLLM instance is being set up.
 	desiredLauncherPod.Annotations = utils.MapSet(desiredLauncherPod.Annotations, requesterAnnotationKey, string(requestingPod.UID)+" "+requestingPod.Name)
+	desiredLauncherPod.Labels = utils.MapSet(desiredLauncherPod.Labels, api.DualLabelName, requestingPod.Name)
 	if !slices.Contains(desiredLauncherPod.Finalizers, providerFinalizer) {
 		desiredLauncherPod.Finalizers = append(desiredLauncherPod.Finalizers, providerFinalizer)
 	}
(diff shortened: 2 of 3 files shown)


pkg/controller/dual-pods/inference-server.go
```diff
@@ -42,8 +42,8 @@ import (
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	k8sserializer "k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/sets"
-	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
```
waltforme, line 46: Some linter suggested this reordering.

MikeSpreitzer: I am surprised that this disorder even got merged. I thought that we had stuff both preventing and checking that. Worth an Issue.

If this fix were a separate PR then I would have already merged it.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -75,6 +75,34 @@ type launcherSyncResult struct {
...
+		return inst, nil
+	}
+	if !IsInstanceNotFoundError(err) {
+		return nil, fmt.Errorf("get vLLM instance %q: %w", instanceID, err)
+	}
+
+	result, err := lClient.CreateNamedInstance(ctx, instanceID, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("create vLLM instance: %w", err)
+	}
```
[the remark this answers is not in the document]
MikeSpreitzer: It is not just a matter of noise. In case of controller restart or concurrent copies of the controller, this condition will keep getting hit. For functional correctness, the 409 MUST be recognized and handled.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,6 +422,34 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return fmt.Errorf("unable to wake up server because port not known: %w", err), true
 			}
 		}
+		// A launcher pod may be created in a bound state (requester annotation set at
+		// creation) before its vLLM instance is set up. Handle that here.
+		// The absence of DualLabelName distinguishes this from a controller restart
+		// where InstanceID is unknown but the instance already exists on the launcher.
+		if launcherBased && serverDat.InstanceID == "" {
+			if _, hasDual := providingPod.Labels[api.DualLabelName]; !hasDual {
```
MikeSpreitzer: (A) The dual label is defined to be FYI --- written by the controller but not read. Do not sneak in a change to that. If a change must be made then make it properly: change the statements that it is FYI, and add it to the set of labels that are protected against change by other parties.

(B) This is an inadequate defense against the condition of the instance existing but not being recorded in `serverDat`. The controller could crash right after sending the request to create the instance. The proper defense has to be in properly reacting to the launcher revealing that the instance already exists.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,6 +422,34 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+				if err != nil {
+					return err, true
+				}
+				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
+				result, err := ensureNamedLauncherInstance(ctx, lClient, launcherDat, iscHash, *cfg)
+				if err != nil {
+					return err, true
+				}
+				logger.V(5).Info("Ensured vLLM instance for pre-bound launcher Pod", "instance_id", result.InstanceID, "status", result.Status)
+				return ctl.bind(ctx, serverDat, requestingPod, providingPod, &iscHash, int16(isc.Spec.ModelServerConfig.Port),
```
MikeSpreitzer: +1 on code re-use, although some things that `bind` does are redundant in this case. However, `bind` calls `ctl.wakeSleeper`, which will probably fail because the vllm instance is not ready to handle that call yet. It would be better to not suffer this predictable failure.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -649,6 +705,14 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 	// Sleeper budget is met. Make a new launcher Pod.
 
+	// Bind at creation time so the launcher-populator cannot delete this pod
+	// while the vLLM instance is being set up. The dual label is added
+	// later by bind() once the instance is ready.
+	desiredLauncherPod.Annotations = utils.MapSet(desiredLauncherPod.Annotations, requesterAnnotationKey, string(requestingPod.UID)+" "+requestingPod.Name)
+	if !slices.Contains(desiredLauncherPod.Finalizers, providerFinalizer) {
+		desiredLauncherPod.Finalizers = append(desiredLauncherPod.Finalizers, providerFinalizer)
+	}
+
```
MikeSpreitzer, line 660: Once it is restored to its write-only treatment by this controller, the dual label could be set here. This is not necessary, since other code will ensure its presence later, but it is nice to get the information out as soon as it is clear to the controller.

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: I opened #447 for the tooling issue.

pkg/controller/dual-pods/inference-server.go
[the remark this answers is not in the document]

pkg/controller/dual-pods/inference-server.go

pkg/controller/dual-pods/inference-server.go

pkg/controller/dual-pods/inference-server.go
waltforme: Done.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -75,6 +75,42 @@ type launcherSyncResult struct {
...
+	instanceID string,
+	cfg VllmConfig,
+) (*InstanceState, error) {
+	inst, err := lClient.GetInstanceState(ctx, instanceID)
+	if err == nil {
+		launcherDat.Instances[instanceID] = time.Now()
+		return inst, nil
+	}
+	if !IsInstanceNotFoundError(err) {
+		return nil, fmt.Errorf("get vLLM instance %q: %w", instanceID, err)
```
MikeSpreitzer: I would write "_failed_ to get ..." here.
Similarly in the other error wraps.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -75,6 +75,42 @@ type launcherSyncResult struct {
...
+	result, err := lClient.CreateNamedInstance(ctx, instanceID, cfg)
+	if err != nil {
+		if IsInstanceAlreadyExistsError(err) {
+			inst, retryErr := lClient.GetInstanceState(ctx, instanceID)
+			if retryErr != nil {
+				return nil, fmt.Errorf("get existing vLLM instance %q after conflict: %w", instanceID, retryErr)
+			}
+			launcherDat.Instances[instanceID] = time.Now()
+			return inst, nil
+		}
```
MikeSpreitzer: In my opinion this special case handling is not worth the trouble. It is only an optimization, right? This situation should be very rare.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,6 +430,35 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return fmt.Errorf("unable to wake up server because port not known: %w", err), true
 			}
 		}
+		// A launcher pod may be created in a pre-bound state, or the controller
+		// may have restarted and lost the in-memory InstanceID. In both cases,
+		// ensureNamedLauncherInstance recovers or creates the instance.
+		if launcherBased && serverDat.InstanceID == "" {
+			if providingPod.Status.PodIP == "" || !utils.IsPodReady(providingPod) {
+				return nil, true
```
MikeSpreitzer: Re-queuing is not necessary in this case. The future changes in the providingPod will cause the infSvrItem to be enqueued.

pkg/controller/dual-pods/inference-server.go
waltforme: Changed all.

pkg/controller/dual-pods/inference-server.go
waltforme: Agreed, changed.

MikeSpreitzer:
On further thought, I think that the problem is a little bigger than fixing #421, and the solution needs to be addressed at the whole problem.

Regarding the problem: note that the dual-pods controller can crash right after waking (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L619) or creating (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L629) a vllm instance in a launcher --- in which case the following call to `ctl.bind` does not happen (before the crash).

Note also this unwritten invariant, designed in milestone 2 and built into the definition of launcher population control: vllm awake implies provider Pod is bound. Put another way: in a wake-up scenario the provider Pod is first bound and then then vllm instance is woken up, and in a go-to-sleep scenario the vllm instance is first put to sleep and then the provider Pod is unbound. But the current milestone 3 code violates this invariant in the causes where a suitable unbound launcher is sought (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L590) and found.

Let me make a couple of observations while warming up to the solution.

Note that in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L400, the `&& providingPod.Status.PodIP != ""` is unnecessary, because `serverDat.InstanceID != ""` implies `providingPod.Status.PodIP != ""` (in the current code). BTW, it would be nice to have invariants like that written in the code; in this case, a comment somewhere in the definition of `serverData`. Also, this particular invariant might go away, because of the following.

Note that the vllm instance ID is known as early as it is computed --- at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L577, which precedes the calls to `bind`.

So here are my current thoughts about solution.

1. Change `serverDat.InstanceID` so that it no longer conveys information about vllm instance existence/discovery, it only holds the instance ID returned as `iscHash` at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L577
2. Add a field to `serverData`, meaningful only when launcher based, `InstanceExists *bool`. It is given a non-`nil` value after the instance sync call that is currently at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L401 but: that sync will be done regardless of `serverDat.InstanceID` and `serverDat.InstanceExists`.
3. Factor `bind` into two parts. The first part just updates `providerPod` in-memory object, and the second part does most of the rest. The first part is called before the API call to create the launcher. In the case of waking or creating a vllm instance in a previously-unbound launcher (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L616-L640), the first part is called and then the kube apiserver call is made to Update the launcher Pod; once that Update returns, this invocation of `infSvrItem.process` completes (no more wake or create a vllm instance here) successfully, relying on the notification of the update to trigger the remaining processing. That will get to https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/9ae47f4d63f1f821322e19add8f92880e4a3e61f/pkg/controller/dual-pods/inference-server.go#L400 and the call to `ctl.syncLauncherInstances`. After the sync, we need handling for the new possibility of a bound launcher with the vllm instance not existing. Here is where the wake or create logic goes. Around here is where the remainder of `bind` goes.
4. Here is a separate problem that is noticeable here but pre-existing and could be addressed in a follow-on PR: while a vllm instance starts up the dual-pods controller should poll for readiness (which it effectively does by querying `/is_sleeping`) at something like every 5 seconds, rather than go through an exponential backoff starting with something very much shorter and growing to something longer.

pkg/controller/dual-pods/inference-server.go
waltforme: This is part of the handling of launcher returned 409. Concurrent copies of the controller could result in this case, right?

MikeSpreitzer: Yes. Concurrent copies of the controller should be a very rare thing. Immediately returning from `infSrItem.process` with a retry request would also be functionally correct, right?

waltforme: I'm not following. If we remove the `if IsInstanceAlreadyExistsError(err)` block, then we return to the initial state that Copilot complained about at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/443#discussion_r3117696427, right?

waltforme:
The force-push to https://github.com/llm-d-incubation/llm-d-fast-model-actuation/commit/83a551482a2dec7941809ee31dfa0d4e863e7c25 was a rebase onto `main` due to small and nonfunctional conflicts introduced by the merge of https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/453.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: Right. That complaint is about "noisy reconcile errors/backoff" in the rare circumstances. That is, not a functional error but more activity (including logging) than is strictly necessary. I suspect that Copilot does not understand how rare these circumstances are.

Again, this code is not wrong; I am just suggesting that it is not worth the effort to develop/maintain. It is not a show-stopper; I can accept it if you really want.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,6 +430,35 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return fmt.Errorf("unable to wake up server because port not known: %w", err), true
 			}
 		}
+		// A launcher pod may be created in a pre-bound state, or the controller
+		// may have restarted and lost the in-memory InstanceID. In both cases,
+		// ensureNamedLauncherInstance recovers or creates the instance.
+		if launcherBased && serverDat.InstanceID == "" {
+			if providingPod.Status.PodIP == "" || !utils.IsPodReady(providingPod) {
+				return nil, false
```
MikeSpreitzer, line 412: It is good to have an explanatory log message on each path out of `process` that does not return an error, so that the log makes it clear which path was taken.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -647,6 +712,14 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	}
 	// Sleeper budget is met. Make a new launcher Pod.
 
+	// Bind at creation time so the launcher-populator cannot delete this pod
+	// while the vLLM instance is being set up.
+	desiredLauncherPod.Annotations = utils.MapSet(desiredLauncherPod.Annotations, requesterAnnotationKey, string(requestingPod.UID)+" "+requestingPod.Name)
+	desiredLauncherPod.Labels = utils.MapSet(desiredLauncherPod.Labels, api.DualLabelName, requestingPod.Name)
```
MikeSpreitzer, line 656: FYI, `BuildLauncherPodFromTemplate` ensures that `desiredLauncherPod.Annotations` and `desiredLauncherPod.Labels` are not `nil`, so these statements can use bare indexing instead of `utils.MapSet`.

pkg/controller/dual-pods/inference-server.go
waltforme: Added.

waltforme:
Tried to implement the idea from [this comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/443#issuecomment-4310725850) by https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/443/commits/045535f1d097382c87c44a865ee67526eb469bde.

Also created https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/455 for point 4 of the comment.

pkg/controller/dual-pods/inference-server.go
waltforme: OK. Let's keep it then.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -205,6 +205,13 @@ func IsInstanceNotFoundError(err error) bool {
 	return errors.As(err, &launcherErr) && launcherErr.StatusCode == http.StatusNotFound
 }
 
+// IsInstanceAlreadyExistsError returns true when the launcher reports that the
+// instance already exists (HTTP 409 Conflict).
+func IsInstanceAlreadyExistsError(err error) bool {
```
MikeSpreitzer: Technically this func only determines whether the error is an HTTP 409 Conflict, and it is the caller that makes the deduction that a 409 implies that the vllm instance already exists.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,50 +430,86 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 			}
-			if _, deletedStopped := syncResult.deletedStoppedInstanceIDs[serverDat.InstanceID]; deletedStopped || !instancePresent {
-				if deletedStopped {
-					logger.V(2).Info("Deleted stopped bound instance from launcher during sync")
-				} else {
-					logger.V(2).Info("Bound instance not found in launcher after sync, treating as deleted")
+
+			_, deletedStopped := syncResult.deletedStoppedInstanceIDs[serverDat.InstanceID]
+			if deletedStopped || !instancePresent {
+				if serverDat.InstanceExists != nil && *serverDat.InstanceExists {
```
MikeSpreitzer: To properly resolve the create/delete ambiguity (is this instance absent because it wasn't created yet or because it was created and then deleted?), we can not rely on the controller's memory --- the controller can crash and restart at any moment. We need to orchestrate clarity through the kube API objects. Following is the first idea that occurs to me. When reacting to a vllm instance being stopped, this controller should relay that to server-requesting Pod deletion **before** commanding deletion of the vllm instance.

This can be addressed in a follow-on PR if you prefer.

Unless and until the above is addressed, should the condition in this `if` statement start with `deletedStopped ||` ?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,50 +430,86 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
-					Preconditions:     &metav1.Preconditions{UID: &item.UID, ResourceVersion: &requestingPod.ResourceVersion}})
-				if err == nil {
-					logger.V(2).Info("Requested deletion of server-requesting Pod because bound instance stopped")
-				} else if apierrors.IsGone(err) || apierrors.IsNotFound(err) {
-					logger.V(5).Info("The server-requesting Pod is already gone")
-				} else {
-					return fmt.Errorf("failed to delete server-requesting Pod for stopped instance: %w", err), true
+				// InstanceExists is nil (unknown) — instance hasn't been created yet
+				// (bind-first path) or controller restarted and lost tracking.
+				// ensureNamedLauncherInstance is idempotent: GET first, create if not found.
```
MikeSpreitzer: That is no longer necessary. We just did `ctl.syncLauncherInstances` above, we know that this controller has not commanded creation of the instance. It is OK to unconditionally command creation of the instance here. In the very rare circumstance that something else concurrently commanded creation of the instance: returning a (error, true) will be adequate handling; alternatively, I think that it might be OK to assume that the concurrent thing was another copy of this controller and the concurrently-created instance is just what this copy wanted to happen.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -574,7 +646,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		return err, false
 	}
 
-	cfg, iscHash, err := ctl.configInferenceServer(isc, serverDat.GPUIDs)
+	_, iscHash, err := ctl.configInferenceServer(isc, serverDat.GPUIDs)
```
MikeSpreitzer: Call this once, right after setting serverDat.GPUIDs, and be done with it. Save both outputs in serverDat.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,50 +430,86 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+					return err, true
+				}
+				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
+				result, err := ensureNamedLauncherInstance(ctx, lClient, launcherDat, serverDat.InstanceID, *cfg)
+				if err != nil {
+					return err, true
+				}
+				logger.V(5).Info("Ensured vLLM instance", "instance_id", result.InstanceID, "status", result.Status)
+				// If ISC tracking annotations are missing (pre-bound pod), complete the bind metadata.
+				if _, bound := providingPod.Annotations[iscLabelKeysAnnotationKey]; !bound {
```
MikeSpreitzer: `bound` is not the right name for the condition here. The launcher is certainly bound. The question is whether the ISC labels and annotations have been propagated yet.

MikeSpreitzer (commented):

pkg/controller/dual-pods/launcherclient.go
waltforme: Revised.

pkg/controller/dual-pods/inference-server.go
waltforme: I tried to address this within this PR, using the suggested approach.

pkg/controller/dual-pods/inference-server.go
waltforme: Changed.

pkg/controller/dual-pods/inference-server.go
waltforme: Removed the redundancy.

pkg/controller/dual-pods/inference-server.go
waltforme: Renamed.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,50 +406,76 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+					// Kubernetes API; the stopped vLLM instance is cleaned up by the
+					// next sync after the server data is removed.
+					if instanceStopped {
+						logger.V(2).Info("Bound instance found stopped on launcher")
+					} else {
+						logger.V(2).Info("Bound instance not found in launcher, treating as dead")
+					}
+					// Mark as sleeping so that ensureUnbound (called during requester deletion)
+					// does not attempt to POST /sleep on the dead instance.
+					serverDat.Sleeping = ptr.To(true)
```
MikeSpreitzer, line 443: This is a sign of cruftiness. What this is really trying to say is that this vllm instance no longer exists, and now there is a way to say that (`serverDat.InstanceExists != nil && !(*serverDat.InstanceExists)`). Can be addressed in a follow-on PR.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -363,6 +362,19 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+	var cfg *VllmConfig
+	var iscHash string
+	if launcherBased {
+		cfg, iscHash, err = ctl.configInferenceServer(isc, serverDat.GPUIDs)
+		if err != nil {
+			return fmt.Errorf("failed to configure inference server config: %w", err), true
+		}
+		if serverDat.InstanceID == "" {
+			serverDat.InstanceID = iscHash
+			serverDat.ServerPort = int16(isc.Spec.ModelServerConfig.Port)
```
MikeSpreitzer, line 374: Why not save `cfg` in the `serverDat` too? That would reducing the calling of `ctl.configInferenceServer` from once per `infSvItem.process` invocation to once per inference server.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1301,6 +1303,8 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
 	}
 	serverDat.ProvidingPodName = ""
 	serverDat.ServerPort = -1
+	serverDat.InstanceID = ""
+	serverDat.InstanceExists = nil
```
MikeSpreitzer: In my opinion all this information removal is misguided. IIRC, this started with a review comment from Copilot --- which did not understand the lifecycle of an inference server. A server-requesting Pod is only ever bound to at most one server-providing Pod. Note that these information removing statements are in the `ensureUnbound` method, which is only called when the server-requesting Pod is absent or in the process of being deleted. Removing information here only impoverishes future diagnostics (e.g., debug loggin).

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1669,26 +1673,37 @@ func (ctl *controller) syncLauncherInstances(ctx context.Context, nodeDat *nodeD
...
+				// delete the requesting Pod first (resolves create/delete ambiguity).
+				stoppedInstanceIDs.Insert(inst.InstanceID)
+				logger.V(2).Info("Found stopped bound instance, deferring cleanup",
 					"instanceID", inst.InstanceID)
-				deletedStoppedInstanceIDs.Insert(inst.InstanceID)
-				continue
+			} else {
+				_, delErr := lClient.DeleteInstance(ctx, inst.InstanceID)
+				if delErr != nil && !IsInstanceNotFoundError(delErr) {
+					logger.V(3).Info("Failed to delete stopped instance from launcher during sync",
```
MikeSpreitzer, line 1699: Perhaps in this case there should be a retry, so that a garbage instance is not left behind.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,50 +406,76 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+				result, err := lClient.CreateNamedInstance(ctx, serverDat.InstanceID, *cfg)
+				if err != nil {
+					return fmt.Errorf("failed to create vLLM instance %q: %w", serverDat.InstanceID, err), true
+				}
+				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
+				launcherDat.Instances[serverDat.InstanceID] = time.Now()
+				logger.V(5).Info("Created vLLM instance", "instance_id", result.InstanceID, "status", result.Status)
+				// If ISC tracking annotations are missing (pre-bound pod), propagate the ISC metadata.
+				if _, propagated := providingPod.Annotations[iscLabelKeysAnnotationKey]; !propagated {
+					return ctl.bind(ctx, serverDat, requestingPod, providingPod, &serverDat.InstanceID, int16(isc.Spec.ModelServerConfig.Port),
```
MikeSpreitzer, line 475: Since `ctl.bind` does more than binding, perhaps it should be renamed to something like `bindAndPropagate`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -983,6 +982,7 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
 		providingPod.Annotations[iscAnnotationKeysAnnotationKey] = strings.Join(annotationKeys, " ")
 	}
 	serverDat.Sleeping = nil
+	serverDat.InstanceExists = nil
```
MikeSpreitzer: This is not accurate, for the call site that is currently at line 474 (`if _, propagated := providingPod.Annotations[iscLabelKeysAnnotationKey]; !propagated`, following successful `lClient.CreateNamedInstance`).

This is not needed, for the other call sites. This field starts out `nil`, and only needs to change once (and that change will be in the positive direction).

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,50 +406,76 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+				logger.V(5).Info("Created vLLM instance", "instance_id", result.InstanceID, "status", result.Status)
+				// If ISC tracking annotations are missing (pre-bound pod), propagate the ISC metadata.
+				if _, propagated := providingPod.Annotations[iscLabelKeysAnnotationKey]; !propagated {
+					return ctl.bind(ctx, serverDat, requestingPod, providingPod, &serverDat.InstanceID, int16(isc.Spec.ModelServerConfig.Port),
+						isc.Spec.ModelServerConfig.Labels, isc.Spec.ModelServerConfig.Annotations, true)
 				}
-				serverDat.RequesterDeleteRequested = true
-				return nil, false
 			}
+			serverDat.InstanceExists = ptr.To(true)
```
MikeSpreitzer: I think that this code will be less crufty if this is done as soon as it is known to be true. That would be as soon as the call to `lClient.CreateNamedInstance` is known to not have erred, and if `!instanceStopped && instancePresent`.

pkg/controller/dual-pods/controller.go
```diff
@@ -305,7 +305,8 @@ type serverData struct {
 	GPUIndicesStr *string
 
 	ProvidingPodName  string
-	InstanceID        string   // if provider launcher-based
+	InstanceID        string // ISC hash; set when computed, independent of instance existence
+	InstanceExists    *bool  // nil=unknown, *true=confirmed present, *false=confirmed absent
```
MikeSpreitzer: Since this field only ever takes on two values (currently `nil` or `&true`), it could be typed as a `bool` and given an appropriate name (e.g., `InstanceKnownToExist`).

pkg/controller/dual-pods/inference-server.go
waltforme: Removed this line.

pkg/controller/dual-pods/inference-server.go
waltforme: Ture. Changed as suggested.

pkg/controller/dual-pods/controller.go
waltforme: Yes. Changed.

waltforme:
I made changes to all comments that concern about `serverDat.InstanceExists`. Maybe other comments can be optional wrt the scope of this PR?

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer, line 993: This is now unneeded, because this field is set earlier. Can be addressed later.

pkg/controller/dual-pods/controller.go
```diff
@@ -304,8 +304,9 @@ type serverData struct {
 	GPUIndices    []string
 	GPUIndicesStr *string
 
-	ProvidingPodName  string
-	InstanceID        string   // if provider launcher-based
+	ProvidingPodName     string
+	InstanceID           string // ISC hash; set when computed, independent of instance existence
+	InstanceKnownToExist bool   // meaningful only for launcher-based providers
 	ISCLabelKeys      []string // keys of ISC labels applied to providingPod
 	ISCAnnotationKeys []string // keys of ISC annotations applied to providingPod
```
MikeSpreitzer, line 311: BTW, the code that recovers these in the case of controller crash and restart is duplicated (appears in two places). A later PR could refactor to remove the duplication.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer, line 985: Why is this statement here?
Can be addressed in a later PR.

MikeSpreitzer, line 1001: Now that this is initialized much earlier in `infSvrItem.process`, the assignment here is unnecessary. The `instPort` parameter of this method is unnecessary.
Can be addressed in a later PR.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -394,50 +406,77 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+				if err != nil {
+					return fmt.Errorf("failed to create vLLM instance %q: %w", serverDat.InstanceID, err), true
+				}
+				serverDat.InstanceKnownToExist = true
+				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
+				launcherDat.Instances[serverDat.InstanceID] = time.Now()
+				logger.V(5).Info("Created vLLM instance", "instance_id", result.InstanceID, "status", result.Status)
+				// If ISC tracking annotations are missing (pre-bound pod), propagate the ISC metadata.
+				if _, propagated := providingPod.Annotations[iscLabelKeysAnnotationKey]; !propagated {
+					return ctl.bind(ctx, serverDat, requestingPod, providingPod, &serverDat.InstanceID, int16(isc.Spec.ModelServerConfig.Port),
```
MikeSpreitzer, line 475: FYI, it took me some time to verify that all the other things that `bind` does can be skipped if `propagated` is `false`. This is part of what I mean about this code being "crufty". There are a lot of important things that are not written down. To be addressed in a later PR.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: To be fair, it does more than "bind" and "propagate", so a truly good name would cover all of it.

MikeSpreitzer (approved):
This is good enough to merge. This needs more work, which can be done in later PR(s).

Issue #408, opened 2026-04-03
[Feature]: Need better control over main memory usage


MikeSpreitzer:
### Feature Area

Other

### Problem Statement

The `maxSleepingInstances` (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/da4a87a444a8289d7e9ecdaf6d8adbf7e4bcefd4/api/fma/v1alpha1/launcherconfig_types.go#L32) of a LauncherConfig and the design of LauncherPopulationPolicy allow the authors of these objects to limit the amount of main memory that will be used by vllm instances. But there are two important gaps that remain.

(1) These limits are in terms of number of vllm instances, not main memory used. The amount of main memory used by a vllm instance can vary A LOT from model to model and depending on other parameters of vllm.

(2) These limits only apply within the scope of one LauncherConfig. If/when/while there are multiple LauncherConfig objects present, there is currently nothing that speaks to the sum of main memory usage over all the LauncherConfig objects (other than, of course, the sum of what is allowed for each LauncherConfig --- which I suspect is not enough control).

### Proposed Solution

TBD

### Willingness to Contribute

Yes, I can submit a PR

MikeSpreitzer:
Unfortunately, Issue #246 has suffered mission creep and now overlaps with this Issue.

MikeSpreitzer:
The following comments about this issue were made in #246:
- https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/246#issuecomment-4379700064
- https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/246#issuecomment-4379716911
- https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/246#issuecomment-4379732992
- https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/246#issuecomment-4380195177
- https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/246#issuecomment-4381030570
- https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/246#issuecomment-4381553223

Let us continue the discussion of this topic in _this_ Issue, cease discussing it in #246.

Issue #246, opened 2026-02-11
`MaxSleepingInstances` for launcher-based server-providing Pods


waltforme:
`MaxSleepingInstances` for launcher-based server-providing Pods is not enforced yet. Creating the issue to track this.

osswangxining:
@MikeSpreitzer @waltforme   It looks like LauncherConfig is NOT the right place for node-level limits. Because LauncherConfig is a global template object with no binding to any specific node. The same LauncherConfig can be referenced by multiple LauncherPopulationPolicy objects and deployed onto different types of nodes. Putting a node-level limit inside LauncherConfig creates a semantic contradiction

osswangxining:
I suggest Per-node limit → belongs in LauncherPopulationPolicy.

```
// New fields in LauncherPopulationPolicySpec
type LauncherPopulationPolicySpec struct {
    EnhancedNodeSelector EnhancedNodeSelector `json:"enhancedNodeSelector"`
    CountForLauncher []CountForLauncher `json:"countForLauncher"`
    
    // NEW: total sleeping instances memory limit on nodes matched by this policy
    // +optional
    NodeSleepingMemoryLimit *resource.Quantity `json:"nodeSleepingMemoryLimit,omitempty"`
    
    // Or: total sleeping instances count limit across all launcher pods on the node
    // +optional  
    NodeMaxSleepingInstances *int32 `json:"nodeMaxSleepingInstances,omitempty"`
}
```

Benefits:
Different node types can have different limits (high-memory nodes vs. low-memory nodes)
Semantically consistent with LPP's "node selection + deployment policy" responsibility
launcher-populator can compute node-level budget while building the desired state

osswangxining:
which controller enforces the per-node limit? 

- launcher-populator manages launcher pod lifecycle (creation/deletion), suitable for static budget (check before creation)
- dual-pods controller manages sleeping instance eviction, suitable for dynamic budget (runtime eviction)

And, The most reasonable division of labor: launcher-populator considers node-level budget when reserving launcher pod counts on nodes; dual-pods controller evicts sleeping instances according to node-level budget before creating new instances.

@waltforme @MikeSpreitzer  does this make sense?

MikeSpreitzer:
I agree that the node-level limits do not belong in LauncherConfig, and that LauncherPopulationPolicy is a plausible place for them (because it is already based on a node selector).

[...]

PR #454, opened 2026-04-24
docs: Address PR #381 review feedback

rubambiza:
## Summary

Follow-up to #381, addressing review feedback from Mike and Ansu on the benchmarking scenarios doc.

### Actuation path and naming changes
- **Rename lukewarm to "Cold Start (with launcher)"**: the path is cold, not warm. Metric renamed from T_luke_warm to T_cold_launcher.
- **Remove M2 actuation path**: focus the matrix entirely on M3 (launcher-based) paths.
- **Rename T_launcher to T_instance_create**: the metric measures instance creation, not launcher behavior (per Ansu's feedback).

### Metric precision (end-instant clarity)
- **T_actuation**: explicitly ends at kubelet readiness probe on the requester pod (Pod Ready condition True).
- **T_cold_launcher, T_instance_create, T_instance_ready**: explicitly end at DPC readiness relay (V5 log: "Successfully relayed the readiness").
- **Rename Hit_rate to Hot_hit_rate**: parallels the new Warm_hit_rate metric.
- **Add Warm_hit_rate**: fraction of requests satisfied by an existing launcher pod.

### Constituent metrics and observability
- **Add constituent duration metrics table**: T_launcher_schedule, T_launcher_startup, T_dpc_react, T_instance_ready as planned sub-metrics with observability sources.
- **T_dpc_react**: replace "negligible" with precise language; add note about later DPC actions (labels/annotations) being captured by T_instance_ready; mark observable as not yet present, link #495.
- **T_instance_create start instant**: link #497 for subsecond launcher logging.
- **Add alternative observability note**: Prometheus histograms and distributed tracing (OpenTelemetry) as alternatives to log parsing.

### Other fixes
- **Add L2 to Resource Scaling and Stress Test**: TTFT cost per requester is low relative to actuation time, and the data is useful at scale.
- **Drop LPC attribution from Warm Start**: DPC does not care whether the pre-existing launcher was created by LPC or by a prior DPC reconciliation.
- **Fix node-level language**: launchers are on Nodes, not GPUs.
- **Change "launcher API responses" to "launcher logs"** in constituent metrics preamble.
- **Align Phase 2 descriptions** with updated end-instant definitions.
- **Remove unused L1+L3 legend entry**; expand L1+L2+L3 description.

## Test plan

- [x] Verify markdown tables render correctly
- [x] Confirm all 4 actuation path columns match between Paths table and Matrix
- [x] Confirm metric names are consistent across definitions, L1 legend, and Integration Phases
- [x] Verify issue links (#495, #497) resolve correctly

## Related issues

- #495 -- Add pre-`CreateNamedInstance` DPC log statement
- #497 -- Add subsecond launcher logging for create request receipt

### DIFF
--- a/inference_server/benchmark/benchmark.md
+++ b/inference_server/benchmark/benchmark.md
@@ -4,10 +4,11 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
 
 ## Purpose
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under four different actuation conditions
+and server-providing pods) becomes available under different actuation conditions
 in order of decreasing latency:
 
 - **Cold start without FMA**: creating a new vLLM instance without using a launcher
+- **Cold start (FMA M2)**: DPC creates a standalone server-providing pod directly (planned)
 - **Cold start (with launcher)**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
 - **Warm start**: creating a new vLLM instance in an existing launcher pod
 - **Hot start**: waking a sleeping vLLM instance on an existing launcher pod
@@ -37,7 +38,7 @@ direct scope but is referenced for completeness and handoff to other frameworks.
 
 **Metric definitions:**
 
-- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or cold start with launcher (T_cold_launcher).
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. For FMA paths, spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or cold start with launcher (T_cold_launcher). For non-FMA and M2 cold starts, T_actuation is measured directly with no FMA-specific sub-components.
 - **T_wake**: Request-response time for the DPC's `/wake_up` call to a sleeping vLLM instance on the server-providing pod. A part of T_actuation when a hot start occurs.
 - **Hit_rate**: Fraction of server-requesting Pods that get satisfied by waking a sleeping vLLM instance.
 - **T_cold_launcher**: Time from the DPC launcher pod creation to the new vLLM instance reporting healthy. Covers the full cold start (with launcher) span: launcher pod scheduling, launcher startup, and vLLM instance creation.
@@ -84,9 +85,9 @@ using the team's established terminology:
 | ------------------------- | ---------------- | ------------ | ---------------------- |
 | **Cold Start (no FMA)**   | No FMA involvement. Raw Kubernetes deploy-to-ready latency. | Non-FMA baseline that all FMA paths should improve upon. | `-t standalone` comparison baseline |
 | **Cold Start (FMA M2)**   | DPC creates a standalone server-providing pod directly (no launcher). The requesting pod uses the `server-patch` annotation. | Measures the DPC's contribution without the launcher. Planned: pending a stable M3 harness. | `-t fma` with `fma.mode: m2` (planned) |
-| **Cold Start (with launcher)** | No suitable launcher pod. The DPC creates a new launcher pod, then DPC commands the launcher to create a new vLLM instance. | Worst-case FMA launcher path, relevant for dynamic situations such as LauncherConfig rollouts or newly added nodes where launchers have not yet been populated. | `-t fma` with LauncherPopulationPolicy that does not cover the target node |
-| **Warm Start**            | A launcher pod exists (pre-created by the LPC) but no sleeping instance is available on the assigned GPU. Launcher creates a new vLLM instance with module preloading. | Measures the launcher's contribution when no sleeping instance is available. | `-t fma` with default `LLMDBENCH_FMA_LAUNCHER_*` env vars |
-| **Hot Start**             | A sleeping vLLM instance exists on the correct GPU. DPC sends `/wake_up`. | Best-case FMA path. Measures sleep-to-wake latency. | `-t fma` with `LLMDBENCH_VLLM_COMMON_ENABLE_SLEEP_MODE=true`, `LLMDBENCH_FMA_DUAL_POD_SLEEPER_LIMIT>=1` |
+| **Cold Start (with launcher)** | No suitable launcher pod exists. The DPC creates a new launcher pod, then DPC commands the launcher to create a new vLLM instance. | Worst-case FMA launcher path, relevant for dynamic situations such as LauncherConfig rollouts or newly added nodes where launchers have not yet been populated. | `-t fma` with LauncherPopulationPolicy that does not cover the target node |
+| **Warm Start**            | A launcher pod already exists but no sleeping instance is available. DPC commands the launcher to create a new vLLM instance with module preloading. | Measures the launcher's contribution when no sleeping instance is available. | `-t fma` with default `LLMDBENCH_FMA_LAUNCHER_*` env vars |
+| **Hot Start**             | A sleeping vLLM instance exists in a suitable launcher pod. DPC sends `/wake_up`. | Best-case FMA path. Measures sleep-to-wake latency. | `-t fma` with `LLMDBENCH_VLLM_COMMON_ENABLE_SLEEP_MODE=true`, `LLMDBENCH_FMA_DUAL_POD_SLEEPER_LIMIT>=1` |
 
 > **Note on simulation:** Any of the above paths can be exercised with mock GPUs
 > (`llm-d-inference-sim` image or launcher `--mock-mode`) for CI pipelines and scenario
@@ -103,15 +104,14 @@ using the team's established terminology:
 Cell annotations indicate which measurement layers apply:
 - **L1** -- Layer 1 actuation metrics (T_actuation, T_wake, Hit_rate, T_cold_launcher, T_launcher)
 - **L1+L2** -- Actuation metrics plus inference readiness (T_first_token, T_e2e)
-- **L1+L3** -- Actuation metrics plus steady-state performance (TPOT, throughput, queue depth, KV cache, replica stability)
-- **L1+L2+L3** -- All three layers
+- **L1+L2+L3** -- Actuation metrics plus inference readiness plus steady-state performance (TPOT, throughput, queue depth, KV cache, replica stability)
 - **--** -- Not applicable to this combination
 
 | Scenario                           | Cold Start (no FMA) | Cold Start (FMA M2) | Cold Start (with launcher) | Warm Start | Hot Start |
 | ---------------------------------- | :-----------------: | :-----------------: | :------------------------: | :--------: | :-------: |
 | **Fast Replica Scale Up**          | L1+L2               | L1+L2 (planned)     | L1+L2                      | L1+L2      | L1+L2     |
 | **Introducing New Variant**        | L1+L2               | L1+L2 (planned)     | L1+L2                      | L1+L2      | --        |
-| **Resource Scaling and Stress Test** | L1+L3             | L1+L3 (planned)     | L1+L3                      | L1+L3      | L1+L3     |
+| **Resource Scaling and Stress Test** | L1+L2+L3          | L1+L2+L3 (planned)  | L1+L2+L3                   | L1+L2+L3   | L1+L2+L3  |
 | **Maintenance Planning**           | L1+L2+L3            | L1+L2+L3 (planned)  | L1+L2+L3                   | L1+L2+L3   | L1+L2+L3  |
 
 
@@ -179,9 +179,9 @@ Once actuation paths are classified, isolate the path-specific timing components
 
 - **T_wake** (hot): measure the `/wake_up` round-trip, approximated by the requester pod's
   transition from creation to Ready on known-hot actuations.
-- **T_launcher** (warm): time between DPC binding and vLLM instance readiness, approximated
-  by (requester dual-label timestamp - launcher pod Ready timestamp). Includes some DPC
-  reconciliation overhead.
+- **T_launcher** (warm and cold start with launcher): time between DPC binding and vLLM
+  instance readiness, approximated by (requester dual-label timestamp - launcher pod Ready
+  timestamp). Includes some DPC reconciliation overhead.
 - **T_cold_launcher** (cold start with launcher): end-to-end from launcher pod creation
   timestamp to vLLM instance healthy.
 
--- a/inference_server/benchmark/benchmark.md
+++ b/inference_server/benchmark/benchmark.md
@@ -7,8 +7,8 @@ The goal is to quantify and compare how quickly a model-serving duo (server-requ
 and server-providing pods) becomes available under four different actuation conditions
 in order of decreasing latency:
 
-- **Cold start**: creating a new vLLM instance without using a launcher
-- **Luke warm start**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
+- **Cold start without FMA**: creating a new vLLM instance without using a launcher
+- **Cold start (with launcher)**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
 - **Warm start**: creating a new vLLM instance in an existing launcher pod
 - **Hot start**: waking a sleeping vLLM instance on an existing launcher pod
 
@@ -31,19 +31,37 @@ direct scope but is referenced for completeness and handoff to other frameworks.
 
 | Layer | Focus | Metrics | Measured By |
 | ----- | ----- | ------- | ----------- |
-| **L1: Actuation** | Requester pod readiness | T_actuation (requester creation to readiness), T_wake (DPC wakes sleeping vLLM instance), Hit_rate (GPU hits), T_luke_warm (DPC creates launcher pod then vLLM instance), T_launcher (launcher creates new vLLM instance) | llm-d-benchmark new harness |
+| **L1: Actuation** | Requester pod readiness | T_actuation (requester creation to readiness), T_wake (DPC wakes sleeping vLLM instance), Hit_rate (GPU hits), T_cold_launcher (DPC creates launcher pod then vLLM instance), T_launcher (launcher creates new vLLM instance) | llm-d-benchmark new harness |
 | **L2: Inference Readiness** | First inference response | T_e2e (requester creation to first inference response), T_first_token (requester ready to first inference response) | llm-d-benchmark nop/inference-perf harness |
 | **L3: Steady-State** | Throughput/latency | TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
 
 **Metric definitions:**
 
-- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or luke warm start (T_luke_warm).
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or cold start with launcher (T_cold_launcher).
 - **T_wake**: Request-response time for the DPC's `/wake_up` call to a sleeping vLLM instance on the server-providing pod. A part of T_actuation when a hot start occurs.
 - **Hit_rate**: Fraction of server-requesting Pods that get satisfied by waking a sleeping vLLM instance.
-- **T_luke_warm**: Time from the DPC requesting launcher pod creation to the new vLLM instance reporting healthy. Covers the full luke warm start span: launcher pod scheduling, launcher readiness, DPC reconciliation, and vLLM instance creation. Measured end-to-end because the boundary between launcher readiness and instance creation is not directly observable from outside the DPC.
-- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to the warm start path, where a launcher pod already exists.
+- **T_cold_launcher**: Time from the DPC launcher pod creation to the new vLLM instance reporting healthy. Covers the full cold start (with launcher) span: launcher pod scheduling, launcher startup, and vLLM instance creation.
+- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to both cold start (with launcher) and warm start paths.
 - **T_e2e**: Total time from requester pod creation to first successful inference response (T_actuation + T_first_token). Spans the full actuation and inference readiness path.
 - **T_first_token**: Time from requester pod readiness to receiving the first streamed token from the server-providing pod's vLLM instance (time-to-first-token, post-actuation). Requires streaming inference requests.
+
+**Constituent duration metrics (planned):**
+
+The following metrics break T_cold_launcher and T_launcher into their constituent durations.
+Collecting them requires correlating Kubernetes object timestamps, DPC log messages, and
+launcher API responses. The fidelity of each (e.g., delay in retrieving/parsing DPC logs)
+needs further evaluation before they are added to the benchmarking harness.
+
+| Metric | Definition | Observable Via |
+| ------ | ---------- | -------------- |
+| **T_launcher_schedule** | Launcher pod `creationTimestamp` to `PodScheduled` condition `lastTransitionTime` | Kube pod status |
+| **T_launcher_startup** | Launcher pod `PodScheduled` to `Ready` condition `lastTransitionTime` | Kube pod status |
+| **T_dpc_react** | Launcher pod `Ready` to DPC issuing `CreateNamedInstance` | DPC logs (V5: "Creating new vLLM instance") |
+| **T_instance_ready** | `CreateNamedInstance` call to vLLM instance serving (requester pod becomes Ready) | DPC logs + Kube pod status |
+
+Relationships:
+- T_cold_launcher ≈ T_launcher_schedule + T_launcher_startup + T_dpc_react + T_instance_ready
+- T_launcher ≈ T_dpc_react + T_instance_ready (launcher already Ready)
 
 ## Benchmarking Scenarios
 
@@ -64,15 +82,12 @@ using the team's established terminology:
 
 | Actuation Path            | What It Measures | Why Included | llm-d-benchmark Config |
 | ------------------------- | ---------------- | ------------ | ---------------------- |
-| **Cold Start**            | No launcher, no sleeping pods. Raw Kubernetes deploy-to-ready latency (non-FMA baseline, or FMA milestone 2 without launcher). | Establishes the baseline that all FMA paths should improve upon. | `-t standalone` comparison baseline |
-| **Luke Warm Start**       | No launcher pod on the assigned GPU. The DPC creates a new launcher pod, then the launcher creates a new vLLM instance. | Worst-case FMA path, relevant for dynamic situations such as LauncherConfig rollouts or newly added nodes where the LPC has not yet populated launchers. | `-t fma` with LauncherPopulationPolicy that does not cover the target GPU node |
+| **Cold Start (no FMA)**   | No FMA involvement. Raw Kubernetes deploy-to-ready latency. | Non-FMA baseline that all FMA paths should improve upon. | `-t standalone` comparison baseline |
+| **Cold Start (FMA M2)**   | DPC creates a standalone server-providing pod directly (no launcher). The requesting pod uses the `server-patch` annotation. | Measures the DPC's contribution without the launcher. Planned: pending a stable M3 harness. | `-t fma` with `fma.mode: m2` (planned) |
+| **Cold Start (with launcher)** | No suitable launcher pod. The DPC creates a new launcher pod, then DPC commands the launcher to create a new vLLM instance. | Worst-case FMA launcher path, relevant for dynamic situations such as LauncherConfig rollouts or newly added nodes where launchers have not yet been populated. | `-t fma` with LauncherPopulationPolicy that does not cover the target node |
 | **Warm Start**            | A launcher pod exists (pre-created by the LPC) but no sleeping instance is available on the assigned GPU. Launcher creates a new vLLM instance with module preloading. | Measures the launcher's contribution when no sleeping instance is available. | `-t fma` with default `LLMDBENCH_FMA_LAUNCHER_*` env vars |
 | **Hot Start**             | A sleeping vLLM instance exists on the correct GPU. DPC sends `/wake_up`. | Best-case FMA path. Measures sleep-to-wake latency. | `-t fma` with `LLMDBENCH_VLLM_COMMON_ENABLE_SLEEP_MODE=true`, `LLMDBENCH_FMA_DUAL_POD_SLEEPER_LIMIT>=1` |
 
-> **Note on naming:** "Cold FMA Start" was considered as an alternative to "Luke Warm
-> Start", but the latter, though informal, was preferred for consistency with the
-> existing hot/warm/cold temperature metaphor.
->
 > **Note on simulation:** Any of the above paths can be exercised with mock GPUs
 > (`llm-d-inference-sim` image or launcher `--mock-mode`) for CI pipelines and scenario
 > prototyping. Simulation is an orthogonal testing mode, not a separate actuation path.
@@ -86,18 +101,18 @@ using the team's established terminology:
 ### Matrix
 
 Cell annotations indicate which measurement layers apply:
-- **L1** -- Layer 1 actuation metrics (T_actuation, T_wake, Hit_rate, T_luke_warm, T_launcher)
+- **L1** -- Layer 1 actuation metrics (T_actuation, T_wake, Hit_rate, T_cold_launcher, T_launcher)
 - **L1+L2** -- Actuation metrics plus inference readiness (T_first_token, T_e2e)
 - **L1+L3** -- Actuation metrics plus steady-state performance (TPOT, throughput, queue depth, KV cache, replica stability)
 - **L1+L2+L3** -- All three layers
 - **--** -- Not applicable to this combination
 
-| Scenario                           | Cold Start | Luke Warm Start | Warm Start | Hot Start |
-| ---------------------------------- | :--------: | :-------------: | :--------: | :-------: |
-| **Fast Replica Scale Up**          | L1+L2      | L1+L2           | L1+L2      | L1+L2     |
-| **Introducing New Variant**        | L1+L2      | L1+L2           | L1+L2      | --        |
-| **Resource Scaling and Stress Test** | L1+L3    | L1+L3           | L1+L3      | L1+L3     |
-| **Maintenance Planning**           | L1+L2+L3   | L1+L2+L3        | L1+L2+L3   | L1+L2+L3  |
+| Scenario                           | Cold Start (no FMA) | Cold Start (FMA M2) | Cold Start (with launcher) | Warm Start | Hot Start |
+| ---------------------------------- | :-----------------: | :-----------------: | :------------------------: | :--------: | :-------: |
+| **Fast Replica Scale Up**          | L1+L2               | L1+L2 (planned)     | L1+L2                      | L1+L2      | L1+L2     |
+| **Introducing New Variant**        | L1+L2               | L1+L2 (planned)     | L1+L2                      | L1+L2      | --        |
+| **Resource Scaling and Stress Test** | L1+L3             | L1+L3 (planned)     | L1+L3                      | L1+L3      | L1+L3     |
+| **Maintenance Planning**           | L1+L2+L3            | L1+L2+L3 (planned)  | L1+L2+L3                   | L1+L2+L3   | L1+L2+L3  |
 
 
 ### Scenario Rationale
@@ -151,7 +166,7 @@ but does not classify which actuation path the DPC took. This phase adds classif
 logic to `fma_functions.py`:
 
 - Compare the launcher pod's creation timestamp against the requester pod's creation
-  timestamp. If the launcher was created *after* the requester, it is a luke warm start.
+  timestamp. If the launcher was created *after* the requester, it is a cold start (with launcher).
 - For remaining cases, check whether the vLLM instance was woken from sleep (hot start)
   or newly created (warm start). The launcher's `/v2/vllm/instances` API returns instance
   status, and sleep/wake metrics are already parsed from launcher logs by `nop_functions.py`.
@@ -167,9 +182,8 @@ Once actuation paths are classified, isolate the path-specific timing components
 - **T_launcher** (warm): time between DPC binding and vLLM instance readiness, approximated
   by (requester dual-label timestamp - launcher pod Ready timestamp). Includes some DPC
   reconciliation overhead.
-- **T_luke_warm** (luke warm): end-to-end from launcher pod creation timestamp to vLLM
-  instance healthy, measured as a single span since the internal DPC boundary is not
-  directly observable from outside.
+- **T_cold_launcher** (cold start with launcher): end-to-end from launcher pod creation
+  timestamp to vLLM instance healthy.
 
 **Phase 3: Multi-replica and scenario coverage**
 
--- a/inference_server/benchmark/benchmark.md
+++ b/inference_server/benchmark/benchmark.md
@@ -4,8 +4,7 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
 
 ## Purpose
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under four different actuation conditions
-in order of decreasing latency:
+and server-providing pods) becomes available under four different actuation conditions:
 
 - **Cold start (no FMA)**: creating a new vLLM instance without using a launcher
 - **Cold start (with launcher)**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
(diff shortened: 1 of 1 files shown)


inference_server/benchmark/benchmark.md
```diff
@@ -4,11 +4,12 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
 ## Purpose
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under four different actuation conditions
+and server-providing pods) becomes available under different actuation conditions
 in order of decreasing latency:
 
-- **Cold start**: creating a new vLLM instance without using a launcher
-- **Luke warm start**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
+- **Cold start without FMA**: creating a new vLLM instance without using a launcher
+- **Cold start (FMA M2)**: DPC creates a standalone server-providing pod directly (planned)
```
aavarghese: What does (planned) mean here?

aavarghese: Why is it that we even need to mention M2 in these docs? I doubt we will ever deploy FMA M2...

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +32,38 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
 | **L3: Steady-State** | Throughput/latency | TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
 
 **Metric definitions:**
 
-- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or luke warm start (T_luke_warm).
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. For FMA paths, spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or cold start with launcher (T_cold_launcher). For non-FMA and M2 cold starts, T_actuation is measured directly with no FMA-specific sub-components.
```
aavarghese: Adding more confusion to new reader...we could remove `and M2 cold starts`

inference_server/benchmark/benchmark.md
```diff
@@ -4,11 +4,12 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
 
 ## Purpose
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under four different actuation conditions
+and server-providing pods) becomes available under different actuation conditions
 in order of decreasing latency:
```
MikeSpreitzer: I think that the order in which the cases actually appear is fine, but I suspect that it is not equal to "decreasing latency". The ordering is: first, code complexity, and second, runtime path length (which I expect will correlate with latency).

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +31,38 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
 - **T_wake**: Request-response time for the DPC's `/wake_up` call to a sleeping vLLM instance on the server-providing pod. A part of T_actuation when a hot start occurs.
 - **Hit_rate**: Fraction of server-requesting Pods that get satisfied by waking a sleeping vLLM instance.
-- **T_luke_warm**: Time from the DPC requesting launcher pod creation to the new vLLM instance reporting healthy. Covers the full luke warm start span: launcher pod scheduling, launcher readiness, DPC reconciliation, and vLLM instance creation. Measured end-to-end because the boundary between launcher readiness and instance creation is not directly observable from outside the DPC.
-- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to the warm start path, where a launcher pod already exists.
+- **T_cold_launcher**: Time from the DPC launcher pod creation to the new vLLM instance reporting healthy. Covers the full cold start (with launcher) span: launcher pod scheduling, launcher startup, and vLLM instance creation.
```
MikeSpreitzer: Is this trying to say that T_cold_launch is T_actuation but restricted to cold start with launcher scenarios? In other words, T_actuation is (a) the time from (1) creation of the server-requesting Pod to (2) requester Pod readiness but (b) is only measured for the code start with launcher cases?

If so, then the text currently here is misleading: it suggests a bit less of a span to me. It is also confusing because it says "full ... span".

If not, then this is a different kind of refinement than T_wake: this one covers less of the full path but T_wake is the full path but restricted by actuation case.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +31,38 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+The following metrics break T_cold_launcher and T_launcher into their constituent durations.
+Collecting them requires correlating Kubernetes object timestamps, DPC log messages, and
+launcher API responses. The fidelity of each (e.g., delay in retrieving/parsing DPC logs)
+needs further evaluation before they are added to the benchmarking harness.
+
+| Metric | Definition | Observable Via |
+| ------ | ---------- | -------------- |
+| **T_launcher_schedule** | Launcher pod `creationTimestamp` to `PodScheduled` condition `lastTransitionTime` | Kube pod status |
+| **T_launcher_startup** | Launcher pod `PodScheduled` to `Ready` condition `lastTransitionTime` | Kube pod status |
+| **T_dpc_react** | Launcher pod `Ready` to DPC issuing `CreateNamedInstance` | DPC logs (V5: "Creating new vLLM instance") |
```
MikeSpreitzer: What is "V5"?

MikeSpreitzer: To remove the difficulties of parsing the controller log, the controller could produce a Prometheus histogram of this duration. Depending on the level of correlation with other measurements intended, the fact that this is inherently an aggregate may or may not be a problem. If it is a problem then we could consider supporting distributed tracing.

Same for T_instance_ready

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +31,38 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+| **T_instance_ready** | `CreateNamedInstance` call to vLLM instance serving (requester pod becomes Ready) | DPC logs + Kube pod status |
+
+Relationships:
+- T_cold_launcher ≈ T_launcher_schedule + T_launcher_startup + T_dpc_react + T_instance_ready
+- T_launcher ≈ T_dpc_react + T_instance_ready (launcher already Ready)
```
MikeSpreitzer: In the warm start actuation path, the right starting gun is not "Launcher pod `Ready`". As noted, the launcher Pod is ready before the server-requesting Pod is created.

MikeSpreitzer (commented):

inference_server/benchmark/benchmark.md
rubambiza: It's been done in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454/commits/c46a9d3e9756d2ab0e4003e0713e74ce02a3688d

inference_server/benchmark/benchmark.md
rubambiza: Note that, based on our Slack convo, I [removed](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454/commits/c46a9d3e9756d2ab0e4003e0713e74ce02a3688d) all mentions of M2.

inference_server/benchmark/benchmark.md
rubambiza: The logging level for the exact log entry that would be used for the observation.

rubambiza: I am not familiar with Prometheus functionality. Just clarifying, by histogram of this duration, do you mean the following:

- Data points = the latencies for the duration across all scale-up operations
- Bins = frequencies for each observed (rounded to some precision) latency

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +31,38 @@ direct scope but is referenced for completeness and handoff to other frameworks.
 
 | Layer | Focus | Metrics | Measured By |
 | ----- | ----- | ------- | ----------- |
-| **L1: Actuation** | Requester pod readiness | T_actuation (requester creation to readiness), T_wake (DPC wakes sleeping vLLM instance), Hit_rate (GPU hits), T_luke_warm (DPC creates launcher pod then vLLM instance), T_launcher (launcher creates new vLLM instance) | llm-d-benchmark new harness |
+| **L1: Actuation** | Requester pod readiness | T_actuation (requester creation to readiness), T_wake (DPC wakes sleeping vLLM instance), Hit_rate (GPU hits), T_cold_launcher (DPC creates launcher pod then vLLM instance), T_launcher (launcher creates new vLLM instance) | llm-d-benchmark new harness |
 | **L2: Inference Readiness** | First inference response | T_e2e (requester creation to first inference response), T_first_token (requester ready to first inference response) | llm-d-benchmark nop/inference-perf harness |
```
aavarghese, line 34: Since `T_cold_launcher` is a condition now, `T_launcher` itself is vague. Do we intend it to mean the subset of time taken to create the new vllm instance in an existing OR new launcher? Or is the time taken to create a new vllm instance in existing launcher only aka `T_warm_launcher`?

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +31,38 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
 | **L3: Steady-State** | Throughput/latency | TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
 
 **Metric definitions:**
 
-- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or luke warm start (T_luke_warm).
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. For FMA paths, spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or cold start with launcher (T_cold_launcher). For the non-FMA cold start, T_actuation is measured directly with no FMA-specific sub-components.
```
aavarghese: Read above....depending on answer above, `T_launcher`--->`T_warm_launcher`?

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +31,38 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
 - **Hit_rate**: Fraction of server-requesting Pods that get satisfied by waking a sleeping vLLM instance.
-- **T_luke_warm**: Time from the DPC requesting launcher pod creation to the new vLLM instance reporting healthy. Covers the full luke warm start span: launcher pod scheduling, launcher readiness, DPC reconciliation, and vLLM instance creation. Measured end-to-end because the boundary between launcher readiness and instance creation is not directly observable from outside the DPC.
-- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to the warm start path, where a launcher pod already exists.
+- **T_cold_launcher**: Time from the DPC launcher pod creation to the new vLLM instance reporting healthy. Covers the full cold start (with launcher) span: launcher pod scheduling, launcher startup, and vLLM instance creation.
+- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to both cold start (with launcher) and warm start paths.
```
aavarghese: This answers my question above I think but now I'm thinking `T_launcher` is not a good name for this sub-component. It has nothing to do with the launcher. It only measure time to "Create" new vllm and for it to get ready. So what should we call that? `T_create_vllm`? `T_launch_newvllm`?

inference_server/benchmark/benchmark.md
rubambiza: Dropped the latency ordering to focus on the conditions.

inference_server/benchmark/benchmark.md
rubambiza: No, `T_cold_launcher` is a constituent piece of `T_actuation` for the cold-start-with-launcher path. I have modified the text to clarify that, like `T_wake`, it does not cover the full path.

inference_server/benchmark/benchmark.md
rubambiza: I think I have a sense of the histograms now. I have added a note with your observability suggestions to the constituent duration table.

inference_server/benchmark/benchmark.md
rubambiza: You're right. In the warm start actuation path, the `T_dpc_react` is negligible because, unlike `T_cold_launcher`, we are not waiting for the launcher pod to become ready. This means `T_launcher` is essentially equal to `T_instance_ready`. I changed the relationship accordingly.

inference_server/benchmark/benchmark.md
rubambiza: That's exactly why I kept it vague. The DPC does not care how the launcher came to be ready. Until we see the early numbers, I would like to keep things simple.

inference_server/benchmark/benchmark.md
rubambiza: See justification above, but also [below](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454#discussion_r3147765790).

inference_server/benchmark/benchmark.md
rubambiza: You got a point that `T_launcher` is misleading as a name for the sub-component. I considered your suggestions, but they miss the mark on the fact that the launcher is still somewhat involved, especially across the two paths. Thus, I came up with a compromise: **`T_instance_create`**. It focuses on what's being measured (instance creation through to readiness) without implying the launcher itself is the subject.

In the cold path, the launcher has to be created. In the warm path, the launcher is ready and the metric practically collapses to `T_instance_ready` as argued [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454#discussion_r3139637608).

I have modified the text accordingly.

inference_server/benchmark/benchmark.md
```diff
@@ -4,11 +4,10 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
 ## Purpose
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under four different actuation conditions
-in order of decreasing latency:
+and server-providing pods) becomes available under four different actuation conditions:
 
-- **Cold start**: creating a new vLLM instance without using a launcher
-- **Luke warm start**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
+- **Cold start (no FMA)**: creating a new vLLM instance without using a launcher
+- **Cold start (with launcher)**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
```
MikeSpreitzer: I still think that it is important to be clear here about the fact that the launcher does not create the vllm instance on the launcher's own initiative but rather in response to a command from the DPC. This document still reads like there is just one DPC reaction time involved, and that is just not true. For example, one bit of path not discussed is from vLLM becoming ready to the server-requesting Pod being ready. That one goes like this: the DPC occasionally asks vLLM whether it is ready; after receiving a "yes", the DPC commands the requester container in the server-requesting Pod to start responding "yes" when queried for readiness; the kubelet occasionally queries that, and upon receiving a "yes" requests a kube apiserver to set that container's readiness; some time after that, something tells the kube apiserver to mark the whole server-requesting Pod as ready.

I would replace

> then the launcher creates a new vLLM instance

with

> then the DPC commands the launcher to create a new vLLM instance

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,44 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
 
 **Metric definitions:**
 
-- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or luke warm start (T_luke_warm).
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. For FMA paths, spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_instance_create), or cold start with launcher (T_cold_launcher). For the non-FMA cold start, T_actuation is measured directly with no FMA-specific sub-components.
 - **T_wake**: Request-response time for the DPC's `/wake_up` call to a sleeping vLLM instance on the server-providing pod. A part of T_actuation when a hot start occurs.
 - **Hit_rate**: Fraction of server-requesting Pods that get satisfied by waking a sleeping vLLM instance.
```
MikeSpreitzer: Don't we want to measure two hit rates, one for hot start and one for warm start?

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,44 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
-- **T_luke_warm**: Time from the DPC requesting launcher pod creation to the new vLLM instance reporting healthy. Covers the full luke warm start span: launcher pod scheduling, launcher readiness, DPC reconciliation, and vLLM instance creation. Measured end-to-end because the boundary between launcher readiness and instance creation is not directly observable from outside the DPC.
-- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to the warm start path, where a launcher pod already exists.
+- **T_cold_launcher**: Time from the DPC launcher pod creation to the new vLLM instance reporting healthy. A constituent of T_actuation for cold start (with launcher) cases, covering launcher pod scheduling, launcher startup, and vLLM instance creation. Does not include the earlier portion of T_actuation (requester scheduling and DPC reconciliation before the launcher pod is created).
```
MikeSpreitzer: The "DPC" in the definition seems out of place.

> Time from the DPC launcher pod creation to the new vLLM instance reporting healthy.

I think that the definition would be clearer without the "DPC" in it.

> Time from the launcher pod creation to the new vLLM instance reporting healthy.

I assume that the starting timestamp would be the creationTimestamp of the Pod. Perhaps instead you meant a little earlier:

> Time from (a) when the DPC sends the request to create the launcher Pod to (b) the new vLLM instance reporting healthy.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,44 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
 | **L3: Steady-State** | Throughput/latency | TPOT (time per output token), throughput, queue depth, KV cache usage, replica stability | llm-d-benchmark / WVA |
 
 **Metric definitions:**
 
-- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. Spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_launcher), or luke warm start (T_luke_warm).
+- **T_actuation**: Time from requester pod creation (ReplicaSet scale-up) to requester pod readiness (`/ready` probe passes), which implies the DPC has bound the requester to a server-providing pod and the vLLM instance is serving. For FMA paths, spans different sub-components depending on the actuation path: hot start (T_wake), warm start (T_instance_create), or cold start with launcher (T_cold_launcher). For the non-FMA cold start, T_actuation is measured directly with no FMA-specific sub-components.
```
MikeSpreitzer: The end instant of this interval is unclear to me, because it is stated in the passive voice: we are not told _what_ is requesting `/ready`.

The definition of T_cold_launcher has the same problem. T_instance_create too.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,44 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
-- **T_launcher**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to the warm start path, where a launcher pod already exists.
+- **T_cold_launcher**: Time from the DPC launcher pod creation to the new vLLM instance reporting healthy. A constituent of T_actuation for cold start (with launcher) cases, covering launcher pod scheduling, launcher startup, and vLLM instance creation. Does not include the earlier portion of T_actuation (requester scheduling and DPC reconciliation before the launcher pod is created).
+- **T_instance_create**: Time from the launcher receiving a create request to the new vLLM instance reporting healthy. Includes the benefit of vLLM module preloading. Applies to both cold start (with launcher) and warm start paths.
```
MikeSpreitzer: As far as I know, the stated starting instant of this interval --- the launcher's _receipt of the request_ to create a new instance --- is not observable with the current code. I am not aware of logging that happens then, nor am I aware of this timestamp being stored anywhere. Following is an excerpt of verbose launcher logging, from https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/24859440518/job/72782102149:

```
INFO:     10.131.2.234:40080 - "GET /v2/vllm/instances HTTP/1.1" 200 OK
[2026-04-23 14:41:45] INFO launcher.py:97: Translated GPU UUIDs ['GPU-edd65826-8501-ebf1-d4e8-34140b865ac5'] to indices ['7'].
[2026-04-23 14:41:45] INFO launcher.py:104: Set CUDA_VISIBLE_DEVICES to                     7 based on UUIDs.
INFO:     10.131.2.234:40080 - "PUT /v2/vllm/instances/IxRoyPYOYX0UXpa458qa30ynJIgctRLjnFopEnerbr8gi HTTP/1.1" 201 Created
```

Note that the timestamps here have only whole-second precision.
One might reasonably infer that the "Translated GPU UUIDs" method happens soon after receipt of the request.

It would be easy to add logging of the receipt of the request, with whatever precision is desired.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,44 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
 - **T_e2e**: Total time from requester pod creation to first successful inference response (T_actuation + T_first_token). Spans the full actuation and inference readiness path.
 - **T_first_token**: Time from requester pod readiness to receiving the first streamed token from the server-providing pod's vLLM instance (time-to-first-token, post-actuation). Requires streaming inference requests.
 
+**Constituent duration metrics (planned):**
+
+The following metrics break T_cold_launcher and T_instance_create into their constituent durations.
+Collecting them requires correlating Kubernetes object timestamps, DPC log messages, and
+launcher API responses. The fidelity of each (e.g., delay in retrieving/parsing DPC logs)
```
MikeSpreitzer: What records launcher API responses?

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,44 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+needs further evaluation before they are added to the benchmarking harness.
+
+| Metric | Definition | Observable Via |
+| ------ | ---------- | -------------- |
+| **T_launcher_schedule** | Launcher pod `creationTimestamp` to `PodScheduled` condition `lastTransitionTime` | Kube pod status |
+| **T_launcher_startup** | Launcher pod `PodScheduled` to `Ready` condition `lastTransitionTime` | Kube pod status |
+| **T_dpc_react** | Launcher pod `Ready` to DPC issuing `CreateNamedInstance`. Applies to cold start (with launcher) only; in warm start, the launcher is already Ready before the requester exists, so this duration is negligible. | DPC logs (V5: "Creating new vLLM instance") |
```
MikeSpreitzer: Note that the interval here involves more than just one DPC reaction. The DPC also updates the launcher Pod with the labels and annotations specified by the InferenceServerConfig. This is done after the launcher Pod's creation, not in the creation request. At least, in the current code.

MikeSpreitzer: In warm start, "negligible" is not what this duration is. The word "negligible" usually means "so small that it is not important". What is going on here is simply that we do not care about this duration in the warm start case; this duration plays no role in latency in the warm start case.

MikeSpreitzer: The end point is not currently observable in the log; there is no "Creating new vLLM instance" log message. What we _do_ have is https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/23dd879b9b7db69ce286acd2f38f86e5be64e591/pkg/controller/dual-pods/inference-server.go#L472 , which happens after the launcher finishes the creation and the DPC gets the reply. We can add needed log statements to the DPC.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,44 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+| **T_instance_ready** | `CreateNamedInstance` call to vLLM instance serving (requester pod becomes Ready) | DPC logs + Kube pod status |
+
+Relationships:
+- T_cold_launcher ≈ T_launcher_schedule + T_launcher_startup + T_dpc_react + T_instance_ready
+- T_instance_create ≈ T_instance_ready (warm start; launcher already Ready, DPC react time is negligible)
```
MikeSpreitzer: If T_instance_create has only one constituent, why break it out?

MikeSpreitzer (commented):

inference_server/benchmark/benchmark.md
```diff
@@ -4,11 +4,10 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
 The goal is to quantify and compare how quickly a model-serving duo (server-requesting
-and server-providing pods) becomes available under four different actuation conditions
-in order of decreasing latency:
+and server-providing pods) becomes available under four different actuation conditions:
 
-- **Cold start**: creating a new vLLM instance without using a launcher
-- **Luke warm start**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
+- **Cold start (no FMA)**: creating a new vLLM instance without using a launcher
+- **Cold start (with launcher)**: DPC creates a new launcher pod, then the launcher creates a new vLLM instance
 - **Warm start**: creating a new vLLM instance in an existing launcher pod
```
MikeSpreitzer, line 11: There are two variants of warm start: one just creates a vllm instance, the other first deletes a sleeping instance and then creates the new instance.
The implementation does not yet do the second, but it is coming (this is Issue #246).

inference_server/benchmark/benchmark.md
aavarghese: Resolving because the new names and breakdown are less vague - thanks!!

inference_server/benchmark/benchmark.md
rubambiza: Replacement is done! I appreciate the extra clarification, but I am trying to be mindful of this document's scope growing too much. In particular, I don't want to recreate the [latency outline](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/38). Perhaps this doc can reference a companion doc that meets [this need](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/38#issuecomment-4217304230). I'd love to review a PR from you with doc-wide changes after this one is merged **or** modify this one myself once you have the latency outline completed. What do you think?

inference_server/benchmark/benchmark.md
rubambiza: I don't see why not. I created https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/496 to track this.

inference_server/benchmark/benchmark.md
rubambiza: It is latter. Fixed!

inference_server/benchmark/benchmark.md
rubambiza: I tried to clarify the components doing the observing and being probed for all three metrics. `T_actuation` now specifies the kubelet's readiness probe on the requester pod; `T_cold_launcher` and `T_instance_create` specify the DPC's readiness relay as their end instant.

inference_server/benchmark/benchmark.md
rubambiza: Agreed. I created https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/497  to track this.

inference_server/benchmark/benchmark.md
rubambiza: The launcher. You referenced an example of it [above](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454#discussion_r3151696079). I have changed from `launcher API responses` to `launcher logs`.

inference_server/benchmark/benchmark.md
rubambiza: Good point. I added a clarifying note to `T_dpc_react` that it ends at `CreateNamedInstance` and does not include later DPC actions (e.g., updating launcher pod labels/annotations), which are captured by `T_instance_ready`.

rubambiza: On the same page. I modified to be a bit more precise.

rubambiza: I have created https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/495 to track this.

inference_server/benchmark/benchmark.md
rubambiza: We break it out because collapsing it into `T_instance_ready` would make the metric definitions ambiguous. `T_instance_create` is a primary metric that applies to both warm start and cold start (with launcher) paths. `T_instance_ready` is a constituent of the `T_cold_launcher` decomposition. In the warm start case they converge, but keeping `T_instance_create` as the top-level metric gives warm start a named metric without forcing it into the `T_cold_launcher` decomposition table.

inference_server/benchmark/benchmark.md
rubambiza: Are you suggesting that we break them up into two distinct actuation paths or lump the cases but with an extended description (i.e., `creating a new vLLM instance *or* delete an unsuitable sleeping instance and then create a new vLLM instance in an existing launcher pod`?

MikeSpreitzer: I think that maybe we could go either way. If we call these both the same "actuation path" then, for purposes of understandable results, we would need a further level of distinction --- something like different _cases_ within one _actuation path_. I suspect this might be the way to go, addressing #408 is going to also split the cold-start-with-FMA actuation path into two cases. Actually both could theoretically involve more than two cases (e.g., delete 0, delete 1, and delete 2) --- but I expect that deleting more than 1 will be very rare in practice, so am fine with not benchmarking them.

inference_server/benchmark/benchmark.md
```diff
@@ -4,18 +4,17 @@ latency of model-serving pods within the LLM-D Fast Model Actuation workflow.
...
+- **Cold start (no FMA)**: creating a new vLLM instance without using a launcher
+- **Cold start (with launcher)**: DPC creates a new launcher pod, then the DPC commands the launcher to create a new vLLM instance
 - **Warm start**: creating a new vLLM instance in an existing launcher pod
 - **Hot start**: waking a sleeping vLLM instance on an existing launcher pod
 
 These metrics will guide future optimizations for the **Dual-Pods Controller (DPC)**. Ultimately, the goal
-is *high predictability*, which is defined as achieving close to 100% hit rate of awakening
-available, sleeping pods on cluster GPUs as a function of total inference server
-requests for common user scenarios.
+is *high predictability*, which is defined as achieving close to 100% hot-start hit rate
```
MikeSpreitzer, line 15: Not a problem introduced in this PR, but: I see no reason to think that 100% hot-start hit rate is achievable. That would require always getting very lucky. And clearly, different usage patterns will have different hit rates.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,45 @@ direct scope but is referenced for completeness and handoff to other frameworks.
 
 | Layer | Focus | Metrics | Measured By |
 | ----- | ----- | ------- | ----------- |
-| **L1: Actuation** | Requester pod readiness | T_actuation (requester creation to readiness), T_wake (DPC wakes sleeping vLLM instance), Hit_rate (GPU hits), T_luke_warm (DPC creates launcher pod then vLLM instance), T_launcher (launcher creates new vLLM instance) | llm-d-benchmark new harness |
+| **L1: Actuation** | Requester pod readiness | T_actuation (requester creation to readiness), T_wake (DPC wakes sleeping vLLM instance), Hot_hit_rate (hot-start hits), Warm_hit_rate (warm-start hits), T_cold_launcher (DPC creates launcher pod then vLLM instance), T_instance_create (DPC commands new vLLM instance creation) | llm-d-benchmark new harness |
```
MikeSpreitzer: This list starts out with things that are clearly quantitative: time intervals and hit rates. But it finishes with two things whose description reads like a scenario, not a particular quantitative thing.

> T_cold_launcher (DPC creates launcher pod then vLLM instance)

> T_instance_create (DPC commands new vLLM instance creation)

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,45 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+- **Warm_hit_rate**: Fraction of server-requesting Pods that get satisfied by an existing launcher pod (warm start), avoiding the cost of creating a new launcher pod.
+- **T_cold_launcher**: Time from (a) when the DPC sends the request to create the launcher Pod to (b) the DPC successfully relaying readiness to the requester pod (DPC V5 log: "Successfully relayed the readiness"). A constituent of T_actuation for cold start (with launcher) cases, covering launcher pod scheduling, launcher startup, and vLLM instance creation. Does not include the earlier portion of T_actuation (requester scheduling and DPC reconciliation before the launcher pod is created).
+- **T_instance_create**: Time from the launcher receiving a create request ([#497](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/497) tracks adding subsecond launcher logging for this instant) to the DPC successfully relaying readiness to the requester pod (DPC V5 log: "Successfully relayed the readiness"). Includes the benefit of vLLM module preloading. Applies to both cold start (with launcher) and warm start paths.
```
MikeSpreitzer, line 44: This is not the first interval to end at readiness, but it is the first one with a remark about how that is observed. At the earlier mention I assumed that the observation plan was the last-update timestamp on the readiness Condition in the Pod's status.

The observation plan should be stated at first mention of the moment, and/or in a list of observation plans.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,45 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+Collecting them requires correlating Kubernetes object timestamps, DPC log messages, and
+launcher logs. The fidelity of each (e.g., delay in retrieving/parsing DPC logs)
+needs further evaluation before they are added to the benchmarking harness.
+
+| Metric | Definition | Observable Via |
+| ------ | ---------- | -------------- |
+| **T_launcher_schedule** | Launcher pod `creationTimestamp` to `PodScheduled` condition `lastTransitionTime` | Kube pod status |
+| **T_launcher_startup** | Launcher pod `PodScheduled` to `Ready` condition `lastTransitionTime` | Kube pod status |
+| **T_dpc_react** | Launcher pod `Ready` to DPC issuing `CreateNamedInstance`. Applies to cold start (with launcher) only; in warm start, the launcher is already Ready before the requester exists, so this duration does not significantly contribute to latency. Does not include later DPC actions (e.g., updating launcher pod labels/annotations), which occur during instance creation and are captured by T_instance_ready. | DPC logs (not yet observable; [#495](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/495) tracks adding a pre-`CreateNamedInstance` log statement) |
```
MikeSpreitzer: Re

> in warm start, the launcher is already Ready before the requester exists, so this duration does not significantly contribute to latency

When "the launcher is already Ready before the requester exists", warm start involves no waiting for Ready at all. "does not significantly contribute to latency" is misleading. The interval from launcher Pod ready to DPC issuing CreateNamedInstance is well defined, but it is LONG and IRRELEVANT.

For warm start, you want something like the interval from (a) when the requester container starts running to (b) when the DPC sends the `CreateNamedInstance` request.

In all actuation paths there is latency from (a) requester Pod creation to (b) requester container start.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,45 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+| **T_instance_ready** | `CreateNamedInstance` call to DPC successfully relaying readiness to the requester pod (DPC V5 log: "Successfully relayed the readiness") | DPC logs + Kube pod status |
+
+Relationships:
+- T_cold_launcher ≈ T_launcher_schedule + T_launcher_startup + T_dpc_react + T_instance_ready
```
MikeSpreitzer, line 63: You want to be careful about each instant involved. For example, I see two different starting instants.

T_cold_launcher says it starts when "the DPC sends the request to create the launcher Pod".

T_launcher_schedule says it starts at "Launcher pod `creationTimestamp`".

inference_server/benchmark/benchmark.md
```diff
@@ -64,15 +88,11 @@ using the team's established terminology:
...
-
-> **Note on naming:** "Cold FMA Start" was considered as an alternative to "Luke Warm
-> Start", but the latter, though informal, was preferred for consistency with the
-> existing hot/warm/cold temperature metaphor.
->
+| **Cold Start (no FMA)**   | No FMA involvement. Raw Kubernetes deploy-to-ready latency. | Non-FMA baseline that all FMA paths should improve upon. | `-t standalone` comparison baseline |
+| **Cold Start (with launcher)** | No suitable launcher pod exists. The DPC creates a new launcher pod, then DPC commands the launcher to create a new vLLM instance. | Worst-case FMA launcher path, relevant for dynamic situations such as LauncherConfig rollouts or newly added nodes where launchers have not yet been populated. | `-t fma` with LauncherPopulationPolicy that does not cover the target node |
+| **Warm Start**            | A launcher pod already exists but no sleeping instance is available. DPC commands the launcher to create a new vLLM instance with module preloading. | Measures the launcher's contribution when no sleeping instance is available. | `-t fma` with default `LLMDBENCH_FMA_LAUNCHER_*` env vars |
```
MikeSpreitzer: "A launcher pod already exists _and is 'Ready'_ ".

While each launcher pod spends some time in the state of existing but not being ready, I expect this will be rare enough that we do not need to bother benchmarking scenarios where the DPC latches on to such a launcher and waits for it to become ready.

MikeSpreitzer (commented):
I have finished a round of review.

inference_server/benchmark/benchmark.md
MikeSpreitzer: > | **T_instance_ready** | `CreateNamedInstance` call to DPC successfully relaying readiness to the requester pod (DPC V5 log: "Successfully relayed the readiness") | DPC logs + Kube pod status |

That does not say "warm start but not cold start" to me. Also, it is not precise about the starting instant.

inference_server/benchmark/benchmark.md
rubambiza: I think it makes sense to keep the lumped but insert an appropriate signal in the results metadata about which case applied (i.e., as simple as an extra label in the result string or something). I'd like to leave the implementation details for later.

inference_server/benchmark/benchmark.md
rubambiza: Hence the `close` part. I think it would be rare, if not impossible, to hit 100% on active clusters like ours.

inference_server/benchmark/benchmark.md
rubambiza: Modified as follows:

>T_cold_launcher (launcher creation to vLLM readiness relay)

>T_instance_create (DPC instance creation request receipt to instance readiness relay)

I tried to convey the timed/quantitative part without repeating the metrics definitions.

inference_server/benchmark/benchmark.md
rubambiza: I don't quite follow. What do you consider to be the remark about how it is observed? I stated it as `(DPC V5 logs: ...)`, which is also shared with the earlier mention on `T_cold_launcher`. Perhaps you meant the observation for `T_actuation`? There, too, the readiness is quite descriptive (`the kubelet's readiness probe on the requester pod succeeds, marking the Pod's Ready condition True`), which aligns with your assumed observation plan about the last-update of the readiness condition.

inference_server/benchmark/benchmark.md
rubambiza: I agree on two points: (1) that "does not significantly contribute to latency" is misleading and (2) that the interval from launcher Pod ready to DPC issuing CreateNamedInstance is well defined, but it is LONG and IRRELEVANT for **warm start**. However, note that `T_dpc_react` applies to cold start (with launcher) where the launcher Pod ready _is_ relevant.

I also agree that in all actuation paths there is latency from (a) requester Pod creation to (b) requester container start, but I don't see its relevance for the actuation path benchmarking because it's a latency that I presume is primarily driven by the kube scheduler and resource availability, not necessarily FMA components.

Therefore, I have decided to simply remove the following part:

> in warm start, the launcher is already Ready before the requester exists, so this duration does not significantly contribute to latency.

to avoid any misleading statements.

inference_server/benchmark/benchmark.md
rubambiza: Good catch. I've aligned both `T_cold_launcher` and `T_launcher_schedule` to start from the launcher Pod's `creationTimestamp`. The alternative would be a pre-creation DPC log statement (tracked in #495), but since Kubernetes condition timestamps (lastTransitionTime) are at second precision, using creationTimestamp (also second precision) keeps the decomposition internally consistent. The DPC-to-API-server gap is nanoseconds, so no meaningful accuracy is lost.

inference_server/benchmark/benchmark.md
rubambiza: Fixed in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/454/commits/52a48ae69bc786981ef54e587d3fe6c41626b178. When there is no sleeping instance available, I had it implied it in my head that other instances had previously been created on the launcher - just not for the requested instance. In other words, it had already reached `Ready` state at some instant in the past.

inference_server/benchmark/benchmark.md
rubambiza: You're right on both counts. `T_instance_ready` applies to both cold start (with launcher) and warm start. It's a constituent of `T_cold_launcher` in the former, and approximately equals `T_instance_create` in the latter. I'll clarify this in the definition. For the starting instant, "CreateNamedInstance call" means when the DPC issues the HTTP request to the launcher. This isn't precisely observable yet (tracked in #495); the closest current proxy is the launcher's HTTP access log timestamp for the PUT request.

inference_server/benchmark/benchmark.md
MikeSpreitzer: OK with me.

inference_server/benchmark/benchmark.md
MikeSpreitzer: I do not think that "close to" does the job here. How good the hot-start hit rate is will depend on the overall usage pattern (over time) and how lucky is the Pod and GPU scheduling. These could easily be **far** from getting 100% lucky.

inference_server/benchmark/benchmark.md
MikeSpreitzer: Yes, the earliest mention of ending at requester readiness is on new line 39 with the text "the kubelet's readiness probe on the requester pod succeeds, marking the Pod's Ready condition True". That text has two different things in it, the second one is a process (a call from the kubelet to a kube apiserver) that covers a span of time, and both are different from what is said here ("DPC V5 log ...").

inference_server/benchmark/benchmark.md
MikeSpreitzer: > The DPC-to-API-server gap is nanoseconds

Do you have evidence of that? (Beware timestamps collected from different clocks. IIRC ntp typically gets you within about 1 ms.)

But regardless of whether it is milliseconds or nanoseconds, the point about information lost is that it is some unknown fraction of a second.

inference_server/benchmark/benchmark.md
```diff
@@ -31,20 +30,45 @@ direct scope but is referenced for completeness and handoff to other frameworks.
...
+| **T_instance_ready** | DPC issuing the `CreateNamedInstance` HTTP request to the launcher ([#495](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/495) tracks adding a pre-call log) to DPC successfully relaying readiness to the requester pod (DPC V5 log: "Successfully relayed the readiness"). Applies to both cold start (with launcher) and warm start paths. | DPC logs + Kube pod status |
+
+Relationships:
+- T_cold_launcher ≈ T_launcher_schedule + T_launcher_startup + T_dpc_react + T_instance_ready
+- T_instance_create ≈ T_instance_ready (warm start; launcher already Ready, so T_dpc_react does not apply)
+
+**Alternative observability approaches:** As an alternative to DPC log parsing for
+T_dpc_react and T_instance_ready, the DPC could emit Prometheus histograms for these
+durations, which would be more reliable than log parsing but only provide aggregate
```
MikeSpreitzer, line 68: Prometheus histograms are not strictly an **alternative**: they could be done _additionally_.

inference_server/benchmark/benchmark.md
```diff
@@ -164,12 +188,13 @@ Once actuation paths are classified, isolate the path-specific timing components
...
-  by (requester dual-label timestamp - launcher pod Ready timestamp). Includes some DPC
-  reconciliation overhead.
-- **T_luke_warm** (luke warm): end-to-end from launcher pod creation timestamp to vLLM
-  instance healthy, measured as a single span since the internal DPC boundary is not
-  directly observable from outside.
+- **T_instance_create** (warm and cold start with launcher): time from the launcher receiving
+  a create request to DPC relaying readiness. Initially approximated by (requester dual-label
+  timestamp - launcher pod Ready timestamp), which is an upper bound that includes DPC
+  reconciliation overhead. A tighter measurement requires DPC log parsing or Prometheus
+  histograms (see alternative observability approaches above).
```
MikeSpreitzer, line 195: See comment on "alternative" above.

inference_server/benchmark/benchmark.md
MikeSpreitzer: I will come back to this after we get this one merged.

MikeSpreitzer (approved):
This still needs some work; I would not complain if it were done in follow-on PRs. Substantial progress has been made in this PR.

inference_server/benchmark/benchmark.md
rubambiza: The [Huygens clock synchronization](https://www.usenix.org/system/files/conference/nsdi18/nsdi18-geng.pdf) work demonstrates the possibility to synchronize hosts to within tens of nanoseconds. However, I do not have such evidence for the DPC-to-API-server gap in particular.

PR #458, opened 2026-04-27
FMA-HPA integration in OCP cluster

diegocastanibm:
## Summary

Adds end-to-end demo scripts that deploy FMA integrated with HPA autoscaling on an OCP cluster with real GPUs. These scripts automate the full flow validated during the PoC on pokprod001 (H100 GPUs):

```
Client → Agentgateway → EPP → vLLM (on launcher Pod, GPU)
                         ↓
                    Prometheus → prometheus-adapter → External Metrics API → HPA → FMA
```

## What's included

Four scripts in `test/e2e/demo-fma-hpa/`:

| Script | Purpose |
|--------|---------|
| `demo-fma-hpa-ocp.sh` | **Main deployment** — idempotent, 8-step setup: namespace/RBAC, CRDs, EPP+Gateway, FMA controllers, FMA objects (ISC, LauncherConfig, LPP, ReplicaSet), HPA with External Metrics, prometheus-adapter rules, validation. Checks each component before deploying and skips if already present. |
| `demo-fma-hpa-loadgen.sh` | **Traffic generator** — runs an in-cluster pod with configurable concurrent workers (default: 300) sending `/v1/completions` requests through the Gateway for a configurable duration (default: 120s). |
| `demo-fma-hpa-monitor.sh` | **Live dashboard** — watch loop showing HPA status, requester/launcher pod states with DUAL/SLEEPING labels, and External Metrics values. |
| `demo-fma-hpa-cleanup.sh` | **Teardown** — removes FMA objects, HPA, controllers, and RBAC. Supports `FULL_CLEANUP=true` to also delete the namespace. |

## Validated on pokprod001 (OCP, H100)

- **Scale-up**: 300 concurrent workers → `running_requests ≈ 300` → HPA scaled from 1 to 4 requesters in ~15s → FMA created and bound 4 launchers in ~90s
- **Scale-down**: traffic stopped → HPA stabilization window (5 min) → HPA reduced to 1 replica → FMA put excess launchers to sleep

## Key design decisions from the PoC

- **Single ISC on port 8000** with InferencePool discovery labels (`llm-d.ai/inference-serving`, `llm-d.ai/guide`, `llm-d.ai/model`) — required for EPP backend discovery
- **No labels in LauncherConfig** — avoids label collision with ISC propagation (PR #437) that blocks binding
- **LPP restricted to a single labeled node** (`fma-hpa-poc=true`) — prevents populating launchers on all GPU nodes in a shared cluster
- **prometheus-adapter rules filter by `job` + `namespace`** — the `top_level_controller_name` label suggested in the llm-d guide does not exist in EPP v0.5.x metrics
- **`maxSleepingInstances: 0`** — InferencePool's fixed `targetPortNumber: 8000` limits us to 1 active vLLM instance per launcher

## Known issues discovered during the PoC

| Issue | Severity | Workaround |
|-------|----------|------------|
| ISC label collision with LauncherConfig prevents binding | Medium | Put discovery labels only in ISC, not in LC |
| Launcher populator creates extras during bind/unbind transitions | Low-Medium | Delete extras manually |
| `maxSleepingInstances=0` does not GC sleeping launchers after scale-down | Medium | Delete sleeping pods manually |
| prometheus-adapter guide references non-existent `top_level_controller_name` label | Medium | Use `job` + `namespace` labels instead |
| Agentgateway needs `anyuid` + `privileged` SCC on OCP | Low | Grant SCCs to gateway SA |


test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,570 @@
...
+#   - helm, helmfile, kubectl, jq, yq installed
+#   - Container images already pushed to registry (see CONTAINER_IMG_REG)
+#   - llm-d/guides repo cloned locally (for EPP/Gateway via helmfile)
+#
+# Required environment variables:
+#   LLM_D_GUIDES_DIR  - path to llm-d/guides repo (contains inference-scheduling helmfile)
+#
+# Optional environment variables (with defaults):
+#   NAMESPACE          - target namespace (default: fma-hpa)
+#   CONTAINER_IMG_REG  - image registry (default: quay.io/diego_castan)
```
diegocastanibm: I built and used my images to add the last changes in the repo for the PoC

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,570 @@
...
+    annotations:
+      description: "FMA ISC for HPA demo - ${MODEL}"
+  launcherConfigName: lc-hpa
+---
+apiVersion: fma.llm-d.ai/v1alpha1
+kind: LauncherConfig
+metadata:
+  name: lc-hpa
+spec:
+  maxSleepingInstances: 0
```
aavarghese: no sleepers?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,570 @@
...
+          - name: XDG_CACHE_HOME
+            value: "/tmp"
+          - name: XDG_CONFIG_HOME
+            value: "/tmp"
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef: { fieldPath: spec.nodeName }
+          - name: NAMESPACE
+            valueFrom:
+              fieldRef: { fieldPath: metadata.namespace }
```
aavarghese: Don't think you need NODE_NAME and NAMESPACE if you are using the real launcher on GPUs.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,570 @@
...
+    CONTAINER_IMG_REG="$CONTAINER_IMG_REG" \
+    IMAGE_TAG="$IMAGE_TAG" \
+    NODE_VIEW_CLUSTER_ROLE=create/please \
+    RUNTIME_CLASS_NAME=nvidia \
+    HELM_EXTRA_ARGS="--set launcherPopulator.enabled=true" \
+    "$SCRIPT_DIR/deploy_fma.sh"
+fi
+
+# =========================================================================
+# Step 5: FMA objects (ISC, LauncherConfig, LPP, ReplicaSet)
```
aavarghese, line 272: Can we re-use mkobjs-openshift.sh from FMA for Step 5?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,570 @@
...
+    echo "  Created RBAC roles and bindings"
+fi
+
+if kubectl get cm gpu-map -n "$NAMESPACE" &>/dev/null; then
+    echo "  ConfigMap gpu-map exists"
+else
+    kubectl create cm gpu-map -n "$NAMESPACE"
+    echo "  Created ConfigMap gpu-map"
+fi
+
```
aavarghese: We dont need gpu-map anymore for real GPUs...

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: That parameter does not work (or is not doing what, IMO, it supposed to do). What I think it has to do is, when  scaling down, keep a max number of instances sleeping that is determine by the parameter `maxSleepingInstances`. If we have more instances than `maxSleepingInstances`, they should be killed. However, all the instances are sleeping after scaling down. No effect what so ever. 
I reported this issue in the `Known issues discovered during the PoC` in the description.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: In this case, and just for the PoC, I needed it because, otherwise and by default, the launcher populator sets a launcher in all GPU nodes in a shared cluster. I described this behavior in the `Key design decisions from the PoC` in the description.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Maybe yes, but not sure.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Good point. The gpu-map ConfigMap is only needed by the dual-pods controller for non-launcher-based server-providing pods. Since this demo is fully launcher-based, that code path is never hit, so you're right that the mapping data itself is not used. Hwever, it is there because it matches what run-launcher-based.sh does (always creates it). Plus a creating an empty ConfigMap is harmless and defensive. I'm happy to remove it if you prefer.

aavarghese: yes, please remove it to avoid future confusion. `run-launcher-based.sh` has it because it is a Kind e2e test which does not use real gpus...

diegocastanibm: Sure, no problem

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
```diff
@@ -0,0 +1,89 @@
...
+
+cleanup() {
+    echo ""
+    echo "Cleaning up loadgen pod..."
+    kubectl delete pod "$POD_NAME" -n "$NAMESPACE" --ignore-not-found --wait=false 2>/dev/null
+}
+trap cleanup EXIT
+
+# Find the gateway service ClusterIP
+GW_SVC=$(kubectl get svc -n "$NAMESPACE" -o name 2>/dev/null | grep -m1 gateway | sed 's|service/||' || true)
```
rubambiza: **Nit**: For human readability, I would like to suggest switching the sed delimiter to something like `#` so that, unlike `|`,  it stands out from the piping happening around it.

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
```diff
@@ -0,0 +1,93 @@
...
+# Uninstall FMA controllers (Helm)
+echo "--- Uninstalling FMA controllers ---"
+helm uninstall fma -n "$NAMESPACE" 2>/dev/null || true
+
+# Delete RBAC
+echo "--- Deleting RBAC ---"
+kubectl delete rolebinding testreq testlauncher -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+kubectl delete role testreq testlauncher -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+kubectl delete sa testreq testlauncher -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+kubectl delete cm gpu-map -n "$NAMESPACE" --ignore-not-found 2>/dev/null
```
rubambiza: Am I recalling incorrectly that we moved away from GPU mapping in M3?

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
```diff
@@ -0,0 +1,93 @@
...
+
+# Delete ClusterRole created by deploy_fma.sh
```
rubambiza: `deploy_fma.sh` is referenced, but I don't see it in the set of committed files 👀

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
```diff
@@ -0,0 +1,93 @@
...
+kubectl delete clusterrole fma-node-view --ignore-not-found 2>/dev/null
```
rubambiza: Perhaps for consistency to the steps above the `clusterrolebinding` should be deleted first?

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
```diff
@@ -0,0 +1,89 @@
...
+#
+# Prerequisites:
+#   - demo-fma-hpa-ocp.sh has been run (Gateway + EPP + FMA deployed)
+#   - oc/kubectl authenticated
+#
+# Optional environment variables:
+#   NAMESPACE   - target namespace (default: fma-hpa)
+#   WORKERS     - number of concurrent workers (default: 300)
+#   DURATION    - load duration in seconds (default: 120)
+#   MODEL       - model name for completions (default: HuggingFaceTB/SmolLM2-360M-Instruct)
```
rubambiza, line 16: Is the model gated on `HuggingFace`. If so, then exporting the `HF_TOKEN` would also be a pre-requisite - ideally with instructions on how to create one (e.g., permissions).

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
```diff
@@ -0,0 +1,89 @@
...
+kubectl delete pod "$POD_NAME" -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+
+echo "Starting loadgen pod..."
+kubectl run "$POD_NAME" -n "$NAMESPACE" --rm -i --restart=Never \
+    --image=curlimages/curl:latest -- sh -c "
+GW=\"${GW_URL}\"
+MODEL=\"${MODEL}\"
+WORKERS=${WORKERS}
+DURATION=${DURATION}
+MAX_TOKENS=${MAX_TOKENS}
```
rubambiza, line 66: Is there a missing `"` to close the `-c` part? The lack of syntax highlighting for the subsequent function indicates that this may be the case.

test/e2e/demo-fma-hpa/demo-fma-hpa-monitor.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+# Live monitoring dashboard for the FMA + HPA demo.
+#
+# Refreshes every N seconds showing HPA status, pod states, and metrics.
```
rubambiza, line 5: **Nit**: I think every `INTERVAL` seconds (instead of every `N` seconds) might be more descriptive to match the arguments below. Wdyt?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
rubambiza: Is this referring to the `fma` images?  If so, it might be worth pinning to a particular release cut (alpha should be okay) for consistency and reproducibility. But, this can be addressed in a subsequent PR.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,560 @@
...
+if kubectl get crd inferencepools.inference.networking.x-k8s.io &>/dev/null; then
+    echo "  GAIE CRDs: present"
+else
+    echo "  Installing GAIE CRDs..."
+    kubectl apply -f https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v0.4.0/manifests.yaml
+fi
+
+if kubectl get crd inferenceserverconfigs.fma.llm-d.ai &>/dev/null; then
```
rubambiza: Is this implying `inferenceserverconfigs.fma.llm-d.ai` is a stand-in for all CRDs? In other words, there is no possibility of any other CRDs being already installed in the namespace until `inferenceserverconfigs.fma.llm-d.ai` is?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,560 @@
...
+        }'
+        kubectl rollout restart deployment -n "$NAMESPACE" \
+            -l app.kubernetes.io/name=inference-scheduler
+        echo "  Waiting for EPP rollout..."
+        kubectl rollout status deployment -n "$NAMESPACE" \
+            -l app.kubernetes.io/name=inference-scheduler --timeout=120s
+    fi
+fi
+
+# SCC for gateway service account
```
rubambiza: I've seen it mentioned in earlier files. What is `SCC`? Can you kindly spell it out on first use?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,560 @@
...
+GW_SA=$(kubectl get sa -n "$NAMESPACE" -o name 2>/dev/null | grep -m1 gateway | sed 's|serviceaccount/||' || true)
```
rubambiza: See earlier comment on the delimiter here.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,560 @@
...
+      labels:
+        app: fma-hpa-requester
+      annotations:
+        dual-pods.llm-d.ai/admin-port: "8081"
+        dual-pods.llm-d.ai/inference-server-config: "isc-smol"
+    spec:
+      runtimeClassName: nvidia
+      serviceAccountName: testreq
+      containers:
+        - name: inference-server
```
rubambiza, line 385: Is there any functionality that depends on both the [launcher](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/458/changes#diff-3ece2588db0331a78b7c430368b0fe04da16c443b1e4d1471f94112015f2eed5R304) and requester containers being named `inference-server`? If not, then I'd like to suggest naming the launcher container to something more aligned with its function.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,560 @@
...
+# =========================================================================
+# Step 6: HPA
+# =========================================================================
+
+step "HPA"
+
+if kubectl get hpa fma-hpa -n "$NAMESPACE" &>/dev/null; then
+    echo "  HPA fma-hpa already exists"
+else
+    echo "  Creating HPA targeting ReplicaSet/fma-requester..."
```
rubambiza, line 417: Should `fma-requester` as the ReplicaSet be parameterized (similar to `HPA_MAX_REPLICAS` below) for re-use in the ReplicaSet spec above and HPA spec below? Perhaps this also reverts back to @aavarghese 's [comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/458/changes#r3150321135) above about re-using some existing functionality.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,560 @@
...
+    - seriesQuery: 'wva_desired_replicas{variant_name!="",exported_namespace!=""}'
+      resources:
+        overrides:
+          exported_namespace:
+            resource: namespace
+          variant_name:
+            resource: deployment
+      name:
+        matches: "^wva_desired_replicas"
+        as: "wva_desired_replicas"
```
rubambiza: Where do the WVA stuff come in here? I thought that was going to be a separate/second push?

rubambiza (commented):
Leaving individual comments with some requested changes that are not blockers.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,560 @@
...
+# Deploys FMA + HPA + EPP + Gateway on an OCP cluster with real GPUs.
+#
+# Idempotent: checks each component before deploying, skips if already present.
+# Run from the root of the llm-d-fast-model-actuation repository.
+#
+# Prerequisites:
+#   - oc CLI authenticated to an OCP cluster with GPU nodes
+#   - helm, helmfile, kubectl, jq, yq installed
+#   - Container images already pushed to registry (see CONTAINER_IMG_REG)
+#   - llm-d/guides repo cloned locally (for EPP/Gateway via helmfile)
```
MikeSpreitzer, line 14: Another prerequisite is that this repo (ll-d-incubation/llm-d-fast-model-actuation) is cloned locally. That is not necessarily so for a bash script, which could be used by `curl | bash`.

I think it is worth being clear up front that this script deploys the _version_ of FMA that is checked out in the local repo.

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
aavarghese: It's an existing FMA script file (not new)

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
MikeSpreitzer: Yes, the dual-pods technique requires that particular name so that the relevant container can be identified. This allows side-cars to also be present without confusing the controllers.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
MikeSpreitzer: "Security Context Constraint", an OpenShift thing. See https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/authentication_and_authorization/managing-pod-security-policies

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
MikeSpreitzer: We factor the collection of image references into three parts: a common prefix (CONTAINER_IMG_REG), the particular image name (e.g., launcher, requester), and a common tag (IMAGE_TAG). This is a simple way to get a set of consistent image references.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
MikeSpreitzer: Yes, the name of this parameter is not consistent with what it actually means. I opened an Issue suggesting a fix for this, see #471 .

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
MikeSpreitzer: Remember that a LauncherPopulationPolicy can discriminate between nodes.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
MikeSpreitzer: Small generalizations of that existing script are not out of the question.

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
MikeSpreitzer: No, GPU mapping functionality is needed in both milestones. However, perhaps your question is specifically about the hack of doing it with a ConfigMap?

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
MikeSpreitzer: What do you mean by "gated on"? If we are using a public model, is an HF_TOKEN needed?

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
diegocastanibm: Sure, it's a small change

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
diegocastanibm: Sure, no problem. Also, I'm deleting line 61 since I removed the `gpu-map` based on [this comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/458#discussion_r3154852777)

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
diegocastanibm: The model is public and no token is needed it

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
diegocastanibm: Nop. The `"` are there (in line 86) to close the arg `-c`. All good.

test/e2e/demo-fma-hpa/demo-fma-hpa-monitor.sh
diegocastanibm: In this case, IMO, `INTERVAL` is more confusing. The word `every` is already defining a loop implicitly and `N` means a natural number. If I read the word `INTERVAL`, it came to my mind something like `3-5`. I really think is more clear with `N`.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: My understanding is that they are always installed together from the same `config/crd/` directory in a single loop, so it's effectively all-or-nothing. If `ISC` is present, the others will be too.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: OK, no problem. `#` it is.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: `fma-requester` appears in three scripts and the parametrization would need to be a shared env var across all scripts. For a demo script suite this adds complexity with limited benefit, since the name is self-descriptive and unlikely to need changing.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Good catch. The `wva_desired_replicas` rule crept in from the PoC environment where we were also experimenting with WVA alongside FMA+HPA. It is not needed for this demo. The HPA demo only uses `epp_queue_size` and `epp_running_requests`. Removed.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: OK. I have added `Deploys the version of FMA that is checked out locally` to the header and added the repo clone as an explicit prerequisite.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,551 @@
...
+#   REQUESTER_IMAGE    - requester image (default: $CONTAINER_IMG_REG/requester:$IMAGE_TAG)
+#   MODEL              - vLLM model (default: HuggingFaceTB/SmolLM2-360M-Instruct)
+#   GPU_NODE           - node for LPP (default: first node with nvidia.com/gpu.present=true)
+#   HPA_MAX_REPLICAS   - max HPA replicas (default: 4)
+#   PROM_ADAPTER_NS    - prometheus-adapter namespace (default: openshift-user-workload-monitoring)
+
+set -euo pipefail
+
+NAMESPACE="${NAMESPACE:-fma-hpa}"
+CONTAINER_IMG_REG="${CONTAINER_IMG_REG:-quay.io/diego_castan}"
```
aavarghese: why not point to our official img location at ghcr.io?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,551 @@
...
+IMAGE_TAG="${IMAGE_TAG:-f323a8f}"
```
aavarghese: Is this a special build `f323a8f` that has something HPA specific? Or could this point to our latest FMA release tag?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,551 @@
...
+    annotations:
+      description: "FMA ISC for HPA demo - ${MODEL}"
+  launcherConfigName: lc-hpa
+---
+apiVersion: fma.llm-d.ai/v1alpha1
+kind: LauncherConfig
+metadata:
+  name: lc-hpa
+spec:
+  maxSleepingInstances: 0
```
aavarghese: This line can be removed since the addition of a new field `maxInstances` can be used now...

test/e2e/demo-fma-hpa/demo-fma-hpa-monitor.sh
```diff
@@ -0,0 +1,54 @@
...
+    kubectl get hpa fma-hpa -n "$NAMESPACE" 2>/dev/null || echo "  (not found)"
+
+    echo ""
+    echo "--- Pods (requesters + launchers) ---"
+    # Show all requester and launcher pods with FMA labels in one table
+    kubectl get pods -n "$NAMESPACE" \
+        -L dual-pods.llm-d.ai/dual,dual-pods.llm-d.ai/sleeping \
+        --field-selector=status.phase!=Succeeded \
+        2>/dev/null \
+        | grep -E "^NAME|requester|^launcher" || echo "  (no FMA pods)"
```
aavarghese, line 38: Do we need the `grep` here? Isn't the label `dual-pods.llm-d.ai/dual` sufficient to find requester+launcher pods?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
aavarghese: We should move the image references in these scripts to use the official FMA images instead of a specific tag in quay

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: I explained it in this comment above: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/458#discussion_r3149593339

I will use the new image FMA and test it again.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Yes. And I'll test the behavior of the new field too

test/e2e/demo-fma-hpa/demo-fma-hpa-monitor.sh
diegocastanibm: Using `dual-pods.llm-d.ai/dual` would only show already-bound pairs. The grep on pod names captures all requesters and launchers regardless of binding state (which is more useful for a live demo where we want to observe that transition).

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: No HPA specific. As in the previous comment, I'll change it and test it with the new FMA release.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: We'll wait until a confirmation form @MikeSpreitzer

diegocastanibm:
## Changes in this update

### Fix CUDA/driver incompatibility — bump to v0.6.0-alpha.12

The default `IMAGE_TAG` now points to `v0.6.0-alpha.12`, which ships with vLLM v0.20.2 (CUDA 12.10). This resolves the `cudaErrorSystemDriverMismatch` (error 803) on clusters running NVIDIA driver 580.x (Open Kernel Module).

### Add EPP metrics scraping support (ServiceMonitor + RBAC)

GAIE v1.5.0 does not create a ServiceMonitor or the necessary RBAC for Prometheus to scrape EPP metrics. The deploy script now creates:

- A **ClusterRole + ClusterRoleBinding** granting the EPP ServiceAccount permissions to perform `tokenreviews` and `subjectaccessreviews` (required because EPP validates incoming scrape tokens via the K8s API).
- A **Secret** of type `kubernetes.io/service-account-token` for the EPP SA.
- A **ServiceMonitor** with bearer token auth pointing at the EPP's `http-metrics` port.

Without these, the HPA shows `<unknown>` because prometheus-adapter has no data.

### Update metric name for GAIE v1.5.0

The queue size metric was renamed upstream:
- **Before** (GAIE v1.4.x): `inference_extension_flow_control_queue_size{inference_pool=...}`
- **After** (GAIE v1.5.0): `inference_pool_average_queue_size{name=...}`

The prometheus-adapter rule for `epp_queue_size` now queries the correct series.

### Cleanup script: remove EPP RBAC on teardown

`demo-fma-hpa-cleanup.sh` now also deletes the cluster-scoped ClusterRole and ClusterRoleBinding created for EPP metrics auth.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,551 @@
...
+            resource: "namespace"
+        namespaced: false
+      name:
+        as: "epp_running_requests"
+      metricsQuery: 'sum(inference_objective_running_requests{namespace="${NAMESPACE}",job="${EPP_JOB}"})'
+VALEOF
+        helm upgrade prometheus-adapter prometheus-community/prometheus-adapter \
+            -n "$PROM_ADAPTER_NS" --reuse-values \
+            -f /tmp/adapter-epp-values.yaml
+        rm -f /tmp/adapter-epp-values.yaml
```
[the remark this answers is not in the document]
diegocastanibm: solved

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,551 @@
...
+    echo "  Deploying FMA controllers via deploy_fma.sh..."
+    cd "$REPO_ROOT"
+    FMA_NAMESPACE="$NAMESPACE" \
+    FMA_CHART_INSTANCE_NAME="$FMA_CHART" \
+    CONTAINER_IMG_REG="$CONTAINER_IMG_REG" \
+    IMAGE_TAG="$IMAGE_TAG" \
+    NODE_VIEW_CLUSTER_ROLE=create/please \
+    RUNTIME_CLASS_NAME=nvidia \
+    HELM_EXTRA_ARGS="--set launcherPopulator.enabled=true" \
+    "$SCRIPT_DIR/deploy_fma.sh"
```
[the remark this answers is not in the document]
diegocastanibm: Done

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
```diff
@@ -0,0 +1,205 @@
...
+    # 5. Strip finalizers — must happen BEFORE we delete controllers, otherwise
+    # finalizer removal is never processed and pods (and the namespace) hang
+    echo "--- Stripping dual-pods finalizers from pods ---"
+    strip_dual_pods_finalizers
+
+    # 6. Delete launcher pods explicitly (matches old + new label conventions)
+    echo "--- Deleting launcher pods ---"
+    kubectl delete pods -n "$NAMESPACE" -l app.kubernetes.io/component=launcher --ignore-not-found 2>/dev/null
+    kubectl delete pods -n "$NAMESPACE" -l dual-pods.llm-d.ai/sleeping --ignore-not-found 2>/dev/null
+    kubectl delete pods -n "$NAMESPACE" -l dual-pods.llm-d.ai/launcher-config-name --ignore-not-found 2>/dev/null
```
aavarghese: This step #6 is not needed. Deleting the launcherpopulationpolicy below will remove all the launcher pods...

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
```diff
@@ -0,0 +1,205 @@
...
+
+    # 7. FMA objects (CRs)
+    echo "--- Deleting FMA objects ---"
+    kubectl delete launcherpopulationpolicy lpp-hpa -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+    kubectl delete launcherconfig lc-hpa -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+    kubectl delete inferenceserverconfig isc-smol -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+
```
aavarghese, line 91: I would add a `sleep 10` here before uninstalling the Helm FMA release so that we give some time to the controller to delete all the resoruces...

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
```diff
@@ -0,0 +1,205 @@
...
+    # 8. FMA controllers (Helm release)
+    echo "--- Uninstalling FMA controllers ---"
+    helm uninstall fma -n "$NAMESPACE" 2>/dev/null || true
+
+    # 9. RBAC
```
aavarghese: deploy_fma.sh creates a clusterrole, and not a role. So `clusterrole` will need to be deleted. Lines 107-109 can be removed because deleting the namespace will delete these namespace-scoped resources..

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+
+if kubectl get crd inferenceserverconfigs.fma.llm-d.ai &>/dev/null; then
+    echo "  FMA CRDs: present"
+else
+    echo "  Installing FMA CRDs..."
+    for crd_file in "$REPO_ROOT"/config/crd/*.yaml; do
+        kubectl apply --server-side --force-conflicts -f "$crd_file"
+    done
+fi
+
```
aavarghese: CRDs won't get created until we call deploy_fma.sh

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+step "Namespace, ServiceAccounts, RBAC"
+
+if kubectl get ns "$NAMESPACE" &>/dev/null; then
+    echo "  Namespace $NAMESPACE exists"
+else
+    kubectl create ns "$NAMESPACE"
+    echo "  Created namespace $NAMESPACE"
+fi
+
+for sa in testreq testlauncher; do
```
aavarghese: The `testreq` SA isn't needed in OCP. That's a Kind resource..

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+  resourceNames: [gpu-allocs]
+  resources: [configmaps]
+  verbs: [update, patch, get, list, watch]
+- apiGroups: [""]
+  resources: [configmaps]
+  verbs: [create]
+- apiGroups: [""]
+  resources: [pods]
+  verbs: [get, list, watch]
+---
```
aavarghese: Similarly, this can also be removed

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+metadata:
+  name: testlauncher
+rules:
+- apiGroups: [""]
+  resources: [pods]
+  verbs: [get, patch]
+EOF
+    kubectl create rolebinding testreq \
+        --role=testreq --serviceaccount="${NAMESPACE}:testreq" \
+        -n "$NAMESPACE" 2>/dev/null || true
```
aavarghese: and this can be removed...

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+  template:
+    metadata:
+      labels:
+        app: fma-hpa-requester
+      annotations:
+        dual-pods.llm-d.ai/admin-port: "8081"
+        dual-pods.llm-d.ai/inference-server-config: "isc-smol"
+    spec:
+      runtimeClassName: nvidia
+      serviceAccountName: testreq
```
aavarghese: Requester in OCP can use the default SA. No need for line 430

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+          - name: XDG_CACHE_HOME
+            value: "/tmp"
+          - name: XDG_CONFIG_HOME
+            value: "/tmp"
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef: { fieldPath: spec.nodeName }
+          - name: NAMESPACE
+            valueFrom:
+              fieldRef: { fieldPath: metadata.namespace }
```
aavarghese: We dont need these...those were for Kind.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+---
+apiVersion: fma.llm-d.ai/v1alpha1
+kind: LauncherPopulationPolicy
+metadata:
+  name: lpp-hpa
+spec:
+  enhancedNodeSelector:
+    labelSelector:
+      matchLabels:
+        fma-hpa-poc: "true"
```
aavarghese, line 358: Shoudl we also add `nvidia.com/gpu.present: "true"`?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,648 @@
...
+    options: "--model ${MODEL} --enable-sleep-mode"
+    env_vars:
+      VLLM_USE_V1: "1"
+      VLLM_LOGGING_LEVEL: "DEBUG"
+      HF_HOME: "/tmp"
+      VLLM_CACHE_ROOT: "/tmp"
+      FLASHINFER_WORKSPACE_BASE: "/tmp"
+      TRITON_CACHE_DIR: "/tmp"
+      XDG_CACHE_HOME: "/tmp"
+      XDG_CONFIG_HOME: "/tmp"
```
aavarghese: From vllm docs:
```
To enable sleep mode in a vLLM server you need to initialize it with the flag VLLM_SERVER_DEV_MODE=1 and pass --enable-sleep-mode to the vLLM server.
```
We are missing `VLLM_SERVER_DEV_MODE=1` in env_vars above

aavarghese: These are set in `LauncherConfig`. We dont need them in ISC

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
rubambiza: @MikeSpreitzer Yes, I meant the hack of doing it with a ConfigMap (which I recall was circa M2).

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
rubambiza: This is now complete: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/516. Leaving open for @aavarghese to resolve.

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
rubambiza: Thanks for clarifying that the model is public. I was looking ahead in case we build on this for the benchmarking. The current model being pulled for the benchmarking, `meta-llama-3.1-8B-Instruct` does require a `HF_TOKEN`

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
rubambiza: Thank you!

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
aavarghese: Doing it with a CM is now only used on simulated GPUs on Kind. Not needed on real clusters with M3.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
aavarghese: Yes, let's remove maxSleepingInstances and use maxInstances instead.

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
```diff
@@ -0,0 +1,89 @@
...
+echo "  Duration:  ${DURATION}s"
+echo "  MaxTokens: $MAX_TOKENS"
+echo ""
+
+# Delete any leftover pod
+kubectl delete pod "$POD_NAME" -n "$NAMESPACE" --ignore-not-found 2>/dev/null
+
+echo "Starting loadgen pod..."
+kubectl run "$POD_NAME" -n "$NAMESPACE" --rm -i --restart=Never \
+    --image=curlimages/curl:latest -- sh -c "
```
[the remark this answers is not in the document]
diegocastanibm: Add version to `--image=curlimages/curl:8.12.1`

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Adding `maxInstances: 1`

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
diegocastanibm: Resolve @rubambiza ?

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
diegocastanibm: OK

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
diegocastanibm: Sure, no problem

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
diegocastanibm: Done. I have deleted the lines that removed Roles/RoleBindings/SAs namespace-scoped (they are gone with the namespace with the `FULL_CLEANUP` option). ClusterRoles cluster-scoped are eliminated below

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: OK

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: OK

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: yes, removed with the previous comment

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Also removed in the previous comment

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Same thing as above

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Right. If real GPUs, then no need for these.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: It's an extra security layer but can be useful in case we label a node without GPU's. I have added it

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: mmmmm, this is weird because I'm sure it worked without it, but it is better to use the official doucmentation and adding it

diegocastanibm: Sure

diegocastanibm:
## Changes since last review

### Fix: prometheus-adapter rules not being discovered by HPA

**Root cause:** The cluster has two prometheus-adapter instances. The deploy script was updating the one in `openshift-user-workload-monitoring`, but the `external.metrics.k8s.io` APIService is actually served by a different adapter in `workload-variant-autoscaler-monitoring`. Additionally, the rules used `namespaced: false` which prevented namespace-scoped HPA queries from working.

**Fix (deploy script - Step 7 rewritten):**
- Auto-detect the adapter namespace from the `v1beta1.external.metrics.k8s.io` APIService instead of hardcoding it
- Patch the adapter's ConfigMap directly instead of relying on Helm (the adapter may not be Helm-managed)
- Use `<<.LabelMatchers>>` in metricsQuery for multi-tenant compatibility
- Remove `namespaced: false` from resource definitions so metrics are namespace-scoped
- Remove `PROM_ADAPTER_NS` as a user-facing parameter (now auto-detected)

**Fix (cleanup script):**
- Same approach: discover adapter via APIService, patch ConfigMap directly, restart pods

### Verified end-to-end
- HPA correctly reads `epp_queue_size` and `epp_running_requests` metrics
- Autoscaling triggers under load (scaled from 1 to 2+ replicas)

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
```diff
@@ -0,0 +1,191 @@
...
+    echo ""
+    echo "--- Removing agentgateway control plane ---"
+    if ! kubectl get ns agentgateway-system &>/dev/null; then
+        echo "  agentgateway-system not present — skipping."
+    elif ! command -v helmfile &>/dev/null; then
+        echo "  WARNING: helmfile not installed. Run manually:"
+        echo "    helmfile destroy -f \${LLM_D_GUIDES_DIR}/prereq/gateway-provider/agentgateway.helmfile.yaml"
+    elif [ -z "${LLM_D_GUIDES_DIR:-}" ]; then
+        echo "  WARNING: LLM_D_GUIDES_DIR not set. Run manually:"
+        echo "    helmfile destroy -f \${LLM_D_GUIDES_DIR}/prereq/gateway-provider/agentgateway.helmfile.yaml"
```
rubambiza: Two things here:

- This seems to be the first expansion of `LLM_D_GUIDES_DIR`, but I don't see it defined anywhere in the file
- We're expanding `LLM_D_GUIDES_DIR` and passing it in the invocation on line 182, but its existence is not even checked until like 185.

Maybe I'm missing some context on the variable because I haven't read the rest of the files where it's set as an `env` variable.

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
```diff
@@ -0,0 +1,89 @@
...
+MODEL=\"${MODEL}\"
+WORKERS=${WORKERS}
+DURATION=${DURATION}
+MAX_TOKENS=${MAX_TOKENS}
+
+send_request() {
+  while true; do
+    curl -s -o /dev/null -w '' \"\$GW/v1/completions\" \
+      -H 'Content-Type: application/json' \
+      -d '{\"model\":\"\$MODEL\",\"prompt\":\"Write a very long and detailed essay about the history of artificial intelligence from the 1950s to the present day, covering all major breakthroughs, researchers, institutions, and technological advances in great detail.\",\"max_tokens\":'\$MAX_TOKENS',\"temperature\":0.9,\"stream\":false}' 2>/dev/null
```
rubambiza: I think the LLM gets it at `very long and detailed`, no need for `in great detail` at the end 😆

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,601 @@
...
+#
+# Prerequisites:
+#   - This repo (llm-d-incubation/llm-d-fast-model-actuation) cloned locally
+#   - oc CLI authenticated to an OCP cluster with GPU nodes
+#   - helm, helmfile, kubectl, jq, yq installed
+#   - Container images already pushed to registry (see CONTAINER_IMG_REG)
+#   - llm-d/guides repo cloned locally (for EPP/Gateway via helmfile)
+#
+# Required environment variables:
+#   LLM_D_GUIDES_DIR  - path to llm-d/guides repo (contains optimized-baseline Helm values and prereqs)
```
rubambiza, line 17: I think this is the `env` I was looking for in my previous comment. It may be nice to indicate in the other script that this `env-var` is already pre-set in another script.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,601 @@
...
+#
+# Optional environment variables (with defaults):
+#   NAMESPACE          - target namespace (default: fma-hpa)
+#   CONTAINER_IMG_REG  - image registry (default: ghcr.io/llm-d-incubation/llm-d-fast-model-actuation)
+#   IMAGE_TAG          - image tag (default: v0.6.0-alpha.12, latest release)
```
rubambiza: The comment of `latest release` may go out of style/date when we cut a new release. I think, for whatever version that a reader is looking at for the script, having `v0.6.0-alpha.12` is a clear-cut point in time unless/until the script changes. In other words, I'd suggest removing `latest release` to avoid confusion.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,601 @@
...
+step() {
+    step_num=$((step_num + 1))
+    echo ""
+    echo "========================================"
+    echo "  Step ${step_num}/${total_steps}: $*"
+    echo "========================================"
+    echo ""
+}
+
+check_skip() {
```
rubambiza: This does not seem to be invoked anywhere in the file. Suggestion: pass a parameter to the function (e.g., component/object) so that it just echoes

`$component already exists, skipping`

and then invoke it on lines like 66, 73, etc.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,601 @@
...
+# =========================================================================
+# Step 2: CRDs (Gateway API, GAIE, FMA)
+# =========================================================================
+
+step "CRDs"
+
+if kubectl get crd gateways.gateway.networking.k8s.io &>/dev/null; then
+    echo "  Gateway API CRDs: present"
+else
+    echo "  Installing Gateway API CRDs v1.5.1..."
```
rubambiza: I don't know how often the CRDs change, if at all. If they do so frequently, might be worth parameterizing this. Otherwise, feel free to ignore. I noticed similar thing for `GAIE` on line 115 and 116.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,601 @@
...
+
+echo "  FMA CRDs: installed by deploy_fma.sh in Step 4"
+
+# =========================================================================
+# Step 3: EPP + Gateway
+# =========================================================================
+
+step "EPP + Gateway (optimized-baseline)"
+
+if [ -z "${LLM_D_GUIDES_DIR:-}" ]; then
```
rubambiza: Shouldn't this come much sooner to avoid doing unnecessary work if everything hinges on it?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
rubambiza: Seems this is already the case on [line 133](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/458/changes#diff-3ece2588db0331a78b7c430368b0fe04da16c443b1e4d1471f94112015f2eed5R133). Maybe move it here?

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
```diff
@@ -0,0 +1,601 @@
...
+                echo "  EPP rules present but stale — replacing..."
+            else
+                echo "  Adding EPP External Metrics rules to prometheus-adapter..."
+            fi
+
+            # Remove any old FMA rules, then append fresh ones
+            yq eval 'del(.externalRules[] | select(.name.as == "epp_queue_size" or .name.as == "epp_running_requests"))' \
+                "$CURRENT_CONFIG" > "$NEW_CONFIG"
+
+            # Append our two rules
```
rubambiza: Can you please say more in the top line comment about what the `two rules` are doing? I was lost a bit reading the next 15 lines or so.

rubambiza (dismissed):
I left some request for changes, but most of them are minor. I should be good to approve once they are addressed.

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
diegocastanibm: This is actually a very valid comment. 
First, some context:  `LLM_D_GUIDES_DIR` is an environment variable that the user must set externally (pointing to their local clone of the llm-d/guides repo). It's documented as a required env var in the deploy script header. 
Secondly, you are right. The order should be improved. I have reordered the checks so `LLM_D_GUIDES_DIR`  is validated before checking for helmfile. This way the logic short-circuits earlier if the variable isn't set.

test/e2e/demo-fma-hpa/demo-fma-hpa-loadgen.sh
diegocastanibm: Not critical, but I removed the redundant "in great detail" at the end

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: added LLM_D_GUIDES_DIR to the cleanup script's header

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: I removed "latest release"

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Ah! Good one. it was a leftover and I simply removed it because the existing per-component messages are already descriptive enough so a generic helper doen not add much

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Addressed both: moved `GAIE_VERSION` to the top of the file alongside a new `GATEWAY_API_VERSION` variable. Both CRD install commands now use these variables, so bumping versions is a single-line change. Removed the duplicate `GAIE_VERSION` declaration that was further down.

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: moved the LLM_D_GUIDES_DIR validation to the top of the script

test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh
diegocastanibm: Added a comment explaining what each rule does: `epp_queue_size` drives scale-up  and `epp_running_requests` prevents premature scale-down

MikeSpreitzer:
@diegocastanibm: this needs to be rebased onto `main` so that we can graduate from stage B of #471 .

test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh
rubambiza: @diegocastanibm: Given @aavarghese's comment

>  Doing it with a CM is now only used on simulated GPUs on Kind. Not needed on real clusters with M3.

Perhaps it makes sense to remove the ConfigMap creation/deletion? This can be done in a follow-on PR.

rubambiza (commented):
LGTM. The remaining comment on the ConfigMap hack can be addressed in a follow-up PR. Leaving a general comment so that Mike's [request for a rebase](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/458#issuecomment-4590084992) can be addressed before approval.

test/e2e/deploy_fma.sh
```diff
@@ -1,8 +1,5 @@
 #!/usr/bin/env bash
 
-# Usage: $0
```
MikeSpreitzer, line 3: This usage description is still accurate.

test/e2e/deploy_fma.sh
```diff
@@ -1,8 +1,5 @@
...
-# Current working directory must be the root of the Git repository.
```
MikeSpreitzer, line 4: The user still needs to know what is expected regarding working directory. The comment should be updated to remain accurate.

test/e2e/deploy_fma.sh
diegocastanibm: Response below

test/e2e/deploy_fma.sh
diegocastanibm: Restored the `# Usage: $0` line and added a note that the script can be run from any directory (it auto-detects the repository root).

aavarghese (approved):
LGTM - One thing missing might be a super short document on steps 1, 2 and 3 to run FMA with HPA so a reader does not have search through the comments in these scripts. Can be part of another PR. Thanks Diego!

diegocastanibm:
> LGTM - One thing missing might be a super short document on steps 1, 2 and 3 to run FMA with HPA so a reader does not have search through the comments in these scripts. Can be part of another PR. Thanks Diego!

Sure, I'll do it in a separate PR. This one is already big enough

Issue #435, opened 2026-04-16
[Bug]: Image build is too big for runner


MikeSpreitzer:
### What happened?

This Issue is about a flaky test.

An E2E test on OpenShift workflow run failed in the build-image job (which does not even run on the OpenShift cluster), running out of disk space. This does not always happen. A retry has been known to succeed.

For an example, see https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/24527702495/job/71702332146 . The Annotations on the build-image job say the following.

```
System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.333.1/_diag/Worker_20260416-184233-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.333.1/_diag/Worker_20260416-184233-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Common.Tracing.Error(Exception exception)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
Unhandled exception. System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.333.1/_diag/Worker_20260416-184233-utc.log'
[...]

PR #465, opened 2026-04-28
fix: populator retries after per-node reconciliation errors

### LINKED ISSUE #462: [Bug]: populator missing some retries
### Contact Details

_No response_

### What happened?

In reviewing #374, I noticed a couple of cases where the launcher population controller just logs an error and moves on to the next Node but instead should retry.

https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3155919001
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3155949339

### Version

non-main branch (please specify below)

### Branch name

osswangxining:handleLauncherConfigUpdate

### Commit SHA

9c06feacfa604ff492b504d2bacfb4e1159ac022

### Relevant log output

```shell

```

MikeSpreitzer:
## Summary

- When `reconcileLaunchersOnSingleNode` returned an error (e.g. a pod deletion failed due to a transient API error), `reconcileAllLaunchers` logged it and moved on without triggering a retry, leaving the cluster potentially inconsistent.
- Fix: set `anyRequeueNeeded = true` whenever a per-node error is returned, so the work item is requeued.
- Also collapse the `nodeLister.Get` error handling: the lister only ever returns NotFound in practice, so the non-NotFound branch is unreachable. It now logs distinctively rather than propagating an error.

## Test plan

- [x] Existing unit/integration tests pass
- [ ] Manual review: verify that a simulated pod-deletion error causes the item to be requeued

Fixes #462

### DIFF
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -345,10 +345,10 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			logger.Info("Node no longer exists, skipping reconciliation", "node", nodeName)
-		} else {
-			logger.Error(err, "Unexpected error from node lister (should be impossible), skipping reconciliation", "node", nodeName)
+			return false, nil
 		}
-		return false, nil
+		logger.Error(err, "Unexpected error from node lister (should be impossible), will retry", "node", nodeName)
+		return true, nil
 	}
 
 	didDelete := false
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -321,6 +321,7 @@ func (ctl *controller) reconcileAllLaunchers(ctx context.Context, desired map[No
 		needsRequeue, err := ctl.reconcileLaunchersOnSingleNode(ctx, nodeName, keys, desired)
 		if err != nil {
 			logger.Error(err, "Failed to reconcile launchers on node", "node", nodeName)
+			anyRequeueNeeded = true
 			continue
 		}
 		anyRequeueNeeded = anyRequeueNeeded || needsRequeue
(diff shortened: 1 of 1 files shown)


pkg/controller/launcher-populator/populator.go
```diff
@@ -344,9 +345,10 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			logger.Info("Node no longer exists, skipping reconciliation", "node", nodeName)
-			return false, nil
+		} else {
+			logger.Error(err, "Unexpected error from node lister (should be impossible), skipping reconciliation", "node", nodeName)
 		}
-		return false, fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		return false, nil
```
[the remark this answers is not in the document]
MikeSpreitzer: changed

pkg/controller/launcher-populator/populator.go
```diff
@@ -346,7 +347,8 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 			logger.Info("Node no longer exists, skipping reconciliation", "node", nodeName)
 			return false, nil
 		}
-		return false, fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		logger.Error(err, "Unexpected error from node lister (should be impossible), will retry", "node", nodeName)
+		return true, nil
```
diegocastanibm, line 351: Why are we returning `true` instead of `false` as before?

MikeSpreitzer: This is handling for an error that we think is impossible. If ever it becomes possible, returning `true` here is better because it will lead to either a successful retry or spamming the log so that the problem hopefully gets noticed.

diegocastanibm: ah, I see. Thanks

diegocastanibm:
Can we solve the space issue, and run the test again before approving and merging?
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/25076202661

MikeSpreitzer:
See #435 for the runner disk space issue, and the GitHub documentation about "larger runners". If I understand the latter correctly, we need help from an admin of the llm-d-incubation organization on GitHub.

osswangxining (approved):
LGTM  Keeping retry if cannot get node

PR #466, opened 2026-04-29
Fixes #456 refine the status of LauncherPopulationPolicy

### LINKED ISSUE #456: Refine the status for LauncherPopulationPolicy if user uses wrong labelselector
If the error is from `metav1.LabelSelectorAsSelector(&selector.LabelSelector)` then this is a user error that should be reported in the `Status` of the `LauncherPopulationPolicy`. Can be done in a follow-on PR.

_Originally posted by @MikeSpreitzer in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3145072076_

osswangxining:
Fixes #456 

## Background:

If the error is from metav1.LabelSelectorAsSelector(&selector.LabelSelector) then this is a user error that should be reported in the Status of the LauncherPopulationPolicy. Can be done in a follow-on PR.

Originally posted by @MikeSpreitzer in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3145072076

### DIFF
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -246,19 +246,21 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
 			// This is a user error: the LabelSelector in the policy's EnhancedNodeSelector is invalid.
 			// Report it in the policy's Status.Errors so the user can see it via kubectl.
 			logger.Error(selectorErr, "Invalid LabelSelector in policy, reporting in Status", "policy", lpp.Name)
-			if statusErr := ctl.reportLPPSelectorError(ctx, lpp, selectorErr); statusErr != nil {
+			if statusErr := ctl.setLPPStatusErrors(ctx, lpp, []string{selectorErr.Error()}); statusErr != nil {
 				logger.Error(statusErr, "Failed to update Status for policy", "policy", lpp.Name)
 			}
 			continue
+		}
+		// Clear any previously reported selector errors now that the selector is valid.
+		// This is done as soon as selectorErr == nil, before checking err, so that a
+		// transient infrastructure error does not prevent stale Status errors from being cleared.
+		if statusErr := ctl.setLPPStatusErrors(ctx, lpp, nil); statusErr != nil {
+			logger.Error(statusErr, "Failed to clear Status errors for policy", "policy", lpp.Name)
 		}
 		if err != nil {
 			// This is an infrastructure error: the lister failed to list nodes.
 			logger.Error(err, "Failed to get matching nodes for policy", "policy", lpp.Name)
 			continue
-		}
-		// Clear any previously reported selector errors now that the selector is valid.
-		if statusErr := ctl.clearLPPSelectorError(ctx, lpp); statusErr != nil {
-			logger.Error(statusErr, "Failed to clear Status errors for policy", "policy", lpp.Name)
 		}
 
 		for _, countRule := range lpp.Spec.CountForLauncher {
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -37,6 +37,7 @@ import (
 	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
 	genctlr "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/generic"
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
+	fmaclientv1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/generated/clientset/versioned/typed/fma/v1alpha1"
 	fmainformers "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/generated/informers/externalversions"
 	fmalisters "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/generated/listers/fma/v1alpha1"
 )
@@ -52,6 +53,7 @@ type Controller interface {
 func NewController(
 	logger klog.Logger,
 	coreClient coreclient.CoreV1Interface,
+	fmaClient fmaclientv1alpha1.FmaV1alpha1Interface,
 	namespace string,
 	corev1PreInformers corev1preinformers.Interface,
 	fmaInformerFactory fmainformers.SharedInformerFactory,
@@ -59,6 +61,7 @@ func NewController(
 	ctl := &controller{
 		enqueueLogger: logger.WithName(ControllerName),
 		coreclient:    coreClient,
+		fmaclient:     fmaClient,
 		namespace:     namespace,
 		podInformer:   corev1PreInformers.Pods().Informer(),
 		podLister:     corev1PreInformers.Pods().Lister(),
@@ -97,6 +100,7 @@ func NewController(
 type controller struct {
 	enqueueLogger klog.Logger
 	coreclient    coreclient.CoreV1Interface
+	fmaclient     fmaclientv1alpha1.FmaV1alpha1Interface
 	namespace     string
 	podInformer   cache.SharedIndexInformer
 	podLister     corev1listers.PodLister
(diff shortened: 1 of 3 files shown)


pkg/controller/launcher-populator/populator.go
```diff
@@ -237,11 +241,25 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
...
+		}
 		if err != nil {
+			// This is an infrastructure error: the lister failed to list nodes.
 			logger.Error(err, "Failed to get matching nodes for policy", "policy", lpp.Name)
 			continue
 		}
+		// Clear any previously reported selector errors now that the selector is valid.
+		if statusErr := ctl.clearLPPSelectorError(ctx, lpp); statusErr != nil {
+			logger.Error(statusErr, "Failed to clear Status errors for policy", "policy", lpp.Name)
+		}
```
MikeSpreitzer: Shouldn't this be done as soon as it is known that `selectorErr == nil`, so that the `err != nil` handling does not prevent it?

pkg/controller/launcher-populator/populator.go
```diff
@@ -283,16 +301,20 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
...
-	// Use label selector to filter nodes
-	labelSelector, err := metav1.LabelSelectorAsSelector(&selector.LabelSelector)
-	if err != nil {
-		return nil, fmt.Errorf("failed to convert label selector: %w", err)
+// getMatchingNodes returns nodes that match the EnhancedNodeSelector.
+// It returns three values: the matched nodes, a user-facing selector error (non-nil when the
+// LabelSelector itself is malformed — this is a user configuration error), and an internal
+// error (non-nil for unexpected infrastructure failures such as lister errors).
+// Callers should handle selectorErr and err independently.
+func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha1.EnhancedNodeSelector) ([]corev1.Node, error, error) {
```
MikeSpreitzer: It would be simpler to return the selector error, or lack thereof, as a `[]string`, since that is what the ultimate consumer wants.

pkg/controller/launcher-populator/populator_status_handler.go
```diff
@@ -0,0 +1,40 @@
...
+
+	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
+	applyfmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/generated/applyconfiguration/fma/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// reportLPPSelectorError writes a selector validation error into the LauncherPopulationPolicy's
+// Status.Errors field using Server-Side Apply so the user can observe it via kubectl.
+func (ctl *controller) reportLPPSelectorError(ctx context.Context, lpp *fmav1alpha1.LauncherPopulationPolicy, selectorErr error) error {
+	apply := applyfmav1alpha1.LauncherPopulationPolicy(lpp.Name, lpp.Namespace).
```
MikeSpreitzer: As in the case of `clearLPPSelectorError`, this should first check whether `lpp.Status.Errors` already has the desired content.

In fact, we do not really need two distinct methods here. Just one that takes a `desiredErrors []string` and stores that if it is different from what is already in `lpp.Status.Errors`.

Setting the Status should include setting the Status.ObservedGeneration.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/populator.go
osswangxining: Make sense, I missed this scenario. Fixed now.   As long as selectorErr == nil (indicating the selector is valid), the historical error record should be cleared immediately. Subsequent transient infrastructure failures (such as nodeLister errors) should not prevent this cleanup.

pkg/controller/launcher-populator/populator.go
osswangxining: The current design keeps getMatchingNodes unaware of the Status data format — converting the error to a string is the responsibility of reportLPPSelectorError. Returning error also preserves structured logging context that would be lost with []string.

pkg/controller/launcher-populator/populator_status_handler.go
osswangxining: Make sense, it's simpler. I refined this.

pkg/controller/launcher-populator/populator_status_handler.go
```diff
@@ -0,0 +1,30 @@
...
+		return nil
+	}
+	apply := applyfmav1alpha1.LauncherPopulationPolicy(lpp.Name, lpp.Namespace).
+		WithStatus(applyfmav1alpha1.LauncherPopulationPolicyStatus().
+			WithObservedGeneration(lpp.Generation).
+			WithErrors(desiredErrors...))
+	_, err := ctl.fmaclient.LauncherPopulationPolicies(lpp.Namespace).ApplyStatus(ctx, apply, metav1.ApplyOptions{
+		FieldManager: ControllerName,
+		Force:        true,
+	})
```
MikeSpreitzer: The following suggestion does the update with simpler coding, and adds a uniform log message. Logging the ResourceVersion helps with debugging concurrency issues. Doing the logging here removes the need to do it in the caller, leading to more code simplification.

```suggestion
	lpp = lpp.DeepCopy()
	lpp.Status = LauncherPopulationPolicyStatus{
	    ObservedGeneration: lpp.Generation,
	    Errors: desiredErrors}
	echo, err := ctl.fmaclient.LauncherPopulationPolicies(lpp.Namespace).Update(ctx, lpp, metav1.UpdateOptions{FieldManager: ControllerName})
	klog.FromContext(ctx).V(4).Info("Updated LauncherPopulationPolicyStatus", "name", lpp.Name, "observedGeneration", lpp.Generation, "errors", desiredErrors, "resourceVersion", echo.ResourceVersion)
```

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Structured logging can log a slice of strings. See the suggested log statement in my other comment.

There is nothing wrong with getMatchingNodes returning user errors as a slice of strings. I see no reason to make an inconvenient choice when we could make a convenient choice.

pkg/controller/launcher-populator/populator.go
```diff
@@ -237,8 +241,24 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
...
+				logger.Error(statusErr, "Failed to update Status for policy", "policy", lpp.Name)
+			}
+			continue
+		}
+		// Clear any previously reported selector errors now that the selector is valid.
+		// This is done as soon as selectorErr == nil, before checking err, so that a
+		// transient infrastructure error does not prevent stale Status errors from being cleared.
+		if statusErr := ctl.setLPPStatusErrors(ctx, lpp, nil); statusErr != nil {
+			logger.Error(statusErr, "Failed to clear Status errors for policy", "policy", lpp.Name)
+		}
```
MikeSpreitzer:
```suggestion
		nodes, selectorErrs, err := ctl.getMatchingNodes(ctx, lpp.Spec.EnhancedNodeSelector)
		// Ensure proper status for lpp.
		if statusErr := ctl.setLPPStatusErrors(ctx, lpp, selectorErrs); statusErr != nil {
			logger.Error(statusErr, "Failed to set Status for policy", "policy", lpp.Name)
		}
```

MikeSpreitzer (commented):

pkg/controller/launcher-populator/populator.go
osswangxining: fine, I changed to []string instead of errors, thanks.

pkg/controller/launcher-populator/populator.go
osswangxining: resolved as this suggestion

pkg/controller/launcher-populator/populator_status_handler.go
osswangxining: fixed

osswangxining:
@MikeSpreitzer thanks for your review, I updated based on your comments above. Please help review again.

pkg/controller/launcher-populator/populator_status_handler.go
```diff
@@ -0,0 +1,33 @@
...
+	lppCopy.Status = fmav1alpha1.LauncherPopulationPolicyStatus{
+		ObservedGeneration: lpp.Generation,
+		Errors:             desiredErrors,
+	}
+	echo, err := ctl.fmaclient.LauncherPopulationPolicies(lppCopy.Namespace).Update(ctx, lppCopy, metav1.UpdateOptions{FieldManager: ControllerName})
+	resourceVersion := ""
+	if echo != nil {
+		resourceVersion = echo.ResourceVersion
+	}
+	klog.FromContext(ctx).V(4).Info("Updated LauncherPopulationPolicyStatus", "name", lppCopy.Name, "observedGeneration", lppCopy.Generation, "errors", desiredErrors, "resourceVersion", resourceVersion)
```
MikeSpreitzer, line 31: This would be better if it revealed whether `err != nil`.

MikeSpreitzer (approved):
Suggested one improvement. Can be made in follow-on PR.

PR #467, opened 2026-04-29
Fixes #451 refine the status of LauncherConfig during populating

### LINKED ISSUE #451: Refine the status for LauncherConfig
Since the only error that actually reaches here is a user error (not putting an inference server container in the launcher Pod template), the proper handling of this error is to, in addition to ignoring the LauncherConfig here, expose the error to users in the `Status` section of the `LauncherConfig` (which, I see, does not yet have any place to report problems). Can be handled in other PR(s).

_Originally posted by @MikeSpreitzer in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3124765727_

osswangxining:
This PR addresses the part of #451 (reporting spec errors in the status of a LauncherPopulationPolicy) that falls on the launcher population controller.

## background

Since the only error that actually reaches here is a user error (not putting an inference server container in the launcher Pod template), the proper handling of this error is to, in addition to ignoring the LauncherConfig here, expose the error to users in the Status section of the LauncherConfig (which, I see, does not yet have any place to report problems). Can be handled in other PR(s).

Originally posted by @MikeSpreitzer in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3124765727


api/fma/v1alpha1/launcherconfig_types.go
```diff
@@ -57,6 +57,9 @@ type LauncherConfigSpec struct {
 
 // LauncherConfigStatus represents the current status
 type LauncherConfigStatus struct {
```
MikeSpreitzer, line 59: A Status section should also have the following.

```golang
	// `observedGeneration` is the `metadata.generation` last seen by the controller.
    // In other words, the version of the spec that the rest of the status is about.
	// +optional
	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```

pkg/controller/launcher-populator/populator_status_handler.go
```diff
@@ -38,3 +38,32 @@ func (ctl *controller) clearLPPSelectorError(ctx context.Context, lpp *fmav1alph
 	})
 	return err
 }
+
+// reportLCTemplateError writes a PodTemplate validation error into the LauncherConfig's
+// Status.Errors field using Server-Side Apply so the user can observe it via kubectl.
+func (ctl *controller) reportLCTemplateError(ctx context.Context, lc *fmav1alpha1.LauncherConfig, templateErr error) error {
```
MikeSpreitzer: Note that one LauncherConfig object can be processed many times by this controller in one call to `reconcileFromPolicies`. We really do not want a round trip with the apiserver every time.

As in #466, these two methods can be simplified into one that takes `desiredErrors []string` and makes an update if `lc.Status` currently has the wrong `ObservedGeneration` or `Errors`.

Because of the way that one read of the LauncherConfig is followed by potentially many detections of the same user error, we will want an additional `sets.Set` of LauncherConfig names to keep track of the ones already reported.

pkg/controller/launcher-populator/populator.go
```diff
@@ -406,12 +406,23 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
 			nominalPod, err := utils.BuildLauncherPodFromTemplate(
 				entry.LauncherConfigSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
 			if err != nil {
```
MikeSpreitzer, line 428: There should be an unconditional call to make sure that the LauncherConfig's Status holds the right ObservedGeneration and Errors.

MikeSpreitzer (commented):

Also, note https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/451#issuecomment-4340994783 . Because of that, this PR does not entirely resolve the Issue.

To fix the IDL consumption oversights:

```shell
make manifests generate generate_client
```

api/fma/v1alpha1/launcherconfig_types.go
osswangxining: added.

pkg/controller/launcher-populator/populator_status_handler.go
osswangxining: agree, make sense. thanks for reminder

pkg/controller/launcher-populator/populator.go
osswangxining: I added the observedGeneration now , please help check if this answers the question  or not.

osswangxining:
> I left some individual comments.
> 
> Also, note [#451 (comment)](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/451#issuecomment-4340994783) . Because of that, this PR does not entirely resolve the Issue.
> 
> To fix the IDL consumption oversights:
> 
> ```shell
> make manifests generate generate_client
> ```

THanks, I run this now

MikeSpreitzer:
> Also, note https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/451#issuecomment-4340994783 . Because of that, this PR does not entirely resolve the Issue.

I noticed that the opening comment of this PR is formatted in a way that might make GitHub automatically close Issue 451. I edited the opening comment to avoid that and be clear about the reason.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: The addition of ObservedGeneration does not respond to my previous comment here. My point was that we want to make sure that the status is correct when `err` is `nil` as well as when it is not. I see that the latest revision does that, but I think that it could be accomplished with code that is more simple and uniform (see my comments on #466 for a way of doing that).

pkg/controller/launcher-populator/populator_status_handler.go
MikeSpreitzer: It still looks like this code is going to do a round trip with a kube apiserver for every Node that the policy applies to. That multiplies the work by the number of those Nodes. It also does unnecessary work if the policy's status already holds the desired content.

pkg/controller/launcher-populator/populator.go
osswangxining: I submit one new version, please help check if this catchesr your comment, thanks.

pkg/controller/launcher-populator/populator.go
```diff
@@ -265,6 +265,23 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
 				return nil, fmt.Errorf("failed to get LauncherConfig %s: %w", countRule.LauncherConfigName, err)
 			}
 
+			// Validate the PodTemplate once per LauncherConfig, not once per node.
```
MikeSpreitzer, line 273: This is getting closer to once per LauncherConfig per reconcileFromPolicies, but there are still two levels multiplicity possible.
- There can be multiple LauncherPopulationPolicy objects that reference the same LauncherConfig.
- In one LauncherPopulationPolicy object, there can be multiple CountForLauncher structs that reference the same LauncherConfig (I see little motivation for a user to write such a thing, but it is possible).

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Yes, the latest revision responds to my original comment in this thread.

pkg/controller/launcher-populator/populator_status_handler.go
osswangxining: Add lcStatusReported tracks LauncherConfigs whose Status has already been updated

pkg/controller/launcher-populator/populator.go
osswangxining: Refined.  Now  lcStatusReported tracks LauncherConfigs whose Status has already been updated

pkg/controller/launcher-populator/populator.go
```diff
@@ -239,6 +239,11 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
 		return nil, fmt.Errorf("failed to list LauncherPopulationPolicies: %w", err)
 	}
 
+	// lcStatusReported tracks LauncherConfigs whose Status has already been updated this
+	// reconcile cycle. A single LauncherConfig may be referenced by multiple policies or
+	// multiple count rules; we only need one Status update per LC per cycle.
+	lcStatusReported := make(map[string]struct{})
```
MikeSpreitzer, line 245: This is not wrong, but in this project we usually use the abstraction https://github.com/kubernetes/apimachinery/blob/v0.34.7/pkg/util/sets/set.go#L25

pkg/controller/launcher-populator/populator.go
```diff
@@ -265,6 +270,28 @@ func (ctl *controller) buildDesiredStateFromPolicies(ctx context.Context) (map[N
 				return nil, fmt.Errorf("failed to get LauncherConfig %s: %w", countRule.LauncherConfigName, err)
 			}
 
+			// Validate the PodTemplate once per LauncherConfig, not once per node.
```
MikeSpreitzer, line 273: This comment is not quite right. There is a validation being done for each CountPerLauncher, which can happen multiple times for a given LauncherConfig.

MikeSpreitzer (approved):
This is good enough to merge. I left a couple of independent comments about things that could be improved.

PR #477, opened 2026-05-04
Revise test-cases to always check GPU pin; randomize test-requester GPU pick

MikeSpreitzer:
The GPU assignment logic works regardless of OpenShift or not.

Also extends the test-requester's fake GPU allocator to behave more
like the real Kubernetes scheduler + NVIDIA device plugin:

- Pick among unused GPUs at random rather than from the head of the
  sorted list.
- If `NVIDIA_VISIBLE_DEVICES` is set on the requester container,
  restrict the candidate set to those UUIDs (`all` means no
  restriction; `void`/`none` means no GPUs visible and fails
  allocation).

Existing tests pin the picked UUID via `pin_gpu` before scale-up,
so randomizing the initial pick does not affect them.

Fixes #409

### DIFF
--- a/test/e2e/test-cases.sh
+++ b/test/e2e/test-cases.sh
@@ -13,7 +13,6 @@
 #   FMA_CHART_INSTANCE_NAME - Helm release name prefix (default: fma)
 #   READY_TARGET            - minimum ready launchers before proceeding (default: 2)
 #   POLICIES_ENABLED        - "true"/"false"; auto-detected if unset
-#   E2E_PLATFORM            - "openshift" or "kind" (default: openshift)
 #   POLL_LIMIT_SECS         - polling timeout seconds (default: 600)
 #   FMA_DEBUG               - "true" to enable shell tracing (set -x)
 
@@ -28,7 +27,6 @@ fi
 POLL_LIMIT_SECS="${POLL_LIMIT_SECS:-600}"
 READY_TARGET="${READY_TARGET:-2}"
 FMA_CHART_INSTANCE_NAME="${FMA_CHART_INSTANCE_NAME:-fma}"
-E2E_PLATFORM="${E2E_PLATFORM:-openshift}"
 
 NS="$FMA_NAMESPACE"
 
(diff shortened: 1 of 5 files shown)


test/e2e/test-cases.sh
```diff
@@ -226,15 +226,10 @@ date
...
-# the pod is Ready.
-if [ "$E2E_PLATFORM" = "openshift" ]; then
-    expect '[ -n "$(kubectl get pod -n '"$NS"' $req1 -o jsonpath={.metadata.annotations.dual-pods\\.llm-d\\.ai/accelerators})" ]'
-    assigned_gpu_uuids=$(kubectl get pod "$req1" -n "$NS" -o jsonpath='{.metadata.annotations.dual-pods\.llm-d\.ai/accelerators}')
-    echo "Assigned GPU UUID(s) on OpenShift: $assigned_gpu_uuids"
-fi
+# Discover and remember the assigned GPUs.
+expect '[ -n "$(kubectl get pod -n '"$NS"' $req1 -o jsonpath={.metadata.annotations.dual-pods\\.llm-d\\.ai/accelerators})" ]'
+assigned_gpu_uuids=$(kubectl get pod "$req1" -n "$NS" -o jsonpath='{.metadata.annotations.dual-pods\.llm-d\.ai/accelerators}')
+echo "Assigned GPU UUID(s) on OpenShift: $assigned_gpu_uuids"
```
waltforme: This 'on OpenShift' should be removed.

MikeSpreitzer: fixed

MikeSpreitzer:
I turned on debug logging because a local run of the test failed.

MikeSpreitzer:
This is flaky on the test requester. The replacement Pod is created before the dying old Pod is fully gone from the apiserver.

MikeSpreitzer:
The flakiness will be removed if #409 is resolved.

MikeSpreitzer:
Oops, forgot to make the pin unconditional.

MikeSpreitzer:
The [E2E test on OpenShift](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/25356299663) succeeded.

waltforme (approved):
lgtm

Issue #471, opened 2026-04-29
[Feature]: Replace MaxSleepingInstances with MaxInstances


MikeSpreitzer:
### Feature Area

Other

### Problem Statement

MaxSleepingInstances (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/1de2c365a566f57aae9175a9ac341573878ad257/api/fma/v1alpha1/launcherconfig_types.go#L53-L55) is poorly named. We allow 1 more sleeping instance when there is no awake instance, because really they all have the same bound on memory usage.

So `MaxSleepingInstances` is actually `MaxInstances - 1`. The API would be clearer if we simply had "MaxInstances".

### Proposed Solution

Delete MaxSleepingInstances and add MaxInstances.

Do it now, while we have no data migration issue.

However, we _do_ have a CRD migration issue in the shared cluster. So the change will have to be done appropriately.

### Alternatives Considered

Stick with the current API.

### Willingness to Contribute

Yes, I can submit a PR

MikeSpreitzer:
Here is a proposed resolution that I developed with Claude Code (maybe mostly Haiku 4.5? It was mostly in planning mode). This was produced with `main` checked out at a time when that meant commit 2e9fbc141ef7fe62fab42e60048a0b7dc5232c4c .

[smooth-imagining-sutherland.md](https://github.com/user-attachments/files/27423802/smooth-imagining-sutherland.md)

Broken into three staged sub-issues to keep the shared OpenShift test cluster's single CRD compatible with all in-flight PRs throughout the transition (per `docs/cluster-sharing.md`). The staging keeps everything inside `v1alpha1`, avoiding a conversion webhook.

- [ ] #482 — Stage A: additively introduce `maxInstances` alongside `maxSleepingInstances`; CEL rule enforces exactly-one-of.
- [ ] #483 — Stage B: migrate fixtures and docs to `maxInstances`. Do not open until Stage A is pervasive on `main` (every in-flight PR touching the shared cluster has rebased past it).
- [ ] #484 — Stage C: remove `maxSleepingInstances`, drop the CEL rule, mark `maxInstances` required. Do not open until Stage B is pervasive.

MikeSpreitzer:
Here is a `/export` of the conversation that I had with Claude Code to produce that plan and introspect on the process.

[...]

PR #485, opened 2026-05-06
[0.7] Update LauncherPopulationPolicy CRD

osswangxining:
Following #246 , this PR is firstly to update LPP CRD.

I suggest Per-node limit → belongs in LauncherPopulationPolicy.

```
// New fields in LauncherPopulationPolicySpec
type LauncherPopulationPolicySpec struct {
    EnhancedNodeSelector EnhancedNodeSelector `json:"enhancedNodeSelector"`
    CountForLauncher []CountForLauncher `json:"countForLauncher"`
    
    // NEW: total sleeping instances memory limit on nodes matched by this policy
    // +optional
    NodeSleepingMemoryLimit *resource.Quantity `json:"nodeSleepingMemoryLimit,omitempty"`
    
    // Or: total sleeping instances count limit across all launcher pods on the node
    // +optional  
    NodeMaxSleepingInstances *int32 `json:"nodeMaxSleepingInstances,omitempty"`
}
```

Benefits:
Different node types can have different limits (high-memory nodes vs. low-memory nodes)
Semantically consistent with LPP's "node selection + deployment policy" responsibility
launcher-populator can compute node-level budget while building the desired state


MikeSpreitzer:
This PR should also cite Issue #408 , which is the directly relevant issue. #246 was opened for a related but distinct topic and later drifted into this territory.

As I noted in #246, taking on this issue implies, to me, merging the two controllers. That is a big enough change that we should first stop and consider whether we want to do it for release 0.6.0 vs. later.

osswangxining:
> This PR should also cite Issue #408 , which is the directly relevant issue. #246 was opened for a related but distinct topic and later drifted into this territory.
> 
> As I noted in #246, taking on this issue implies, to me, merging the two controllers. That is a big enough change that we should first stop and consider whether we want to do it for release 0.6.0 vs. later.


Sure.  And I think we can break this down into multiple PRs to achieve the goal. This PR is the base step one.

MikeSpreitzer:
I would not want release 0.6 to have an interface that will not be implemented until release 0.7.

osswangxining:
@MikeSpreitzer  Per your request within slack, I rebase this onto main branch, please help review this again, thanks a lot.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
```diff
@@ -80,6 +80,13 @@ type LauncherPopulationPolicySpec struct {
 	// relevant Node, for various LauncherConfigs.
 	// +required
 	CountForLauncher []CountForLauncher `json:"countForLauncher"`
+
+	// NodeSleepingBudget defines the resource budget for sleeping inference
+	// engine instances across all launcher pods on a node matched by this policy.
+	// When multiple policies match the same node, the most restrictive budget
+	// (smallest MaxInstances and MaxMemory) applies.
+	// +optional
+	NodeSleepingBudget *NodeSleepingBudget `json:"nodeSleepingBudget,omitempty"`
```
MikeSpreitzer: Given that `NodeSleepingBudget{}` means "unlimited" --- in other words, no budget --- there is no need for a level of pointer indirection here.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
```diff
@@ -120,6 +127,20 @@ type CountForLauncher struct {
 	LauncherCount int32 `json:"launcherCount"`
 }
 
+// NodeSleepingBudget defines the resource budget for sleeping inference
+// engine instances across all launcher pods on a node.
+type NodeSleepingBudget struct {
+	// MaxInstances limits the total number of sleeping instances across
```
MikeSpreitzer: What is the word "sleeping" doing in here?
The CPU memory usage of an awake vllm can be as great as it was when asleep --- the model tensors are left in CPU memory when doing a wake-up.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
```diff
@@ -120,6 +127,20 @@ type CountForLauncher struct {
...
+	// all launcher pods on the node. Zero means unlimited.
+	// +optional
+	MaxInstances int32 `json:"maxInstances,omitempty"`
+
+	// MaxMemory limits the total accelerator memory used by sleeping
+	// instances across all launcher pods on the node. Zero means unlimited.
+	// +optional
+	MaxMemory *resource.Quantity `json:"maxMemory,omitempty"`
```
MikeSpreitzer: (A) I thought that this PR was intended to address #408 (which got some discussion in #246). The concerns CPU memory, not GPU memory.

(B) There is no point in putting a limit on actual memory usage (either CPU or GPU) without a corresponding ability of the code to read the memory usage. This is currently a problem area, and for that reason I suggest deferring work on limiting what we have no good way to measure yet. (If you look carefully you will find that the dual-pods controller has an option to debug GPU memory usage, but it is based on a sensing technology that is hacky and expensive --- not something that I would recommend llm-d to adopt).

pkg/controller/common/interface.go
```diff
@@ -36,4 +36,11 @@ const (
 	// for the management of vLLM instances.
 	// This is a contract between the controllers and the launcher implementation.
 	LauncherServicePort = 8001
+
+	// NodeSleepingBudgetAnnotationKey is the key of an annotation on a launcher Pod.
+	// The value is a JSON-serialized NodeSleepingBudget that defines the resource
+	// budget for sleeping instances across all launcher pods on the node.
+	// The launcher-populator writes this annotation when creating launcher pods;
+	// the dual-pods controller reads it to enforce node-level sleeper budgets.
+	NodeSleepingBudgetAnnotationKey = "dual-pods.llm-d.ai/node-sleeping-budget"
```
MikeSpreitzer: What is the dual-pods controller to do if different launchers on the same Node have different values for this annotation?

It would be more appropriate to put this annotation on a Node, rather than the launcher Pods. However, a Node is a cluster-scoped resource and so modifying one requires an exceptionally high level of access.

It could work if the annotation goes on launcher Pods and the dual-pods controller combines all of those for a node by MAX, but as an exception treats the empty set as imposing no budget. But I suspect that this would be a more complex solution than combining the controllers.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
osswangxining: You  mean "The zero value means no budget constraint (unlimited)."?

api/fma/v1alpha1/launcherpopulationpolicy_types.go
osswangxining: It's fine and Agreed — dropped MaxMemory. 
MaxInstances (per-pod on LauncherConfig, plus this new per-node NodeInstanceBudget.MaxInstances) serves as a method for CPU memory until we have a reliable way to measure actual memory usage. We can introduce a real MaxMemory budget once memory sensing lands.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
osswangxining: agree. refined

pkg/controller/common/interface.go
osswangxining: Good point.  Since dual-pods doesn't consume this annotation yet, I've removed the NodeInstanceBudgetAnnotationKey tentatively — the NodeInstanceBudget field on the LPP spec stays as the user contract,  until there's an actual consumer to design against.

osswangxining:
@MikeSpreitzer  Please give your further review after I refined based on the previous comments above. Thanks in advance.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
MikeSpreitzer: Yes

api/fma/v1alpha1/launcherpopulationpolicy_types.go
```diff
@@ -80,6 +80,17 @@ type LauncherPopulationPolicySpec struct {
...
 	// +required
 	CountForLauncher []CountForLauncher `json:"countForLauncher"`
+
+	// NodeInstanceBudget defines the budget for inference engine instances
+	// across all launcher pods on a node matched by this policy.
+	// This limits total instances regardless of their state (sleeping or awake),
+	// because an awake instance's CPU memory footprint is the same as when it
+	// was sleeping (model tensors remain resident in CPU memory after wake-up).
+	// When multiple policies match the same node, the most restrictive budget
+	// (smallest MaxInstances) applies.
```
MikeSpreitzer: +1
The identity element for MIN is +infinity, which means that when no policy objects apply to a given node then there is no limit --- which is a desirable result.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
```diff
@@ -120,6 +131,18 @@ type CountForLauncher struct {
...
+// NodeInstanceBudget defines the budget for inference engine instances
+// across all launcher pods on a node. It limits total instances regardless
+// of their state (sleeping or awake), because an awake instance's CPU memory
+// footprint is the same as when it was sleeping (model tensors remain
+// resident in CPU memory after wake-up).
+type NodeInstanceBudget struct {
+	// MaxInstances limits the total number of inference engine instances across
+	// all launcher pods on the node. Zero means unlimited.
+	// +optional
+	MaxInstances int32 `json:"maxInstances,omitempty"`
```
MikeSpreitzer: You might want to consider putting a level of pointer indirection in _here_, so that it is possible to say "no instances allowed" as well as to say "unconstrained".

MikeSpreitzer (commented):
I left independent comments.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
osswangxining: thanks, when no LauncherPopulationPolicy matches a node, the per-node budget should be unconstrained, which falls out naturally as the identity element of MIN (+∞).

api/fma/v1alpha1/launcherpopulationpolicy_types.go
osswangxining: agree. I'll switch MaxInstances to *int32 and related codes.

osswangxining:
@MikeSpreitzer  please help review again, thanks.

The main change is :  Switched MaxInstances from int32 to *int32

MikeSpreitzer (commented):
This revision looks good.

But I recommend holding off merging it until we are ready to do the implementation; it would be confusing to users if there is an unimplemented part of the API.

osswangxining:
> This revision looks good.
> 
> But I recommend holding off merging it until we are ready to do the implementation; it would be confusing to users if there is an unimplemented part of the API.

So will we plan to merge this for next release version - 0.7？

MikeSpreitzer:
> So will we plan to merge this for next release version - 0.7

Yes, I think that's the plan.

osswangxining:
> > So will we plan to merge this for next release version - 0.7
> 
> Yes, I think that's the plan.

Thanks.  I mark this as ready for merge for 0.7 now.

MikeSpreitzer:
@osswangxining: this PR needs to be rebased onto `main`, so that we can graduate from stage B of #471 .

Issue #472, opened 2026-04-30
[Bug]: populator needs to be made incremental


MikeSpreitzer:
### What happened?

The launcher population controller does a full-sweep reconciliation in reaction to any notification. This is way more expensive than necessary. Even while doing reads from informer caches rather than apiservers, this is still a lot of unnecessary work.

### Version

main (please specify commit below)

### Commit SHA

1de2c365a566f57aae9175a9ac341573878ad257

### Relevant log output

```shell

```

osswangxining:
I will update this once #486 is merged. @MikeSpreitzer

Issue #470, opened 2026-04-29
[Bug]: populator does not handle pod, node events


MikeSpreitzer:
### What happened?

While inspecting the code I noticed a bug. The launcher population controller subscribes to notifications about changes to Pod and Node objects but ignores those notifications. The controller should react to those changes. Following are some example scenarios where this matters.

- When the number of launchers is greater than the number called for by the policy objects, and properly so because of a greater number of server-requesting Pods, and then one of those server-requesting Pods gets deleted and consequently one of the launchers gets unbound, this controller should react by deleting that launcher.
- When a GPU-bearing Node is added to the cluster, or its labels change in a way that affects whether it passes a policy's node selector, this controller should react.

### Version

main (please specify commit below)

### Commit SHA

1de2c365a566f57aae9175a9ac341573878ad257

### Relevant log output

```shell

```

MikeSpreitzer:
See https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/469#issuecomment-4349459258 for a plan for resolving both issues together.

osswangxining:
> See [#469 (comment)](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/469#issuecomment-4349459258) for a plan for resolving both issues together.

In. #486 , I add the event handle for both pod and node, does this cover the issue?

osswangxining:
#510  is ready for review to handle this issue.

MikeSpreitzer:
@osswangxining: I think that PR #510 addresses Issue #472 .

**This** Issue, handling Pod and Node events, has been resolved by PR #486 (which also resolved #469).

Issue #469, opened 2026-04-29
[Bug]: populator not accounting for its own actions


MikeSpreitzer:
### What happened?

While reviewing the code I realized a bug. Each invocation of `controller.reconcileLaunchersOnSingleNode` reads a collection of launcher Pods from the pod informer's local cache (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/1de2c365a566f57aae9175a9ac341573878ad257/pkg/controller/launcher-populator/populator.go#L370) --- which might not have been updated yet with all (or any!) of the consequences of the previous invocation for the same Node. This could cause more launcher deletions or creations than is proper.

### Version

main (please specify commit below)

### Commit SHA

1de2c365a566f57aae9175a9ac341573878ad257

### Relevant log output

```shell

```

MikeSpreitzer:
The nature of Kubernetes API machinery makes this impossible, as far as I can tell, to do this correctly using only informers. I think that the query for existing launchers has to be made to an apiserver. This is concerning because it can lead to a _lot_ of calls on the apiservers. Fixing this bug will really ramp up the pressure to make this controller as incremental as possible (currently it is very much _not_).

The last time I looked at the existing Pod set controllers for the `apps` API group built into Kubernetes, they tried to minimize calls on the apiservers by keeping track of a count of creation or delete notifications expected to be received. And this almost always works, but is not 100% guaranteed to be correct. For example, it is possible that soon after the controller creates a Pod, something deletes that Pod. In this case controller's informer is not guaranteed to ever be notified of anything about that Pod.

MikeSpreitzer:
Recent changes in the contract for "ResourceVersion" strings might provide a better way. My previous analysis was done with the previous contract, which forbade clients to treat ResoureVersion strings as anything other than opaque tokens. Now perhaps they can be compared lexicographically, with successive versions required to be increasing. With the controller tracking the ResourceVersion from its latest update (call it U_M), that would enable the controller to know---upon receiving a notification of another update U_N that follows U_M---that its informer has seen U_M.

MikeSpreitzer:
After further consideration, I think that tracking the pending changes with a fallback to direct apiserver query after a timeout of waiting for the expected changes is the best approach. Following is a plan for resolving both this Issue and #470 that I got Claude Code to develop, after some slapping around.

[...]

PR #486, opened 2026-05-06
Feature/add pod node event handler 

### LINKED ISSUE #448: [Bug]: Launcher population controller ignores Pod create/delete
### Contact Details

_No response_

### What happened?

While looking at the code for the launcher population controller, I noticed that it does not react to Pod create or delete events. This is bad, since this controller's job is to maintain the proper population of launcher Pods.

### Version

### Branch name

_No response_

### Commit SHA

f323a8fd498e1a07cff02250d661ca302ceb87bf

### Relevant log output

```shell

```

osswangxining:
Fixes #448 

Currently Launcher population controller ignores Pod create/delete event, this PR is fixing this issue.

### DIFF
--- a/pkg/controller/launcher-populator/pending_expectations.go
+++ b/pkg/controller/launcher-populator/pending_expectations.go
@@ -1,0 +1,165 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package launcherpopulator
+
+import (
+	"sync"
+	"time"
+)
+
+const (
+	// expectationTimeout is how long to wait for the informer cache to reflect
+	// pending mutations before falling back to a direct apiserver query.
+	// This covers the normal watch propagation delay while bounding how long
+	// the controller will defer reconciliation.
+	expectationTimeout = 5 * time.Second
+)
+
+// ExpectationStatus represents the state of expectations for a given key.
+type ExpectationStatus int
+
+const (
+	// ExpectationsSatisfied means no pending mutations remain; the informer
+	// cache is considered up-to-date and safe to read.
+	ExpectationsSatisfied ExpectationStatus = iota
+	// ExpectationsWaiting means pending mutations exist but the timeout has
+	// not yet been reached. The caller should requeue and try again later.
+	ExpectationsWaiting
+	// ExpectationsTimedOut means pending mutations exist and the timeout has
+	// passed. The caller should fall back to querying the apiserver directly.
+	ExpectationsTimedOut
+)
+
+// pendingExpectations tracks Pod create/delete mutations that the controller
+// has performed but whose effects have not yet been observed via informer
+// notifications. This prevents the controller from making incorrect decisions
+// based on stale informer cache state.
+//
+// The typical lifecycle:
+//  1. Controller creates/deletes Pods → calls expectCreations/expectDeletions.
+//  2. Informer fires OnAdd/OnDelete → calls observeCreation/observeDeletion.
+//  3. Next reconcile calls check() to determine if the cache is safe to use.
+//
+// If expected notifications never arrive (e.g., a Pod was created and
+// immediately deleted externally), the timeout ensures the controller
+// eventually falls back to an authoritative apiserver query.
+type pendingExpectations struct {
+	mu      sync.Mutex
+	entries map[NodeLauncherKey]*expectationEntry
+}
+
+type expectationEntry struct {
+	// adds is the number of Pod creations not yet reflected in the cache.
+	adds int
+	// dels is the number of Pod deletions not yet reflected in the cache.
+	dels int
+	// deadline is the wall-clock time after which we consider the expectations
+	// stale and fall back to querying the apiserver directly.
+	deadline time.Time
+}
+
+func newPendingExpectations() *pendingExpectations {
+	return &pendingExpectations{
+		entries: make(map[NodeLauncherKey]*expectationEntry),
+	}
+}
+
+// expectCreations records that `count` Pod creations are pending for the key.
+func (pe *pendingExpectations) expectCreations(key NodeLauncherKey, count int) {
+	pe.mu.Lock()
+	defer pe.mu.Unlock()
+	e := pe.getOrCreate(key)
+	e.adds += count
+	e.deadline = time.Now().Add(expectationTimeout)
+}
+
+// expectDeletions records that `count` Pod deletions are pending for the key.
+func (pe *pendingExpectations) expectDeletions(key NodeLauncherKey, count int) {
+	pe.mu.Lock()
+	defer pe.mu.Unlock()
+	e := pe.getOrCreate(key)
+	e.dels += count
+	e.deadline = time.Now().Add(expectationTimeout)
+}
+
+// observeCreation is called when the informer notifies of a launcher Pod
+// creation for the given key, reducing the pending creation count.
+func (pe *pendingExpectations) observeCreation(key NodeLauncherKey) {
+	pe.mu.Lock()
+	defer pe.mu.Unlock()
+	e, ok := pe.entries[key]
+	if !ok {
+		return
+	}
+	e.adds--
+	pe.cleanupIfSatisfied(key, e)
+}
+
+// observeDeletion is called when the informer notifies of a launcher Pod
+// deletion for the given key, reducing the pending deletion count.
+func (pe *pendingExpectations) observeDeletion(key NodeLauncherKey) {
+	pe.mu.Lock()
+	defer pe.mu.Unlock()
+	e, ok := pe.entries[key]
+	if !ok {
+		return
+	}
+	e.dels--
+	pe.cleanupIfSatisfied(key, e)
+}
+
+// check returns the current status of expectations for the given key.
+func (pe *pendingExpectations) check(key NodeLauncherKey) ExpectationStatus {
+	pe.mu.Lock()
+	defer pe.mu.Unlock()
+	e, ok := pe.entries[key]
+	if !ok {
+		return ExpectationsSatisfied
+	}
+	if e.adds <= 0 && e.dels <= 0 {
+		delete(pe.entries, key)
+		return ExpectationsSatisfied
+	}
+	if time.Now().After(e.deadline) {
+		return ExpectationsTimedOut
+	}
+	return ExpectationsWaiting
+}
+
+// reset clears all expectations for the given key. This is called after
+// falling back to an apiserver query, since the controller now has
+// authoritative state and no longer needs to track pending changes.
+func (pe *pendingExpectations) reset(key NodeLauncherKey) {
+	pe.mu.Lock()
+	defer pe.mu.Unlock()
+	delete(pe.entries, key)
+}
+
+func (pe *pendingExpectations) getOrCreate(key NodeLauncherKey) *expectationEntry {
+	if e, ok := pe.entries[key]; ok {
+		return e
+	}
+	e := &expectationEntry{}
+	pe.entries[key] = e
+	return e
+}
+
+func (pe *pendingExpectations) cleanupIfSatisfied(key NodeLauncherKey, e *expectationEntry) {
+	if e.adds <= 0 && e.dels <= 0 {
+		delete(pe.entries, key)
+	}
+}
--- a/pkg/controller/launcher-populator/pending_expectations.go
+++ b/pkg/controller/launcher-populator/pending_expectations.go
@@ -59,10 +59,20 @@ const (
 //     With UID-based tracking, a "lost" observe is harmless: the expectation
 //     will simply time out and fall back to an authoritative apiserver query.
 //
+// Concurrency model. The controller goroutine and the informer goroutine race:
+// the informer's OnAdd/OnDelete handler can fire before the corresponding
+// Create/Delete API call returns to the controller goroutine. To make
+// expect/observe order-independent, each entry keeps two extra UID sets,
+// earlyCreations and earlyDeletions, which buffer observations that arrived
+// before the matching expectXxx call. expectXxx and observeXxx cancel matching
+// UIDs across these sets, so the final state is the same regardless of which
+// goroutine wins the race.
+//
 // The typical lifecycle:
-//  1. Controller creates/deletes a Pod → calls expectCreation/expectDeletion with the Pod UID.
-//  2. Informer fires OnAdd/OnDelete → calls observeCreation/observeDeletion with the Pod UID.
-//  3. Next reconcile calls check() to determine if the cache is safe to use.
+//  1. Controller creates/deletes a Pod and the informer is notified, in some
+//     order. expectCreation/expectDeletion and observeCreation/observeDeletion
+//     are each called once with the Pod UID; matching pairs cancel out.
+//  2. Next reconcile calls check() to determine if the cache is safe to use.
 //
 // If expected notifications never arrive (e.g., a Pod was created and
 // immediately deleted externally), the timeout ensures the controller
@@ -82,8 +92,19 @@ type expectationEntry struct {
 	// pendingDeletions tracks UIDs of Pods whose deletion has been confirmed
 	// by the apiserver but not yet observed via informer notification.
 	pendingDeletions sets.Set[types.UID]
+	// earlyCreations buffers UIDs whose informer add notification arrived
+	// before the controller had a chance to call expectCreation (the watch
+	// fired while the Create response was still in flight). The next
+	// expectCreation with the same UID cancels out instead of adding to
+	// pendingCreations, so expect/observe order is irrelevant.
+	earlyCreations sets.Set[types.UID]
+	// earlyDeletions buffers UIDs whose informer delete notification arrived
+	// before expectDeletion. Mirrors earlyCreations.
+	earlyDeletions sets.Set[types.UID]
 	// deadline is the wall-clock time after which we consider the expectations
-	// stale and fall back to querying the apiserver directly.
+	// stale and fall back to querying the apiserver directly. The same
+	// deadline also bounds how long early-arrival buffers may sit unused
+	// before being garbage-collected.
 	deadline time.Time
 }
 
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
#448 is part of #470, which interacts strongly with #469. As you see there, I recommend addressing them both together; I put an outline in #469 .

osswangxining:
> #448 is part of #470, which interacts strongly with #469. As you see there, I recommend addressing them both together; I put an outline in #469 .

@MikeSpreitzer  Below is my solution for this issue, please help review again, thanks a lot.

Let me summary the planning solution: Fix informer cache staleness bug with pending expectations tracker

Details:
Introduce a `pendingExpectations` mechanism (inspired by the approach used by Kubernetes built-in ReplicaSet/StatefulSet controllers) that tracks mutations the controller has performed but whose effects have not yet been observed in the informer cache.

The reconcile flow now checks expectations before reading the cache:
- Satisfied: use the informer cache (fast path, zero apiserver cost).
- Waiting: skip this key and request requeue (cache not yet up-to-date).
- Timed out: fall back to a direct apiserver List query and reset expectations (handles edge cases where informer notifications are lost).

Key changes:
- Add `pending_expectations.go` with thread-safe expectation tracking.
- Integrate expectations into OnAdd/OnDelete event handlers to observe (fulfill) pending mutations as informer notifications arrive.
- Refactor `getCurrentLaunchersOnNode` into a three-state dispatcher with `listPodsFromCache` and `listPodsFromApiserver` backends.
- Record expectations in `createLaunchers` (also fixes logging  pod.GenerateName → actual pod name) and both deletion loops.
- Add `expectationsWaiting` flag to the requeue decision logic.

The default timeout is 5 seconds, which covers normal watch propagation delay. In steady state (no pending mutations), the controller uses only the informer cache with zero additional apiserver calls.

pkg/controller/launcher-populator/pending_expectations.go
```diff
@@ -0,0 +1,165 @@
...
+	"sync"
+	"time"
+)
+
+const (
+	// expectationTimeout is how long to wait for the informer cache to reflect
+	// pending mutations before falling back to a direct apiserver query.
+	// This covers the normal watch propagation delay while bounding how long
+	// the controller will defer reconciliation.
+	expectationTimeout = 5 * time.Second
```
MikeSpreitzer: This should be configurable at startup time.
Not a critical problem.

pkg/controller/launcher-populator/populator.go
```diff
@@ -129,8 +135,43 @@ type lcItem struct {
...
+	cache.ObjectName
+}
+
+type nodeItem struct {
+	cache.ObjectName
+}
+
+// isLauncherPod returns true if the Pod is a launcher pod managed by this controller.
+func isLauncherPod(pod *corev1.Pod) bool {
+	return pod.Labels[common.ComponentLabelKey] == common.LauncherComponentLabelValue
```
MikeSpreitzer: This is not wrong, but it would be better if this func examined a label that is protected from mischief.
Right now we have no label that is appears exclusively on launchers (throughout their lifecycle) and is protected from mishief.
I think that a better choice would be to look for the presence of the label `common.LauncherConfigNameLabelKey` and prevent modification of that label's value.

pkg/controller/launcher-populator/populator.go
```diff
@@ -203,6 +270,21 @@ func (item lcItem) process(ctx context.Context, ctl *controller) (error, bool) {
...
+	return ctl.reconcileFromPolicies(ctx)
+}
+
+func (item nodeItem) process(ctx context.Context, ctl *controller) (error, bool) {
+	// Node events trigger a full reconciliation because changes to node labels
+	// or allocatable resources may affect which nodes match policies, altering
+	// the desired launcher population across the cluster.
+	return ctl.reconcileFromPolicies(ctx)
+}
+
```
MikeSpreitzer: As you noted in another PR: since all nodes and pods are treated the same, it is spurious to enqueue unequal items for them. No point in introducing the extra complexity and cost of `nodeItem` and `podItem` here; just enqueue and process `int(0)`.

pkg/controller/launcher-populator/populator.go
```diff
@@ -582,27 +688,83 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
...
 }
 
-// getCurrentLaunchersOnNode returns launcher pods for a specific config on a specific node
-func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLauncherKey) ([]*corev1.Pod, error) {
+// getCurrentLaunchersOnNode returns launcher pods for a specific config on a specific node.
+// It checks pending expectations to decide whether to use the informer cache
+// or fall back to a direct apiserver query. The returned ExpectationStatus
+// indicates which path was taken; ExpectationsWaiting means the caller should
+// requeue without acting on this key.
+func (ctl *controller) getCurrentLaunchersOnNode(ctx context.Context, key NodeLauncherKey) ([]*corev1.Pod, ExpectationStatus, error) {
```
MikeSpreitzer, line 727: FYI, you could save some conversion by making this return `iter.Seq[*corev1.Pod]` rather than `[]*corev1.Pod`. In the `ctl.listPodsFromCache` case, just return https://pkg.go.dev/slices@go1.26.3#Values of the `[]*corev1.Pod`. In `ctl.listPodsFromApiserver`, return an `iter.Seq` that iterates over the indices of the `[]corev1.Pod` and yields the pointer to each `corev1.Pod`.

pkg/controller/launcher-populator/pending_expectations.go
```diff
@@ -0,0 +1,165 @@
...
+type pendingExpectations struct {
+	mu      sync.Mutex
+	entries map[NodeLauncherKey]*expectationEntry
+}
+
+type expectationEntry struct {
+	// adds is the number of Pod creations not yet reflected in the cache.
+	adds int
+	// dels is the number of Pod deletions not yet reflected in the cache.
+	dels int
```
MikeSpreitzer: I think that it is better to track individuals expected rather than count expected. Following are a couple of reasons.

1. Other actors --- including another running copy of this same code --- could also be doing creates and deletes. Tracking only counts could cause expectations to be satisfied while the local cache is still stale, and it would still be wrong to read from the stale cache.
2. Think carefully about concurrency, sequencing, and errors. For example: did you know that sometimes the notification of a write arrives before the reply to the write request? If you track counts, the only correct way to increment the count of expected creations is to make that increment part of the same atomic transaction as actually does the creation. That is, of course, impossible. OTOH, presence or absence of individuals is much safer to expect and evaluate.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/pending_expectations.go
osswangxining: Agree, I make the `expectationTimeout` in the launcher-populator controller configurable at startup time via the `--expectation-timeout` flag, instead of being a hardcoded constant.

## Changes

- **pending_expectations.go**: Rename `expectationTimeout` (unexported const)
  to `DefaultExpectationTimeout` (exported). Add a `timeout` field to
  `pendingExpectations` and accept it in `newPendingExpectations(timeout)`.
- **populator.go**: Add `expectationTimeout time.Duration` parameter to
  `NewController`, pass it through to `newPendingExpectations`.
- **main.go**: Register `--expectation-timeout` pflag with a default of 5s,
  pass the value to `NewController`.
- **pending_expectations_test.go**: Update all test call sites to use the
  new `newPendingExpectations(DefaultExpectationTimeout)` signature.

pkg/controller/launcher-populator/populator.go
osswangxining: Yes, agree to Use launcher-config-name label for launcher Pod identification and updated as below:

Replace the generic `app.kubernetes.io/component` label check in isLauncherPod with the project-specific `dual-pods.llm-d.ai/launcher-config-name`label presence check.  

Additionally, protect the launcher-config-name label from external modification by adding it to the fma-immutable-fields ValidatingAdmissionPolicy.

pkg/controller/launcher-populator/populator.go
osswangxining: Yes,  it should Simplify work queue by replacing per-type items with a single sentinel.

I Replaced the `queueItem` interface and four concrete item types(`lppItem`, `lcItem`, `podItem`, `nodeItem`) with a single `int(0)` sentinel constant enqueued for all event types.

 

## Changes

- Remove `queueItem` interface and all four item types.
- Add `reconcileSentinel = 0` constant; all event handlers enqueue it.
- Change the queue type parameter from `queueItem` to `int`.
- Simplify `process` to directly call `reconcileFromPolicies`.

pkg/controller/launcher-populator/populator.go
osswangxining: Good point — switching to iter.Seq[*corev1.Pod] would avoid the intermediate slice allocation in the apiserver fallback path. Since the practical savings are modest (the apiserver path is a rare fallback, and the caller still categorizes pods into slices), I'll note this as a follow-up optimization. Thanks for the suggestion!

pkg/controller/launcher-populator/pending_expectations.go
osswangxining: Agree,  Fix by tracking individual Pod UIDs in sets.Set[types.UID] instead of
integer counts:
- expectCreation/expectDeletion take a specific UID
- observeCreation/observeDeletion remove a specific UID
- Satisfied when both sets are empty

osswangxining: Replace `expectationEntry.adds/dels` (int counters) with `pendingCreations/pendingDeletions` (`sets.Set[types.UID]`):

| Before | After |
|--------|-------|
| `expectCreations(key, count)` | `expectCreation(key, uid)` |
| `expectDeletions(key, count)` | `expectDeletion(key, uid)` |
| `observeCreation(key)` | `observeCreation(key, uid)` |
| `observeDeletion(key)` | `observeDeletion(key, uid)` |

- **Creations**: Record expectation immediately after each successful `Create` returns with the Pod's UID
- **Deletions**: Record expectation immediately after each successful `Delete` with the Pod's UID
- **Satisfied**: When both UID sets are empty for a given key

osswangxining:
@MikeSpreitzer  I updated following the comments, please give your review again, thanks in advance.

pkg/controller/launcher-populator/populator.go
```diff
@@ -110,35 +113,59 @@ type controller struct {
...
 var _ Controller = &controller{}
 
-type queueItem interface {
-	// process returns (err error, retry bool).
-	// There will be a retry iff `retry`, error logged if `err != nil`.
-	process(ctx context.Context, ctl *controller) (error, bool)
-}
-
-type lppItem struct {
-	cache.ObjectName
```
MikeSpreitzer, line 125: Since we also have #472 on our plate, it is a step in the wrong direction to remove the existing more specific queue item types (I am not concerned about the related inefficiencies in the meantime). My recommendation is to leave the existing ones in place, introduce a simple sentinel for the new cases, and when resolving #472 replace the sentinel with actual object references.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: The deletion expectation should be recorded here (and in the other case like this) too, right?

pkg/controller/launcher-populator/pending_expectations.go
```diff
@@ -0,0 +1,186 @@
...
+//     create or delete pods for the same key. Count-based tracking could be
+//     satisfied by unrelated mutations while our own pods remain invisible.
+//   - Watch notifications can arrive before the API write response returns.
+//     With UID-based tracking, a "lost" observe is harmless: the expectation
+//     will simply time out and fall back to an authoritative apiserver query.
+//
+// The typical lifecycle:
+//  1. Controller creates/deletes a Pod → calls expectCreation/expectDeletion with the Pod UID.
+//  2. Informer fires OnAdd/OnDelete → calls observeCreation/observeDeletion with the Pod UID.
+//  3. Next reconcile calls check() to determine if the cache is safe to use.
```
MikeSpreitzer: There is a concurrency bug here. For example, consider this possibility: controller calls generated stub method to create a Pod, gets success result, then records expectation of new Pod --- but the informer called the Add handler (which is done in a separate goroutine, upon the receipt of an independent message) before the Pod create method returned.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/populator.go
osswangxining: OK.  I follow this to update now.

pkg/controller/launcher-populator/populator.go
osswangxining: Yes, you are right, thanks for your reminder.

pkg/controller/launcher-populator/pending_expectations.go
osswangxining: OK  let me check and fix this.

osswangxining:

I updated based on the comments above, please help review again, thanks a lot! @MikeSpreitzer

pkg/controller/launcher-populator/pending_expectations.go
```diff
@@ -0,0 +1,244 @@
...
+	pendingDeletions sets.Set[types.UID]
+	// earlyCreations buffers UIDs whose informer add notification arrived
+	// before the controller had a chance to call expectCreation (the watch
+	// fired while the Create response was still in flight). The next
+	// expectCreation with the same UID cancels out instead of adding to
+	// pendingCreations, so expect/observe order is irrelevant.
+	earlyCreations sets.Set[types.UID]
+	// earlyDeletions buffers UIDs whose informer delete notification arrived
+	// before expectDeletion. Mirrors earlyCreations.
+	earlyDeletions sets.Set[types.UID]
```
MikeSpreitzer: I think that this is more complex and expensive than necessary. In addition to general complexity, it introduces the problem of another time constant (how long to wait from a potentially early observation to decide that it was immaterial?) and requires one time constant variable to serve two purposes. Remember that the informer's local caches already store the state of what has been added and removed. I think that it would be simpler to remove the `observeXXX` methods and instead have `check` enumerate the remaining expectations and prune ones that are now met.

pkg/controller/launcher-populator/populator.go
```diff
@@ -147,14 +202,22 @@ func (ctl *controller) OnAdd(obj any, isInInitialList bool) {
 
 func (ctl *controller) OnUpdate(prev, obj any) {
 	switch typed := obj.(type) {
+	case *corev1.Pod:
+		if !isLauncherPod(typed) {
+			ctl.enqueueLogger.V(5).Info("Ignored update of non-launcher Pod", "name", typed.Name)
+			return
+		}
+		ctl.enqueueLogger.V(5).Info("Enqueuing reconciliation due to launcher Pod update", "name", typed.Name)
```
MikeSpreitzer: FYI: when investigating concurrency bugs, I appreciate log messages that include the ResouceVersion(s) involved.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/pending_expectations.go
osswangxining: Update as suggested and check now enumerates remaining expectations and prunes them based on the informer cache's current state.  Please help check again, thanks.

pkg/controller/launcher-populator/populator.go
osswangxining: it's fine to add this for debugging

pkg/controller/launcher-populator/populator.go
```diff
@@ -600,22 +715,89 @@ func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeN
...
+	pods, err := ctl.listPodsFromCache(key)
+	if err != nil {
+		return nil, ExpectationsSatisfied, err
+	}
+	presentUIDs := sets.New[types.UID]()
+	for _, p := range pods {
+		presentUIDs.Insert(p.UID)
+	}
+
+	status := ctl.expectations.check(key, presentUIDs)
```
MikeSpreitzer, line 739: A Pod object does not change its name either.
We could avoid the expense of this listing and conversion by making each expectation record both name and UID.

pkg/controller/launcher-populator/pending_expectations.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func newPendingExpectations(timeout time.Duration) *pendingExpectations {
+	return &pendingExpectations{
+		entries: make(map[NodeLauncherKey]*expectationEntry),
+		timeout: timeout,
+	}
+}
+
+// expectCreation records that a Pod creation (identified by UID) is pending
```
MikeSpreitzer, line 79: The creation itself is not pending, that already happened by the time this call is made (as stipulated in the next sentence). What (might or might not!) be pending is the corresponding update in the Pod informer's local cache.

pkg/controller/launcher-populator/populator.go
```diff
@@ -147,14 +197,28 @@ func (ctl *controller) OnAdd(obj any, isInInitialList bool) {
...
+		prevRV := ""
+		if prevPod, ok := prev.(*corev1.Pod); ok {
+			prevRV = prevPod.ResourceVersion
+		}
+		ctl.enqueueLogger.V(5).Info("Enqueuing reconciliation due to launcher Pod update",
+			"name", typed.Name, "uid", typed.UID,
+			"prevResourceVersion", prevRV, "resourceVersion", typed.ResourceVersion)
+		ctl.Queue.Add(reconcileAllSentinel{})
+	case *corev1.Node:
+		ctl.enqueueLogger.V(5).Info("Enqueuing reconciliation due to Node update", "name", typed.Name)
```
MikeSpreitzer, line 214: Including the ResourceVersion in these messages is good for all kinds of objects.
Not a critical problem (yet).

pkg/controller/launcher-populator/pending_expectations_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// uidSet is a small helper to build the presentUIDs argument tersely.
+func uidSet(uids ...types.UID) sets.Set[types.UID] {
+	return sets.New[types.UID](uids...)
+}
+
+func TestPendingExpectations_BasicCreation(t *testing.T) {
+	pe := newPendingExpectations(DefaultExpectationTimeout)
```
MikeSpreitzer, line 33: Experience has shown that testing machines are sometimes overloaded and run surprisingly slow. Maybe a longer timeout here would be helpful to avoid flakes.

To make time-related tests non-flaky and run fast without undue intimacy with the code being tested, the Kubernetes project defined a variant sort of clock that gives unit tests precise control over time. The idea is that the non-test code is written to use a `PassiveClock` (https://github.com/kubernetes/utils/blob/master/clock/clock.go#L24-L27) rather than a real clock. The normal constructor creates an implementation of `PassiveClock` (https://github.com/kubernetes/utils/blob/master/clock/clock.go#L85) that delegates to the normal Go clock. But unit tests instead create and use an implementation (https://github.com/kubernetes/utils/blob/master/clock/testing/fake_clock.go#L32-L36) that gives the test case explicit precise control over time.

pkg/controller/launcher-populator/pending_expectations_test.go
```diff
@@ -0,0 +1,238 @@
...
+func TestPendingExpectations_Timeout(t *testing.T) {
+	pe := newPendingExpectations(DefaultExpectationTimeout)
+	key := NodeLauncherKey{NodeName: "node-1", LauncherConfigName: "lc-1"}
+
+	pe.expectCreation(key, types.UID("uid-timeout"))
+
+	// Manually set the deadline to the past to simulate timeout.
+	pe.mu.Lock()
+	pe.entries[key].deadline = time.Now().Add(-1 * time.Second)
+	pe.mu.Unlock()
```
MikeSpreitzer, line 106: See my comment above for a better way to test timeout.

pkg/controller/launcher-populator/pending_expectations_test.go
```diff
@@ -0,0 +1,238 @@
...
+	if status := pe.check(key, uidSet()); status != ExpectationsSatisfied {
+		t.Errorf("expected ExpectationsSatisfied after reset, got %v", status)
+	}
+}
+
+// TestPendingExpectations_ExpectAfterCacheAlreadyCaughtUp covers the race
+// where the informer cache reflects the new Pod before the controller calls
+// expectCreation (the watch fired while the Create response was still in
+// flight). The next check immediately reconciles the expectation against
+// the cache and reports Satisfied, with no extra bookkeeping needed.
```
MikeSpreitzer, line 134: This comment looks outdated.
It is no longer a good description of this test case.
The scenario tested here is already covered by TestPendingExpectations_BasicCreation.

pkg/controller/launcher-populator/pending_expectations_test.go
```diff
@@ -0,0 +1,238 @@
...
+	key := NodeLauncherKey{NodeName: "node-1", LauncherConfigName: "lc-1"}
+	uid := types.UID("uid-race-c")
+
+	// Controller records the expectation after the cache already shows it.
+	pe.expectCreation(key, uid)
+
+	if status := pe.check(key, uidSet(uid)); status != ExpectationsSatisfied {
+		t.Errorf("expected ExpectationsSatisfied when cache already reflects the UID, got %v", status)
+	}
+	pe.mu.Lock()
```
MikeSpreitzer, line 146: FYI, the locking is not needed, since it is statically known that there is no concurrent access.

pkg/controller/launcher-populator/pending_expectations_test.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+}
+
+// TestPendingExpectations_CheckWithoutExpectations confirms check() is a
+// no-op when nothing is pending, regardless of what the cache contains.
+func TestPendingExpectations_CheckWithoutExpectations(t *testing.T) {
+	pe := newPendingExpectations(DefaultExpectationTimeout)
+	key := NodeLauncherKey{NodeName: "node-1", LauncherConfigName: "lc-1"}
+
+	if status := pe.check(key, uidSet("uid-foreign")); status != ExpectationsSatisfied {
```
MikeSpreitzer, line 229: This same scenario is tested at the beginning of TestPendingExpectations_BasicCreation

MikeSpreitzer (approved):
I do not see bugs, so I think that this is good enough to merge.
This could be improved further, in this PR or follow-on(s).

Issue #484, opened 2026-05-06
[Feature]: (#471 Stage C) Remove maxSleepingInstances from LauncherConfig v1alpha1


MikeSpreitzer:
Sub-issue of #471 — Stage C of the staged rename of `MaxSleepingInstances` → `MaxInstances`. Blocked on #483 (Stage B).

## Goal

Delete `MaxSleepingInstances` from the `v1alpha1` API, drop the transitional CEL rule, and make `MaxInstances` required.

## Scope

- `api/fma/v1alpha1/launcherconfig_types.go`
  - Delete the `MaxSleepingInstances` field.
  - Drop the `+kubebuilder:validation:XValidation` "exactly one of" rule added in Stage A.
  - On `MaxInstances`: replace `+optional` with `+kubebuilder:validation:Required` and add `+kubebuilder:validation:Minimum=1`. Update the doc comment to drop the transition language. The Go type stays `int32`; `required` is a schema concern.

- `pkg/controller/dual-pods/inference-server.go`
  - Delete `effectiveMaxInstances`; both callsites become `int(lc.Spec.MaxInstances) - 1`.

- Regenerate: `make manifests generate generate_client`. The resulting CRD lists only `maxInstances` with `minimum: 1` and `required: [maxInstances]`, and no CEL rule.

## Timing

Do not even open the Stage C PR until Stage B (#483) is pervasive: it must be merged to `main` **and** every other PR that touches the shared test cluster must have rebased past it, so that no in-flight commit still sets `maxSleepingInstances` in any fixture or doc. Because the CRD in Stage B still lists `maxSleepingInstances`, old fixtures continue to work during this waiting period.

## Verification

- `make manifests generate generate_client`; `go build ./...`; `go test ./...`.
- E2E recipe passes with only `maxInstances` set.
- On a `kind` cluster, applying a LauncherConfig with `maxSleepingInstances` is rejected under strict decoding (or silently dropped by non-strict clients), since the CRD schema no longer lists it.
- Applying a LauncherConfig that omits `maxInstances` (or sets it to 0) is rejected by `required` / `minimum: 1`.

Issue #483, opened 2026-05-06
[Feature]: (#471 Stage B) Migrate fixtures and docs from maxSleepingInstances to maxInstances


MikeSpreitzer:
Sub-issue of #471 — Stage B of the staged rename of `MaxSleepingInstances` → `MaxInstances`. Blocked on #482 (Stage A).

## Goal

Migrate all in-repo YAML and docs that currently set `maxSleepingInstances` to use the new `maxInstances` field. This stage has no Go or CRD changes.

## Scope

At the time of the PR, `grep -rn maxSleepingInstances` should reveal only the v1alpha1 type declaration (kept until Stage C) and the generated code. Currently the fixtures/docs to migrate are:

- `test/e2e/mkobjs.sh:116` — `maxSleepingInstances: 1` → `maxInstances: 2`.
- `test/e2e/mkobjs-openshift.sh:173` — `maxSleepingInstances: 3` → `maxInstances: 4`.
- `docs/e2e-recipe.md:545` — `maxSleepingInstances: 3` → `maxInstances: 4`.

Re-run the grep before opening the PR and fold in anything new.

## Out of scope

- Removing `maxSleepingInstances` from the API (Stage C — separate sub-issue).

## Timing

Do not even open the Stage B PR until Stage A (#482) is pervasive: it must be merged to `main` **and** every other PR that touches the shared test cluster must have rebased past it, so that no in-flight commit still relies on `maxSleepingInstances` being the required field. Because Stage A is additive, the shared-cluster CRD will continue to admit `maxSleepingInstances` throughout this waiting period.

## Verification

- Run the updated `docs/e2e-recipe.md` and `test/e2e/mkobjs*.sh` on `kind` and the shared OpenShift cluster. Confirm instance counts match previous behavior (`mkobjs.sh`: up to 2 total instances; `mkobjs-openshift.sh`: up to 4).

PR #487, opened 2026-05-07
Add maxInstances alongside maxSleepingInstances in LauncherConfig v1alpha1 (#482)

### LINKED ISSUE #482: [Feature]: (#471 Stage A) Add maxInstances alongside maxSleepingInstances in LauncherConfig v1alpha1
Sub-issue of #471 — Stage A of the staged rename of `MaxSleepingInstances` → `MaxInstances`.

## Goal

Additively introduce `maxInstances` in `LauncherConfigSpec` without breaking any in-flight PR or the shared OpenShift test cluster. This stage is purely additive to `v1alpha1` and compatible with existing YAML that sets `maxSleepingInstances` (including `maxSleepingInstances: 0`).

## Scope

- `api/fma/v1alpha1/launcherconfig_types.go`
  - Add `MaxInstances int32` with JSON tag `json:"maxInstances,omitempty"` and kubebuilder markers `+optional` and `+kubebuilder:validation:Minimum=0`. Zero encodes "unset".
  - Remove `+kubebuilder:validation:Required` from `MaxSleepingInstances`; add `+optional` and `+kubebuilder:validation:Minimum=0` (negatives were never meaningful; the existing schema failed to forbid them). Rewrite the doc comment to state the observable semantics without naming an enforcing component:

    > MaxSleepingInstances caps the number of sleeping inference-engine instances per launcher pod when an awake instance is also present. When no instance is awake, one additional sleeping instance is permitted — i.e., a launcher pod can hold up to `MaxSleepingInstances + 1` instances in total. Deprecated: use MaxInstances, which simply caps the total number of instances per launcher pod.

  - Add a CRD-level CEL rule on `LauncherConfigSpec` that forbids only the genuinely ambiguous case — both fields positive — using only `has()` and integer comparison. This form is supported in the CRD CEL environment from its GA in Kubernetes 1.25, so there is no apiserver-version risk. Using `has()` here is acceptable because the rule is transient (removed in Stage C):
    ```
    +kubebuilder:validation:XValidation:rule="!(has(self.maxInstances) && has(self.maxSleepingInstances) && self.maxInstances > 0 && self.maxSleepingInstances > 0)",message="maxInstances and maxSleepingInstances must not both be positive"
    ```

- `pkg/controller/dual-pods/inference-server.go`
  - Add a helper `effectiveMaxInstances(lc) int` returning `int(lc.Spec.MaxInstances)` when non-zero, else `int(lc.Spec.MaxSleepingInstances) + 1`.
  - Update the two existing callsites (around the current lines 629 and 651) to pass `effectiveMaxInstances(lc) - 1`. Numeric behavior is unchanged when only `maxSleepingInstances` is set.

- Regenerate: `make manifests generate generate_client`.

- Unit tests for `effectiveMaxInstances`: only-`MaxInstances`, only-`MaxSleepingInstances` (including zero), both-zero (degenerate; returns 1), both-set defensive case (picks `MaxInstances`).

## Semantics of the rule

- `{maxSleepingInstances: N}` for any `N ≥ 0` → admitted; controller uses `N + 1`. Preserves every currently-valid client, including `maxSleepingInstances: 0`.
- `{maxInstances: M}` for any `M ≥ 1` → admitted; controller uses `M`.
- `{}` or both-zero → admitted; controller degenerates to `effectiveMaxInstances = 1` (one instance only). Small permissiveness regression from today's `required` schema, but harmless and tightened in Stage C.
- Both positive → rejected by CEL.
- Negative in either field → rejected by the schema's `minimum: 0`.

## Out of scope

- Migrating fixtures/docs to `maxInstances` (that's Stage B — #483).
- Removing `maxSleepingInstances` (that's Stage C — #484).

## Context

See #471 for motivation. The staging approach avoids introducing `v1alpha2` (and thus a conversion webhook) by keeping the change inside `v1alpha1` per `docs/cluster-sharing.md`. `MaxSleepingInstances` is misnamed: the true semantics is "max sleeping instances when there is also an awake instance", so the total cap per launcher is `MaxSleepingInstances + 1`. `MaxInstances` renames that cap to match reality.

## Verification

- `make manifests generate generate_client` is clean.
- `go build ./...`; `go test ./...`.
- On a `kind` cluster, `kubectl apply` the new CRD and confirm the bullets in the "Semantics of the rule" section above. (A `kind` test is not a proxy for the shared OpenShift cluster; rely on the GHA E2E workflow's conditional `kubectl apply` against the shared cluster at PR time for that assurance.)
- Existing E2E recipe (`docs/e2e-recipe.md`) still works unchanged with `maxSleepingInstances`.

MikeSpreitzer:
## Summary

Stage A of the staged rename of `MaxSleepingInstances` → `MaxInstances` (tracked by #471). Purely additive within `v1alpha1`; no conversion webhook, no break for in-flight PRs or the shared OpenShift test cluster.

- `LauncherConfigSpec`: add `MaxInstances` (optional, `minimum: 0`). Relax `MaxSleepingInstances` to optional with `minimum: 0` and rewrite its doc to describe the observable `+1` semantics. Deprecate `MaxSleepingInstances` in favor of `MaxInstances`.
- Add a CRD-level CEL rule forbidding only the ambiguous case where both fields are positive. A single English statement of the mutual exclusion lives on the type's doc comment.
- `pkg/controller/dual-pods/inference-server.go`: add `effectiveMaxInstances(lc)` returning `MaxInstances` when positive, else `MaxSleepingInstances + 1`. Both existing callsites now pass `effectiveMaxInstances(lc) - 1`. Numeric behavior is unchanged when only `maxSleepingInstances` is set.
- Regenerated CRD manifest and applyconfiguration via `make manifests generate generate_client`.

Out of scope (later stages): migrating fixtures/docs to `maxInstances` (#483), removing `maxSleepingInstances` (#484).

Resolves #482.

## Test plan

- [x] `make manifests generate generate_client` is clean
- [x] `go build ./...`
- [x] `go test ./...`
- [x] On a `kind` cluster, `kubectl apply` the new CRD and verify:
  - [x] `{maxSleepingInstances: N}` for `N ≥ 0` admitted
  - [x] `{maxInstances: M}` for `M ≥ 1` admitted
  - [x] `{}` admitted
  - [x] both positive → rejected by CEL
  - [x] negative in either field → rejected by `minimum: 0`
- [ ] Existing E2E recipe (`docs/e2e-recipe.md`) still works unchanged with `maxSleepingInstances`

### DIFF
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -626,7 +626,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		// then those with capacity for new instances.
 		// Note that multiple vLLM instances could exist in one launcher Pod, but at most one instance could be awake at a time.
 
-		launcherPod, hasSleepingInstance, someNotReady, err := ctl.selectBestLauncherPod(ctx, launcherPodAnys, iscHash, desiredPort, int(lc.Spec.MaxSleepingInstances), nodeDat)
+		launcherPod, hasSleepingInstance, someNotReady, err := ctl.selectBestLauncherPod(ctx, launcherPodAnys, iscHash, desiredPort, effectiveMaxInstances(lc)-1, nodeDat)
 		if err != nil {
 			return err, true
 		}
@@ -648,7 +648,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	// Remains: Zero matching launcher Pods, or the matching launcher Pod cannot host more instances to fulfill the request.
 
 	// TODO(waltforme): enforceSleeperBudget should be revised for launcher-based server-providing Pods
-	err, retry := ctl.enforceSleeperBudget(ctx, serverDat, requestingPod, int(lc.Spec.MaxSleepingInstances))
+	err, retry := ctl.enforceSleeperBudget(ctx, serverDat, requestingPod, effectiveMaxInstances(lc)-1)
 	if err != nil || retry {
 		return err, retry
 	}
@@ -794,6 +794,16 @@ func (ctl *controller) selectBestLauncherPod(
 	// No suitable launchers found
 	logger.V(4).Info("No suitable launcher Pod found with sleeping instance or necessary capacity")
 	return nil, false, false, nil
+}
+
+// effectiveMaxInstances returns the cap on the total number of inference-engine instances
+// per launcher pod. If MaxInstances is positive, it is used directly; otherwise the legacy
+// MaxSleepingInstances semantics apply and the cap is MaxSleepingInstances + 1.
+func effectiveMaxInstances(lc *fmav1alpha1.LauncherConfig) int {
+	if lc.Spec.MaxInstances > 0 {
+		return int(lc.Spec.MaxInstances)
+	}
+	return int(lc.Spec.MaxSleepingInstances) + 1
 }
 
 // configInferenceServer computes the VllmConfig.
--- a/api/fma/v1alpha1/launcherconfig_types.go
+++ b/api/fma/v1alpha1/launcherconfig_types.go
@@ -45,14 +45,28 @@ type EmbeddedPodTemplateSpec struct {
 }
 
 // LauncherConfigSpec defines the configuration to manage the nominal server-providing pod definition.
+// At most one of `maxSleepingInstances` and `maxInstances` may be positive; setting both to positive
+// values is rejected. Either may be zero or omitted.
+// +kubebuilder:validation:XValidation:rule="!(has(self.maxInstances) && has(self.maxSleepingInstances) && self.maxInstances > 0 && self.maxSleepingInstances > 0)",message="maxInstances and maxSleepingInstances must not both be positive"
 type LauncherConfigSpec struct {
 	// PodTemplate defines the pod specification for the server-providing pod.
 	// +optional
 	PodTemplate EmbeddedPodTemplateSpec `json:"podTemplate,omitempty"`
 
-	// MaxSleepingInstances is the maximum number of sleeping inference engine instances allowed per launcher pod.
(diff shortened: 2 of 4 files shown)


pkg/controller/dual-pods/inference-server.go
```diff
@@ -796,6 +796,16 @@ func (ctl *controller) selectBestLauncherPod(
...
 }
 
+// effectiveMaxInstances returns the cap on the total number of inference-engine instances
+// per launcher pod. If MaxInstances is positive, it is used directly; otherwise the legacy
+// MaxSleepingInstances semantics apply and the cap is MaxSleepingInstances + 1.
+func effectiveMaxInstances(lc *fmav1alpha1.LauncherConfig) int {
+	if lc.Spec.MaxInstances > 0 {
+		return int(lc.Spec.MaxInstances)
+	}
+	return int(lc.Spec.MaxSleepingInstances) + 1
```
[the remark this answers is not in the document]
MikeSpreitzer: This repo does not support migrating existing API objects from one version of the repo to another.

osswangxining, line 806: Can you help me recall  the meaning  about  "MaxSleepingInstances + 1" instead of direct "MaxSleepingInstances"?

osswangxining (approved):
No other comments except for : 
Can you help me recall the meaning about "MaxSleepingInstances + 1" instead of direct "MaxSleepingInstances"?

pkg/controller/dual-pods/inference-server.go
aavarghese: The +1 is for the one instance that can be awake at a time.  So 1 (awake) + MaxSleepingInstances (sleeping) is the total instances

aavarghese (approved):

aavarghese:
@rubambiza leaving reminder for us to update benchmarking to use new field..

PR #492, opened 2026-05-07
Consistent ISC usage during the life of a vLLM instance

### LINKED ISSUE #475: [Bug]: dual-pods controller not coping with changes to InferenceServerConfig
### Contact Details

_No response_

### What happened?

While reviewing #473, I noticed a bug.

This controller is not properly accounting for the fact that an InferenceServerConfig object can be mutated at any time and in any way. In particular, the controller is mistakenly taking the current server port from that object even for a vllm instance that was created from an earlier version of the InferenceServerConfig in which the port was different. Similarly: apart from the two copies of the recovery code, serverDat.ISCLabelKeys and serverDat.ISCAnnotationKeys are set in bind, which can happen before the vllm instance is created. In fact, in case of controller crash and recovery, even the instance ID can be wrong.

On the down-swing, only the port, label+annotation keys, and instance ID come from the InferenceServerConfig object --- and they need to come from the version used to create the vllm instance. On the upswing: the whole InferenceServerConfigSpec and instance ID need to all come from the same version of the InferenceServerConfig object.

### Version

### Branch name

_No response_

### Commit SHA

b771f873718f78bb3b1f0fac67e10d99f8272d45

### Relevant log output

```shell

```

waltforme:
This PR fixes #475 (which was also raised in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/473#discussion_r3175276220).

### DIFF
--- a/api/fma/v1alpha1/inferenceserverconfig_types.go
+++ b/api/fma/v1alpha1/inferenceserverconfig_types.go
@@ -35,6 +35,8 @@ type InferenceServerConfigSpec struct {
 type ModelServerConfig struct {
 	// Port is the port on which the vLLM server will listen
 	// Particularly, management of vLLM instances' sleep state is done through this port
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=32766
 	Port int32 `json:"port"`
 
 	// Options are the vLLM startup options, excluding Port
--- a/api/fma/v1alpha1/inferenceserverconfig_types.go
+++ b/api/fma/v1alpha1/inferenceserverconfig_types.go
@@ -36,7 +36,7 @@ type ModelServerConfig struct {
 	// Port is the port on which the vLLM server will listen
 	// Particularly, management of vLLM instances' sleep state is done through this port
 	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=32766
+	// +kubebuilder:validation:Maximum=65535
 	Port int32 `json:"port"`
 
 	// Options are the vLLM startup options, excluding Port
--- a/charts/fma-controllers/templates/rbac/role.yaml
+++ b/charts/fma-controllers/templates/rbac/role.yaml
@@ -15,3 +15,6 @@ rules:
 - apiGroups: [ "" ]
   resources: [ pods ]
   verbs: [ get, list, watch, create, update, patch, delete ]
+- apiGroups: [ "" ]
+  resources: [ events ]
+  verbs: [ create, patch ]
--- a/charts/fma-controllers/templates/rbac/role.yaml
+++ b/charts/fma-controllers/templates/rbac/role.yaml
@@ -15,6 +15,3 @@ rules:
 - apiGroups: [ "" ]
   resources: [ pods ]
   verbs: [ get, list, watch, create, update, patch, delete ]
-- apiGroups: [ "" ]
-  resources: [ events ]
-  verbs: [ create, patch ]
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -308,7 +308,10 @@ type serverData struct {
 	GPUIndicesStr *string
 
 	ProvidingPodName string
-	// Snapshot of the bound launcher-based vLLM instance. Not mutated after being set.
+	// The next five fields form a snapshot of the bound launcher-based
+	// vLLM instance's ISC-derived state. Each field may be reassigned
+	// over time (e.g. on rebind), but whichever value (pointer, slice)
+	// is currently stored is deeply immutable for as long as it is stored.
 	InstanceID           string
 	InstanceConfig       *VllmConfig
 	InstanceKnownToExist bool     // meaningful only for launcher-based providers
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -36,9 +36,11 @@ import (
 	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	corev1preinformers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
 	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 
 	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
@@ -167,9 +169,12 @@ func (config ControllerConfig) NewController(
 	corev1PreInformers corev1preinformers.Interface,
 	fmaInformerFactory fmainformers.SharedInformerFactory,
 ) (*controller, error) {
+	eventBroadcaster := record.NewBroadcaster()
 	ctl := &controller{
 		enqueueLogger:       logger.WithName(ControllerName),
 		coreclient:          coreClient,
+		eventBroadcaster:    eventBroadcaster,
+		eventRecorder:       eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: ControllerName}),
 		namespace:           namespace,
 		podInformer:         corev1PreInformers.Pods().Informer(),
 		podLister:           corev1PreInformers.Pods().Lister(),
@@ -232,9 +237,11 @@ func (config ControllerConfig) NewController(
 }
 
 type controller struct {
-	enqueueLogger klog.Logger
-	coreclient    coreclient.CoreV1Interface
-	namespace     string
+	enqueueLogger    klog.Logger
+	coreclient       coreclient.CoreV1Interface
+	eventBroadcaster record.EventBroadcaster
+	eventRecorder    record.EventRecorder
+	namespace        string
 	podInformer   cache.SharedIndexInformer
 	podLister     corev1listers.PodLister
 	cmInformer    cache.SharedIndexInformer
@@ -618,6 +625,12 @@ func (ctl *controller) Start(ctx context.Context) error {
 	if !cache.WaitForNamedCacheSync(ControllerName, ctx.Done(), ctl.cmInformer.HasSynced, ctl.podInformer.HasSynced, ctl.nodeInformer.HasSynced, ctl.iscInformer.HasSynced, ctl.lcInformer.HasSynced) {
 		return fmt.Errorf("caches not synced before end of Start context")
 	}
+	ctl.eventBroadcaster.StartStructuredLogging(0)
+	ctl.eventBroadcaster.StartRecordingToSink(&coreclient.EventSinkImpl{Interface: ctl.coreclient.Events("")})
+	go func() {
+		<-ctx.Done()
+		ctl.eventBroadcaster.Shutdown()
+	}()
 	err := ctl.StartWorkers(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to start workers: %w", err)
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -36,11 +36,9 @@ import (
 	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	corev1preinformers "k8s.io/client-go/informers/core/v1"
-	"k8s.io/client-go/kubernetes/scheme"
 	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 
 	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
@@ -169,12 +167,9 @@ func (config ControllerConfig) NewController(
 	corev1PreInformers corev1preinformers.Interface,
 	fmaInformerFactory fmainformers.SharedInformerFactory,
 ) (*controller, error) {
-	eventBroadcaster := record.NewBroadcaster()
 	ctl := &controller{
 		enqueueLogger:       logger.WithName(ControllerName),
 		coreclient:          coreClient,
-		eventBroadcaster:    eventBroadcaster,
-		eventRecorder:       eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: ControllerName}),
 		namespace:           namespace,
 		podInformer:         corev1PreInformers.Pods().Informer(),
 		podLister:           corev1PreInformers.Pods().Lister(),
@@ -237,11 +232,9 @@ func (config ControllerConfig) NewController(
 }
 
 type controller struct {
-	enqueueLogger    klog.Logger
-	coreclient       coreclient.CoreV1Interface
-	eventBroadcaster record.EventBroadcaster
-	eventRecorder    record.EventRecorder
-	namespace        string
+	enqueueLogger klog.Logger
+	coreclient    coreclient.CoreV1Interface
+	namespace     string
 	podInformer      cache.SharedIndexInformer
 	podLister        corev1listers.PodLister
 	cmInformer       cache.SharedIndexInformer
@@ -626,12 +619,6 @@ func (ctl *controller) Start(ctx context.Context) error {
 	if !cache.WaitForNamedCacheSync(ControllerName, ctx.Done(), ctl.cmInformer.HasSynced, ctl.podInformer.HasSynced, ctl.nodeInformer.HasSynced, ctl.iscInformer.HasSynced, ctl.lcInformer.HasSynced) {
 		return fmt.Errorf("caches not synced before end of Start context")
 	}
-	ctl.eventBroadcaster.StartStructuredLogging(0)
-	ctl.eventBroadcaster.StartRecordingToSink(&coreclient.EventSinkImpl{Interface: ctl.coreclient.Events("")})
-	go func() {
-		<-ctx.Done()
-		ctl.eventBroadcaster.Shutdown()
-	}()
 	err := ctl.StartWorkers(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to start workers: %w", err)
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -242,16 +242,16 @@ type controller struct {
 	eventBroadcaster record.EventBroadcaster
 	eventRecorder    record.EventRecorder
 	namespace        string
-	podInformer   cache.SharedIndexInformer
-	podLister     corev1listers.PodLister
-	cmInformer    cache.SharedIndexInformer
-	cmLister      corev1listers.ConfigMapLister
-	nodeInformer  cache.SharedIndexInformer
-	nodeLister    corev1listers.NodeLister
-	iscInformer   cache.SharedIndexInformer
-	iscLister     fmalisters.InferenceServerConfigLister
-	lcInformer    cache.SharedIndexInformer
-	lcLister      fmalisters.LauncherConfigLister
+	podInformer      cache.SharedIndexInformer
+	podLister        corev1listers.PodLister
+	cmInformer       cache.SharedIndexInformer
+	cmLister         corev1listers.ConfigMapLister
+	nodeInformer     cache.SharedIndexInformer
+	nodeLister       corev1listers.NodeLister
+	iscInformer      cache.SharedIndexInformer
+	iscLister        fmalisters.InferenceServerConfigLister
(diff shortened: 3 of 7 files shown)


pkg/controller/dual-pods/inference-server.go
```diff
@@ -466,19 +450,19 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				if err != nil {
 					return err, true
 				}
-				result, err := lClient.CreateNamedInstance(ctx, serverDat.InstanceID, *cfg)
+				if serverDat.InstanceConfig == nil {
+					return ctl.ensureReqStatus(ctx, requestingPod, serverDat,
+						fmt.Sprintf("unable to create vLLM instance %q because bound launcher Pod %q is missing annotation %q",
```
MikeSpreitzer: This is an odd way to handle this situation. The problem here is not related to the responsibilities of the author of the serve-requesting Pod.

Kubernetes controllers can publish Kubernetes `Event` objects.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -829,6 +819,133 @@ func (ctl *controller) configInferenceServer(isc *fmav1alpha1.InferenceServerCon
...
 
+func (ctl *controller) computeDesiredInstanceState(isc *fmav1alpha1.InferenceServerConfig, gpuUUIDs []string) (*vllmInstanceState, error) {
+	cfg, instanceID, err := ctl.configInferenceServer(isc, gpuUUIDs)
+	if err != nil {
+		return nil, err
+	}
+	return &vllmInstanceState{
+		cfg:            cfg,
+		instanceID:     instanceID,
+		serverPort:     int16(isc.Spec.ModelServerConfig.Port),
```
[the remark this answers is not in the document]
MikeSpreitzer: IMO it would be better to update the IDL to require the port to be in the range [1, 32766].

pkg/controller/dual-pods/inference-server.go
```diff
@@ -829,6 +819,133 @@ func (ctl *controller) configInferenceServer(isc *fmav1alpha1.InferenceServerCon
...
+	state.annotationKeys = annotationKeys
+
+	cfgJSON, err := json.Marshal(state.cfg)
+	if err != nil {
+		return []string{fmt.Sprintf("failed to marshal launcher instance config: %s", err)}
+	}
+	providingPod.Annotations = utils.MapSet(providingPod.Annotations, launcherInstanceIDAnnotationKey, state.instanceID)
+	providingPod.Annotations[launcherServerPortAnnotationKey] = strconv.Itoa(int(state.serverPort))
+	providingPod.Annotations[launcherVllmConfigAnnotationKey] = string(cfgJSON)
+	providingPod.Annotations[iscLabelKeysAnnotationKey] = strings.Join(labelKeys, " ")
```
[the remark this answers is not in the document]
MikeSpreitzer: The environment variables are already being held outside of Secret objects, this is not a big concern.

I am more concerned about the data volume, but let's address that in follow-on work.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: These are never mutated. Might as well mark them as immutable.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -74,6 +74,16 @@ type launcherSyncResult struct {
...
 }
 
+type vllmInstanceState struct {
+	cfg            *VllmConfig
+	instanceID     string
+	serverPort     int16
+	iscLabels      map[string]string
+	iscAnnotations map[string]string
+	labelKeys      []string
+	annotationKeys []string
```
MikeSpreitzer: These are never mutated. Might as well mark them as immutable.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -829,6 +819,133 @@ func (ctl *controller) configInferenceServer(isc *fmav1alpha1.InferenceServerCon
...
+	providingPod.Annotations[iscAnnotationKeysAnnotationKey] = strings.Join(annotationKeys, " ")
+	return nil
+}
+
+func commitInstanceState(serverDat *serverData, state *vllmInstanceState) {
+	serverDat.InstanceID = state.instanceID
+	serverDat.InstanceConfig = state.cfg
+	serverDat.ServerPort = state.serverPort
+	serverDat.ISCLabelKeys = slices.Clone(state.labelKeys)
+	serverDat.ISCAnnotationKeys = slices.Clone(state.annotationKeys)
```
MikeSpreitzer: With `serverDat.ISCLabelKeys` and `state.labelKeys` commented as being immutable, the `Clone` would not be needed here.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: Doesn't the call to `recoverInstanceStateFromLauncherPod` above obviate this copy of recovery logic?

pkg/controller/dual-pods/controller.go
MikeSpreitzer: The comment on this field is outdated now.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -829,6 +819,133 @@ func (ctl *controller) configInferenceServer(isc *fmav1alpha1.InferenceServerCon
...
+}
+
+func recoverInstanceStateFromLauncherPod(serverDat *serverData, requestingPod, providingPod *corev1.Pod) error {
+	instanceID := providingPod.Annotations[launcherInstanceIDAnnotationKey]
+	if instanceID == "" && requestingPod != nil {
+		instanceID = requestingPod.Labels[api.InstanceLabelName]
```
MikeSpreitzer: See https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/a8111c667c1fc30e57e65ae7eb86baf1f25ea360/pkg/api/interface.go#L117-L119

pkg/controller/dual-pods/inference-server.go
```diff
@@ -829,6 +819,133 @@ func (ctl *controller) configInferenceServer(isc *fmav1alpha1.InferenceServerCon
...
+
+func commitInstanceState(serverDat *serverData, state *vllmInstanceState) {
+	serverDat.InstanceID = state.instanceID
+	serverDat.InstanceConfig = state.cfg
+	serverDat.ServerPort = state.serverPort
+	serverDat.ISCLabelKeys = slices.Clone(state.labelKeys)
+	serverDat.ISCAnnotationKeys = slices.Clone(state.annotationKeys)
+}
+
+func recoverInstanceStateFromLauncherPod(serverDat *serverData, requestingPod, providingPod *corev1.Pod) error {
```
MikeSpreitzer: The logic in this function is much more complex than it needs to be.

First, note that the five annotations relied on here need to be among those that the ValidatingAdmissionPolicy allows only the FMA controllers to write.

Note also that `commitInstanceState` sets all five fields at once.

Given the above, the logic here can be simple. If any of the five fields is unset then they all are and they all must be read from the launcher Pod.

MikeSpreitzer (commented):

waltforme:
Force-push to https://github.com/llm-d-incubation/llm-d-fast-model-actuation/commit/999f6d2a69f66cb86bd4c22ccc53bd1916ee9168:
- rebases onto `main` and resolves merge conflicts;
- addresses @MikeSpreitzer's all review comments except https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/492#discussion_r3206860403

api/fma/v1alpha1/inferenceserverconfig_types.go
```diff
@@ -35,6 +35,8 @@ type InferenceServerConfigSpec struct {
 type ModelServerConfig struct {
 	// Port is the port on which the vLLM server will listen
 	// Particularly, management of vLLM instances' sleep state is done through this port
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=32766
```
MikeSpreitzer: I am sorry, I was mistaken about the upper bound. It seems that port numbers above 32766 are usable. If I understand correctly, the upper bound should be 65534 (it could maybe be lowered some, but I wouldn't bother).

charts/fma-controllers/templates/rbac/role.yaml
```diff
@@ -15,3 +15,6 @@ rules:
 - apiGroups: [ "" ]
   resources: [ pods ]
   verbs: [ get, list, watch, create, update, patch, delete ]
+- apiGroups: [ "" ]
+  resources: [ events ]
+  verbs: [ create, patch ]
```
MikeSpreitzer: Is `patch` needed here?

pkg/controller/dual-pods/controller.go
```diff
@@ -93,6 +93,9 @@ import (
 
 const requesterAnnotationKey = "dual-pods.llm-d.ai/requester"
 const nominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
+const launcherInstanceIDAnnotationKey = "dual-pods.llm-d.ai/instance-id"
+const launcherServerPortAnnotationKey = "dual-pods.llm-d.ai/server-port"
+const launcherVllmConfigAnnotationKey = "dual-pods.llm-d.ai/vllm-config"
 const iscLabelKeysAnnotationKey = "dual-pods.llm-d.ai/isc-label-keys"
 const iscAnnotationKeysAnnotationKey = "dual-pods.llm-d.ai/isc-annotation-keys"
```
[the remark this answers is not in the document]
MikeSpreitzer: done

pkg/controller/dual-pods/controller.go
MikeSpreitzer: To clarify: every slice that is ever stored here is immutable from the moment it is store here forward. This field may hold different slices over time.

pkg/controller/dual-pods/controller.go
```diff
@@ -304,8 +314,9 @@ type serverData struct {
 	GPUIndices    []string
 	GPUIndicesStr *string
 
-	ProvidingPodName     string
-	InstanceID           string // ISC hash; set when computed, independent of instance existence
+	ProvidingPodName string
+	// Snapshot of the bound launcher-based vLLM instance. Not mutated after being set.
```
MikeSpreitzer: I think that this comment is intended to apply to the following five fields. It would be better if that scoping were clearer.

pkg/controller/dual-pods/controller.go
```diff
@@ -164,9 +169,12 @@ func (config ControllerConfig) NewController(
 	corev1PreInformers corev1preinformers.Interface,
 	fmaInformerFactory fmainformers.SharedInformerFactory,
 ) (*controller, error) {
+	eventBroadcaster := record.NewBroadcaster()
```
MikeSpreitzer: This may be a future problem for unit testing (of which there is currently very little, and IMO this is not a critical deficiency, but we may want more in the future --- IMO behavioral unit tests are not bad). The formal parameters are all things that can be provided by the standard fake clients that get generated to support testing.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -74,6 +74,18 @@ type launcherSyncResult struct {
 	stoppedInstanceIDs sets.Set[string] // bound instances found stopped (not deleted by sync)
 }
 
+// vllmInstanceState holds a snapshot of the ISC-derived state for a
+// launcher-based vLLM instance. Fields are not mutated after being set.
```
MikeSpreitzer: Note that there are multiple depths of mutation possible. The shallowest regards assignments to the fields in this struct. The next concerns mutations of the values that these fields reference. This comment, as currently written, speaks to me of only the shallowest depth.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -466,19 +452,21 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				if err != nil {
 					return err, true
 				}
-				result, err := lClient.CreateNamedInstance(ctx, serverDat.InstanceID, *cfg)
+				if serverDat.InstanceConfig == nil {
+					ctl.eventRecorder.Eventf(providingPod, corev1.EventTypeWarning, "MissingInstanceConfig",
+						"unable to create vLLM instance %q because bound launcher Pod is missing annotation %q",
```
MikeSpreitzer: There are many possible causes for `serverDat.InstanceConfig == nil` at the end of line 389 --- and they all cause line 390 to be taken, so `serverDat.InstanceConfig == nil` is guaranteed to be `false` here, right?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -74,6 +74,18 @@ type launcherSyncResult struct {
 	stoppedInstanceIDs sets.Set[string] // bound instances found stopped (not deleted by sync)
 }
 
+// vllmInstanceState holds a snapshot of the ISC-derived state for a
+// launcher-based vLLM instance. Fields are not mutated after being set.
+type vllmInstanceState struct {
+	cfg            *VllmConfig
+	instanceID     string
+	serverPort     int16
```
MikeSpreitzer: For simplicity I would type this the same as the corresponding field in the API object.

pkg/controller/dual-pods/inference-server.go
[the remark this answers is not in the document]
MikeSpreitzer: On second thought, I recommend a different bound, as I commented there.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1003,16 +1089,15 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
 	}
 	serverDat.ProvidingPodName = providingPod.Name
 	if launcherBased {
-		serverDat.InstanceID = *instanceID
+		commitInstanceState(serverDat, launcherState)
 	}
 	logger.V(2).Info("Bound server-providing Pod", "name", providingPod.Name, "node", requestingPod.Spec.NodeName, "gpus", serverDat.GPUIDsStr, "newResourceVersion", echo.ResourceVersion, "instanceID", serverDat.InstanceID)
 	var serverPort int16
```
MikeSpreitzer: There is too much chaos regarding the type for port numbers. I was mistaken to use `int16` earlier. I think it would be better to use `int32` throughout, since the fundamental APIs in the Go stdlib use strings anyway and Kube API guidelines disfavor `uint16`. Doesn't have to all be done in this PR.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: Since the other threshings of these five have procedural abstractions, it would not be bad for this one too. Not a big deal, could be done later.

MikeSpreitzer (commented):
I have done another round of review.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -829,6 +819,133 @@ func (ctl *controller) configInferenceServer(isc *fmav1alpha1.InferenceServerCon
...
+	if portS := providingPod.Annotations[launcherServerPortAnnotationKey]; portS != "" {
+		port, err := strconv.ParseInt(portS, 10, 16)
+		if err != nil {
+			return fmt.Errorf("bound launcher Pod %q has invalid annotation %q value %q: %w", providingPod.Name, launcherServerPortAnnotationKey, portS, err)
+		}
+		serverPort = int16(port)
+	}
+	if serverPort <= 0 {
+		return fmt.Errorf("bound launcher Pod %q is missing annotation %q", providingPod.Name, launcherServerPortAnnotationKey)
+	}
```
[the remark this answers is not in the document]
MikeSpreitzer: Revised.

api/fma/v1alpha1/inferenceserverconfig_types.go
waltforme: Changed to 65535.

charts/fma-controllers/templates/rbac/role.yaml
waltforme: Removed.

pkg/controller/dual-pods/controller.go
waltforme: Addressed

pkg/controller/dual-pods/controller.go
waltforme: Removed this since the recorder seems to be unnecessary.

pkg/controller/dual-pods/inference-server.go
waltforme: Addressed

pkg/controller/dual-pods/inference-server.go
waltforme: Yes, removed this dead block.

pkg/controller/dual-pods/inference-server.go
waltforme: Addressed

pkg/controller/dual-pods/inference-server.go
waltforme: Addressed

pkg/controller/dual-pods/controller.go
```diff
@@ -304,8 +307,12 @@ type serverData struct {
 	GPUIndices    []string
 	GPUIndicesStr *string
 
-	ProvidingPodName     string
-	InstanceID           string // ISC hash; set when computed, independent of instance existence
+	ProvidingPodName string
+	// The next five fields form a snapshot of the bound launcher-based
```
MikeSpreitzer: A blank line before this one would be a welcome break.
Not a big deal.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -74,6 +74,21 @@ type launcherSyncResult struct {
 	stoppedInstanceIDs sets.Set[string] // bound instances found stopped (not deleted by sync)
 }
 
+// vllmInstanceState holds a snapshot of the ISC-derived state for a
+// launcher-based vLLM instance. Once a *vllmInstanceState has been
+// returned from computeDesiredInstanceState, neither its fields nor
+// the values they reference (pointed-to VllmConfig, map entries,
+// slice elements) are mutated.
```
MikeSpreitzer: That's not exactly true, there is an exception for ISCLabelKeys and ISCAnnotationKeys.

This can be remedied by simplification. There is no longer a need for those two slices anywhere. Since the launcher Pod carries the full maps of labels and annotations, there is no need for duplicate storage of the keys in those maps.

pkg/controller/dual-pods/controller.go
```diff
@@ -304,8 +307,12 @@ type serverData struct {
...
+	// vLLM instance's ISC-derived state. Each field may be reassigned
+	// over time (e.g. on rebind), but whichever value (pointer, slice)
+	// is currently stored is deeply immutable for as long as it is stored.
+	InstanceID           string
 	InstanceConfig       *VllmConfig
 	InstanceKnownToExist bool     // meaningful only for launcher-based providers
```
MikeSpreitzer: Oops, sorry, I missed this one. This is not like the other four.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -362,47 +364,34 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+		}
+		isc, err = ctl.iscLister.InferenceServerConfigs(ctl.namespace).Get(iscName)
+		if err != nil {
+			return ctl.ensureReqStatus(ctx, requestingPod, serverDat,
+				fmt.Sprintf("failed to get InferenceServerConfig %q: %v", iscName, err),
+			)
+		}
+		desiredInstanceState, err = ctl.computeDesiredInstanceState(isc, serverDat.GPUIDs)
+		if err != nil {
+			return fmt.Errorf("failed to configure inference server config: %w", err), true
```
MikeSpreitzer: If I traced the code correctly, the only error that appears here is a marshaling failure, which is not a transient error. It should be reported as a non-transient problem in the server-requesting Pod's FMA status.

MikeSpreitzer (commented):

waltforme:
I addressed the latest review comments.

pkg/controller/dual-pods/controller.go
```diff
@@ -304,12 +311,16 @@ type serverData struct {
...
-	InstanceConfig       *VllmConfig
-	InstanceKnownToExist bool     // meaningful only for launcher-based providers
-	ISCLabelKeys         []string // keys of ISC labels applied to providingPod
-	ISCAnnotationKeys    []string // keys of ISC annotations applied to providingPod
+	ProvidingPodName string
+
+	// The next two fields form a snapshot of the bound launcher-based
+	// vLLM instance's ISC-derived state. Each field may be reassigned
+	// over time (e.g. on rebind), but whichever value (pointer) is
+	// currently stored is deeply immutable for as long as it is stored.
```
MikeSpreitzer, line 319: The limitation "for as long as it is stored" is a bit too short. Any code that reads such a field and then proceeds to _use_ what it read, no longer has a guarantee. I think that the more useful statement would be more like the following.

> ..., but once a reference is stored in one of these fields, the referent is henceforth deeply immutable.

Can be fixed in a follow-on PR.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: I was a bit wrong about the simplification, but I see that a good simplification was done.

MikeSpreitzer (approved):
This is good enough to merge.
Thanks!

Issue #447, opened 2026-04-22
[Feature]: Go imports section canonicalization


MikeSpreitzer:
### Feature Area

Other

### Problem Statement

There should be something checking that Go imports are in canonical order and grouping. It should be easy for a contributor to put the Go imports in a file into canonical order and grouping.

Canonical grouping is the following groups, in this order.

1. Imports from Go's stdlib.
2. Imports of packages that are independent of Kubernetes.
3. Imports of Kubernetes packages.
4. Imports of dependencies that depend on Kubernetes.
5. Imports from other llm-d repos (there are none of these, so far).
6. Imports of other packages in this repo.

Within each group, imports should be in lexicographic order.

### Proposed Solution

CI should include a gating check of this.

We should publish recommendations for how contributors establish, or at least check, compliance before submitting a PR.

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

The standard tools `gofmt` and `goimports` do ordering within the groups.

`goimports -local` will handle three groups, not the five that we want.

I usually use VS Code, which is maintaining order within the groups but does not do anything about proper grouping. I find that not terribly difficult to do by hand.

MikeSpreitzer:
KubeStellar uses https://github.com/openshift-eng/openshift-goimports, see https://github.com/kubestellar/kubestellar/blob/v0.29.0/Makefile#L321

MikeSpreitzer:
openshift-goimports does not seem to be actively maintained. The last activity was in spring of 2023. I was able to make a change at that time.

A more actively maintained tool is https://github.com/incu6us/goimports-reviser, which has releases and was last released in Feb of this year. It does not specifically deal with Kubernetes. It _does_ support a user-defined "company" group in addition to the local "project" group.

There is a standard tool, `goimports`, but it is pretty crude. It supports just three groups, in a fixed order: stdlib, everything else, and a user-defined "local".

PR #498, opened 2026-05-13
Requester tcp proxy v2

delavet:
This PR continues the unfinished work from #423, further simplifying and refactoring the TCP proxy within the requester.

Additionally, the experiments have been redesigned and re-run to measure the overhead introduced by the TCP proxy. For the experimental report, please refer to: https://docs.google.com/document/d/1qX4KtcTJEfdatgVmtsveOMfjm883MG-HW-e7ddmlqkA

## Main Changes since #423

- **Proxy concurrency model rewritten.** Replaced the `lazyTarget` placeholder + always-running listener with a two-channel design (`configCh` from `Configure` to `Run`, `startedCh` the other way). References: [Configure waits until listening](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#discussion_r3230008633), [two-channel simplification](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#discussion_r3230478593).

- **`AddFlags` now takes `*pflag.FlagSet`.** Updated the receiver in both `pkg/server/requester/config/config.go` and `pkg/server/requester/proxy/server.go`; call sites in `cmd/requester/main.go` and `cmd/test-requester/main.go` updated accordingly. References: [`config.go`](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#discussion_r3229648923), [`proxy/server.go`](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#discussion_r3229820658).

- **Imports regrouped per #447.** Six-group canonical order in `cmd/requester/main.go`: stdlib → non-k8s third-party → k8s → k8s-dependent → other llm-d repos → this repo. References: [import grouping](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#discussion_r3229612646).

- **Example yaml gets the proxy port; readiness probe targets the proxy.** `config/examples/requester.yaml` now declares `containerPort: 8082` and `--proxy-port=8082`, and the `readinessProbe` is `httpGet /health:8082`. The same readiness change is reflected in `cmd/requester/README.md`'s example PodSpecs and the curl walk-through. The probes server itself is left in place — cleaning it up is intentionally deferred to a follow-on PR. References: [point readinessProbe at proxy port instead of adding another proxy layer](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#discussion_r3229785147).

- **README Overview rewritten** to frame the requester as the inference server's stand-in plus a relay to/from the dual-pods controller, and to note that the TCP proxy makes the requester look *more* like an inference server. References: [README framing](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/423#discussion_r3229738522).

cc @MikeSpreitzer @aavarghese

### DIFF
--- a/cmd/requester/README.md
+++ b/cmd/requester/README.md
@@ -2,6 +2,74 @@ This document shows the steps to exercise the requester and dual-pods controller
 in a local k8s environment with model `ibm-granite/granite-3.3-2b-instruct`
 cached on local PV in the cluster.
 
+## Requester Overview
+
+The requester executable runs in the requester container of a server-requesting
+Pod. Its purpose is two-fold:
+
+1. **Look to the rest of llm-d like an inference server.** The requester listens
+   on the same set of ports an inference server would and forwards client traffic
+   to the bound vLLM instance via a built-in TCP reverse proxy.
+2. **Relay controller-bound state to and from the dual-pods controller.** The
+   requester exposes a small SPI surface so that the dual-pods controller can
+   read GPU IDs, push the bound server's address, and update readiness state.
+
+Adding the reverse TCP proxy makes the requester look *more* like an inference
+server than it otherwise would: client requests can be sent directly to the
+requester Pod's IP and the proxy forwards them to the actual server-providing
+Pod after binding.
+
+The server-requesting Pod runs three servers:
+
+| Server | Default Port | Purpose |
+|--------|-------------|---------|
+| Probes | 8080 | Readiness relay (`/ready` endpoint) — reflects the health status of the bound inference server |
+| SPI | 8081 | Dual-pods controller interface (`/v1/dual-pods/accelerators`, `/v1/proxy/config`) — exposes GPU info and accepts proxy configuration from the controller |
+| Proxy | 8082 | TCP reverse proxy listen port — accepts client connections and forwards them to the bound vLLM instance; rejects connections until configured |
+
+### TCP Proxy
+
+The requester includes a **TCP proxy** that forwards traffic to the bound inference server.
+After the dual-pods controller binds the requester to a server-providing Pod, it configures
+the proxy target via a `PUT` to the SPI endpoint at `/v1/proxy/config`. You can query the
+current proxy configuration with `GET /v1/proxy/config`.
+
+## Command-line flags
+
+### Port Configuration
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--probes-port` | `8080` | Port number for readiness/liveness probes |
+| `--spi-port` | `8081` | Port for dual-pods SPI requests |
+| `--proxy-port` | `8082` | TCP proxy listen port for forwarding client connections |
+
+### Proxy Configuration
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--proxy-dial-timeout` | `10s` | Timeout for dialing the backend inference server |
+
+### Logging (klog)
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--logtostderr` | `true` | Log to standard error instead of files |
+| `--alsologtostderr` | — | Log to standard error as well as files (no effect when `--logtostderr=true`) |
+| `--log_dir` | — | If non-empty, write log files in this directory (no effect when `--logtostderr=true`) |
+| `--log_file` | — | If non-empty, use this log file (no effect when `--logtostderr=true`) |
+| `--log_file_max_size` | `1800` | Maximum size a log file can grow to, in MB (no effect when `--logtostderr=true`) |
+| `--log_backtrace_at` | `:0` | When logging hits line file:N, emit a stack trace |
+| `--v` | `0` | Number for the log level verbosity |
+| `--vmodule` | — | Comma-separated list of pattern=N settings for file-filtered logging |
+| `--stderrthreshold` | `2` | Logs at or above this threshold go to stderr (no effect when `--logtostderr=true`) |
+| `--one_output` | — | Only write logs to their native severity level (no effect when `--logtostderr=true`) |
+| `--skip_headers` | — | Avoid header prefixes in log messages |
+| `--skip_log_headers` | — | Avoid headers when opening log files (no effect when `--logtostderr=true`) |
+| `--add_dir_header` | — | Add the file directory to the header of log messages |
+
+## Local Testing
+
 Build and push the requester container image (use your favorate
 `CONTAINER_IMG_REG`) with a command like the following. You can omit
 the `TARGETARCH` if the runtime ISA matches your build time ISA.
@@ -114,10 +182,12 @@ spec:
             containerPort: 8080
           - name: spi
             containerPort: 8081
+          - name: proxy
+            containerPort: 8082
           readinessProbe:
             httpGet:
-              path: /ready
-              port: 8080
+              path: /health
+              port: 8082
             initialDelaySeconds: 2
             periodSeconds: 5
           resources:
@@ -198,10 +268,12 @@ spec:
             containerPort: 8080
           - name: spi
             containerPort: 8081
+          - name: proxy
+            containerPort: 8082
           readinessProbe:
             httpGet:
-              path: /ready
-              port: 8080
+              path: /health
+              port: 8082
             initialDelaySeconds: 2
             periodSeconds: 5
           resources:
@@ -268,6 +340,9 @@ ports:
   - containerPort: 8081
     name: spi
     protocol: TCP
+  - containerPort: 8082
+    name: proxy
+    protocol: TCP
 readinessProbe:
   failureThreshold: 3
   httpGet:
@@ -299,11 +374,13 @@ Check the relayed readiness.
 ```console
 $ kubectl wait pod/my-request-5n2m6-dual-2wn7w --for=condition=Ready --timeout=120s
 pod/my-request-5n2m6-dual-2wn7w condition met
-$ curl $REQ_IP:8080/ready
-OK
+$ curl $REQ_IP:8082/health
+{"status":"ok"}
 ```
 
 Make an inference request.
+
+### Direct to inference server
 ```console
 $ kubectl get po -owide
 NAME                          READY   STATUS    RESTARTS   AGE     IP           NODE               NOMINATED NODE   READINESS GATES
@@ -320,12 +397,30 @@ $ curl -s http://10.0.0.145:8000/v1/completions \
 {"id":"cmpl-cfe04f79eb904748891561c76ae29986","object":"text_completion","created":1763768447,"model":"ibm-granite/granite-3.3-2b-instruct","choices":[{"index":0,"text":" Paris, which is known for its rich history, cultural landmarks, and iconic architecture like the Eiffel Tower and Notre","logprobs":null,"finish_reason":"length","stop_reason":null,"token_ids":null,"prompt_logprobs":null,"prompt_token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":5,"total_tokens":35,"completion_tokens":30,"prompt_tokens_details":null},"kv_transfer_params":null}
 ```
 
+### Via the TCP proxy
+```console
+$ curl -s http://$REQ_IP:8082/v1/completions \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "ibm-granite/granite-3.3-2b-instruct",
+    "prompt": "The capital of France is",
+    "max_tokens": 30
+  }'
+{"id":"...","object":"text_completion","created":1763768447,"model":"ibm-granite/granite-3.3-2b-instruct",...}
+
+$ # Verify proxy configuration via SPI endpoint
+$ curl -s http://$REQ_IP:8081/v1/proxy/config
+{"address":"10.0.0.145","port":8000}
+```
+
 Check the log of the server-requesting pod.
 ```console
 $ kubectl logs my-request-5n2m6
 I1121 23:22:48.119370       1 server.go:64] "starting server" logger="probes-server" port="8080"
 I1121 23:22:48.142001       1 server.go:84] "Got GPU UUIDs" logger="spi-server" uuids=["GPU-0d1d8df2-4bc7-98fe-1d41-a5d13a5866d1"]
 I1121 23:22:48.142119       1 server.go:171] "starting server" logger="spi-server" port="8081"
+I1121 23:22:48.142200       1 server.go:67] "Starting TCP proxy server" logger="proxy-server" port=8082 dialTimeout="10s"
+I1121 23:22:48.142300       1 server.go:90] "TCP proxy server listening" logger="proxy-server" port=8082 target="10.0.0.145:8000"
 I1121 23:30:32.039243       1 server.go:139] "Setting ready" logger="spi-server" newReady=false
 I1121 23:37:13.904292       1 server.go:139] "Setting ready" logger="spi-server" newReady=true
 ```
--- a/cmd/test-requester/main.go
+++ b/cmd/test-requester/main.go
@@ -30,12 +30,14 @@ import (
 	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
 
-	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/coordination"
-	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/probes"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
-	"k8s.io/klog/v2"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/config"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/coordination"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/probes"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/proxy"
 )
 
 // This variant requester emulates the GPU allocation behavior of
@@ -58,8 +60,7 @@ func main() {
 	overrides := &clientcmd.ConfigOverrides{}
 	var nodeName, podUID string
 	numGPUs := uint(1)
-	probesPort := int16(8080)
-	spiPort := int16(8081)
+	cfg := config.NewDefault()
 
 	klog.InitFlags(nil)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
@@ -67,8 +68,7 @@ func main() {
 	pflag.CommandLine.StringVar(&nodeName, "node", nodeName, "name of this Pod's Node")
 	pflag.CommandLine.StringVar(&podUID, "pod-uid", podUID, "UID of this Pod")
 	pflag.CommandLine.UintVar(&numGPUs, "num-gpus", numGPUs, "number of GPUs to allocate")
-	pflag.CommandLine.Int16Var(&probesPort, "probes-port", probesPort, "port number for /ready")
-	pflag.CommandLine.Int16Var(&spiPort, "spi-port", spiPort, "port for dual-pods requests")
+	cfg.AddFlags(pflag.CommandLine)
 
 	pflag.Parse()
 
@@ -108,26 +108,38 @@ func main() {
 	gpuUUIDs := allocateGPUs(ctx, kubeClient.CoreV1(), nodeName, overrides.Context.Namespace, apitypes.UID(podUID), numGPUs)
 
 	var ready atomic.Bool
+	proxySrv := proxy.New(cfg.Proxy)
 
 	var wg sync.WaitGroup
-	wg.Add(2)
-
+	wg.Add(1)
 	go func() {
 		defer wg.Done()
 
-		err := coordination.RunWithGPUUUIDs(ctx, strconv.FormatInt(int64(spiPort), 10), &ready, os.Stdout, gpuUUIDs)
+		err := coordination.RunWithGPUUUIDs(ctx, strconv.FormatUint(uint64(cfg.SPIPort), 10), &ready, os.Stdout, gpuUUIDs, proxySrv.Configure)
 		if err != nil {
-			logger.Error(err, "failed to start requester SPI server")
+			logger.Error(err, "failed to run requester SPI server")
 		}
 	}()
 
 	// Start the readiness probe server
+	wg.Add(1)
 	go func() {
 		defer wg.Done()
 
-		err := probes.Run(ctx, strconv.FormatInt(int64(probesPort), 10), &ready)
+		err := probes.Run(ctx, strconv.FormatUint(uint64(cfg.ProbesPort), 10), &ready)
 		if err != nil {
-			logger.Error(err, "failed to start requester probes server")
+			logger.Error(err, "failed to run requester probes server")
+		}
+	}()
+
+	// Start the reverse proxy server
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		err := proxySrv.Run(ctx)
+		if err != nil {
+			logger.Error(err, "failed to run requester proxy server")
 		}
 	}()
 
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -510,6 +510,14 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		if err != nil {
 			return err, true
 		}
+		// After a controller restart the binding may have been persisted
+		// but the proxy config never sent. Ensure it is configured now.
+		if !serverDat.ProxyConfigured {
+			if err := ensureProxyConfigured(ctx, requestingPod, providingPod, adminPort, serverPort); err != nil {
+				return err, true
+			}
+			serverDat.ProxyConfigured = true
+		}
 		// Relay readiness if not already done.
 		// For launcher-based providers, readiness follows the bound instance's
 		// sleeping state rather than the launcher's Pod readiness.
@@ -1305,6 +1313,24 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
 			return fmt.Errorf("unable to wake up server because port not known: %w", err), true
 		}
 	}
+	// Initialize a reverse proxy between the launcher Pod and the requester Pod.
+	// Requests can be proxied to the launcher Pod from the requester Pod.
+	adminPort := requestingPod.Annotations[api.AdminPortAnnotationName]
+	if adminPort == "" {
+		adminPort = api.AdminPortDefaultValue
+	}
+	requesterAddr := fmt.Sprintf("%s:%s", requestingPod.Status.PodIP, adminPort)
+	launcherAddr := fmt.Sprintf("%s:%d", providingPod.Status.PodIP, serverPort)
+	url := fmt.Sprintf("http://%s%s", requesterAddr, stubapi.ProxyConfigPath)
+	proxyConfig, _ := json.Marshal(stubapi.ProxyTargetConfig{
+		Address: providingPod.Status.PodIP,
+		Port:    uint16(serverPort),
+	})
+
+	if err := doPut(ctx, url, bytes.NewReader(proxyConfig)); err != nil {
+		return fmt.Errorf("failed to initialize proxy (requester %s -> launcher %s): %w",
+			requesterAddr, launcherAddr, err), true
+	}
 	if !skipWake {
 		err = ctl.wakeSleeper(ctx, serverDat, requestingPod, providingPod, serverPort, "freshly-bound")
 		if err != nil {
@@ -1895,6 +1921,72 @@ func (ctl *controller) ensureReqState(ctx context.Context, requestingPod *corev1
 		logger.V(2).Info("Failed to set status/finalizers", "serverRequestingPod", requestingPod.Name, "status", status, "accelerators", desiredAccelerators, "boundName", serverDat.ProvidingPodName, "instanceID", desiredInstanceID, "finalizers", requestingPod.Finalizers, "resourceVersion", requestingPod.ResourceVersion, "k8sCallStartTime", updStartStr, "err", err.Error())
 	}
 	return err, err != nil
+}
+
+func doPut(ctx context.Context, url string, data io.Reader) error {
+	return doHTTPRequest(ctx, http.MethodPut, url, data)
+}
+
+func ensureProxyConfigured(ctx context.Context, requestingPod, providingPod *corev1.Pod, adminPort string, serverPort int32) error {
+	logger := klog.FromContext(ctx)
+	proxyURL := fmt.Sprintf("http://%s:%s%s", requestingPod.Status.PodIP, adminPort, stubapi.ProxyConfigPath)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, proxyURL, nil)
+	if err != nil {
+		return fmt.Errorf("ensureProxyConfigured: create GET request: %w", err)
+	}
+	client := &http.Client{Timeout: 5 * time.Second}
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("ensureProxyConfigured: GET %s: %w", proxyURL, err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+	if resp.StatusCode != http.StatusNotFound {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("ensureProxyConfigured: GET %s returned %d: %s", proxyURL, resp.StatusCode, string(body))
+	}
+
+	proxyConfig, _ := json.Marshal(stubapi.ProxyTargetConfig{
+		Address: providingPod.Status.PodIP,
+		Port:    uint16(serverPort),
+	})
+	if err := doPut(ctx, proxyURL, bytes.NewReader(proxyConfig)); err != nil {
+		return fmt.Errorf("ensureProxyConfigured: PUT %s: %w", proxyURL, err)
+	}
+	logger.V(2).Info("Configured proxy after controller restart",
+		"requester", requestingPod.Name, "provider", providingPod.Name,
+		"target", fmt.Sprintf("%s:%d", providingPod.Status.PodIP, serverPort))
+	return nil
+}
+
+func doHTTPRequest(ctx context.Context, method, url string, data io.Reader) error {
+	req, err := http.NewRequestWithContext(ctx, method, url, data)
+	if err != nil {
+		return fmt.Errorf("http %s %q: %w", method, url, err)
+	}
+	if data != nil {
+		req.Header.Set("Content-Type", "application/json")
+	}
+	client := &http.Client{
+		Timeout: 5 * time.Second,
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("http %s %q: %w", method, url, err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("http %s %q returned unexpected status %d; response body=%s", method, url, resp.StatusCode, string(body))
+	}
+
+	return nil
 }
 
 var coreScheme *k8sruntime.Scheme
--- a/pkg/server/requester/config/config.go
+++ b/pkg/server/requester/config/config.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/proxy"
+)
+
+// Config holds all configurable runtime parameters for the requester.
+type Config struct {
+	ProbesPort uint16
+	SPIPort    uint16
+	Proxy      proxy.ProxyConfig
+}
+
+// NewDefault returns a Config with sensible defaults.
+func NewDefault() Config {
+	return Config{
+		ProbesPort: 8080,
+		SPIPort:    8081,
+		Proxy:      proxy.DefaultProxyConfig,
+	}
+}
+
+// AddFlags registers command-line flags for all Config fields.
+func (cfg *Config) AddFlags(fs *pflag.FlagSet) {
+	fs.Uint16Var(&cfg.ProbesPort, "probes-port", cfg.ProbesPort, "port number for readiness/liveness probes")
+	fs.Uint16Var(&cfg.SPIPort, "spi-port", cfg.SPIPort, "port for dual-pods SPI requests")
+	cfg.Proxy.AddFlags(fs)
+}
--- a/pkg/server/requester/proxy/server.go
+++ b/pkg/server/requester/proxy/server.go
@@ -1,0 +1,225 @@
+/*
+Copyright 2026 The llm-d Authors.
+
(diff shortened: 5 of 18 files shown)


cmd/requester/README.md
```diff
@@ -320,12 +397,30 @@ $ curl -s http://10.0.0.145:8000/v1/completions \
...
+{"address":"10.0.0.145","port":8000}
+```
+
 Check the log of the server-requesting pod.
 ```console
 $ kubectl logs my-request-5n2m6
 I1121 23:22:48.119370       1 server.go:64] "starting server" logger="probes-server" port="8080"
 I1121 23:22:48.142001       1 server.go:84] "Got GPU UUIDs" logger="spi-server" uuids=["GPU-0d1d8df2-4bc7-98fe-1d41-a5d13a5866d1"]
 I1121 23:22:48.142119       1 server.go:171] "starting server" logger="spi-server" port="8081"
+I1121 23:22:48.142200       1 server.go:67] "Starting TCP proxy server" logger="proxy-server" config={"Port":8082,"DialTimeout":10000000000}
```
MikeSpreitzer: The timeout is displayed with a painful number of zeros. Doesn't `time.Duration` have a friendlier conversion to string? If not, maybe "DialTimeoutSeconds" showing a floating-point number?

cmd/test-requester/main.go
```diff
@@ -31,8 +31,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/config"
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/coordination"
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/probes"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/proxy"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
 	"k8s.io/klog/v2"
```
MikeSpreitzer: See the import grouping in the non-test requester.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1068,44 +928,103 @@ func (ctl *controller) enforceSleeperBudget(ctx context.Context, serverDat *serv
...
+	launcherAddr := fmt.Sprintf("%s:%d", providingPod.Status.PodIP, serverPort)
+	url := fmt.Sprintf("http://%s%s", requesterAddr, stubapi.ProxyConfigPath)
+	proxyConfig, _ := json.Marshal(stubapi.ProxyTargetConfig{
+		Address: providingPod.Status.PodIP,
+		Port:    uint16(serverPort),
+	})
+
+	if err := doPut(url, bytes.NewReader(proxyConfig)); err != nil {
+		logger.Error(err, "Failed to initialize requester proxy",
+			"requesterAddr", requesterAddr, "proxyConfig", string(proxyConfig))
```
MikeSpreitzer: FYI, it is not strictly necessary to log this here, since the caller will log the error returned just below.

pkg/server/requester/config/config.go
```diff
@@ -0,0 +1,46 @@
...
+
+package config
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/server/requester/proxy"
+)
+
+// Config holds all configurable port parameters for the requester binary.
```
MikeSpreitzer: In "configurable port parameters", the qualifier "port" seems unnecessary and even confusing. It seems a stretch to me to say that the proxy's dial timeout is a port parameter.

The word "binary" seems out of place. Everything in the code goes into the binary, it seems odd to mention the binary in one place.

pkg/server/requester/config/config.go
```diff
@@ -0,0 +1,46 @@
...
+type Config struct {
+	ProbesPort int16
+	SPIPort    int16
```
MikeSpreitzer: These should be `uint16`.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+	startedCh chan struct{}
+
+	stateMu  sync.Mutex
+	target   *stubapi.ProxyTargetConfig // set by the first successful Configure PUT
+	startErr error                      // set by Run before closing startedCh on a Start failure
+)
+
+// initState lazily initializes the module-level coordination channels.
+func initState() {
+	initOnce.Do(func() {
```
MikeSpreitzer: I see no reason to programmatically delay this initialization. It would be simpler to do in `func init()`. Even simpler to do in the variable declarations.

```go
var configCh = make(chan *stubapi.ProxyTargetConfig, 1)
var startedCh = make(chan struct{})
```

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// Module-level coordination state. Run and Configure may run on different
+// goroutines and in either order; the channels handle the hand-off between
+// them. configCh carries the ProxyTargetConfig from Configure to Run;
+// startedCh is closed by Run once the proxy is actively listening (or has
+// failed to start), so Configure does not return until that point.
+var (
+	initOnce  sync.Once
+	configCh  chan *stubapi.ProxyTargetConfig
```
MikeSpreitzer: Using a _pointer_ in the channel's element type here is unnecessary complexity. What if the receiver receives a `nil`? Currently the receiver does not check for that, but should; put another way, the correctness of not checking for nil involves a long-distance dependency that is not checked by the compiler. It would be simpler if the channel's element type were simply the config struct.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+	logger.Info("Starting TCP proxy server", "config", cfg)
+
+	initState()
+
+	var tgt *stubapi.ProxyTargetConfig
+	select {
+	case tgt = <-configCh:
+		logger.V(2).Info("Received proxy target from Configure", "address", tgt.Address, "port", tgt.Port)
+	case <-ctx.Done():
+		logger.V(2).Info("Context cancelled before proxy was configured")
```
MikeSpreitzer, line 100: Don't you need to `close(startedCh)` here? Every channel should be closed exactly once, in code that can be re-used multiple times in a binary (e.g., is called from multiple test cases).

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+		return nil
+	}
+
+	targetAddr := net.JoinHostPort(tgt.Address, fmt.Sprintf("%d", tgt.Port))
```
MikeSpreitzer: Maybe define a `String` method on `tgt`.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+		http.Error(w, "proxy already configured", http.StatusConflict)
+		return
+	}
+	target = &cfg
+	stateMu.Unlock()
+
+	logger.V(2).Info("Delivering proxy target to Run", "address", cfg.Address, "port", cfg.Port)
+
+	select {
+	case configCh <- &cfg:
```
MikeSpreitzer: Since this channel has a non-zero storage capacity, this case will always be enabled.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+	case <-r.Context().Done():
```
MikeSpreitzer: The text "before proxy could receive target" is confusing to me. At this point in the code, the request body has certainly been received and decoded. Perhaps the intent is to say something about the channel-based transfer to `Run`. But that will never wait on anything, because the channel has a non-zero storage capacity.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+// startedCh is closed by Run once the proxy is actively listening (or has
+// failed to start), so Configure does not return until that point.
+var (
+	initOnce  sync.Once
+	configCh  chan *stubapi.ProxyTargetConfig
+	startedCh chan struct{}
+
+	stateMu  sync.Mutex
+	target   *stubapi.ProxyTargetConfig // set by the first successful Configure PUT
+	startErr error                      // set by Run before closing startedCh on a Start failure
```
MikeSpreitzer: The ordering of write and read of this variable wrt the closing of `startedCh` is sufficient synchronization; holding the `stateMu` is not necessary for reading nor writing this variable.

Because of the way that access to this variable is synchronized, I would list it after `startedCh` and NOT after `stateMu`.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+
+	targetAddr := net.JoinHostPort(tgt.Address, fmt.Sprintf("%d", tgt.Port))
+	p := &tcpproxy.Proxy{}
+	p.AddRoute(fmt.Sprintf(":%d", cfg.Port), &tcpproxy.DialProxy{
+		Addr:        targetAddr,
+		DialTimeout: cfg.DialTimeout,
+	})
+
+	if err := p.Start(); err != nil {
+		stateMu.Lock()
```
MikeSpreitzer: It is enough that writing `startErr` happens before closing `startedCh` and reading `startErr` happens after closing `startedCh`; locking `stateMu` here is unnecessary.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+		sErr := startErr
+		stateMu.Unlock()
+		if sErr != nil {
+			http.Error(w, fmt.Sprintf("proxy failed to start: %v", sErr), http.StatusInternalServerError)
+			return
+		}
+	case <-r.Context().Done():
+		http.Error(w, "request cancelled before proxy started listening", http.StatusServiceUnavailable)
+		return
+	}
```
MikeSpreitzer, line 221: There is an additional `context` whose cancellation should cause a return here. The one given to `proxy.Run` and `coordination.RunWithGPUUUIDs` (https://github.com/delavet/llm-d-fast-model-actuation/blob/20a93c60efd42702605ddbe9ef9dbdcca4144174/pkg/server/requester/coordination/server.go#L208).

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+	"testing"
+	"time"
+
+	stubapi "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/spi"
+)
+
+// resetInstance clears module-level state so tests don't interfere with each other.
+func resetInstance(t *testing.T) {
+	t.Helper()
+	initOnce = sync.Once{}
```
MikeSpreitzer: Each test case needs to call `resetInstance` before exercising either `proxy.Run` or `proxy.Configure`. Thus, using a `sync.Once` is not helpful here. Simply define a func that does the initialization of the global variables, and call it from `resetInstance` and from the `init` func of server.go.

I still think that it would be simpler to move the proxy's global variables into a struct and define a `New` func for those struct values. That removes the need to define a "reset" function and also allows test cases to proceed in parallel (provided that they get configured to use different ports).

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,228 @@
...
+	case <-r.Context().Done():
+		http.Error(w, "request cancelled before proxy could receive target", http.StatusServiceUnavailable)
+		return
+	}
+
+	select {
+	case <-startedCh:
+		stateMu.Lock()
+		sErr := startErr
+		stateMu.Unlock()
```
MikeSpreitzer: This is much of the implementation of a channel receive. Remember that this is intended to be executed at most once. It would be simpler to have `startedCh` carry the error value.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+// startTestEchoServer starts a TCP server that echoes back any data it receives.
+func startTestEchoServer(t *testing.T) (addr string, port uint16, closer func()) {
+	t.Helper()
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("failed to start echo server: %v", err)
+	}
+	tcpAddr := ln.Addr().(*net.TCPAddr)
```
MikeSpreitzer: This (unchecked) cast would be unnecessary if the listen had been done with https://pkg.go.dev/net@go1.26.3#ListenTCP .

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+				return
+			}
+			go func(c net.Conn) {
+				_, _ = io.Copy(c, c)
+				_ = c.Close()
+			}(conn)
+		}
+	}()
+
+	return addr, port, func() { _ = ln.Close() }
```
MikeSpreitzer: Does the stop function need to close the connections too?

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func TestProxy_EchoRoundTrip(t *testing.T) {
+	resetInstance(t)
+
+	_, backendPort, backendCloser := startTestEchoServer(t)
+	defer backendCloser()
+
+	proxyPort := findFreePort(t)
+	stop := startProxy(t, context.Background(), ProxyConfig{Port: proxyPort, DialTimeout: 2 * time.Second})
```
MikeSpreitzer: Use the test's context (https://pkg.go.dev/testing@go1.26.3#T.Context) rather than `context.Background()`.

MikeSpreitzer: Why strive for such a short connection timeout? What goes wrong with 10 seconds or more?

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+	defer stop()
+
+	// Configure does not return until the proxy is listening.
```
MikeSpreitzer: Doesn't this comment belong on the call to `Configure`?

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	defer func() { _ = conn.Close() }()
+
+	testMsg := "hello proxy\n"
+	if _, err := conn.Write([]byte(testMsg)); err != nil {
+		t.Fatalf("failed to write to proxy: %v", err)
+	}
+
+	reader := bufio.NewReader(conn)
+	resp, err := reader.ReadString('\n')
```
MikeSpreitzer: Note: this is not testing whether anything else appears after that `'\n'`.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+func TestProxy_NoListenerBeforeConfigure(t *testing.T) {
+	resetInstance(t)
+
+	proxyPort := findFreePort(t)
+	stop := startProxy(t, context.Background(), ProxyConfig{Port: proxyPort, DialTimeout: DefaultProxyConfig.DialTimeout})
+	defer stop()
+
+	// Run is blocked waiting for Configure, so the proxy port has no listener.
+	time.Sleep(100 * time.Millisecond)
```
MikeSpreitzer: I would use larger timeouts in this test. Remember that the machines running CI workflows can be surprisingly overloaded.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+	Configure(w3, req3)
+	if w3.Code != http.StatusOK {
+		t.Errorf("GET after configure should return 200, got %d", w3.Code)
+	}
+
+	var config stubapi.ProxyTargetConfig
+	if err := json.Unmarshal(w3.Body.Bytes(), &config); err != nil {
+		t.Fatalf("failed to parse response JSON: %v", err)
+	}
+	if config.Address != "127.0.0.1" {
```
MikeSpreitzer: I would compare with fields of `body` from above.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	var cfg stubapi.ProxyTargetConfig
+	if err := json.Unmarshal(wGet.Body.Bytes(), &cfg); err != nil {
+		t.Fatalf("failed to parse GET response: %v", err)
+	}
+	if cfg.Address != "127.0.0.1" || cfg.Port != backendPort {
+		t.Errorf("GET returned wrong config: want {127.0.0.1, %d}, got {%s, %d}", backendPort, cfg.Address, cfg.Port)
+	}
+
+	req2 := httptest.NewRequest(http.MethodPut, stubapi.ProxyConfigPath, bytes.NewReader(jsonBody))
```
MikeSpreitzer: We probably want to allow writing the same configuration.
I think that it would be better to test writing a different configuration.

MikeSpreitzer (commented):

MikeSpreitzer:
Please rebase this onto `main` so that we can proceed to stage B of resolving #471 (which we planned to do only after all open PRs include stage A).

MikeSpreitzer:
@delavet: we want to progress on #471. The next step, #483, cannot be started until all current work incorporates the schema change (Issue #482, PR #487). As a specific demonstration, note that the CI test of THIS PR on OpenShift, launched just above, builds and uses the HEAD of this PR, currently commit 20a93c60efd42702605ddbe9ef9dbdcca4144174 --- you can see that in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/26109505229/job/76782612321 . The head of THIS PR does NOT include the effects of #487, and so we can not yet progress to stage B (#483) of resolving #471.

To enable this progress, can you please rebase this PR onto `main`? Just doing that rebase, no other work, will be sufficient to enable progress on #471.

MikeSpreitzer:
On the overall advisability of this approach (proxying the inference requests), I have realized a problem. Other stuff in llm-d (e.g., EPP, autoscalers) rely on Prometheus metrics from the inference server Pods --- the Pods that match the pod selector predicate in the InferencePool object. I do not think that trying to proxy the Prometheus scraping is a good idea.

cmd/requester/README.md
delavet: Switched the proxy log line to named fields (port / dialTimeout), so the duration now prints as "10s" instead of 10000000000.

cmd/test-requester/main.go
delavet: Reordered imports to match the cmd/requester grouping (stdlib · pflag · k8s.io+klog · sigs.k8s.io · local).

pkg/controller/dual-pods/inference-server.go
delavet: Dropped the inline logger.Error;

pkg/server/requester/config/config.go
delavet: Reworded to Config holds all configurable runtime parameters for the requester. (dropped both "port" and "binary").

pkg/server/requester/config/config.go
delavet: done

pkg/server/requester/proxy/server.go
delavet: Resolved by the struct refactor: there are no more global variables — New(cfg) initializes all channels/state inline

pkg/server/requester/proxy/server.go
delavet: done, configCh is now chan stubapi.ProxyTargetConfig (value type, no pointer).

pkg/server/requester/proxy/server.go
delavet: Restructured so Run no longer needs to send/close `startedCh` in the early-exit path: Configure now also selects on `s.currentRunCtx().Done()`, so a process shutdown unblocks an in-flight PUT regardless of how Run exits.

pkg/server/requester/proxy/server.go
delavet: done

pkg/server/requester/proxy/server.go
delavet: Dropped the select; it is just configCh <- cfg now

pkg/server/requester/proxy/server.go
delavet: Resolved by removing that select entirely (see comment https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/498#discussion_r3301933256). The confusing error string is gone too.

pkg/server/requester/proxy/server.go
delavet: `startErr` is gone (the value is delivered through `startedCh` directly)

pkg/server/requester/proxy/server.go
delavet: same to https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/498#discussion_r3301940843 , `startErr` no longer exists now

pkg/server/requester/proxy/server.go
delavet: `Configure` now selects on `s.currentRunCtx().Done()` as the third case.

pkg/server/requester/proxy/server_test.go
delavet: Done — refactored in the final commit. `proxy.Server` struct with `New(cfg)` constructor

pkg/server/requester/proxy/server.go
delavet: `startedCh` is now chan error

pkg/server/requester/proxy/server_test.go
delavet: Switched the listeners to `net.ListenTCP`, which returns a typed `*net.TCPAddr `— no unchecked cast.

pkg/server/requester/proxy/server_test.go
delavet: The echo server's closer now also closes any connections it accepted, so no goroutines / sockets linger between tests.

pkg/server/requester/proxy/server_test.go
delavet: done

delavet: Indeed, the timeout should be longer. Bumped to `testDialTimeout = 10 * time.Second` (constant shared by all tests).

pkg/server/requester/proxy/server_test.go
delavet: yes, moved the comment

pkg/server/requester/proxy/server_test.go
delavet: Switched to `io.ReadFull` for exact length, then a follow-up `Read` with a 100ms deadline that asserts no trailing byte arrives

pkg/server/requester/proxy/server_test.go
delavet: Thanks for point that out again. I’ve now adjusted it to 500 ms, as I don’t want the test to take too long—Is this acceptable?

pkg/server/requester/proxy/server_test.go
delavet: done

pkg/server/requester/proxy/server_test.go
delavet: Renamed to `TestConfigure_DifferentConfigReturnsConflict` and PUT a different `(address, port)` on the second call. On the API side — kept "same config also returns 409" as-is for now; happy to revisit if you want PUT to be idempotent on identical configs (would need to update the controller's `doPut` to treat 200 vs. 409 the same way).

delavet:
> On the overall advisability of this approach (proxying the inference requests), I have realized a problem. Other stuff in llm-d (e.g., EPP, autoscalers) rely on Prometheus metrics from the inference server Pods --- the Pods that match the pod selector predicate in the InferencePool object. I do not think that trying to proxy the Prometheus scraping is a good idea.

Sorry for the delay! Rebased now.

Regarding Prometheus scraping through the proxy: since this is a transparent L4 (TCP) proxy, metrics requests to the requester Pod's proxy port are forwarded byte-for-byte to vLLM — Prometheus sees the same /metrics response as if it connected directly. The instance label naturally carries the requester Pod identity, which aligns with the InferencePool selector.
Is there a specific failure mode you're concerned about? I considered: 

- proxy-not-ready (pod won't be Ready either, should be OK)
- vllm server crash (requester pod  becomes not-ready, align with the real server status, should be OK)
- identity confusion (actually correct since requester pod IS the inference server from EPP's perspective).

Happy to discuss further or open a design issue if there's something we're missing.

MikeSpreitzer:
@delavet: you're right about the Prometheus proxying, I had overlooked the fact that it is already part of the proxying functionality.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1106,6 +1106,24 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
 			return fmt.Errorf("unable to wake up server because port not known: %w", err), true
 		}
 	}
+	// Initialize a reverse proxy between the launcher Pod and the requester Pod.
```
MikeSpreitzer, line 1316: What happens if the dual-pods controller process crashes at this point --- after the binding has been recorded in the launcher Pod but before the requester is given the proxy config? Kubernetes will create a replacement dual-pods controller, which will find that the launcher is already bound to the requester. What gets the proxy config established in this case?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,241 @@
...
+	stateMu sync.Mutex
+	target  *stubapi.ProxyTargetConfig
+}
+
+// New creates a Server with the given configuration.
+func New(cfg ProxyConfig) *Server {
+	return &Server{
+		cfg:       cfg,
+		configCh:  make(chan stubapi.ProxyTargetConfig, 1),
+		startedCh: make(chan error, 1),
```
MikeSpreitzer, line 82: Both of these channels should be `close()`ed on every possible execution path.

I think it is reasonable to stipulate that a call to `New` is certainly followed by a call to `Run`.

Currently I see no `close` of `configCh`, and the early out of `Run` skips closing `startedCh`.

pkg/controller/dual-pods/inference-server.go
delavet: **Fix:** Add an `ensureProxyConfigured` helper that is called in the "already bound" reconciliation path. It GETs `/v1/proxy/config` on the requester first — if 200, proxy is already configured (normal case, no extra cost beyond one GET); if 404, it PUTs the config to recover from the crash scenario.

pkg/server/requester/proxy/server.go
delavet: **`startedCh`:** Agreed — added `close(s.startedCh)` on the early-exit path of `Run` (context cancelled before config). `Run` is the sole writer, so this is safe with no additional synchronization.

**`configCh`:** I have not yet implemented its shutdown.. Reason: `Configure` sends on `configCh` from a different goroutine. Safely closing `configCh` from `Run`'s early-exit path requires a mutex to prevent a send-on-closed-channel panic (the send in `Configure` and the close in `Run` can race). This adds a `sync.Mutex` + `bool` to the struct and lock/unlock calls on both the send and close paths.
I believe adding this complexity does not yield any additional benefits, because:

- Go channels don't need `close()` for GC — they're collected when unreferenced.
- No goroutine blocks on `configCh` after `Run` exits. `Configure` either already buffered its value (cap=1) or detects `Run`'s exit via `currentRunCtx().Done()`.
- `configCh` is single-use (at most one send, at most one receive). There's no `range` or "wait for close" reader that needs the close signal.

MikeSpreitzer: I stand corrected. It is not necessary to close these channels on every possible execution.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -494,6 +494,11 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		if err != nil {
 			return err, true
 		}
+		// After a controller restart the binding may have been persisted
+		// but the proxy config never sent. Ensure it is configured now.
+		if err := ensureProxyConfigured(ctx, requestingPod, providingPod, adminPort, serverPort); err != nil {
```
MikeSpreitzer: As currently written, `ensureProxyConfigured` will do an HTTP GET on the requester every time. This seems wasteful. Wouldn't it be better if the GET were done only if `serverDat` did not indicate that the proxy had already been configured?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,242 @@
...
+	logger := klog.FromContext(r.Context()).WithName("proxy-server")
+
+	switch r.Method {
+	case http.MethodGet:
+		s.stateMu.Lock()
+		cur := s.target
+		s.stateMu.Unlock()
+		if cur == nil {
+			w.WriteHeader(http.StatusNotFound)
+			_, _ = w.Write([]byte("proxy not configured"))
```
MikeSpreitzer: Could this be shortened by using `http.Error` ?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,242 @@
...
+	select {
+	case startErr := <-s.startedCh:
+		if startErr != nil {
+			http.Error(w, fmt.Sprintf("proxy failed to start: %v", startErr), http.StatusInternalServerError)
+			return
+		}
+	case <-r.Context().Done():
+		http.Error(w, "request cancelled before proxy started listening", http.StatusServiceUnavailable)
+		return
+	case <-s.currentRunCtx().Done():
```
MikeSpreitzer: There is a concurrency bug here. The expression `s.currentRunCtx().Done()` is evaluated once, at the start of the `select` statement, to yield a channel. If `s.runCtx` was not already set by that time then the channel will be that of `context.Background().Done()`. So, for example, if `Run` starts executing after this moment then its setting of `s.runCtx` will not succeed in passing that context's `.Done()` into this `select` statement.

test/e2e/test-cases.sh
```diff
@@ -208,7 +208,7 @@ launcher2=${launcher2:-}
 launcher3=${launcher3:-}
 launcher4=${launcher4:-}
 testnode=${testnode:-}
-if [ -n "${pfpid:-}" ]; then kill $pfpid; fi
+if [ -n "${pfpid:-}" ]; then kill ${pfpid:-}; fi
```
MikeSpreitzer: This change is unnecessary because the `kill` command will only be executed if `pfpid` is defined to be something non-empty.

MikeSpreitzer (commented):

Is https://docs.google.com/document/d/1qX4KtcTJEfdatgVmtsveOMfjm883MG-HW-e7ddmlqkA up to date?

delavet:
> 
> Is https://docs.google.com/document/d/1qX4KtcTJEfdatgVmtsveOMfjm883MG-HW-e7ddmlqkA up to date?

The content in this document was prepared at the time this PR was submitted. Although the proxy server’s build approach has undergone some changes since then, the overall architecture remains unchanged. Therefore, the conclusion remains valid: use `inetaf/tcpproxy`.

pkg/controller/dual-pods/inference-server.go
delavet: Done. I added a `ProxyConfigured` field to store the status of the proxy configuration; the `GET` request is only issued for further checks when `!ProxyConfigured` is true.

pkg/server/requester/proxy/server.go
delavet: Indeed, reformed

test/e2e/test-cases.sh
delavet: done

pkg/server/requester/proxy/server.go
delavet: Thanks for the catch. I’ve further optimized the synchronization behavior here—specifically, I’ve removed `runCtx` from the `Server`, and `Configure` can now listen directly to `startedCh`. Additionally, I’ve ensured that every execution path within `Run` writes to `startedCh`. As a result, the `case <-s.currentRunCtx().Done()` branch is no longer needed, thereby eliminating this problem entirely.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1838,26 +1864,71 @@ func (ctl *controller) ensureReqState(ctx context.Context, requestingPod *corev1
...
+		"requester", requestingPod.Name, "provider", providingPod.Name,
+		"target", fmt.Sprintf("%s:%d", providingPod.Status.PodIP, serverPort))
+	return nil
+}
+
+func doHTTPRequest(ctx context.Context, method, url string, data io.Reader) error {
+	req, err := http.NewRequestWithContext(ctx, method, url, data)
 	if err != nil {
-		return fmt.Errorf("http post %q: %w", url, err)
+		return fmt.Errorf("http %s %q: %w", method, url, err)
```
MikeSpreitzer, line 1969: The wrapper here and in the `Do` handling should be different, so that someone reading a log can tell the two apart.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// DefaultProxyConfig provides sensible defaults.
+var DefaultProxyConfig = ProxyConfig{
+	Port:        8082,
+	DialTimeout: 10 * time.Second,
+}
+
+// AddFlags registers command-line flags for all proxy configuration fields.
+func (cfg *ProxyConfig) AddFlags(fs *pflag.FlagSet) {
```
MikeSpreitzer, line 54: FYI, this method's name is not wrong, but I would be a little more specific. Something like "AddFlagsToSet" or "AddToFlagSet".

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,225 @@
...
+	cfg ProxyConfig
+
+	// configCh carries the target from Configure to Run; sent on at most once.
+	configCh chan stubapi.ProxyTargetConfig
+	// startedCh carries the result of starting the listener (nil on success).
+	// Run sends exactly one value then closes: nil on success, an error on
+	// Start failure or shutdown-before-configuration.
+	startedCh chan error
+
+	// target is the most recently delivered backend target; guarded by stateMu
```
MikeSpreitzer, line 71: The words "most recently" here could be deleted, because only one is ever delivered, right?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,225 @@
...
+	logger := klog.FromContext(ctx).WithName("proxy-server")
+	logger.Info("Starting TCP proxy server", "port", s.cfg.Port, "dialTimeout", s.cfg.DialTimeout)
+
+	var tgt stubapi.ProxyTargetConfig
+	select {
+	case tgt = <-s.configCh:
+		logger.V(2).Info("Received proxy target from Configure", "target", tgt)
+	case <-ctx.Done():
+		logger.V(2).Info("Context cancelled before proxy was configured")
+		s.startedCh <- fmt.Errorf("proxy shutdown cancelled before configuration")
```
MikeSpreitzer, line 101: "shutdown cancelled" seems a bit off. It was the startup that was interrupted, not the shudown. Maybe something like "proxy shut down before configuration"?

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,225 @@
...
+		_ = json.NewEncoder(w).Encode(cur)
+		return
+	case http.MethodPut:
+		// fall through to body handling below
+	default:
+		http.Error(w, "method not allowed, use GET or PUT", http.StatusMethodNotAllowed)
+		return
+	}
+
+	defer func() { _ = r.Body.Close() }()
```
MikeSpreitzer, line 173: This is not a bug, but it strikes me as a bit odd. Why do this for PUT but not the other methods? Did you know that it is not necessary to do at all? https://pkg.go.dev/net/http@go1.26.4#Request says the following.

```
	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
```

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,225 @@
...
+	body, err := io.ReadAll(r.Body)
```
MikeSpreitzer, line 174: This is an unnecessary opportunity for a hostile client to cause an out-of-memory failure here. This could use `json.NewDecoder(r.Body)` instead, right?

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,300 @@
...
+			}(conn)
+		}
+	}()
+
+	return addr, port, func() {
+		_ = ln.Close()
+		connsMu.Lock()
+		defer connsMu.Unlock()
+		for _, c := range conns {
+			_ = c.Close()
```
MikeSpreitzer, line 100: Is it OK to`Close` a `Conn` that has already been `Close`d? Is it OK to have two concurrent calls to `Close`?

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	_ = conn.SetReadDeadline(time.Now().Add(testReadTimeout))
+	got := make([]byte, len(testMsg))
+	if _, err := io.ReadFull(conn, got); err != nil {
+		t.Fatalf("failed to read echo: %v", err)
+	}
+	if string(got) != testMsg {
+		t.Errorf("expected echo of %q, got %q", testMsg, got)
+	}
+	_ = conn.SetReadDeadline(time.Now().Add(100 * time.Millisecond))
```
MikeSpreitzer, line 156: I would wait a bit longer for misbehavior. Remember that CI servers can be overloaded.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,300 @@
...
+	var got stubapi.ProxyTargetConfig
+	if err := json.Unmarshal(w3.Body.Bytes(), &got); err != nil {
+		t.Fatalf("failed to parse response JSON: %v", err)
+	}
+	if got != body {
+		t.Errorf("GET returned %+v, want %+v", got, body)
+	}
+}
+
+func TestConfigure_DifferentConfigReturnsConflict(t *testing.T) {
```
MikeSpreitzer, line 225: This test overlaps substantially with TestConfigure_GetStatus; I would combine the two.

test/e2e/test-cases.sh
```diff
@@ -532,6 +532,71 @@ check_gpu_pin $req4
...
+# Reverse Proxy Initialization and Forwarding
+# ---------------------------------------------------------------------------
+
+intro_case Reverse Proxy Initialization and Forwarding
+
+# This test verifies that the dual-pods controller initializes the TCP proxy
+# on the requester pod after binding, and that the proxy configuration points
+# to the correct launcher pod IP.
+#
+# Starting state: $req4 is bound to $launcher1, both Ready.
```
MikeSpreitzer, line 566: Also assumed: the ISC of `$req4` is `$isc`.

test/e2e/test-cases.sh
```diff
@@ -532,6 +532,71 @@ check_gpu_pin $req4
...
+fi
+
+echo "Proxy is configured: $proxy_resp"
+
+# Verify proxy config matches expected JSON using jq
+launcher1_ip=$(kubectl get pod "$launcher1" -n "$NS" -o jsonpath='{.status.podIP}')
+launcher1_port=$(kubectl get inferenceserverconfig "$isc" -n "$NS" -o jsonpath='{.spec.modelServerConfig.port}')
+
+expected=$(printf '{"address":"%s","port":%d}' "$launcher1_ip" "$launcher1_port")
+if ! echo "$proxy_resp" | jq -e --argjson expected "$expected" '. == $expected' >/dev/null 2>&1; then
```
MikeSpreitzer, line 595: FYI, in `bash` you can write `&> /dev/null`.

pkg/server/requester/proxy/server_test.go
MikeSpreitzer: It is not necessary for correctness, but it would help with latency. In the case of the controller being unsure of whether the proxy has been configured, the controller could just PUT the config without having to first do a GET.

MikeSpreitzer (dismissed):
I think that this is good enough to merge.
I left a few independent comments.
I would prefer to see the performance report updated with measurements of this edition.

This feature is not part of our original plan for Milestone 3, so I am not merging it without further discussion.

delavet:
> I think that this is good enough to merge. I left a few independent comments. I would prefer to see the performance report updated with measurements of this edition.
> 
> This feature is not part of our original plan for Milestone 3, so I am not merging it without further discussion.

Thanks! I will update the report in this week

MikeSpreitzer:
Holding until after we cut the 0.6.0 release.

delavet:
FYI. I re-ran the experiments described in https://docs.google.com/document/d/1qX4KtcTJEfdatgVmtsveOMfjm883MG-HW-e7ddmlqkA using the latest code implementation, and appended the new results (version v4.1) to the end of that document. The conclusion remains unchanged—TCP Proxy introduces no observable overhead.

MikeSpreitzer:
We want this PR.
@delavet, please rebase.

delavet:
> We want this PR. @delavet, please rebase.

As suggested, I have split this PR into two parts. The first part is located at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/719, and the second part will be submitted after that PR is merged.

PR #503, opened 2026-05-17
avoid to reshape

### LINKED ISSUE #460: Avoid to reshape
BTW, it is wasteful to have buildDesiredStateFromPolicies create a data structure that is then immediately reshaped. It would be better to have buildDesiredStateFromPolicies directly create the useful shape of data structure. Can be dealt with in a later PR.

_Originally posted by @MikeSpreitzer in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3145098534_

osswangxining:
fixes #460 

Previous comments:

 it is wasteful to have buildDesiredStateFromPolicies create a data structure that is then immediately reshaped. It would be better to have buildDesiredStateFromPolicies directly create the useful shape of data structure.

### DIFF
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -216,7 +216,7 @@ func (ctl *controller) reconcileFromPolicies(ctx context.Context) (error, bool) 
 		return err, true
 	}
 
-	logger.Info("Final population policy", "policy", MapToLoggable(populationPolicy))
+	logger.Info("Final population policy", "policy", populationPolicy)
 
 	// Adjust launcher pods according to final requirements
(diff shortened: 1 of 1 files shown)


osswangxining:
@MikeSpreitzer  this is fixing previous comments: avoid to reshape, pls give your review again, thanks in advance.

MikeSpreitzer (approved):
This is an improvement.

I suggest a couple of further improvements, which you could do in this PR or later.

(A) It is unnecessary to accumulate `nodeDesiredGroup.keys`, because that is entirely redundant with the content of `nodeDesiredGroup.desired`.

(B) Those keys do not need to contain the node name, since it is fixed for every `nodeDesiredGroup`. It is positively confusing to use keys that have a fixed field.

PR #504, opened 2026-05-19
Enforce per-launcher capacity for vLLM instances

### LINKED ISSUE #246: `MaxSleepingInstances` for launcher-based server-providing Pods

waltforme:
This PR enforces a per-launcher capacity for vLLM instances.

Closes #246.

### DIFF
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -661,6 +661,15 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	return ctl.ensureReqStatus(ctx, requestingPod, serverDat)
 }
 
+type launcherReclaimPlan struct {
+	launcherPod *corev1.Pod
+	launcherDat *launcherData
+	victims     []string
+	lruID       string
+	lruKnown    bool
+	lruTime     time.Time
+}
+
 // selectOrReclaimLauncherPod evaluates matching launcher Pods and selects one
 // for fulfilling a request. Priority 1 is a launcher with a sleeping vLLM
 // instance matching targetInstanceID. Priority 2 is a launcher with capacity
@@ -681,12 +690,7 @@ func (ctl *controller) selectOrReclaimLauncherPod(
 
 	var candidateWithCapacity *corev1.Pod
 	var somePodsNotReady bool
-	type reclaimPlan struct {
-		launcherPod *corev1.Pod
-		launcherDat *launcherData
-		victims     []string
-	}
-	var bestReclaimPlan *reclaimPlan
+	var bestReclaimPlan *launcherReclaimPlan
 
 	for _, podAny := range launcherPodAnys {
 		launcherPod := podAny.(*corev1.Pod)
@@ -716,8 +720,8 @@ func (ctl *controller) selectOrReclaimLauncherPod(
 
 		launcherDat := ctl.getLauncherData(nodeDat, launcherPod.Name)
 		hasSleepingInstance := false
-		portConflictID := ""
-		candidateVictims := make([]string, 0, len(insts.Instances))
+		portConflictVictims := make([]string, 0, 1)
+		otherVictims := make([]string, 0, len(insts.Instances))
 		hasUnusablePort := false
 		for _, inst := range insts.Instances {
 			instPort, err := getVLLMInstancePort(inst)
@@ -737,9 +741,10 @@ func (ctl *controller) selectOrReclaimLauncherPod(
 				}
 				continue
 			}
-			candidateVictims = append(candidateVictims, inst.InstanceID)
 			if instPort == desiredPort {
-				portConflictID = inst.InstanceID
+				portConflictVictims = append(portConflictVictims, inst.InstanceID)
+			} else {
+				otherVictims = append(otherVictims, inst.InstanceID)
 			}
 		}
 		if hasUnusablePort {
@@ -756,7 +761,7 @@ func (ctl *controller) selectOrReclaimLauncherPod(
 		}
 
 		// Check if this launcher has capacity for a new instance
-		if portConflictID == "" && insts.TotalInstances <= maxOthers {
+		if len(portConflictVictims) == 0 && insts.TotalInstances <= maxOthers {
 			if candidateWithCapacity == nil {
 				// Priority 2: Has capacity for new instance
 				logger.V(5).Info("Found launcher with capacity for new instance",
@@ -769,23 +774,30 @@ func (ctl *controller) selectOrReclaimLauncherPod(
 		}
 
 		toDelete := max(insts.TotalInstances-maxOthers, 1)
-		if portConflictID != "" {
-			// Deleting a different vLLM instance would leave the desired port
-			// unavailable, so only the conflicting vLLM instance is eligible.
-			candidateVictims = []string{portConflictID}
+		victims := slices.Clone(portConflictVictims)
+		if len(victims) > 0 {
+			// The conflicting vLLM instance must be deleted; deleting only
+			// other vLLM instances would leave the desired port unavailable.
+			toPick := toDelete - len(victims)
+			victims = append(victims, pickInstanceVictims(otherVictims, launcherDat.Instances, toPick)...)
+		} else {
+			victims = pickInstanceVictims(otherVictims, launcherDat.Instances, toDelete)
 		}
-		victims := pickInstanceVictims(candidateVictims, launcherDat.Instances, toDelete)
 		if len(victims) < toDelete {
 			logger.V(5).Info("Not enough deletion candidates to reclaim launcher capacity",
 				"name", launcherPod.Name, "need", toDelete, "have", len(victims))
 			continue
 		}
-		plan := &reclaimPlan{
+		lruID, lruKnown, lruTime := reclaimPlanLRU(victims, launcherDat.Instances)
+		plan := &launcherReclaimPlan{
 			launcherPod: launcherPod,
 			launcherDat: launcherDat,
 			victims:     victims,
+			lruID:       lruID,
+			lruKnown:    lruKnown,
+			lruTime:     lruTime,
 		}
-		if bestReclaimPlan == nil || compareInstanceLastUsed(victims[0], launcherDat.Instances, bestReclaimPlan.victims[0], bestReclaimPlan.launcherDat.Instances) < 0 {
+		if bestReclaimPlan == nil || compareReclaimPlanLRU(plan, bestReclaimPlan) < 0 {
 			bestReclaimPlan = plan
 		}
 	}
@@ -839,7 +851,7 @@ func pickInstanceVictims(
 	}
 	candidates = slices.Clone(candidates)
 	slices.SortFunc(candidates, func(a, b string) int {
-		return compareInstanceLastUsed(a, knownLastUsed, b, knownLastUsed)
+		return compareInstanceLastUsed(a, b, knownLastUsed)
 	})
 	if limit > len(candidates) {
 		limit = len(candidates)
@@ -847,20 +859,45 @@ func pickInstanceVictims(
 	return candidates[:limit]
 }
 
-func compareInstanceLastUsed(a string, aLastUsed map[string]time.Time, b string, bLastUsed map[string]time.Time) int {
-	ta, oka := aLastUsed[a]
-	tb, okb := bLastUsed[b]
+func reclaimPlanLRU(victims []string, knownLastUsed map[string]time.Time) (string, bool, time.Time) {
+	if len(victims) == 0 {
+		return "", false, time.Time{}
+	}
+	lruID := victims[0]
+	lruTime, lruKnown := knownLastUsed[lruID]
+	for _, victim := range victims[1:] {
+		victimTime, victimKnown := knownLastUsed[victim]
+		if compareLastUsed(victim, victimKnown, victimTime, lruID, lruKnown, lruTime) < 0 {
+			lruID = victim
+			lruKnown = victimKnown
+			lruTime = victimTime
+		}
+	}
+	return lruID, lruKnown, lruTime
+}
+
+func compareReclaimPlanLRU(a, b *launcherReclaimPlan) int {
+	return compareLastUsed(a.lruID, a.lruKnown, a.lruTime, b.lruID, b.lruKnown, b.lruTime)
+}
+
+func compareInstanceLastUsed(a, b string, knownLastUsed map[string]time.Time) int {
+	ta, oka := knownLastUsed[a]
+	tb, okb := knownLastUsed[b]
+	return compareLastUsed(a, oka, ta, b, okb, tb)
+}
+
+func compareLastUsed(a string, aKnown bool, aTime time.Time, b string, bKnown bool, bTime time.Time) int {
 	switch {
-	case oka && okb:
-		if ta.Before(tb) {
+	case aKnown && bKnown:
+		if aTime.Before(bTime) {
 			return -1
 		}
-		if tb.Before(ta) {
+		if bTime.Before(aTime) {
 			return 1
 		}
-	case oka && !okb:
+	case aKnown && !bKnown:
 		return -1
-	case !oka && okb:
+	case !aKnown && bKnown:
 		return 1
 	}
 	return strings.Compare(a, b)
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -818,10 +818,10 @@ func (ctl *controller) selectOrReclaimLauncherPod(
 		}
 		for _, victim := range bestReclaimPlan.victims {
 			if _, err := lClient.DeleteInstance(ctx, victim); err != nil && !IsInstanceNotFoundError(err) {
-				return nil, false, true, fmt.Errorf("delete instance %q from launcher Pod %q: %w", victim, bestReclaimPlan.launcherPod.Name, err)
+				return nil, false, true, fmt.Errorf("failed to delete instance %q from launcher Pod %q: %w", victim, bestReclaimPlan.launcherPod.Name, err)
 			}
 			delete(bestReclaimPlan.launcherDat.Instances, victim)
-			logger.V(4).Info("Deleted vLLM instance to reclaim launcher capacity",
+			logger.V(4).Info("Ensured vLLM instance absent to reclaim launcher capacity",
 				"launcherPod", bestReclaimPlan.launcherPod.Name, "instanceID", victim, "maxOthers", maxOthers)
 		}
 		return bestReclaimPlan.launcherPod, false, false, nil
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -609,35 +609,22 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		// then those with capacity for new instances.
 		// Note that multiple vLLM instances could exist in one launcher Pod, but at most one instance could be awake at a time.
 
-		launcherPod, hasSleepingInstance, someNotReady, err := ctl.selectBestLauncherPod(ctx, launcherPodAnys, desiredInstanceState.instanceID, desiredPort, effectiveMaxInstances(lc)-1, nodeDat)
+		launcherPod, hasSleepingInstance, retry, err := ctl.selectOrReclaimLauncherPod(ctx, launcherPodAnys, desiredInstanceState.instanceID, desiredPort, effectiveMaxInstances(lc)-1, nodeDat)
 		if err != nil {
 			return err, true
 		}
-		if someNotReady {
+		if retry {
 			logger.V(4).Info("Launcher Pods exist but some are not ready yet, will retry later")
 			return nil, true
 		}
-		if launcherPod == nil {
-			logger.V(5).Info("No suitable launcher Pod found with sleeping instance or necessary capacity, attempting to reclaim capacity")
-			reclaimed, retry, err := ctl.reclaimLauncherCapacity(ctx, launcherPodAnys, desiredInstanceState.instanceID, desiredPort, effectiveMaxInstances(lc)-1, nodeDat)
-			if err != nil {
-				return err, true
-			}
-			if retry {
-				return nil, true
-			}
-			if reclaimed != nil {
-				logger.V(4).Info("Reclaimed launcher capacity, binding to launcher Pod", "name", reclaimed.Name)
-				return ctl.bind(ctx, serverDat, requestingPod, reclaimed, desiredInstanceState, true)
-			}
-			// Fall through to create new launcher Pod
-		} else {
+		if launcherPod != nil {
 			// Bind first, then rely on informer notification to trigger re-reconciliation.
 			// The "bound provider" path will handle instance creation/waking.
 			// This ensures the invariant: vllm awake implies provider Pod is bound.
 			logger.V(5).Info("Selected launcher Pod, binding first", "name", launcherPod.Name, "hasSleepingInstance", hasSleepingInstance)
 			return ctl.bind(ctx, serverDat, requestingPod, launcherPod, desiredInstanceState, true)
 		}
+		// Fall through to create new launcher Pod.
 	}
 	// Remains: Zero matching launcher Pods, or the matching launcher Pod cannot host more instances to fulfill the request.
 	// Make a new launcher Pod.
@@ -674,18 +661,18 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	return ctl.ensureReqStatus(ctx, requestingPod, serverDat)
 }
 
-// selectBestLauncherPod evaluates all matching launcher Pods and selects the 'best' one for fulfilling a request.
-// Currently the definition of 'best' is radically simple:
-// Priority 1: Launcher with a sleeping instance matching iscHash (fastest - just wake up);
-// Priority 2: Launcher with capacity for a new instance (slower - need to create);
-// Otherwise, no launcher Pod is selected.
-// Returns (selectedPod, hasSleepingInstance, somePodsNotReady, error).
-// Returns (nil, false, false, nil) if no suitable launcher found and all pods are ready or failed.
-// Returns (nil, false, true, nil) if there are pods not ready yet - caller should retry later.
-func (ctl *controller) selectBestLauncherPod(
+// selectOrReclaimLauncherPod evaluates matching launcher Pods and selects one
+// for fulfilling a request. Priority 1 is a launcher with a sleeping vLLM
+// instance matching targetInstanceID. Priority 2 is a launcher with capacity
+// for a new vLLM instance. Priority 3 is reclaiming capacity by deleting the
+// least-recently-used eligible vLLM instance across all scanned launcher Pods.
+// Returns (nil, false, false, nil) if no suitable launcher is found and all
+// launcher Pods are ready or failed. Returns (nil, false, true, nil) when
+// there are launcher Pods not ready yet and the caller should retry later.
+func (ctl *controller) selectOrReclaimLauncherPod(
 	ctx context.Context,
 	launcherPodAnys []interface{},
-	iscHash string,
+	targetInstanceID string,
 	desiredPort int32,
 	maxOthers int,
 	nodeDat *nodeData,
@@ -694,6 +681,12 @@ func (ctl *controller) selectBestLauncherPod(
 
 	var candidateWithCapacity *corev1.Pod
 	var somePodsNotReady bool
+	type reclaimPlan struct {
+		launcherPod *corev1.Pod
+		launcherDat *launcherData
+		victims     []string
+	}
+	var bestReclaimPlan *reclaimPlan
 
 	for _, podAny := range launcherPodAnys {
 		launcherPod := podAny.(*corev1.Pod)
@@ -701,9 +694,8 @@ func (ctl *controller) selectBestLauncherPod(
 		if launcherPod.Status.Phase == corev1.PodFailed || launcherPod.DeletionTimestamp != nil {
 			continue
 		}
-		requesterParts := strings.Split(launcherPod.Annotations[requesterAnnotationKey], " ")
-		if len(requesterParts) == 2 {
-			logger.V(5).Info("Launcher Pod already bound to another requester, skipping", "name", launcherPod.Name, "boundRequester", requesterParts[1])
+		if requesterValue := launcherPod.Annotations[requesterAnnotationKey]; requesterValue != "" {
+			logger.V(5).Info("Launcher Pod already bound to another requester, skipping", "name", launcherPod.Name, "boundRequester", requesterValue)
 			continue
 		}
 
@@ -722,9 +714,11 @@ func (ctl *controller) selectBestLauncherPod(
 		}
 		insts := syncResult.instances
 
-		// Check if this launcher has a sleeping instance matching the iscHash
+		launcherDat := ctl.getLauncherData(nodeDat, launcherPod.Name)
 		hasSleepingInstance := false
-		hasPortConflict := false
+		portConflictID := ""
+		candidateVictims := make([]string, 0, len(insts.Instances))
+		hasUnusablePort := false
 		for _, inst := range insts.Instances {
 			instPort, err := getVLLMInstancePort(inst)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/dual-pods/inference-server.go
```diff
@@ -629,13 +640,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		}
 	}
 	// Remains: Zero matching launcher Pods, or the matching launcher Pod cannot host more instances to fulfill the request.
-
-	// TODO(waltforme): enforceSleeperBudget should be revised for launcher-based server-providing Pods
-	err, retry := ctl.enforceSleeperBudget(ctx, serverDat, requestingPod, effectiveMaxInstances(lc)-1)
```
MikeSpreitzer, line 634: We're not destroying milestone 2 functionality yet, right?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -785,6 +790,159 @@ func (ctl *controller) selectBestLauncherPod(
...
+
+	for _, podAny := range launcherPodAnys {
+		launcherPod := podAny.(*corev1.Pod)
+
+		if launcherPod.Status.Phase == corev1.PodFailed || launcherPod.DeletionTimestamp != nil {
+			continue
+		}
+		// Skip launcher Pods already bound to another requester. Same filter as
+		// selectBestLauncherPod.
+		requesterParts := strings.Split(launcherPod.Annotations[requesterAnnotationKey], " ")
```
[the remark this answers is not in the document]
MikeSpreitzer: Not an issue, because these annotations are reliably produced by the same package, not sloppy users.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -785,6 +790,159 @@ func (ctl *controller) selectBestLauncherPod(
...
+		if len(requesterParts) == 2 {
```
MikeSpreitzer: I would make the test simpler, removing dependence on unnecessary detail (the syntax of the value of this annotation). I would test only for the presence (i.e., non-emtpy value) of this annotation.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -785,6 +790,159 @@ func (ctl *controller) selectBestLauncherPod(
...
+			continue
+		}
+
+		launcherBaseURL := fmt.Sprintf("http://%s:%d", launcherPod.Status.PodIP, ctlrcommon.LauncherServicePort)
+		lClient, err := NewLauncherClient(launcherBaseURL)
+		if err != nil {
+			return nil, true, err
+		}
+		for _, victim := range victims {
+			if _, err := lClient.DeleteInstance(ctx, victim); err != nil && !IsInstanceNotFoundError(err) {
```
MikeSpreitzer, line 803: Is there a need to specially handle an "already gone" error here?

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+kubectl scale rs $rs -n "$NS" --replicas=0
+
+expect "kubectl get pods -n $NS -o name -l app=dp-example,instance=$inst | wc -l | grep -w 0"
+
+# Capture launcher pod count so we can later assert reclaim did not spawn a new one.
+launcher_count_before_reclaim=$(kubectl get pods -n "$NS" -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | wc -l)
+echo launcher_count_before_reclaim = $launcher_count_before_reclaim
+
+# Target launcher is still around and unbound.
+kubectl get pods -n "$NS" -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | grep -x "pod/$launcher1"
```
MikeSpreitzer: Wouldn't it be simpler to just get the pod, since its name is known?


```suggestion
kubectl get pod -n "$NS" "$launcher1"
```

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "" ]'
+
+# Confirm launcher is at the cap before reclaim.
+launcher_instances_at_cap=$(kubectl exec -n "$NS" $launcher1 -- python3 -c 'import json,urllib.request; print(json.load(urllib.request.urlopen("http://127.0.0.1:8001/v2/vllm/instances"))["total_instances"])')
```
MikeSpreitzer: This variable name assumes that the test passes. What the variable actually holds is the current count of instances, which the rest of the test does not ASSUME equals the cap.

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+# than create a new one.
+kubectl scale rs $rs -n "$NS" --replicas=1
+
+expect "kubectl get pods -n $NS -o name -l app=dp-example,instance=$inst | wc -l | grep -w 1"
+req_reclaim=$(kubectl get pods -n "$NS" -o name -l app=dp-example,instance=$inst | sed s%pod/%%)
+echo "Server-requesting Pod (post-reclaim) is $req_reclaim"
+
+# Launcher pod set is unchanged — no new launcher was created.
+expect "kubectl get pods -n $NS -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | wc -l | grep -w $launcher_count_before_reclaim"
+kubectl get pods -n "$NS" -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | grep -x "pod/$launcher1"
```
MikeSpreitzer:
```suggestion
kubectl get pods -n "$NS" "$launcher1"
```

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+
+# Bidirectional binding to the same launcher.
+expect '[ "$(kubectl get pod -n '"$NS"' $req_reclaim -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$launcher1" ]'
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$req_reclaim" ]'
+
+# Requester is annotated for isc3.
+expect '[ "$(kubectl get pod -n '"$NS"' $req_reclaim -o jsonpath={.metadata.annotations.dual-pods\\.llm-d\\.ai/inference-server-config})" == "'$isc3'" ]'
```
MikeSpreitzer: Since this is established by the ReplicaSet, it could be checked as soon as the Pod exists.

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+
+date
+kubectl wait --for condition=Ready pod/$req_reclaim -n "$NS" --timeout=120s
```
MikeSpreitzer: We are finding that waiting 120s is making tests flaky, I recommend at least 180s.

test/e2e/test-cases.sh
```diff
@@ -565,8 +627,12 @@ intro_case Delete Obsolete Sleeping Instances After ISC Update
...
 echo "Launcher had $old_total_instances instance(s) before updating ISC to make a sleeping instance obsolete"
 
-# Mutate isc in a hash-relevant way so its sleeping instance becomes obsolete.
-kubectl patch inferenceserverconfig "$isc" -n "$NS" --type=merge -p='{"spec":{"modelServerConfig":{"options":"--model HuggingFaceTB/SmolLM2-360M-Instruct --served-model-name obsolete-after-update"}}}'
+# At this point isc3 is the sleeping instance (isc2 is the awake one bound to
+# req_post_restart). Mutate isc3 in a hash-relevant way so its sleeping
+# instance becomes obsolete. We append --served-model-name to the original
+# options (rather than hardcoding the model) so the model itself stays valid.
+original_isc3_options=$(kubectl get inferenceserverconfig "$isc3" -n "$NS" -o jsonpath='{.spec.modelServerConfig.options}')
+kubectl patch inferenceserverconfig "$isc3" -n "$NS" --type=merge -p='{"spec":{"modelServerConfig":{"options":"'"$original_isc3_options"' --served-model-name obsolete-after-update"}}}'
```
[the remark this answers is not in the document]
MikeSpreitzer: I think that this is a legitimate complaint. I think that a simpler fix would be to use `jq` to produce the proper JSON.


```suggestion
enhanced_options=$(kubectl get inferenceserverconfig "$isc3" -n "$NS" -o jsonpath='{.spec.modelServerConfig}' | jq '.options + " --served-model-name obsolete-after-update"')
kubectl patch inferenceserverconfig "$isc3" -n "$NS" --type=merge -p='{"spec":{"modelServerConfig":{"options":'"$enhanced_options"' }}}'
```

pkg/controller/dual-pods/inference-server.go
```diff
@@ -618,7 +618,18 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 			return nil, true
 		}
 		if launcherPod == nil {
-			logger.V(5).Info("No suitable launcher Pod found with sleeping instance or necessary capacity")
+			logger.V(5).Info("No suitable launcher Pod found with sleeping instance or necessary capacity, attempting to reclaim capacity")
+			reclaimed, retry, err := ctl.reclaimLauncherCapacity(ctx, launcherPodAnys, desiredInstanceState.instanceID, desiredPort, effectiveMaxInstances(lc)-1, nodeDat)
```
MikeSpreitzer: Rather than make this a separate function, with its own call to `ctl.syncLauncherInstances`, I would combine this functionality into `selectBestLauncherPod` (and probably change its name to reflect its expanded mission).

pkg/controller/dual-pods/inference-server.go
```diff
@@ -785,6 +790,159 @@ func (ctl *controller) selectBestLauncherPod(
...
+		if err != nil || retry {
+			somePodsNotReady = true
+			continue
+		}
+		insts := syncResult.instances
+
+		// Skip launchers where a non-target instance already uses desiredPort.
+		// Reclaim doesn't try to resolve port collisions; it would have to
+		// force-delete the conflicting instance, and selectBestLauncherPod's
+		// own port filter already declined this launcher for the same reason.
```
MikeSpreitzer: I don't understand the reasoning here.
If we are reclaiming at all, selectBestLauncherPod did not find a sleeper to waken and did not find a launcher with available capacity. So some sleeping instance has to be deleted. I see nothing wrong with deleting a sleeping instance that happens to be using any particular port.

In a related matter: I see that the current draft picks the first usable launcher in `launcherPodAnys` without regard to the LRU preference among vllm instances. With reclaimLauncherCapacity combined into selectBestLauncherPod, the one scan can produce a list of all vllm instances among all the launchers and pick the least recently used one. With one wrinkle regarding port number: for a launcher that has an instance using the target port, only that instance should be a candidate.

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
waltforme: Right. The removed lines were only on the launcher-based code path.

pkg/controller/dual-pods/inference-server.go
waltforme: If instance is already gone, we don't need to report an error.

pkg/controller/dual-pods/inference-server.go
waltforme: Refactored as suggested.

pkg/controller/dual-pods/inference-server.go
waltforme: Merged as suggested.

test/e2e/test-cases.sh
waltforme: Used 180s.

test/e2e/test-cases.sh
waltforme: Moved this check ahead.

test/e2e/test-cases.sh
waltforme: Used the suggestion.

test/e2e/test-cases.sh
waltforme: Renamed.

test/e2e/test-cases.sh
waltforme: Simplified.

pkg/controller/dual-pods/inference-server.go
waltforme: Simplified.

test/e2e/test-cases.sh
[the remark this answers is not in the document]
waltforme: Simplified.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -609,33 +609,25 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 		// Note that multiple vLLM instances could exist in one launcher Pod, but at most one instance could be awake at a time.
 
-		launcherPod, hasSleepingInstance, someNotReady, err := ctl.selectBestLauncherPod(ctx, launcherPodAnys, desiredInstanceState.instanceID, desiredPort, effectiveMaxInstances(lc)-1, nodeDat)
+		launcherPod, hasSleepingInstance, retry, err := ctl.selectOrReclaimLauncherPod(ctx, launcherPodAnys, desiredInstanceState.instanceID, desiredPort, effectiveMaxInstances(lc)-1, nodeDat)
 		if err != nil {
 			return err, true
 		}
-		if someNotReady {
+		if retry {
 			logger.V(4).Info("Launcher Pods exist but some are not ready yet, will retry later")
```
MikeSpreitzer: Since the variable name is less specific now, I wonder if the log message should be too.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -669,18 +661,18 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
-// Returns (nil, false, true, nil) if there are pods not ready yet - caller should retry later.
-func (ctl *controller) selectBestLauncherPod(
+// selectOrReclaimLauncherPod evaluates matching launcher Pods and selects one
+// for fulfilling a request. Priority 1 is a launcher with a sleeping vLLM
+// instance matching targetInstanceID. Priority 2 is a launcher with capacity
+// for a new vLLM instance. Priority 3 is reclaiming capacity by deleting the
+// least-recently-used eligible vLLM instance across all scanned launcher Pods.
+// Returns (nil, false, false, nil) if no suitable launcher is found and all
+// launcher Pods are ready or failed. Returns (nil, false, true, nil) when
+// there are launcher Pods not ready yet and the caller should retry later.
```
MikeSpreitzer: This does not describe the first `bool` in the return tuple.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +728,144 @@ func (ctl *controller) selectBestLauncherPod(
 					"annotations", inst.Annotations,
 					"options", inst.Options,
 					"err", err)
-				hasPortConflict = true
+				hasUnusablePort = true
 				break
```
MikeSpreitzer: This same controller provides the port annotations in question here, so normally `err` will always be `nil`. If ever `err != nil` (which will only happen in case of bugs or security breach), the instance will not be used for anything. I think the better response in this situation is to delete the instance. Can be adjusted in a later PR.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: I am sorry, I was not careful enough in my previous remark --- I assumed that certainly only one instance needed to be deleted. As you are clearly aware, this is not a justified assumption.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +728,144 @@ func (ctl *controller) selectBestLauncherPod(
...
+			// Don't return yet - keep looking for sleeping instances (higher priority).
+			continue
+		}
+
+		toDelete := max(insts.TotalInstances-maxOthers, 1)
+		if portConflictID != "" {
+			// Deleting a different vLLM instance would leave the desired port
+			// unavailable, so only the conflicting vLLM instance is eligible.
+			candidateVictims = []string{portConflictID}
+		}
```
MikeSpreitzer: I am sorry, my previous comment on the topic incorrectly assumed that the number instances that need to be deleted is certainly 1. After reviewing the latest draft of this PR, I think the logic should be as follows, for cases that get this far.

- Accumulate two lists of candidates: one list has at most one member --- the one with the conflicting inference port number --- and the other list has all the rest.
- the chosen victims should start with the first list and continue with the needed number from the sort of the second list.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +728,144 @@ func (ctl *controller) selectBestLauncherPod(
...
+		victims := pickInstanceVictims(candidateVictims, launcherDat.Instances, toDelete)
+		if len(victims) < toDelete {
+			logger.V(5).Info("Not enough deletion candidates to reclaim launcher capacity",
+				"name", launcherPod.Name, "need", toDelete, "have", len(victims))
+			continue
+		}
+		plan := &reclaimPlan{
```
MikeSpreitzer: I think that it would save a lot of complexity at comparison time if the LRU timestamp for this plan were computed and stored here.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +728,144 @@ func (ctl *controller) selectBestLauncherPod(
...
-	// Found sleeper nor capable ones, but there are pods not ready yet, signal caller to retry later
+	if bestReclaimPlan != nil {
+		launcherBaseURL := fmt.Sprintf("http://%s:%d", bestReclaimPlan.launcherPod.Status.PodIP, ctlrcommon.LauncherServicePort)
+		lClient, err := NewLauncherClient(launcherBaseURL)
+		if err != nil {
+			return nil, false, true, err
+		}
+		for _, victim := range bestReclaimPlan.victims {
+			if _, err := lClient.DeleteInstance(ctx, victim); err != nil && !IsInstanceNotFoundError(err) {
+				return nil, false, true, fmt.Errorf("delete instance %q from launcher Pod %q: %w", victim, bestReclaimPlan.launcherPod.Name, err)
```
MikeSpreitzer: I would make the error message say "_failed to_ delete ..."

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +728,144 @@ func (ctl *controller) selectBestLauncherPod(
...
+			}
+			delete(bestReclaimPlan.launcherDat.Instances, victim)
+			logger.V(4).Info("Deleted vLLM instance to reclaim launcher capacity",
```
MikeSpreitzer: A definite "Deleted" is not quite warranted here, since this statement is also reached if this controller did not do the deletion (something else did concurrently).

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+
+# Capture launcher pod count so we can later assert reclaim did not spawn a new one.
+launcher_count_before_reclaim=$(kubectl get pods -n "$NS" -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | wc -l)
+echo launcher_count_before_reclaim = $launcher_count_before_reclaim
+
+# Target launcher is still around and unbound.
+kubectl get pod -n "$NS" "$launcher1"
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "" ]'
+
+# Confirm launcher is at the cap before reclaim.
```
MikeSpreitzer: s/launcher/launcher1/

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+# Scale up. Controller must reclaim a slot in the existing launcher rather
+# than create a new one.
+kubectl scale rs $rs -n "$NS" --replicas=1
+
+expect "kubectl get pods -n $NS -o name -l app=dp-example,instance=$inst | wc -l | grep -w 1"
+req_reclaim=$(kubectl get pods -n "$NS" -o name -l app=dp-example,instance=$inst | sed s%pod/%%)
+echo "Server-requesting Pod (post-reclaim) is $req_reclaim"
+
+# Requester is annotated for isc3.
+expect '[ "$(kubectl get pod -n '"$NS"' $req_reclaim -o jsonpath={.metadata.annotations.dual-pods\\.llm-d\\.ai/inference-server-config})" == "'$isc3'" ]'
```
MikeSpreitzer: This is something to test once without any waiting, it will not change over time.

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+
+# Launcher pod set is unchanged — no new launcher was created.
+expect "kubectl get pods -n $NS -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | wc -l | grep -w $launcher_count_before_reclaim"
```
MikeSpreitzer: (A) Before checking the launcher count, this test should give the controller time to misbehave. I recommend first waiting for `$req_reclaim` to be bound.

(B) For ease of understanding by someone reading the output, I recommend formulating the check as first fetching the current number and then failing with an error message if the number is wrong.

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,68 @@ check_gpu_pin $req4
...
+kubectl get pod -n "$NS" "$launcher1"
+
+# Bidirectional binding to the same launcher.
+expect '[ "$(kubectl get pod -n '"$NS"' $req_reclaim -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$launcher1" ]'
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "$req_reclaim" ]'
+
+date
+kubectl wait --for condition=Ready pod/$req_reclaim -n "$NS" --timeout=180s
```
MikeSpreitzer: In #491 I made all these timeouts 5 minutes because I observed vllm startup take more than 3 minutes.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -785,6 +790,159 @@ func (ctl *controller) selectBestLauncherPod(
...
+		}
+		// Skip launcher Pods already bound to another requester. Same filter as
+		// selectBestLauncherPod.
+		requesterParts := strings.Split(launcherPod.Annotations[requesterAnnotationKey], " ")
+		if len(requesterParts) == 2 {
+			continue
+		}
+
+		syncResult, err, retry := ctl.syncLauncherInstances(ctx, nodeDat, launcherPod)
+		if err != nil || retry {
```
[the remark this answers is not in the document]
MikeSpreitzer: moot now

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
waltforme: Changed.

pkg/controller/dual-pods/inference-server.go
waltforme: Added.

pkg/controller/dual-pods/inference-server.go
waltforme: Changed as suggested.

pkg/controller/dual-pods/inference-server.go
waltforme: Changed as suggested.

pkg/controller/dual-pods/inference-server.go
waltforme: Changed.

pkg/controller/dual-pods/inference-server.go
waltforme: Used the verb 'Ensured' instead.

test/e2e/test-cases.sh
waltforme: Done.

test/e2e/test-cases.sh
waltforme: Done.

test/e2e/test-cases.sh
waltforme: Changed as suggested.

test/e2e/test-cases.sh
waltforme: Changed to 300s.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -669,18 +661,29 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+	lruID       string
+	lruKnown    bool
+	lruTime     time.Time
+}
+
+// selectOrReclaimLauncherPod evaluates matching launcher Pods and selects one
+// for fulfilling a request. Priority 1 is a launcher with a sleeping vLLM
+// instance matching targetInstanceID. Priority 2 is a launcher with capacity
+// for a new vLLM instance. Priority 3 is reclaiming capacity by deleting the
+// least-recently-used eligible vLLM instance across all scanned launcher Pods.
```
MikeSpreitzer: Since the number that needs to be deleted might be more than 1, there is a two level structure in priority 3, and a choice to be made here. Priority 3 is one chosen launcher and a set (having the right size) of chosen instance deletions for that launcher. What if that number of deletions is more than one and there is great diversity in LRU timestamp among the instances on each launcher?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +734,177 @@ func (ctl *controller) selectBestLauncherPod(
 					"annotations", inst.Annotations,
 					"options", inst.Options,
 					"err", err)
-				hasPortConflict = true
+				hasUnusablePort = true
 				break
```
MikeSpreitzer: IMO it would be simpler to use `continue Label` than encoding it with a variable and `break` and `if ... continue`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -717,9 +720,11 @@ func (ctl *controller) selectBestLauncherPod(
 		}
 		insts := syncResult.instances
 
-		// Check if this launcher has a sleeping instance matching the iscHash
+		launcherDat := ctl.getLauncherData(nodeDat, launcherPod.Name)
 		hasSleepingInstance := false
-		hasPortConflict := false
+		portConflictVictims := make([]string, 0, 1)
+		otherVictims := make([]string, 0, len(insts.Instances))
+		hasUnusablePort := false
```
MikeSpreitzer: If I understand correctly, this means something more like "has unknown port" or "has no defined port". But see also my comment about using `continue Label` instead.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +734,177 @@ func (ctl *controller) selectBestLauncherPod(
...
+					"name", launcherPod.Name,
+					"totalInstances", insts.TotalInstances)
+				candidateWithCapacity = launcherPod
+			}
+			// Don't return yet - keep looking for sleeping instances (higher priority).
+			continue
+		}
+
+		toDelete := max(insts.TotalInstances-maxOthers, 1)
+		victims := slices.Clone(portConflictVictims)
```
MikeSpreitzer: I see no need for `Clone` here. `portConflictVictims` is freshly created for this invocation of this func and will not be used for anything else.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +734,177 @@ func (ctl *controller) selectBestLauncherPod(
...
+		if len(victims) > 0 {
+			// The conflicting vLLM instance must be deleted; deleting only
+			// other vLLM instances would leave the desired port unavailable.
+			toPick := toDelete - len(victims)
+			victims = append(victims, pickInstanceVictims(otherVictims, launcherDat.Instances, toPick)...)
+		} else {
+			victims = pickInstanceVictims(otherVictims, launcherDat.Instances, toDelete)
+		}
```
MikeSpreitzer: This could all be simplified to

```suggestion
		victims := append(portConflictVictims, pickInstanceVictims(otherVictims, launcherDat.Instances, toDelete - len(portConflictVictims) )...)
```

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +734,177 @@ func (ctl *controller) selectBestLauncherPod(
...
+		if len(victims) < toDelete {
+			logger.V(5).Info("Not enough deletion candidates to reclaim launcher capacity",
+				"name", launcherPod.Name, "need", toDelete, "have", len(victims))
+			continue
```
MikeSpreitzer: How could these lines be reached?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +734,177 @@ func (ctl *controller) selectBestLauncherPod(
...
+	})
+	if limit > len(candidates) {
+		limit = len(candidates)
+	}
+	return candidates[:limit]
+}
+
+func reclaimPlanLRU(victims []string, knownLastUsed map[string]time.Time) (string, bool, time.Time) {
+	if len(victims) == 0 {
+		return "", false, time.Time{}
```
MikeSpreitzer: How could this line be reached?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +734,177 @@ func (ctl *controller) selectBestLauncherPod(
...
+	}
+	lruID := victims[0]
+	lruTime, lruKnown := knownLastUsed[lruID]
```
MikeSpreitzer: How could `lruKnown` be `false`?
I see a lot of complexity in this file to deal with the possibility of something missing from the last-used map, but I am dubious that this is even a possibility. If a defense against that possibility is actually needed, I suspect that it would be simpler to have `syncLauncherInstances` ensure that there is an entry for every instance.

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,76 @@ check_gpu_pin $req4
...
+
+expect "kubectl get pods -n $NS -o name -l app=dp-example,instance=$inst | wc -l | grep -w 0"
+
+# Capture launcher pod count so we can later assert reclaim did not spawn a new one.
+launcher_count_before_reclaim=$(kubectl get pods -n "$NS" -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | wc -l)
+echo launcher_count_before_reclaim = $launcher_count_before_reclaim
+
+# Target launcher is still around and unbound.
+kubectl get pod -n "$NS" "$launcher1"
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "" ]'
```
MikeSpreitzer: This can become true if the launcher gets deleted. I would expand the condition tested to include the existence of the launcher Pod.

```console
bash-5.3$ set -euo pipefail

bash-5.3$ if [ "$(kubectl get pod xxx -o jsonpath='{.metadata.labels.fubar}')" == "" ]
> then echo y
> else echo n
> fi
Error from server (NotFound): pods "xxx" not found
y
```

test/e2e/test-cases.sh
MikeSpreitzer: I still see "launcher".

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
waltforme: I somewhat understand the concern. Can you help me to _better_ understand this by an example of 'great diversity in LRU timestamp'?

waltforme:
I revised the PR according to the latest reviews. All comment are addressed except https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/504#discussion_r3291076059 which I need more details.

Also rebased onto `main` to resolve a merge conflict.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: Suppose two candidate launchers there are two candidate launchers for reclamation:
- L1, needing to have two instances removed: I1a, with oldest last-used, and I1b with youngest last-used;
- L2, needing to have one instance removed, L2, with a middling last-used timestamp.

waltforme: I tend to make a choice that deletes as few as possible instances (pick L2 in the example). What do you think?

MikeSpreitzer: So, as I illustrated, there is also another dimension to consider: how many instances to delete.

I suggest that it is preferrable to clean up a launcher that is farther out of compliance.

Another possibility would be to make this function responsible for trimming all launchers that are out of compliance.

MikeSpreitzer: Because this is the only controller that deletes instances, it should not be shy about doing its job.

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,75 @@ check_gpu_pin $req4
...
 
+# ---------------------------------------------------------------------------
+# Per-launcher Instance Cap Enforcement
+# ---------------------------------------------------------------------------
+
+intro_case Per-launcher Instance Cap Enforcement
+
+# Scale requester to 0 so the launcher becomes unbound and both prior instances
+# are sleeping (isc, isc2). The LauncherConfig caps total instances at 2
+# (maxSleepingInstances: 1), so the launcher is at capacity.
```
MikeSpreitzer: The parenthetic remark here is outdated.

waltforme: Fixed.

pkg/controller/dual-pods/inference-server.go
waltforme: I did the 'farther out of compliance' one.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -729,62 +733,148 @@ func (ctl *controller) selectBestLauncherPod(
...
+		return nil
+	}
+	candidates = slices.Clone(candidates)
+	slices.SortFunc(candidates, func(a, b string) int {
+		return compareInstanceLastUsed(a, b, knownLastUsed)
+	})
+	if limit > len(candidates) {
+		limit = len(candidates)
+	}
+	return candidates[:limit]
```
MikeSpreitzer, line 841: FYI, Go now has polymorphic `min` (see https://go.dev/ref/spec#Min_and_max).

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,75 @@ check_gpu_pin $req4
...
+# Confirm launcher1 is at the cap before reclaim.
+launcher_instances_before_reclaim=$(kubectl exec -n "$NS" $launcher1 -- python3 -c 'import json,urllib.request; print(json.load(urllib.request.urlopen("http://127.0.0.1:8001/v2/vllm/instances"))["total_instances"])')
+echo "Launcher $launcher1 has $launcher_instances_before_reclaim instances before reclaim (expected 2)"
+[ "$launcher_instances_before_reclaim" == "2" ]
+
+# Switch the requester to a third ISC for which there is no sleeping instance,
+# so neither the fast wake-up path nor the capacity path applies.
+kubectl patch rs $rs -n "$NS" --type=json -p='[{"op": "replace", "path": "/spec/template/metadata/annotations/dual-pods.llm-d.ai~1inference-server-config", "value": "'$isc3'"}]'
+
+# Scale up. Controller must reclaim a slot in the existing launcher rather
```
MikeSpreitzer, line 517: Maybe "is expected to" would be more clear than "must" here?

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,75 @@ check_gpu_pin $req4
...
+[ "$(kubectl get pod -n "$NS" "$req_reclaim" -o jsonpath='{.metadata.annotations.dual-pods\.llm-d\.ai/inference-server-config}')" == "$isc3" ]
+
+# Wait until the controller binds the requester to a launcher.
+expect '[ -n "$(kubectl get pod -n '"$NS"' $req_reclaim -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" ]'
+
+# Launcher pod set is unchanged — no new launcher was created.
+launcher_count_after_reclaim_bind=$(kubectl get pods -n "$NS" -o name -l dual-pods.llm-d.ai/launcher-config-name=$lc | wc -l | tr -d ' ')
+echo launcher_count_after_reclaim_bind = $launcher_count_after_reclaim_bind
+if [ "$launcher_count_after_reclaim_bind" != "$launcher_count_before_reclaim" ]; then
+    echo "ERROR: expected $launcher_count_before_reclaim launcher pod(s) after reclaim bind, got $launcher_count_after_reclaim_bind" >&2
```
MikeSpreitzer, line 535: Maybe do a `kubectl get pods -l` at this time to list them all. I know that the state dumping later will include launcher pods, but the controller may still be in the midst of a progression of states, so the listing at this moment may be helpful in the case of a bug.

test/e2e/test-cases.sh
```diff
@@ -485,6 +485,75 @@ check_gpu_pin $req4
...
+fi
+kubectl get pod -n "$NS" "$launcher1"
+
+# Bidirectional binding to the same launcher.
+[ "$(kubectl get pod -n "$NS" "$req_reclaim" -o jsonpath='{.metadata.labels.dual-pods\.llm-d\.ai/dual}')" == "$launcher1" ]
+[ "$(kubectl get pod -n "$NS" "$launcher1" -o jsonpath='{.metadata.labels.dual-pods\.llm-d\.ai/dual}')" == "$req_reclaim" ]
+
+date
+kubectl wait --for condition=Ready pod/$req_reclaim -n "$NS" --timeout=300s
+[ "$(kubectl get pod $launcher1 -n "$NS" -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}')" = "True" ]
```
MikeSpreitzer, line 546: This is not wrong, but wouldn't it be simpler to do a kubectl wait for Ready (with a really short timeout)?

MikeSpreitzer (approved):
This is good enough to merge, but I left some independent comments for consideration now or later.
The deletion of an instance with no usable port annotation should get into the upcoming release.

waltforme:
> This is good enough to merge, but I left some independent comments for consideration now or later. The deletion of an instance with no usable port annotation should get into the upcoming release.

Let me merge now so not blocking the release and the planning of demo(s). Will soon submit a follow-up PR to address the latest comments.

PR #505, opened 2026-05-20
Update vllm versions

diegocastanibm:

PR #509, opened 2026-05-22
Update docs and YAML's with new vllm

diegocastanibm:
In #505 , the vllm version was updated for the launcher. Forgot to update the vllm version in the rest of the files.


MikeSpreitzer (approved):
LGTM.
Thanks.

Issue #459, opened 2026-04-28
The order to delete launchers (to be disccussed)


osswangxining:
I think this answer is worth documenting, with a comment here or in a document. The question of which launchers to delete is an interesting one. Can be addressed in a later PR.

_Originally posted by @MikeSpreitzer in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3148620418_

osswangxining:
This is one official projection around llm-d,  which is in incubation phase.  Would have your detailed idea, thanks.

osswangxining:
this has been naturally covered within #510 .  closed for now.

Issue #457, opened 2026-04-27
Refine the reconcile merge process


osswangxining:
Agreed, the two loops could be merged. However, the current structure intentionally separates classification from action for readability, and len(staleUnboundPods) is reused in the log below. Happy to consolidate in a follow-on as suggested.

_Originally posted by @osswangxining in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3145672928_

osswangxining:
this has been naturally covered in #510

Issue #450, opened 2026-04-23
Calling BuildLauncherPodFromTemplate improvement


osswangxining:
Calling BuildLauncherPodFromTemplate here is a lot of wasted work. I see that it is necessary because the only hash available takes the Node name into account. If BuildLauncherPodFromTemplate also returned an additional hash that does NOT take the Node name into account then, for purposes of this consistency checking, it would only be necessary to call BuildLauncherPodFromTemplate once per ResourceVersion of the LauncherConfig. This improvement could be made in a follow-on PR, but should be made.

_Originally posted by @MikeSpreitzer in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/374#discussion_r3125928280_

MikeSpreitzer:
I hope that this can be addressed as part of #510

osswangxining:
> I hope that this can be addressed as part of [#510](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/510)

Yes, this will be covered within #510

PR #510, opened 2026-05-22
make populator incremental handling

### LINKED ISSUE #472: [Bug]: populator needs to be made incremental

osswangxining:
fixes #472 
Replaced the full-sweep reconciliation with targeted, per-event reconciliation.： podItem, nodeItem, lppItem and lcItem.

reconcileFromPolicies is retained as a fallback for edge cases that need global cleanup: LPP deletion (orphaned desired state), LC deletion (orphaned pods). The targeted methods detect IsNotFound and fall back internally.

### DIFF
--- a/pkg/controller/launcher-populator/digest-updater.go
+++ b/pkg/controller/launcher-populator/digest-updater.go
@@ -111,7 +111,7 @@ func (ctl *controller) updateDigestForLC(ctx context.Context, name string) error
 //   - records the per-LPP derived data into ctl.policy.lpps[name];
 //   - applies the LPP to digest entries on every matched node.
 //
-// Other code paths must NOT call setLPPStatusErrors or ValidateLauncherPodTemplate.
+// Other code paths must NOT call setLPPStatusErrors or getMatchingNodes.
 func (ctl *controller) updateDigestForLPP(ctx context.Context, name string) error {
 	logger := klog.FromContext(ctx)
 
@@ -191,9 +191,30 @@ func (ctl *controller) updateDigestForLPP(ctx context.Context, name string) erro
 		}
 	}
 
+	// LC names currently referenced by this LPP. Used to detach the LPP from
+	// digest entries whose lcName has been dropped from CountForLauncher even
+	// though the node still matches the selector.
+	newLCNames := sets.New[string]()
+	for _, cr := range lpp.Spec.CountForLauncher {
+		newLCNames.Insert(cr.LauncherConfigName)
+	}
+
 	// Apply LPP to each matched node and enqueue affected keys.
 	for i := range currentMatchedNodes {
 		node := &currentMatchedNodes[i]
+		// Detach the LPP from entries on this still-matching node whose lcName
+		// is no longer referenced by lpp.Spec.CountForLauncher.
+		if nodeMap, ok := ctl.policy.digest[node.Name]; ok {
+			for lcName, entry := range nodeMap {
+				if _, hasLPP := entry.lpps[name]; !hasLPP {
+					continue
+				}
+				if newLCNames.Has(lcName) {
+					continue
+				}
+				ctl.detachLPPFromEntry(nodeMap, name, node.Name, lcName, entry)
+			}
+		}
 		ctl.applyLPPToDigestForNode(lpp, node)
 		for _, cr := range lpp.Spec.CountForLauncher {
 			ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: node.Name, LauncherConfigName: cr.LauncherConfigName}})
@@ -226,6 +247,16 @@ func (ctl *controller) updateDigestForNode(ctx context.Context, nodeName string)
 // recomputeDigestForNode rebuilds digest entries for a single node by replaying
 // every cached LPP. Pure read of ctl.policy.lpps + ctl.policy.lcs.
 func (ctl *controller) recomputeDigestForNode(node *corev1.Node) {
+	// Snapshot LC names currently in the digest for this node so we can also
+	// enqueue keys for entries that disappear after replay (e.g., a previously
+	// matching LPP no longer matches this node).
+	affected := sets.New[string]()
+	if oldMap, ok := ctl.policy.digest[node.Name]; ok {
+		for lcName := range oldMap {
+			affected.Insert(lcName)
+		}
+	}
+
 	ctl.policy.digest[node.Name] = make(map[string]*digestEntry)
 
 	for _, lppd := range ctl.policy.lpps {
@@ -234,6 +265,9 @@ func (ctl *controller) recomputeDigestForNode(node *corev1.Node) {
 
 	nodeMap := ctl.policy.digest[node.Name]
 	for lcName := range nodeMap {
+		affected.Insert(lcName)
+	}
+	for lcName := range affected {
 		ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: node.Name, LauncherConfigName: lcName}})
 	}
 	if len(nodeMap) == 0 {
--- a/pkg/controller/launcher-populator/digest-updater.go
+++ b/pkg/controller/launcher-populator/digest-updater.go
@@ -293,32 +293,8 @@ func (ctl *controller) applyLPPToDigestForNode(lpp *fmav1alpha1.LauncherPopulati
 			entry = &digestEntry{lpps: make(map[string]*fmav1alpha1.LauncherPopulationPolicy)}
 			ctl.policy.setEntry(node.Name, lcName, entry)
 		}
-
-		lcd := ctl.policy.lcDigestFor(lcName)
-		switch {
-		case lcd == nil || lcd.object == nil:
-			// LC not yet processed or known to be absent: handsOff. The LC's
-			// own update path (updateDigestForLC) will re-enqueue this LPP
-			// when the LC becomes existent.
-			entry.handsOff = true
-			entry.spec = nil
-		case lcd.templateErr != "":
-			// Template invalid; error is already in LC.Status.
-			entry.handsOff = true
-			entry.spec = nil
-		default:
-			entry.handsOff = false
-			entry.spec = &lcd.object.Spec
-			entry.ownerRef = lcd.ownerRef
-		}
-
-		if cr.LauncherCount > entry.count {
-			entry.count = cr.LauncherCount
-		}
-		if entry.lpps == nil {
-			entry.lpps = make(map[string]*fmav1alpha1.LauncherPopulationPolicy)
-		}
 		entry.lpps[lpp.Name] = lpp
+		ctl.recomputeEntryFromLPPs(entry, lcName)
 	}
 }
 
--- a/pkg/controller/launcher-populator/digest-updater.go
+++ b/pkg/controller/launcher-populator/digest-updater.go
@@ -22,6 +22,7 @@ import (
 
 	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -50,23 +51,19 @@ func (ctl *controller) updateDigestForLC(ctx context.Context, name string) error
 		if !apierrors.IsNotFound(err) {
 			return fmt.Errorf("failed to get LC %s: %w", name, err)
 		}
-		// LC deleted.
+		if !prevExists {
+			return nil
+		}
 		logger.Info("LC deleted, marking referencing keys as handsOff", "config", name)
 		delete(ctl.policy.lcs, name)
-		for _, key := range ctl.policy.keysForLC(name) {
-			if entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName); entry != nil {
-				entry.handsOff = true
-				entry.spec = nil
-				entry.count = 0
-				ctl.keyQueue.Queue.Add(keyItem{key})
-			}
+		for key, entry := range ctl.policy.entriesForLC(name) {
+			entry.handsOff = true
+			entry.spec = nil
+			entry.count = 0
+			ctl.keyQueue.Queue.Add(keyItem{key})
 		}
-		// Existence flip: re-enqueue LPPs that reference this LC so they can
-		// recompute missingLCs.
-		if prevExists {
-			for _, lppName := range ctl.policy.lppNamesRefByLC(name) {
-				ctl.digestQueue.Queue.Add(funcItem{kind: kindLPP, name: lppName})
-			}
+		for _, lppName := range ctl.policy.lppNamesRefByLC(name) {
+			ctl.digestQueue.Queue.Add(funcItem{kind: kindLPP, name: lppName})
 		}
 		return nil
 	}
@@ -81,9 +78,11 @@ func (ctl *controller) updateDigestForLC(ctx context.Context, name string) error
 	if templateErr == "" {
 		h, hashErr := utils.ComputeLauncherTemplateHash(lc.Spec.PodTemplate)
 		if hashErr != nil {
-			return fmt.Errorf("failed to compute template hash for LC %s: %w", name, hashErr)
+			templateErr = hashErr.Error()
+			logger.Error(hashErr, "Failed to hash PodTemplate, reporting in Status", "config", name)
+		} else {
+			templateHash = h
 		}
-		templateHash = h
 	}
 	if statusErr := ctl.setLCStatusErrors(ctx, lc, nonNilSlice(templateErr)); statusErr != nil {
 		return fmt.Errorf("failed to set Status for LC %s: %w", name, statusErr)
@@ -110,18 +109,15 @@ func (ctl *controller) updateDigestForLC(ctx context.Context, name string) error
 	}
 
 	// Refresh per-key entries that reference this LC.
-	for _, key := range ctl.policy.keysForLC(name) {
-		entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName)
-		if entry == nil {
-			continue
-		}
+	lcd := ctl.policy.lcs[name]
+	for key, entry := range ctl.policy.entriesForLC(name) {
 		if templateErr != "" {
 			entry.handsOff = true
 			entry.spec = nil
 		} else {
 			entry.handsOff = false
-			entry.spec = &lc.Spec
-			entry.ownerRef = makeLCOwnerRef(lc)
+			entry.spec = &lcd.object.Spec
+			entry.ownerRef = lcd.ownerRef
 		}
 		ctl.keyQueue.Queue.Add(keyItem{key})
 	}
@@ -216,8 +212,9 @@ func (ctl *controller) updateDigestForLPP(ctx context.Context, name string) erro
 	}
 
 	// Apply LPP to each matched node and enqueue affected keys.
-	for _, node := range currentMatchedNodes {
-		ctl.applyLPPToDigestForNode(lpp, node.Name)
+	for i := range currentMatchedNodes {
+		node := &currentMatchedNodes[i]
+		ctl.applyLPPToDigestForNode(lpp, node)
 		for _, cr := range lpp.Spec.CountForLauncher {
 			ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: node.Name, LauncherConfigName: cr.LauncherConfigName}})
 		}
@@ -233,52 +230,41 @@ func (ctl *controller) updateDigestForLPP(ctx context.Context, name string) erro
 func (ctl *controller) updateDigestForNode(ctx context.Context, nodeName string) error {
 	logger := klog.FromContext(ctx)
 
-	_, err := ctl.nodeLister.Get(nodeName)
+	node, err := ctl.nodeLister.Get(nodeName)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			logger.Info("Node deleted, removing from digest", "node", nodeName)
-			removedKeys := ctl.policy.removeNode(nodeName)
-			for _, key := range removedKeys {
-				ctl.keyQueue.Queue.Add(keyItem{key})
-			}
+			delete(ctl.policy.digest, nodeName)
 			return nil
 		}
 		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
 	}
-	return ctl.recomputeDigestForNode(nodeName)
+	ctl.recomputeDigestForNode(node)
+	return nil
 }
 
 // recomputeDigestForNode rebuilds digest entries for a single node by replaying
 // every cached LPP. Pure read of ctl.policy.lpps + ctl.policy.lcs.
-func (ctl *controller) recomputeDigestForNode(nodeName string) error {
-	ctl.policy.digest[nodeName] = make(map[string]*digestEntry)
+func (ctl *controller) recomputeDigestForNode(node *corev1.Node) {
+	ctl.policy.digest[node.Name] = make(map[string]*digestEntry)
 
 	for _, lppd := range ctl.policy.lpps {
-		if lppd == nil || lppd.object == nil {
-			continue
-		}
-		ctl.applyLPPToDigestForNode(lppd.object, nodeName)
+		ctl.applyLPPToDigestForNode(lppd.object, node)
 	}
 
-	nodeMap := ctl.policy.digest[nodeName]
+	nodeMap := ctl.policy.digest[node.Name]
 	for lcName := range nodeMap {
-		ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName}})
+		ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: node.Name, LauncherConfigName: lcName}})
 	}
 	if len(nodeMap) == 0 {
-		delete(ctl.policy.digest, nodeName)
+		delete(ctl.policy.digest, node.Name)
 	}
-	return nil
 }
 
 // applyLPPToDigestForNode evaluates one LPP for one node and updates the digest
 // using ONLY ctl.policy.lcs as the source of truth for LC status. It never
 // validates templates, writes Status, or fetches from listers.
-func (ctl *controller) applyLPPToDigestForNode(lpp *fmav1alpha1.LauncherPopulationPolicy, nodeName string) {
-	node, err := ctl.nodeLister.Get(nodeName)
-	if err != nil {
-		// Node missing or transient lister error: nothing to apply.
-		return
-	}
+func (ctl *controller) applyLPPToDigestForNode(lpp *fmav1alpha1.LauncherPopulationPolicy, node *corev1.Node) {
 	labelSelector, selectorErr := metav1.LabelSelectorAsSelector(&lpp.Spec.EnhancedNodeSelector.LabelSelector)
 	if selectorErr != nil {
 		return
@@ -293,10 +279,10 @@ func (ctl *controller) applyLPPToDigestForNode(lpp *fmav1alpha1.LauncherPopulati
 	for _, cr := range lpp.Spec.CountForLauncher {
 		lcName := cr.LauncherConfigName
 
-		entry := ctl.policy.getEntry(nodeName, lcName)
+		entry := ctl.policy.getEntry(node.Name, lcName)
 		if entry == nil {
 			entry = &digestEntry{lpps: make(map[string]*fmav1alpha1.LauncherPopulationPolicy)}
-			ctl.policy.setEntry(nodeName, lcName, entry)
+			ctl.policy.setEntry(node.Name, lcName, entry)
 		}
 
 		lcd := ctl.policy.lcDigestFor(lcName)
@@ -388,5 +374,3 @@ func (ctl *controller) recomputeEntryFromLPPs(entry *digestEntry, lcName string)
 		entry.ownerRef = lcd.ownerRef
 	}
 }
-
-
--- a/pkg/controller/launcher-populator/digest-updater.go
+++ b/pkg/controller/launcher-populator/digest-updater.go
@@ -250,12 +250,7 @@ func (ctl *controller) recomputeDigestForNode(node *corev1.Node) {
 	// Snapshot LC names currently in the digest for this node so we can also
 	// enqueue keys for entries that disappear after replay (e.g., a previously
 	// matching LPP no longer matches this node).
-	affected := sets.New[string]()
-	if oldMap, ok := ctl.policy.digest[node.Name]; ok {
-		for lcName := range oldMap {
-			affected.Insert(lcName)
-		}
-	}
+	affected := sets.KeySet(ctl.policy.digest[node.Name])
 
 	ctl.policy.digest[node.Name] = make(map[string]*digestEntry)
 
--- a/pkg/controller/launcher-populator/digest-updater.go
+++ b/pkg/controller/launcher-populator/digest-updater.go
@@ -1,0 +1,392 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package launcherpopulator
+
+import (
+	"context"
+	"fmt"
+
+	fmav1alpha1 "github.com/llm-d-incubation/llm-d-fast-model-actuation/api/fma/v1alpha1"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+)
+
+// updateDigestForLC is the SOLE place that:
+//   - validates the LC PodTemplate (utils.ValidateLauncherPodTemplate);
+//   - computes the node-independent template hash;
+//   - writes LauncherConfig.Status (via setLCStatusErrors);
+//   - records the per-LC derived data into ctl.policy.lcs[name].
+//
+// Other code paths must read ctl.policy.lcs[name] instead of recomputing.
+// On LC existence transitions or templateErr changes, all LPPs that reference
+// this LC are enqueued so they can refresh their lppDigest (missingLCs) and
+// re-apply to the digest.
+func (ctl *controller) updateDigestForLC(ctx context.Context, name string) error {
+	logger := klog.FromContext(ctx)
+
+	prev := ctl.policy.lcs[name]
+	prevExists := prev != nil && prev.object != nil
+
+	lc, err := ctl.lcLister.LauncherConfigs(ctl.namespace).Get(name)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get LC %s: %w", name, err)
+		}
+		// LC deleted.
+		logger.Info("LC deleted, marking referencing keys as handsOff", "config", name)
+		delete(ctl.policy.lcs, name)
+		for _, key := range ctl.policy.keysForLC(name) {
+			if entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName); entry != nil {
+				entry.handsOff = true
+				entry.spec = nil
+				entry.count = 0
+				ctl.keyQueue.Queue.Add(keyItem{key})
+			}
+		}
+		// Existence flip: re-enqueue LPPs that reference this LC so they can
+		// recompute missingLCs.
+		if prevExists {
+			for _, lppName := range ctl.policy.lppNamesRefByLC(name) {
+				ctl.digestQueue.Queue.Add(funcItem{kind: kindLPP, name: lppName})
+			}
+		}
+		return nil
+	}
+
+	// LC exists: validate and produce the lcDigest.
+	var templateErr string
+	if vErr := utils.ValidateLauncherPodTemplate(lc.Spec.PodTemplate); vErr != nil {
+		templateErr = vErr.Error()
+		logger.Error(vErr, "Invalid PodTemplate in LC, reporting in Status", "config", name)
+	}
+	var templateHash string
+	if templateErr == "" {
+		h, hashErr := utils.ComputeLauncherTemplateHash(lc.Spec.PodTemplate)
+		if hashErr != nil {
+			return fmt.Errorf("failed to compute template hash for LC %s: %w", name, hashErr)
+		}
+		templateHash = h
+	}
+	if statusErr := ctl.setLCStatusErrors(ctx, lc, nonNilSlice(templateErr)); statusErr != nil {
+		return fmt.Errorf("failed to set Status for LC %s: %w", name, statusErr)
+	}
+
+	prevTemplateErr := ""
+	if prev != nil {
+		prevTemplateErr = prev.templateErr
+	}
+
+	ctl.policy.lcs[name] = &lcDigest{
+		object:       lc,
+		templateErr:  templateErr,
+		templateHash: templateHash,
+		ownerRef:     makeLCOwnerRef(lc),
+	}
+
+	// If the LC just became existent, or its template-validity flipped, the
+	// LPPs that reference it must be re-evaluated.
+	if !prevExists || prevTemplateErr != templateErr {
+		for _, lppName := range ctl.policy.lppNamesRefByLC(name) {
+			ctl.digestQueue.Queue.Add(funcItem{kind: kindLPP, name: lppName})
+		}
+	}
+
+	// Refresh per-key entries that reference this LC.
+	for _, key := range ctl.policy.keysForLC(name) {
+		entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName)
+		if entry == nil {
+			continue
+		}
+		if templateErr != "" {
+			entry.handsOff = true
+			entry.spec = nil
+		} else {
+			entry.handsOff = false
+			entry.spec = &lc.Spec
+			entry.ownerRef = makeLCOwnerRef(lc)
+		}
+		ctl.keyQueue.Queue.Add(keyItem{key})
+	}
+	return nil
+}
+
+// updateDigestForLPP is the SOLE place that:
+//   - runs getMatchingNodes (collecting selector errors);
+//   - determines which referenced LCs are missing (by reading ctl.policy.lcs);
+//   - writes LauncherPopulationPolicy.Status (via setLPPStatusErrors);
+//   - records the per-LPP derived data into ctl.policy.lpps[name];
+//   - applies the LPP to digest entries on every matched node.
+//
+// Other code paths must NOT call setLPPStatusErrors or ValidateLauncherPodTemplate.
+func (ctl *controller) updateDigestForLPP(ctx context.Context, name string) error {
+	logger := klog.FromContext(ctx)
+
+	lpp, err := ctl.lppLister.LauncherPopulationPolicies(ctl.namespace).Get(name)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get LPP %s: %w", name, err)
+		}
+		// LPP deleted: detach from all referencing entries.
+		logger.Info("LPP deleted, updating digest incrementally", "policy", name)
+		for nodeName, nodeMap := range ctl.policy.digest {
+			for lcName, entry := range nodeMap {
+				if _, hasLPP := entry.lpps[name]; !hasLPP {
+					continue
+				}
+				ctl.detachLPPFromEntry(nodeMap, name, nodeName, lcName, entry)
+			}
+		}
+		delete(ctl.policy.lpps, name)
+		return nil
+	}
+
+	// Resolve matched nodes and selector errors.
+	currentMatchedNodes, selectorErrs, matchErr := ctl.getMatchingNodes(ctx, lpp.Spec.EnhancedNodeSelector)
+	if matchErr != nil {
+		return fmt.Errorf("failed to get matching nodes for policy %s: %w", lpp.Name, matchErr)
+	}
+
+	// Compute missing-LC errors by consulting ctl.policy.lcs only.
+	var missingLCs []string
+	for _, cr := range lpp.Spec.CountForLauncher {
+		lcd := ctl.policy.lcDigestFor(cr.LauncherConfigName)
+		if lcd == nil || lcd.object == nil {
+			missingLCs = append(missingLCs, fmt.Sprintf(
+				"LauncherConfig %q referenced in CountForLauncher does not exist", cr.LauncherConfigName))
+		}
+	}
+
+	// Combine all user-facing errors and write LPP.Status (the SOLE writer).
+	allErrs := append([]string(nil), selectorErrs...)
+	allErrs = append(allErrs, missingLCs...)
+	if statusErr := ctl.setLPPStatusErrors(ctx, lpp, allErrs); statusErr != nil {
+		return fmt.Errorf("failed to set Status for policy %s: %w", lpp.Name, statusErr)
+	}
+
+	// Snapshot the LPP digest.
+	ctl.policy.lpps[name] = &lppDigest{
+		object:       lpp,
+		selectorErrs: selectorErrs,
+		missingLCs:   missingLCs,
+	}
+
+	// Determine old matched nodes from existing digest entries.
+	oldNodes := sets.New[string]()
+	for nodeName, nodeMap := range ctl.policy.digest {
+		for _, entry := range nodeMap {
+			if _, hasLPP := entry.lpps[name]; hasLPP {
+				oldNodes.Insert(nodeName)
+				break
+			}
+		}
+	}
+	currentMatched := sets.New[string]()
+	for _, node := range currentMatchedNodes {
+		currentMatched.Insert(node.Name)
+	}
+	removedNodes := oldNodes.Difference(currentMatched).UnsortedList()
+
+	// Detach the LPP from entries on nodes it no longer matches.
+	for _, nodeName := range removedNodes {
+		nodeMap := ctl.policy.digest[nodeName]
+		for lcName, entry := range nodeMap {
+			if _, hasLPP := entry.lpps[name]; !hasLPP {
+				continue
+			}
+			ctl.detachLPPFromEntry(nodeMap, name, nodeName, lcName, entry)
+		}
+	}
+
+	// Apply LPP to each matched node and enqueue affected keys.
+	for _, node := range currentMatchedNodes {
+		ctl.applyLPPToDigestForNode(lpp, node.Name)
+		for _, cr := range lpp.Spec.CountForLauncher {
+			ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: node.Name, LauncherConfigName: cr.LauncherConfigName}})
+		}
+	}
+
+	return nil
+}
+
+// updateDigestForNode handles a Node add/update/delete event.
+// It recomputes digest entries for the affected node by replaying every cached
+// LPP. It does not validate templates or write Status; those are produced by
+// updateDigestForLC / updateDigestForLPP exclusively.
+func (ctl *controller) updateDigestForNode(ctx context.Context, nodeName string) error {
+	logger := klog.FromContext(ctx)
+
+	_, err := ctl.nodeLister.Get(nodeName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("Node deleted, removing from digest", "node", nodeName)
+			removedKeys := ctl.policy.removeNode(nodeName)
+			for _, key := range removedKeys {
+				ctl.keyQueue.Queue.Add(keyItem{key})
+			}
+			return nil
+		}
+		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+	}
+	return ctl.recomputeDigestForNode(nodeName)
+}
+
+// recomputeDigestForNode rebuilds digest entries for a single node by replaying
+// every cached LPP. Pure read of ctl.policy.lpps + ctl.policy.lcs.
+func (ctl *controller) recomputeDigestForNode(nodeName string) error {
+	ctl.policy.digest[nodeName] = make(map[string]*digestEntry)
+
+	for _, lppd := range ctl.policy.lpps {
+		if lppd == nil || lppd.object == nil {
+			continue
+		}
+		ctl.applyLPPToDigestForNode(lppd.object, nodeName)
+	}
+
+	nodeMap := ctl.policy.digest[nodeName]
+	for lcName := range nodeMap {
+		ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName}})
+	}
+	if len(nodeMap) == 0 {
+		delete(ctl.policy.digest, nodeName)
+	}
+	return nil
+}
+
+// applyLPPToDigestForNode evaluates one LPP for one node and updates the digest
+// using ONLY ctl.policy.lcs as the source of truth for LC status. It never
+// validates templates, writes Status, or fetches from listers.
+func (ctl *controller) applyLPPToDigestForNode(lpp *fmav1alpha1.LauncherPopulationPolicy, nodeName string) {
+	node, err := ctl.nodeLister.Get(nodeName)
+	if err != nil {
+		// Node missing or transient lister error: nothing to apply.
+		return
+	}
+	labelSelector, selectorErr := metav1.LabelSelectorAsSelector(&lpp.Spec.EnhancedNodeSelector.LabelSelector)
+	if selectorErr != nil {
+		return
+	}
+	if !labelSelector.Matches(labels.Set(node.Labels)) {
+		return
+	}
+	if !matchesResourceConditions(node.Status.Allocatable, lpp.Spec.EnhancedNodeSelector.AllocatableResources) {
+		return
+	}
+
+	for _, cr := range lpp.Spec.CountForLauncher {
+		lcName := cr.LauncherConfigName
+
+		entry := ctl.policy.getEntry(nodeName, lcName)
+		if entry == nil {
+			entry = &digestEntry{lpps: make(map[string]*fmav1alpha1.LauncherPopulationPolicy)}
+			ctl.policy.setEntry(nodeName, lcName, entry)
+		}
+
+		lcd := ctl.policy.lcDigestFor(lcName)
+		switch {
+		case lcd == nil || lcd.object == nil:
+			// LC not yet processed or known to be absent: handsOff. The LC's
+			// own update path (updateDigestForLC) will re-enqueue this LPP
+			// when the LC becomes existent.
+			entry.handsOff = true
+			entry.spec = nil
+		case lcd.templateErr != "":
+			// Template invalid; error is already in LC.Status.
+			entry.handsOff = true
+			entry.spec = nil
+		default:
+			entry.handsOff = false
+			entry.spec = &lcd.object.Spec
+			entry.ownerRef = lcd.ownerRef
+		}
+
+		if cr.LauncherCount > entry.count {
+			entry.count = cr.LauncherCount
+		}
+		if entry.lpps == nil {
+			entry.lpps = make(map[string]*fmav1alpha1.LauncherPopulationPolicy)
+		}
+		entry.lpps[lpp.Name] = lpp
+	}
+}
+
+// removeLPPFromEntry removes an LPP reference from a digest entry.
+// If the entry has no remaining LPPs, it is reset to a zero state.
+func (ctl *controller) removeLPPFromEntry(entry *digestEntry, lppName string) {
+	delete(entry.lpps, lppName)
+	if len(entry.lpps) == 0 {
+		entry.count = 0
+		entry.spec = nil
+		entry.handsOff = false
+	}
+}
+
+func (ctl *controller) detachLPPFromEntry(nodeMap map[string]*digestEntry, lppName, nodeName, lcName string, entry *digestEntry) {
+	ctl.removeLPPFromEntry(entry, lppName)
+	if len(entry.lpps) == 0 {
+		delete(nodeMap, lcName)
+		if len(nodeMap) == 0 {
+			delete(ctl.policy.digest, nodeName)
+		}
+	} else {
+		ctl.recomputeEntryFromLPPs(entry, lcName)
+	}
+	ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName}})
+}
+
+// recomputeEntryFromLPPs recomputes the (handsOff, count, spec, ownerRef) of a
+// digestEntry from its remaining LPP references and the cached lcDigest.
+// Pure read; no Status writes or template validation.
+func (ctl *controller) recomputeEntryFromLPPs(entry *digestEntry, lcName string) {
+	if len(entry.lpps) == 0 {
+		entry.count = 0
+		entry.spec = nil
+		entry.handsOff = false
+		return
+	}
+
+	var maxCount int32
+	for _, lpp := range entry.lpps {
+		for _, cr := range lpp.Spec.CountForLauncher {
+			if cr.LauncherConfigName == lcName && cr.LauncherCount > maxCount {
+				maxCount = cr.LauncherCount
+			}
+		}
+	}
+
+	lcd := ctl.policy.lcDigestFor(lcName)
+	switch {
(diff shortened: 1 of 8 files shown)


osswangxining:
@MikeSpreitzer  @rubambiza  @waltforme  Please help review this, thanks in advance.

pkg/controller/launcher-populator/populator.go
```diff
@@ -269,28 +287,359 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
 	return item.process(ctx, ctl)
 }
 
-func (lppItem) process(ctx context.Context, ctl *controller) (error, bool) {
-	// Until #472 introduces targeted reconciliation keyed by policy,
-	// any LauncherPopulationPolicy event drives a full reconcile.
-	return ctl.reconcileFromPolicies(ctx)
+func (item podItem) process(ctx context.Context, ctl *controller) (error, bool) {
+	return ctl.reconcilePod(ctx, item.nodeName, item.launcherConfigName)
```
MikeSpreitzer: Why define a function that is called just once and has the same inputs as its caller? In other words, why not inline `ctl.reconcilePod` here?

Same question for the other 3.

pkg/controller/launcher-populator/populator.go
```diff
@@ -269,28 +287,359 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
...
+		if apierrors.IsNotFound(err) {
+			// Policy deleted → full sweep to clean up desired state.
+			logger.Info("LauncherPopulationPolicy deleted, falling back to full reconcile", "policy", name)
+			return ctl.reconcileFromPolicies(ctx)
+		}
+		return err, true
+	}
+
+	// Ensure LPP status is up-to-date.
+	nodes, selectorErrs, err := ctl.getMatchingNodes(ctx, lpp.Spec.EnhancedNodeSelector)
```
MikeSpreitzer: This won't correctly handle a change in an LPP that removes some Nodes from the matching set.

pkg/controller/launcher-populator/populator.go
```diff
@@ -269,28 +287,359 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
...
+		return err, true
+	}
+
+	// Verify LC still exists.
+	lc, err := ctl.lcLister.LauncherConfigs(ctl.namespace).Get(lcName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// LC deleted → full sweep to clean up orphaned pods.
+			logger.Info("LauncherConfig no longer exists, falling back to full reconcile", "config", lcName)
+			return ctl.reconcileFromPolicies(ctx)
```
MikeSpreitzer: I am sorry, I just realized on re-review of `ctl.reconcileFromPolicies` that it does not do what I thought we agreed that it should do. I thought we agreed that in the case of a `CountForLauncher` that refers to a `LauncherConfig` that does not exist at the moment, the launcher population controller will take no action on launchers created from a `LauncherConfig` having the relevant name. I see now that instead the `CountForLauncher` is treated as if it does not exist.

I think that the better reaction, when a `LauncherPopulationPolicy` has a `CountForLauncher` that refers to a non-existent `LauncherConfig`, is to treat this as a user error that is hopefully temporary. The actions would be to note the error in the `Status` of the `LauncherPopulationPolicy` and take no action on launchers created from `LauncherConfig` objects of the given name.

pkg/controller/launcher-populator/populator.go
```diff
@@ -269,28 +287,359 @@ func (ctl *controller) process(ctx context.Context, item queueItem) (error, bool
...
+		}
+		return err, true
+	}
+
+	// Validate PodTemplate; skip reconciliation if invalid.
+	if templateErr := utils.ValidateLauncherPodTemplate(lc.Spec.PodTemplate); templateErr != nil {
+		logger.Error(templateErr, "Invalid PodTemplate in LauncherConfig, skipping pod reconciliation",
+			"config", lcName)
+		if statusErr := ctl.setLCStatusErrors(ctx, lc, []string{templateErr.Error()}); statusErr != nil {
+			logger.Error(statusErr, "Failed to set Status for LauncherConfig", "config", lcName)
```
MikeSpreitzer: As in the case of a missing `LauncherConfig`, I would treat one with problems as a hopefully temporary user error, causing the controller to take no action on the relevant launcher Pods.

MikeSpreitzer:
There is still a lot of duplicated code and scanning. I suspect that we can get a more efficient controller by following the spirit in the existing controller of splitting between (a) digesting all the objects to produce one comprehensive policy data structure and (b) iterating over that data structure and applying it independently for each NodeLauncherKey. The "key" observation is that the digested policy is applied independently for each NodeLauncherKey (that's not quite true in the existing code --- the serialization between launcher deletions and creations is done at the Node level --- but that coarseness is not necessary: that serialization could be done independently per NodeLauncherKey). The basic idea is be to split the controller into two sorts of work queue items: (1) those that lead to updating of the digested policy (if necessary) and enqueuing of the relevant (2) NodeLauncherKeys. Processing of a NodeLauncherKey involves fetching the desired count for that key from the digested data structure, doing the pending expectations dance and finding the relevant existing launcher Pods, and then determining and doing the needed launcher deletes or creates.

I would pay some attention to enabling efficient updates to the digested policy data structure. This could be done to various degrees, with code complexity and memory consumption higher in techniques with lower runtime CPU cost. I suspect that something like the following might strike a good balance, limiting scans of kube API objects while still having some scans in the digested policy data structure.

The idea is that the digested policy data structure would primarily be a map from NodeLauncherKey to a struct that holds (a) both the single digested desired count (i.e., the max over the applicable CountForLauncher structs) and (b) the supporting details from the relevant kube API objects. The digested policy data structure would also hold copies of the input data, to help identifying deltas. Something like the following.

```go
type digestedPolicy struct {
    digested map[string]map[string]*perKey // maps Node name to LC name to *perKey
    lcs      map[string]*LauncherConfig // maps LC name to LC
    lpps     map[string]*LauncherPopulationPolicy // maps LPP name to LPP
}

type perKey struct {
    handsOff bool // true when user error precludes action on this key
    count int // max over relevant (existing, well-formed) CountForLauncher
    lpps map[string]*LauncherPopulationPolicy // maps LPP name to LPP, holding relevant ones
}
```

The handler for an informer event for a Pod would just determine the relevant NodeLauncherKey and enqueue that.

`nodeItem.process` would scan all the LauncherPopulationPolicy objects in the informer cache, collect the `map[name]*LPP` of those that match, then scan through `digestedPolicy.digested[node.Name]` and make any needed updates to `perKey.lpps` and any consequent updates to the (handsOff, count) of the `perKey` and, in those cases of such a consequent update, enqueue the NodeLauncherKey.

`lppItem.process` would first check in `digestedPolicy.lpps` to see if there has been a change in the node selector. If so, this is the relatively expensive case: scan all the Node objects in the informer cache and collect the `map[name]*Node` of those that match. Then scan all of `digestedPolicy.digested` and make any needed additions to or removals from `perKey.lpps`, again following through with any consequent changes to (handsOff, count) and enqueuing of NodeLauncherKey. In the less expensive case the matching does not have to be re-evaluated and the logic can proceed directly to each `perKey.lpps`.

`lcItem.process` would first check `digestedPolicy.lcs` to see if there has been a change in presence or `Spec.Generation`. In case of current presence, `Spec.PodTemplate` has to be evaluated for well-formedness if `Spec.Generation` changed. Then proceed to updating all the `perKey` that have a reference to the LC in `perKey.lpps`, enqueuing NodeLauncherKey in case of change to (handsOff, count) of the `perKey`.

pkg/controller/launcher-populator/populator.go
osswangxining: As for the 4 concrete process methods (podItem, nodeItem, lppItem, lcItem), they are designed to satisfy the queueItem interface. The actual reconciliation logic lives in reconcilePod/reconcileNode/reconcileLPP/reconcileLC which are kept as separate functions for independent unit testability.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+			}
+			return nil
+		}
+		return fmt.Errorf("failed to get LPP %s: %w", name, err)
+	}
+
+	// Report selector errors in LPP Status.
+	currentMatchedNodes, selectorErrs, _ := ctl.getMatchingNodes(ctx, lpp.Spec.EnhancedNodeSelector)
+	if statusErr := ctl.setLPPStatusErrors(ctx, lpp, selectorErrs); statusErr != nil {
+		logger.Error(statusErr, "Failed to set Status for policy", "policy", lpp.Name)
```
MikeSpreitzer: I think that a retry is in order in this situation, where the controller failed to do part of its job.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	// Determine old matched nodes from existing digest entries.
+	oldNodes := make(map[string]bool)
```
MikeSpreitzer: It would be better to use the standard set representation in https://github.com/kubernetes/apimachinery/blob/v0.34.8/pkg/util/sets/set.go#L25; that comes with an implementation of set difference.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+		}
+	}
+
+	currentMatched := make(map[string]bool, len(currentMatchedNodes))
+	for _, node := range currentMatchedNodes {
+		currentMatched[node.Name] = true
+	}
+
+	// Cache the new LPP snapshot.
+	ctl.policy.lpps[name] = lpp.DeepCopy()
```
MikeSpreitzer: I see no need to copy the LPP for every matching Node. The LPP from the informer's local cache is not mutated by anything, no copy is needed here (nor the other lines like this).

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+						ctl.removeLPPFromEntry(entry, name)
+						affectedKeys = append(affectedKeys, NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName})
+					}
+				}
+			}
+			delete(ctl.policy.lpps, name)
+			// Recompute affected entries from remaining LPPs; delete empty entries.
+			for _, key := range affectedKeys {
+				entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName)
+				if entry != nil {
```
MikeSpreitzer: Certainly `entry != nil`, otherwise the key would not be in `affectedKeys`.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+					if len(entry.lpps) == 0 {
+						delete(ctl.policy.digest[key.NodeName], key.LauncherConfigName)
+					} else {
+						ctl.recomputeEntryFromLPPs(entry, key.LauncherConfigName)
+					}
+				}
+				ctl.Queue.Add(keyItem{key})
```
MikeSpreitzer: Why not just move these lines to right after `ctl.removeLPPFromEntry` above? Why the complexity of compiling a list of keys then doing another iteration over it?

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+		for lcName, entry := range nodeMap {
+			if _, hasLPP := entry.lpps[name]; hasLPP {
+				ctl.removeLPPFromEntry(entry, name)
+				if len(entry.lpps) == 0 {
+					delete(nodeMap, lcName)
+				} else {
+					ctl.recomputeEntryFromLPPs(entry, lcName)
+				}
+				ctl.Queue.Add(keyItem{NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName}})
+			}
```
MikeSpreitzer: I see essentially the same logic above, for the case of the LPP being deleted. I think that it would be better to DRY this out.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+
+	// Determine old matched nodes from existing digest entries.
+	oldNodes := make(map[string]bool)
+	for nodeName, nodeMap := range ctl.policy.digest {
+		for _, entry := range nodeMap {
+			if _, hasLPP := entry.lpps[name]; hasLPP {
+				oldNodes[nodeName] = true
+				break
+			}
+		}
```
MikeSpreitzer, line 175: FYI, this could get simpler if `ctl.getMatchingNodes` returned a `map[/*Node name*/string]*Node` instead of a `[]*Node`. With that change, this loop over `ctl.policy.digest` could easily detect the case of a Node that used to match but no longer does, and handle that case right in the loop body rather than messing around with creating a set and doing set difference and iterating over that difference.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,447 @@
...
+					}
+				}
+			}
+			delete(ctl.policy.lpps, name)
+			// Recompute affected entries from remaining LPPs; delete empty entries.
+			for _, key := range affectedKeys {
+				entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName)
+				if entry != nil {
+					if len(entry.lpps) == 0 {
+						delete(ctl.policy.digest[key.NodeName], key.LauncherConfigName)
```
MikeSpreitzer: What if this delete makes `ctl.policy.digest[key.NodeName]` empty? Should `delete(ctl.policy.digest, key.NodeName)` happen in that case?

pkg/controller/launcher-populator/populator.go
```diff
@@ -121,43 +119,40 @@ type controller struct {
...
+// Two kinds of items are enqueued:
+//   - funcItem: carries a closure that updates the digested policy and enqueues
+//     affected NodeLauncherKeys. Used for LPP, LC, and Node events.
+//   - keyItem: carries a NodeLauncherKey for per-key reconciliation. Used for
+//     Pod events and enqueued by funcItem processing.
 type queueItem interface {
-	// process returns (err error, retry bool).
-	// There will be a retry iff `retry`, error logged if `err != nil`.
-	process(ctx context.Context, ctl *controller) (error, bool)
+	isQueueItem()
```
MikeSpreitzer: Why is this method defined? Nothing calls it.

pkg/controller/launcher-populator/populator.go
```diff
@@ -121,43 +119,40 @@ type controller struct {
...
 }
 
-type lppItem struct {
-	cache.ObjectName
+// funcItem wraps a closure that performs digest-level work:
+// updates the digestedPolicy and enqueues affected NodeLauncherKeys.
+type funcItem struct {
+	fn func(ctx context.Context, ctl *controller) error
```
MikeSpreitzer: In the Go language, function values can not be compared nor hashed. That means that `funcItem` here is not a suitable type for a work queue entry. This is borne out by the failure in CI. Look at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/26333080049/job/77522257992?pr=510, and look at the "show launcher-populator log" job step. The displayed log shows a panic because of this problem.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: The two differ only slightly in how their parameters appear, I do not see what difference that makes in unit testability.

MikeSpreitzer:
I still see a lot of more-or-less duplicated code. For example:
- three calls of `ctl.setLPPStatusErrors`
- three calls of `ctl.setLCStatusErrors`
- five calls of `utils.ValidateLauncherPodTemplate`

I think that this can be made less repetitive. The following ideas might help.

In `ctl.policy.lcs[LauncherConfig.Name]` store all that the controller needs to know about that LauncherConfig and all the results of consuming that LauncherConfig. That includes: whether the LauncherConfig API object exists, what are the results of validating its template and all the other processing of the LauncherConfig that is Node-independent (note: the Node-dependent stuff can be factored out and it will depend on this prior stuff). The workqueue function for processing a reference to a LauncherConfig will update this data and consequently set whatever is needed in the LauncherConfig's Status. Nothing else would update the LauncherConfig's Status; this avoids problems with different parts of the controller reporting different errors into the Status. When the LauncherConfig switches between existing and not existing, this would also enqueue references to all the LauncherPopulationPolicy objects in `ctl.policy.lpps` that reference that LauncherConfig.

Similarly, in `ctl.policy.lpps[LPP.Name]` keep everything the controller needs to know about the LPP. Including, for example, any complaints about selector errors. The workqueue function for processing a reference to an LPP updates this, and _reads_ whatever it needs to know about LauncherConfigs from `ctl.policy.lcs`.

The general idea that I am trying to get to here is that each kind of API object has just one copy of the code that digests it and updates its status. To avoid repeating that work (and repeating the code for that work), instead the results of that work are saved in `ctl.policy` so that other functions can simply read those results when needed.

The `ctl.buildInitialDigest` method looks like a fly in this ointment, because it is a separate code path that needs to do all the same processing as the incremental code paths. Here is a suggestion for how to DRY that out. This also fixes another problem: there is no reason to forbid multiple worker goroutines from working on diffferent NodeLauncherKey values. The idea is to have two workqueues: one for references to API objects, and another for NodeLauncherKeys. The workers for the second would be launched only after the initial load of API objects has been processed. How would the code know when its _processing of_ the initial load of objects is finished? That problem has already been solved in the generic controller package: it provides a more advanced kind of controller that provides this functionality. Use https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/v0.6.0-alpha.12/pkg/controller/generic/knows-processed-sync.go#L49 to create such a controller. You will have to define a kind of marker value to put in the first controller's workqueue; one of these so-called "sentinel" values needs to hold the given integer value, so that they are not all equal to each other. You also provide a function that gets called once, when the processing of the initial load of API objects has finished.

pkg/controller/launcher-populator/digest-updater.go
osswangxining: fine

pkg/controller/launcher-populator/digest-updater.go
osswangxining: yes make sense

pkg/controller/launcher-populator/populator.go
osswangxining: fine, remove for now , will add for further function

pkg/controller/launcher-populator/populator.go
osswangxining: got it thanks. I'm fixing this.

osswangxining:
@MikeSpreitzer   Thanks , I adopted your comments and rebase new commit, covering #450   #457  #459 ;

Please help review this big PR again, thanks a lot.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+		logger.Info("LC deleted, marking referencing keys as handsOff", "config", name)
+		delete(ctl.policy.lcs, name)
+		for _, key := range ctl.policy.keysForLC(name) {
+			if entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName); entry != nil {
+				entry.handsOff = true
+				entry.spec = nil
+				entry.count = 0
+				ctl.keyQueue.Queue.Add(keyItem{key})
+			}
+		}
```
MikeSpreitzer: This, too, is not needed when `prevExists` is false.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+	var templateErr string
+	if vErr := utils.ValidateLauncherPodTemplate(lc.Spec.PodTemplate); vErr != nil {
+		templateErr = vErr.Error()
+		logger.Error(vErr, "Invalid PodTemplate in LC, reporting in Status", "config", name)
+	}
+	var templateHash string
+	if templateErr == "" {
+		h, hashErr := utils.ComputeLauncherTemplateHash(lc.Spec.PodTemplate)
+		if hashErr != nil {
+			return fmt.Errorf("failed to compute template hash for LC %s: %w", name, hashErr)
```
MikeSpreitzer: Is this a transient controller failure or a permanent problem with the user input (which should get reported in the LC's status)?

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+		if entry == nil {
+			continue
+		}
+		if templateErr != "" {
+			entry.handsOff = true
+			entry.spec = nil
+		} else {
+			entry.handsOff = false
+			entry.spec = &lc.Spec
+			entry.ownerRef = makeLCOwnerRef(lc)
```
MikeSpreitzer: Why not read the `ownerRef` from `ctl.policy.lcs[name]` (set just above)?
(But: my recommended factoring of BuildLauncherPodFromTemplate makes this moot.)

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+		}
+		ctl.keyQueue.Queue.Add(keyItem{key})
```
MikeSpreitzer: This `keyItem` is not guaranteed to be processed after the `funcItem{kind: kindLPP, name: lppName}` above (different worker pools, no sync between them).

pkg/controller/launcher-populator/digested-policy.go
```diff
@@ -0,0 +1,195 @@
...
+	var keys []NodeLauncherKey
+	for nodeName, nodeMap := range dp.digest {
+		if _, ok := nodeMap[lcName]; ok {
+			keys = append(keys, NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName})
+		}
+	}
+	return keys
+}
+
+// allKeys returns all NodeLauncherKeys in the digest.
```
MikeSpreitzer: This is a pretty expensive thing to do in a large cluster.
I see it used only once, and that is in a place that uses only the length of the result!
It would be much cheaper to compute that count than this data structure.

pkg/controller/launcher-populator/digested-policy.go
```diff
@@ -0,0 +1,195 @@
...
+			keys = append(keys, NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName})
+		}
+	}
+	return keys
+}
+
+// snapshotForKey returns a value-typed snapshot of the digestEntry and its
+// LC's templateHash, taken under RLock. Concurrency-safe; callers must drop
+// the snapshot before treating any pointer fields (spec) as still belonging
+// to the digest.
```
MikeSpreitzer, line 143: I am not sure I understand what this is saying.
Isn't there a simpler story: nothing mutates anything pointed to by anything in the returned struct?

pkg/controller/launcher-populator/helpers.go
```diff
@@ -0,0 +1,165 @@
...
+// getMatchingNodes returns nodes that match the EnhancedNodeSelector.
+// It returns three values: the matched nodes, user-facing selector errors (non-nil when the
+// LabelSelector itself is malformed — this is a user configuration error), and an internal
+// error (non-nil for unexpected infrastructure failures such as lister errors).
+// Callers should handle selectorErrs and err independently.
+func (ctl *controller) getMatchingNodes(ctx context.Context, selector fmav1alpha1.EnhancedNodeSelector) ([]corev1.Node, []string, error) {
+	// Convert the label selector. A failure here is a user error (malformed LabelSelector).
+	labelSelector, selectorErr := metav1.LabelSelectorAsSelector(&selector.LabelSelector)
+	if selectorErr != nil {
+		return nil, []string{fmt.Sprintf("invalid label selector: %v", selectorErr)}, nil
```
MikeSpreitzer: IMO `%#v` is a better format to use when reporting an error in a data structure, due to showing the value more precisely.

pkg/controller/launcher-populator/helpers.go
```diff
@@ -0,0 +1,165 @@
...
+
+// nonNilSlice returns a single-element slice if s is non-empty, or nil otherwise.
+func nonNilSlice(s string) []string {
+	if s == "" {
+		return nil
+	}
+	return []string{s}
+}
+
+// getMatchingNodes returns nodes that match the EnhancedNodeSelector.
```
MikeSpreitzer, line 55: I looked at the call sites. All they need is a `sets.Set[/*Node name*/string]`. The call sites would get simpler if that were what this func returned.

pkg/controller/launcher-populator/helpers.go
```diff
@@ -0,0 +1,165 @@
...
+func (ctl *controller) createLaunchers(ctx context.Context, node corev1.Node, key NodeLauncherKey, count int, lcSpec *fmav1alpha1.LauncherConfigSpec, lcOwnerRef metav1.OwnerReference, templateHash string) error {
+	logger := klog.FromContext(ctx)
+
+	// Create the specified number of launcher pods
+	for i := 0; i < count; i++ {
+		pod, err := utils.BuildLauncherPodFromTemplate(lcSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName, templateHash)
+		if err != nil {
+			return fmt.Errorf("failed to build launcher pod: %w", err)
+		}
+		pod.OwnerReferences = []metav1.OwnerReference{lcOwnerRef}
```
MikeSpreitzer, line 149: This could be hoisted above the loop, right?
(My recommended refactoring of BuildLauncherPodFromTemplate leads to much more being hoisted.)

pkg/controller/utils/pod-helper.go
MikeSpreitzer: What refinement are we talking about here? A launcher is dedicated to a Node but NOT to a set of GPUs. I see no need to add ";gpus=all" into the string being hashed.

MikeSpreitzer: The Node name has already been input to the hash once, due to presence in a label; I see no need to input it again.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+			for _, lppName := range ctl.policy.lppNamesRefByLC(name) {
+				ctl.digestQueue.Queue.Add(funcItem{kind: kindLPP, name: lppName})
+			}
+		}
+		return nil
+	}
+
+	// LC exists: validate and produce the lcDigest.
+	var templateErr string
+	if vErr := utils.ValidateLauncherPodTemplate(lc.Spec.PodTemplate); vErr != nil {
```
MikeSpreitzer, line 67: I see no benefit in having ValidateLauncherPodTemplate and ComputeLauncherTemplateHash be separate functions. The controllers never want a launcher template created from an invalid `lc.Spec.PodTemplate`.

pkg/controller/utils/pod-helper.go
```diff
@@ -191,7 +262,10 @@ func BuildLauncherPodFromTemplate(template v1alpha1.EmbeddedPodTemplateSpec, ns,
 	if pod.Annotations == nil {
 		pod.Annotations = make(map[string]string)
 	}
+	// Legacy node-aware hash for the dual-pods indexer.
 	pod.Annotations = MapSet(pod.Annotations, string(common.LauncherConfigHashAnnotationKey), nominalHash)
```
MikeSpreitzer: The whole point of `MapSet` is to cope with the possibility that the given `map` is `nil`. You do not need _both_ the `if` statement above that prevents that possibility _and_ the complexity of using `MapSet` instead of just `pod.Annotations[common.LauncherConfigHashAnnotationKey] = nominalHash`.

pkg/controller/utils/pod-helper.go
MikeSpreitzer, line 1: I still see a lot of code duplication in this file and unnecessary complexity in its callers. ValidateLauncherPodTemplate and ComputeLauncherTemplateHash just take bites out of the node-independent digestion problem. I think that we can get smaller simpler code if we delete those and factor BuildLauncherPodFromTemplate into the following two pieces. The first piece, NodeIndependentLauncherTemplate, gets called in just one place in each controller; in the launcher population controller, that is the place that stores its results in `ctl.policy.lcs[lc.Name]`. The second piece uses the partially-constructed Pod template that was output by the first piece. This factoring was done by taking BuildLauncherPodFromTemplate and (1) restoring the canonicalization and error handling from ComputeLauncherTemplateHash and hoisting the OwnerReference construction here and then (2) factoring by sweeping all the dependence on the Node name into the second half.

```
// NodeIndependentLauncherTemplate does as much work as possible toward building the Pod
// template for stamping out launchers as can be done without dedicating to a particular node.
// The result is a Pod that needs only the dedication to a Node,
// and a user-facing error (if any).
// The result has the node-independent hash in its LauncherTemplateHashAnnotationKey annotation.
func NodeIndependentLauncherTemplate(lc *v1alpha1.LauncherConfig) (*corev1.Pod, error) {
	template := canonicalizeTemplateForHash(lc.Spec.PodTemplate)
	pod := &corev1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			Labels:      maps.Clone(template.Metadata.Labels),
			Annotations: maps.Clone(template.Metadata.Annotations),
			OwnerReferences: []metav1.OwnerReference{{
				APIVersion:         v1alpha1.SchemeGroupVersion.String(),
				Kind:               "LauncherConfig",
				Name:               lc.Name,
				UID:                lc.UID,
				Controller:         ptr.To(false),
				BlockOwnerDeletion: ptr.To(false),
			}},
			Namespace:    lc.Namespace,
			GenerateName: fmt.Sprintf("launcher-%s-", lc.Name),
		},
		Spec: *DeIndividualize(template.Spec.DeepCopy()),
	}
	// Ensure labels are set
	if pod.Labels == nil {
		pod.Labels = make(map[string]string)
	}
	pod.Labels[common.ComponentLabelKey] = common.LauncherComponentLabelValue
	pod.Labels[common.LauncherConfigNameLabelKey] = lc.Name
	pod.Labels[api.SleepingLabelName] = "false"

	hasher := sha256.New()
	modifiedJSON, err := json.Marshal(pod)
	if err != nil {
		return nil, fmt.Errorf("failed to marshal template for hashing: %w", err)
	}
	hasher.Write(modifiedJSON)
	var modifiedHash [sha256.Size]byte
	modifiedHashSl := hasher.Sum(modifiedHash[:0])
	templateHash := base64.RawStdEncoding.EncodeToString(modifiedHashSl)

	if pod.Annotations == nil {
		pod.Annotations = make(map[string]string)
	}
	// Node-independent template fingerprint for spec-drift detection.
	pod.Annotations[common.LauncherTemplateHashAnnotationKey] = templateHash

	cIdx, err := GetInferenceServerContainerIndex(pod)
	if err != nil {
		return nil, err
	}
	container := &pod.Spec.Containers[cIdx]

	// Configure required environment variables
	configureRequiredEnvVars(container)

	// Set fixed liveness probe
	container.LivenessProbe = &corev1.Probe{
		ProbeHandler: corev1.ProbeHandler{
			HTTPGet: &corev1.HTTPGetAction{
				Path:   "/health",
				Port:   intstr.FromInt(common.LauncherServicePort),
				Scheme: corev1.URISchemeHTTP,
			},
		},
		InitialDelaySeconds: 10,
		PeriodSeconds:       20,
		TimeoutSeconds:      1,
		SuccessThreshold:    1,
		FailureThreshold:    3,
	}

	// Set readiness probe to check if launcher can list instances.
	// This is necessary because otherwise the dual-pods controller will be confused when
	// the launcher Pod is said to be ready but got refused when listing its vLLM instances.
	container.ReadinessProbe = &corev1.Probe{
		ProbeHandler: corev1.ProbeHandler{
			HTTPGet: &corev1.HTTPGetAction{
				Path:   "/v2/vllm/instances",
				Port:   intstr.FromInt(common.LauncherServicePort),
				Scheme: corev1.URISchemeHTTP,
			},
		},
		InitialDelaySeconds: 2,
		PeriodSeconds:       5,
		TimeoutSeconds:      2,
		SuccessThreshold:    1,
		FailureThreshold:    3,
	}

	// Remove nvidia.com/gpu from resource limits
	removeGPUResourceLimits(container)

	// Remove nvidia.com/gpu from Pod-level resource overhead
	if pod.Spec.Overhead != nil {
		delete(pod.Spec.Overhead, corev1.ResourceName("nvidia.com/gpu"))
	}

	addLauncherNotifierSidecar(pod, container.Image, container.ImagePullPolicy)
	return pod, nil
}

// FinishLauncherPodTemplate takes a Pod template that was constructed
// from a v1alpha1.EmbeddedPodTemplateSpec by doing all the relevant
// processing except dedicating it to a Node, and returns a pointer to a Pod
// template that is dedicated to the given node.
// The nodeIndependent input is treated as deeply immutable.
// Its LauncherConfigHashAnnotationKey annotation holds a hash value that
// takes this dedication into account.
func FinishLauncherPodTemplate(nodeIndependent *corev1.Pod, nodeName string) *corev1.Pod {
	hasher := sha256.New()
	hasher.Write([]byte(nodeIndependent.Annotations[common.LauncherTemplateHashAnnotationKey]))
	hasher.Write([]byte(";node="))
	hasher.Write([]byte(nodeName))
	var modifiedHash [sha256.Size]byte
	modifiedHashSl := hasher.Sum(modifiedHash[:0])
	nominalHash := base64.RawStdEncoding.EncodeToString(modifiedHashSl)

	dedicated := *nodeIndependent
	dedicated.ObjectMeta.Labels = maps.Clone(dedicated.ObjectMeta.Labels)
	dedicated.ObjectMeta.Labels[common.NodeNameLabelKey] = nodeName
	dedicated.Spec.NodeSelector = maps.Clone(dedicated.Spec.NodeSelector)
	if dedicated.Spec.NodeSelector == nil {
		dedicated.Spec.NodeSelector = make(map[string]string)
	}
	dedicated.Spec.NodeSelector["kubernetes.io/hostname"] = nodeName
	// Legacy node-aware hash for the dual-pods indexer.
	dedicated.ObjectMeta.Annotations = maps.Clone(dedicated.ObjectMeta.Annotations)
	dedicated.Annotations[common.LauncherConfigHashAnnotationKey] = nominalHash
	return &dedicated
}
```

pkg/controller/launcher-populator/populator.go
```diff
@@ -77,10 +74,30 @@ func NewController(
...
+	// digest mutations sequential and lock-free. KnowsProcessedSync emits the
+	// onDigestSyncProcessed hook after the initial batch drains, which is when
+	// keyQueue's workers start.
+	digestQueue := genctlr.NewKnowsProcessedSync[queueItem](
+		ControllerName+"-digest", 1,
+		ctl.processDigestItem,
+		makeDigestSentinel, isDigestSentinel,
+		ctl.onDigestSyncProcessed,
+	)
+	ctl.digestQueue = &digestQueue
```
MikeSpreitzer, line 88: Why add a level of pointer indirection? It wasn't there before, see https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/v0.6.0-alpha.12/pkg/controller/launcher-populator/populator.go#L83

(Hmm, I see that https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/v0.6.0-alpha.12/pkg/controller/launcher-populator/populator.go#L118 has been using the wrong type, KnowsProcessedSync instead of QueueAndWorkers; that produced no evident errors because KnowsProcessedSync embeds QueueAndWorkers.)

pkg/controller/launcher-populator/populator.go
```diff
@@ -115,49 +132,73 @@ type controller struct {
...
-//     When #472 is resolved, this sentinel can be replaced with concrete
-//     object references analogous to lppItem / lcItem.
-type queueItem interface {
-	// process returns (err error, retry bool).
-	// There will be a retry iff `retry`, error logged if `err != nil`.
-	process(ctx context.Context, ctl *controller) (error, bool)
+// queueItem is the work-queue element type for the digest queue. Concrete
+// elements are funcItem (LPP/LC/Node references) plus sentinel funcItems
+// (kind=kindSentinel) emitted by KnowsProcessedSync. processDigestItem
+// dispatches on the dynamic type. The element type is a value-typed struct so
```
MikeSpreitzer, line 160: The words "the dynamic type", since `processDigestItem` is given an `any` (`interface{}` == `any`), are most likely to be interpreted to mean the type of the value boxed in the `any`. But that is not what I see going on here. I see `processDigestItem` insist that the value boxed inside the `any` be a `funcItem` and then dispatch on the `kind` field of the `funcItem`.
It would be simpler and clearer to explain if either (A) this went back to defining `queueItem` to have a `process` method and making `processDigestItem` simply invoke that method, while defining a specific type with its own specific `process` method for each kind of digest queue item, or (B) make the type parameter of the digest queue simply be `funcItem`. Approach (A) is what this file used to do, and it avoids the possibility of an error at dispatch time (see https://github.com/osswangxining/llm-d-fast-model-actuation/blob/84645997d9837c6a15beb7208e75f4b6d607e1a2/pkg/controller/launcher-populator/populator.go#L266-L270).

pkg/controller/launcher-populator/populator.go
```diff
@@ -175,20 +216,21 @@ func (ctl *controller) OnAdd(obj any, isInInitialList bool) {
...
+		nodeName := typed.Labels[common.NodeNameLabelKey]
+		lcName := typed.Labels[common.LauncherConfigNameLabelKey]
+		ctl.enqueueLogger.V(5).Info("Enqueuing key due to launcher Pod add",
+			"name", typed.Name, "uid", typed.UID, "resourceVersion", typed.ResourceVersion,
+			"node", nodeName, "config", lcName)
+		ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName}})
 	case *corev1.Node:
-		ctl.enqueueLogger.V(5).Info("Enqueuing reconciliation due to Node add", "name", typed.Name)
-		ctl.Queue.Add(reconcileAllSentinel{})
+		ctl.enqueueLogger.V(5).Info("Enqueuing digest update due to Node add", "name", typed.Name)
```
MikeSpreitzer: Technically what is being enqueued is an object reference, not an action.

pkg/controller/launcher-populator/populator.go
```diff
@@ -115,49 +132,73 @@ type controller struct {
...
 	lcLister      fmalisters.LauncherConfigLister
-	genctlr.KnowsProcessedSync[queueItem]
+
+	// digestQueue processes API-object references (LC/LPP/Node). Single worker;
+	// the SOLE writer of ctl.policy.
+	digestQueue *genctlr.KnowsProcessedSync[queueItem]
+
+	// keyQueue processes per-NodeLauncherKey reconciliation. Started by
+	// onDigestSyncProcessed once the initial batch has drained.
+	keyQueue *genctlr.QueueAndWorkers[keyItem]
```
MikeSpreitzer, line 141: Wouldn't it be simpler to make the type parameter here be `NodeLauncherKey`? I do not see the benefit of wrapping into a singleton struct.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
-	// LauncherConfigs are handled inside buildDesiredStateFromPolicies.
-	return ctl.reconcileFromPolicies(ctx)
-}
+	// Enqueue all currently-known LCs/LPPs/Nodes into the digest queue. The
+	// initial batch is processed through the same incremental code paths
+	// (updateDigestForLC/LPP/Node) used at runtime, so there is no separate
+	// "initial digest" code path.
+	if err := ctl.enqueueInitialBatch(); err != nil {
+		return fmt.Errorf("failed to enqueue initial batch: %w", err)
+	}
```
MikeSpreitzer: This is not needed. The call to `WaitForNamedCacheSync(ControllerName, ctx.Done(), ctl.lppInformer.HasSynced, ctl.lcInformer.HasSynced, ctl.podInformer.HasSynced, ctl.nodeInformer.HasSynced)` above already got this job done, without concurrency bugs.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
-				entry.LauncherConfigSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
-			if err != nil {
-				// Should not happen: template was already validated upstream.
-				logger.Error(err, "Unexpected error building nominal pod (template should have been validated)",
-					"node", nodeName, "config", key.LauncherConfigName)
-			} else {
-				nominalHash = nominalPod.Annotations[string(common.LauncherConfigHashAnnotationKey)]
+		if nominalHash != "" {
+			podHash := pod.Annotations[string(common.LauncherTemplateHashAnnotationKey)]
+			// Pre-rollout Pods lack this annotation; treat empty as drift-free to avoid mass replacement.
```
MikeSpreitzer: What "rollout" is this talking about?
This repo does not (yet) attempt to support in-place upgrades.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
 			}
-
-			isBound, _ := ctl.isLauncherBoundToServerRequestingPod(pod)
-			if isBound {
-				liveBoundCount++
+			if apierrors.IsConflict(err) {
+				staleNotDeleted++
 				continue
 			}
+			return fmt.Errorf("failed to delete stale launcher pod %s: %w", pod.Name, err), false
```
MikeSpreitzer: In this case the returned tuple should stipulate a retry. That is done by having the boolean value be `true`.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
 			})
 			if err != nil {
 				if apierrors.IsNotFound(err) || apierrors.IsGone(err) {
-					// Pod is already gone from the apiserver, but the informer cache
-					// may not yet reflect that. Record the deletion expectation so the
-					// next reconcile waits for cache synchronization rather than trying
-					// to act on a Pod that no longer exists.
 					ctl.expectations.expectDeletion(key, pod.UID)
-					logger.Info("Stale launcher pod already deleted",
-						"pod", pod.Name, "uid", pod.UID, "resourceVersion", pod.ResourceVersion)
```
MikeSpreitzer, line 589: I would keep the logging in this case.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
-		if entry.LauncherConfigSpec != nil {
-			nominalPod, err := utils.BuildLauncherPodFromTemplate(
-				entry.LauncherConfigSpec.PodTemplate, ctl.namespace, key.NodeName, key.LauncherConfigName)
-			if err != nil {
-				// Should not happen: template was already validated upstream.
-				logger.Error(err, "Unexpected error building nominal pod (template should have been validated)",
-					"node", nodeName, "config", key.LauncherConfigName)
-			} else {
-				nominalHash = nominalPod.Annotations[string(common.LauncherConfigHashAnnotationKey)]
+		if nominalHash != "" {
```
MikeSpreitzer: Why would nominalHash ever be empty here? In the hands-off case (missing or malformed LauncherConfig), this line should not be reached (there should not be any launchers created or deleted for this NodeLaunchereKey).

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
+			podHash := pod.Annotations[string(common.LauncherTemplateHashAnnotationKey)]
+			// Pre-rollout Pods lack this annotation; treat empty as drift-free to avoid mass replacement.
+			if podHash != "" && podHash != nominalHash {
+				staleUnboundPods = append(staleUnboundPods, pod)
```
MikeSpreitzer, line 433: As before, it would be simpler to do the deletion here rather than build a list and later iterate over it.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
 			}
-			// Record expectation for this specific Pod UID immediately after deletion.
 			ctl.expectations.expectDeletion(key, pod.UID)
-			logger.Info("Deleted stale launcher pod (spec changed)",
-				"pod", pod.Name,
-				"uid", pod.UID,
-				"resourceVersion", pod.ResourceVersion,
-				"node", nodeName,
-				"config", key.LauncherConfigName)
+			logger.Info("Deleted excess launcher pod", "pod", pod.Name, "uid", pod.UID, "node", key.NodeName)
```
MikeSpreitzer, line 493: The name of the LauncherConfig should definitely be included here.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +305,254 @@ func (ctl *controller) Start(ctx context.Context) error {
...
-// For each LauncherConfig, it does deletions immediately as they are identified
-// and remembers creations called for. If any deletions were performed (or are in
-// progress from a previous cycle), it returns true to request a requeue so that
-// creations happen only after deletions have taken effect, minimizing peak resource
-// consumption on the node.
-// So that when a LauncherConfig changes, each corresponding launcher Pod that is
-// not bound to a server-requesting Pod is deleted and replaced.
-func (ctl *controller) reconcileLaunchersOnSingleNode(ctx context.Context, nodeName string, keys []NodeLauncherKey, desired map[NodeLauncherKey]DesiredStateEntry) (bool, error) {
+// reconcileKey adjusts launcher pods for a single NodeLauncherKey to match the desired count.
+func (ctl *controller) reconcileKey(ctx context.Context, key NodeLauncherKey, desiredCount int32, lcSpec *fmav1alpha1.LauncherConfigSpec, ownerRef metav1.OwnerReference, nominalHash string) (error, bool) {
```
MikeSpreitzer: The last parameter here is the node-independent "templateHash", rather than the node-specific "nominalHash", right?

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+	logger := klog.FromContext(ctx)
+
+	_, err := ctl.nodeLister.Get(nodeName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("Node deleted, removing from digest", "node", nodeName)
+			removedKeys := ctl.policy.removeNode(nodeName)
+			for _, key := range removedKeys {
+				ctl.keyQueue.Queue.Add(keyItem{key})
+			}
```
MikeSpreitzer: It is not necessary for this controller to delete the launcher Pods on a Node that does not exist: Kubernetes will take care of that itself.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+// recomputeDigestForNode rebuilds digest entries for a single node by replaying
+// every cached LPP. Pure read of ctl.policy.lpps + ctl.policy.lcs.
+func (ctl *controller) recomputeDigestForNode(nodeName string) error {
+	ctl.policy.digest[nodeName] = make(map[string]*digestEntry)
+
+	for _, lppd := range ctl.policy.lpps {
+		if lppd == nil || lppd.object == nil {
+			continue
```
MikeSpreitzer: This line is not reachable.
This code, unsurprisingly, has the invariant that every (key,value) pair in `ctl.policy.lpps` has a non-nil value that holds a non-nil `object`. I suggest documenting and keeping that invariant, for simplicity.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+			return nil
+		}
+		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+	}
+	return ctl.recomputeDigestForNode(nodeName)
+}
+
+// recomputeDigestForNode rebuilds digest entries for a single node by replaying
+// every cached LPP. Pure read of ctl.policy.lpps + ctl.policy.lcs.
+func (ctl *controller) recomputeDigestForNode(nodeName string) error {
```
MikeSpreitzer: I think that it would be simpler to make this method take the whole non-nil `*corev1.Node` rather than deal with the complexity of nested lookups and the possibility of that returning an error.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+func (ctl *controller) applyLPPToDigestForNode(lpp *fmav1alpha1.LauncherPopulationPolicy, nodeName string) {
+	node, err := ctl.nodeLister.Get(nodeName)
+	if err != nil {
+		// Node missing or transient lister error: nothing to apply.
+		return
+	}
+	labelSelector, selectorErr := metav1.LabelSelectorAsSelector(&lpp.Spec.EnhancedNodeSelector.LabelSelector)
+	if selectorErr != nil {
+		return
+	}
```
MikeSpreitzer, line 280: Cache this in the `lppDigest` rather than repeat it (often!) here.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+// recomputeDigestForNode rebuilds digest entries for a single node by replaying
+// every cached LPP. Pure read of ctl.policy.lpps + ctl.policy.lcs.
+func (ctl *controller) recomputeDigestForNode(nodeName string) error {
+	ctl.policy.digest[nodeName] = make(map[string]*digestEntry)
+
+	for _, lppd := range ctl.policy.lpps {
+		if lppd == nil || lppd.object == nil {
+			continue
+		}
+		ctl.applyLPPToDigestForNode(lppd.object, nodeName)
```
MikeSpreitzer: As it is currently written, `ctl.applyLPPToDigestForNode` only attempts the positive side --- it does not update the `digestEntry` in the case where the LPP _used to_ match the Node but no longer does.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+	ctl.removeLPPFromEntry(entry, lppName)
+	if len(entry.lpps) == 0 {
+		delete(nodeMap, lcName)
+		if len(nodeMap) == 0 {
+			delete(ctl.policy.digest, nodeName)
+		}
+	} else {
+		ctl.recomputeEntryFromLPPs(entry, lcName)
+	}
+	ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: nodeName, LauncherConfigName: lcName}})
```
MikeSpreitzer: The key needs to be enqueued only if there was a change in `entry.handsOff` or `entry.count`.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+			ctl.detachLPPFromEntry(nodeMap, name, nodeName, lcName, entry)
+		}
+	}
+
+	// Apply LPP to each matched node and enqueue affected keys.
+	for _, node := range currentMatchedNodes {
+		ctl.applyLPPToDigestForNode(lpp, node.Name)
+		for _, cr := range lpp.Spec.CountForLauncher {
+			ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: node.Name, LauncherConfigName: cr.LauncherConfigName}})
+		}
```
MikeSpreitzer, line 221: This probably enqueues more keys than necessary.
It is not a functional error to do unnecessary enqueuing, but it is a performance problem.
The only keys that need to be enqueued are those for which the `handsOff` or desired count changed.

Both calls to `ctl.applyLPPToDigestForNode` are followed by excessive enqueuing. It would be better if `ctl.applyLPPToDigestForNode` itself did the enqueuing, and only for keys where the `handsOff` or desired count changed.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+			entry.spec = nil
+		default:
+			entry.handsOff = false
+			entry.spec = &lcd.object.Spec
+			entry.ownerRef = lcd.ownerRef
+		}
+
+		if cr.LauncherCount > entry.count {
+			entry.count = cr.LauncherCount
+		}
```
MikeSpreitzer: I see no implementation of the complementary possibility, which is that `entry.count` should be decreased because `cr.LauncherCount` decreased. Indeed, that direction is trickier: `entry.count` should decrease only when the MAX across all the relevant `CountForLauncher` has decreased.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+	// LPPs that reference it must be re-evaluated.
+	if !prevExists || prevTemplateErr != templateErr {
+		for _, lppName := range ctl.policy.lppNamesRefByLC(name) {
+			ctl.digestQueue.Queue.Add(funcItem{kind: kindLPP, name: lppName})
+		}
+	}
+
+	// Refresh per-key entries that reference this LC.
+	for _, key := range ctl.policy.keysForLC(name) {
+		entry := ctl.policy.getEntry(key.NodeName, key.LauncherConfigName)
```
MikeSpreitzer: Note that every call site uses `keysForLC` in this same way: in the `range` of a loop, whose body starts with `getEntry`. It would be cheaper and simpler to replace `keysForLC` with a `func(lcName string) iter.Seq2[NodeLauncherKey, *digestEntry]` (see https://pkg.go.dev/iter@go1.26.3#Seq2), which is then used in the loop's `range` (using the "function, 2 values" case of https://go.dev/ref/spec#For_range).

pkg/controller/launcher-populator/digest-updater.go
MikeSpreitzer: It is not necessary to do this work; the processing of the LPP triggered by the enqueuing below covers it.

Note that enqueuing the NodeLauncherKey now --- before updating the `digestEntry` --- would be positively wrong if it were not also enqueued by the coming LPP processing.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,392 @@
...
+		templateHash: templateHash,
+		ownerRef:     makeLCOwnerRef(lc),
+	}
+
+	// If the LC just became existent, or its template-validity flipped, the
+	// LPPs that reference it must be re-evaluated.
+	if !prevExists || prevTemplateErr != templateErr {
+		for _, lppName := range ctl.policy.lppNamesRefByLC(name) {
+			ctl.digestQueue.Queue.Add(funcItem{kind: kindLPP, name: lppName})
+		}
```
MikeSpreitzer, line 102: When this enqueuing is done, the work below per `ctl.policy.keysForLC(name)` is unnecessary because the triggered processing per LPP will cover it.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/digest-updater.go
osswangxining: Polish this, while current implementation logic is correct.

pkg/controller/launcher-populator/digest-updater.go
osswangxining: latter

pkg/controller/launcher-populator/digest-updater.go
osswangxining: done

pkg/controller/launcher-populator/digest-updater.go
osswangxining: Done. Replaced keysForLC with entriesForLC returning iter.Seq2[NodeLauncherKey, *digestEntry];

pkg/controller/launcher-populator/digest-updater.go
osswangxining: agreed, dropped the loop and the now-unused removeNode helper

pkg/controller/launcher-populator/digest-updater.go
osswangxining: done

pkg/controller/launcher-populator/digest-updater.go
osswangxining: Done. recomputeDigestForNode now takes a *corev1.Node and returns nothing

pkg/controller/launcher-populator/digest-updater.go
osswangxining: Agreed it should be cached — I'd like to do this in the same follow-up PR as I think that it's not a blocker instead of improvement , but need a lot of refactors things.

pkg/controller/launcher-populator/digest-updater.go
osswangxining: ok

pkg/controller/launcher-populator/digest-updater.go
osswangxining: agree

pkg/controller/launcher-populator/digest-updater.go
osswangxining: agree, done

pkg/controller/launcher-populator/digest-updater.go
osswangxining: done

pkg/controller/launcher-populator/helpers.go
osswangxining: agree and done

pkg/controller/launcher-populator/helpers.go
osswangxining: still need node after refactoring, pls give second overall check.

pkg/controller/utils/pod-helper.go
osswangxining: yes make sense;

osswangxining: ok

pkg/controller/utils/pod-helper.go
osswangxining: Agreed and updated this, thanks.

pkg/controller/launcher-populator/populator.go
osswangxining: agree, refined

pkg/controller/launcher-populator/populator.go
osswangxining: polished

pkg/controller/launcher-populator/populator.go
osswangxining: polished

pkg/controller/launcher-populator/populator.go
osswangxining: agree, typo

pkg/controller/launcher-populator/populator.go
osswangxining: yes, node-independent "templateHash" is preferred.

pkg/controller/launcher-populator/populator.go
```diff
@@ -77,10 +73,30 @@ func NewController(
 		lcLister:      fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Lister(),
 		expectations:  newPendingExpectations(expectationTimeout),
 	}
+	ctl.policy = newDigestedPolicy()
+
+	// digestQueue carries funcItem (LC/LPP/Node references). One worker keeps
+	// digest mutations sequential and lock-free. KnowsProcessedSync emits the
```
MikeSpreitzer: The "lock-free" part is not accurate (the worker method _does_ do locking). If I understand your intent correctly, you are tying to say that no locking is need to ensure mutual exclusion between multiple goroutines doing digestion. I think that is clearly enough established by the other words saying that there is just one goroutine and the digest mutations are serial. IMO we can just delete the words "and lock-free" here.

pkg/controller/launcher-populator/populator.go
```diff
@@ -256,463 +304,213 @@ func (ctl *controller) Start(ctx context.Context) error {
...
+func (ctl *controller) onDigestSyncProcessed(ctx context.Context) {
 	logger := klog.FromContext(ctx)
-
-	// Build desired state from all policies
-	populationPolicy, err := ctl.buildDesiredStateFromPolicies(ctx)
-	if err != nil {
-		logger.Error(err, "Failed to build desired state from policies")
-		return err, true
+	logger.V(1).Info("Initial digest batch processed; starting key workers",
+		"lpps", len(ctl.policy.lpps), "lcs", len(ctl.policy.lcs), "keys", len(ctl.policy.allKeys()))
```
MikeSpreitzer: Did you know that a workqueue has a `Len` function? `ctl.keyQueue.Queue.Len()` gives the number of keys that are in the queue at the moment.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// updateDigestForLPP is the SOLE place that:
+//   - runs getMatchingNodes (collecting selector errors);
+//   - determines which referenced LCs are missing (by reading ctl.policy.lcs);
+//   - writes LauncherPopulationPolicy.Status (via setLPPStatusErrors);
+//   - records the per-LPP derived data into ctl.policy.lpps[name];
+//   - applies the LPP to digest entries on every matched node.
+//
+// Other code paths must NOT call setLPPStatusErrors or ValidateLauncherPodTemplate.
```
MikeSpreitzer: I suspect "ValidateLauncherPodTemplate" is a typo here, "getMatchingNodes" was meant.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,362 @@
...
+				continue
+			}
+			ctl.detachLPPFromEntry(nodeMap, name, nodeName, lcName, entry)
+		}
+	}
+
+	// Apply LPP to each matched node and enqueue affected keys.
+	for i := range currentMatchedNodes {
+		node := &currentMatchedNodes[i]
+		ctl.applyLPPToDigestForNode(lpp, node)
```
MikeSpreitzer, line 218: Correctness error: `ctl.applyLPPToDigestForNode` does not currently update the digest properly for a LauncherConfig name that formerly appeared in `lpp.Spec.CountForLauncher` but does not now. The key enumeration below has the same oversight.

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,362 @@
...
+func (ctl *controller) recomputeDigestForNode(node *corev1.Node) {
+	ctl.policy.digest[node.Name] = make(map[string]*digestEntry)
+
+	for _, lppd := range ctl.policy.lpps {
+		ctl.applyLPPToDigestForNode(lppd.object, node)
+	}
+
+	nodeMap := ctl.policy.digest[node.Name]
+	for lcName := range nodeMap {
+		ctl.keyQueue.Queue.Add(keyItem{NodeLauncherKey{NodeName: node.Name, LauncherConfigName: lcName}})
```
MikeSpreitzer, line 266: Correctness error: no keys enqueued in the case of an LPP that used to match the Node but no longer does.

pkg/controller/launcher-populator/helpers.go
MikeSpreitzer: Yes, I see that the remaining call site uses more than the name of the Node now.

MikeSpreitzer (commented):
I did another round of review. I noted two correctness problems, and a few less critical things.

pkg/controller/launcher-populator/digest-updater.go
osswangxining: yes, copy typo

pkg/controller/launcher-populator/digest-updater.go
osswangxining: make sense

pkg/controller/launcher-populator/digest-updater.go
osswangxining: make sense

pkg/controller/launcher-populator/populator.go
osswangxining: make sense

pkg/controller/launcher-populator/populator.go
osswangxining: updated thanks

MikeSpreitzer:
I suspect that the CI failures are flakes due to transient problems with quay. But I am not going to retry yet, because of https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/529#discussion_r3336295086

pkg/controller/launcher-populator/digest-updater.go
```diff
@@ -0,0 +1,396 @@
...
+// every cached LPP. Pure read of ctl.policy.lpps + ctl.policy.lcs.
+func (ctl *controller) recomputeDigestForNode(node *corev1.Node) {
+	// Snapshot LC names currently in the digest for this node so we can also
+	// enqueue keys for entries that disappear after replay (e.g., a previously
+	// matching LPP no longer matches this node).
+	affected := sets.New[string]()
+	if oldMap, ok := ctl.policy.digest[node.Name]; ok {
+		for lcName := range oldMap {
+			affected.Insert(lcName)
+		}
```
MikeSpreitzer: FYI, this is available as https://github.com/kubernetes/apimachinery/blob/v0.34.8/pkg/util/sets/set.go#L38-L46

MikeSpreitzer (approved):
I do not see any correctness errors now.
This review is very long.
We will continue work in follow-on PRs.

pkg/controller/launcher-populator/digest-updater.go
MikeSpreitzer: I am sorry, my earlier approval overlooked this bug. There is still a correctness problem here.

MikeSpreitzer (changes requested):

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: The comment is gone and the variable has been renamed to podHash but I still see the same code structure.

MikeSpreitzer:
https://github.com/osswangxining/llm-d-fast-model-actuation/pull/3 suggests a fix to the one bug that I am aware of.

pkg/controller/launcher-populator/digest-updater.go
osswangxining: switched to use this sets.KeySet method

MikeSpreitzer (approved):
This looks good enough to merge.
The one CI failure is due to the problem that #529 just fixed.

PR #529, opened 2026-06-01
Apply instead commit for run-launcher-based.sh

diegocastanibm:
To solve this bug:
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/26758603129/job/78865326914?pr=458

### DIFF
--- a/test/e2e/run-launcher-based.sh
+++ b/test/e2e/run-launcher-based.sh
@@ -53,7 +53,6 @@ kubectl get pods -A -o wide
 
 kubectl create clusterrole node-viewer --verb=get,list,watch --resource=nodes
 
-kubectl apply --server-side -f ./config/crd/
 
 kubectl apply -f - <<EOF
 apiVersion: rbac.authorization.k8s.io/v1
--- a/test/e2e/run-launcher-based.sh
+++ b/test/e2e/run-launcher-based.sh
@@ -53,7 +53,7 @@ kubectl get pods -A -o wide
 
 kubectl create clusterrole node-viewer --verb=get,list,watch --resource=nodes
 
-kubectl create -f ./config/crd/
+kubectl apply --server-side -f ./config/crd/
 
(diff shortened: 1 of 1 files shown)


rubambiza (dismissed):
LGTM

test/e2e/run-launcher-based.sh
```diff
@@ -53,7 +53,7 @@ kubectl get pods -A -o wide
 
 kubectl create clusterrole node-viewer --verb=get,list,watch --resource=nodes
 
-kubectl create -f ./config/crd/
+kubectl apply --server-side -f ./config/crd/
```
MikeSpreitzer: I think that the better fix is to delete this line. Later in this script is an invocation of `test/e2e/deploy_fma.sh`, which is the standard way to deploy FMA --- including its CRDs.

MikeSpreitzer: But I also wonder why this is a problem now, even though this file and `test/e2e/deploy_fma.sh` have been this way for months.

diegocastanibm: I am fully in favor of eliminating redundancies

MikeSpreitzer: I am in favor of this bug fix, but something else is bothering me. Why did this inconsistency suddenly become a fatal problem? It has been present for months. I wonder if this signals a supply chain attack.

For example, https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/26738732178/attempts/2 failed but https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/26738732178/attempts/1 succeeded.

MikeSpreitzer:
I have reported the broken runner metadata two ways to GitHub, no response. I am not seeing reports of the runner being compromised. So I am going to proceed on that assumption.

MikeSpreitzer (approved):
LGTM

Issue #385, opened 2026-03-25
Resource usage observability and optimization


waltforme:
This issue captures initial thoughts on observability for resource usage and possible optimizations.

We currently do not have visibility into the resource usage of FMA-managed vLLM instances. That observability could be provided at multiple levels, for example:

- per instance
- per launcher pod
- per node
- cluster-wide / overall

One possible way to close this gap is to add metrics and dashboards to the project.

Once we have that data, the next step is to explore optimization opportunities. There may be several possible objectives, such as:

- minimizing overall latency
- prioritizing the most actively served models
- improving overall resource efficiency

MikeSpreitzer:
@diegocastanibm: I wonder whether the launcher should make a child cgroup for each launched vllm instance, at least for clear reporting if not also for enforcement of limits.

diegocastanibm:
> [@diegocastanibm](https://github.com/diegocastanibm): I wonder whether the launcher should make a child cgroup for each launched vllm instance, at least for clear reporting if not also for enforcement of limits.

Each vLLM instance already isolates itself into its own process group (`os.setpgrp()` at line 815), and the stop path uses `os.killpg(SIGKILL)` as a fallback (lines 246-251). GPUs are isolated via `CUDA_VISIBLE_DEVICES`. In my opinion, adding a child cgroup is gonna make things more complex and I do not see a clear benefit. Am I missing anything here?

MikeSpreitzer:
Also, regarding specifically Prometheus metrics, I created a Discussion for that; see #523

PR #532, opened 2026-06-02
[Draft] resource observability proposal

delavet:
Cluster operators currently have no way to answer basic resource-usage questions about an FMA deployment — GPU memory per instance, launcher pod states, controller health — without `kubectl exec` into individual pods.

This PR proposes first-class Prometheus observability for FMA. Tracking issue: #385

Looking for early community feedback on the overall direction, metric design, and especially the open question in §6 (whether the launcher should federate vLLM's own `vllm:*` metrics). Details are in the proposal doc.

### DIFF
--- a/docs/proposals/resource-observability.md
+++ b/docs/proposals/resource-observability.md
@@ -1,0 +1,415 @@
+# Proposal: Resource Usage Observability for FMA
+
+| Field | Value |
+|---|---|
+| **Status** | Draft |
+| **Tracking issue** | [#385](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/385) |
+| **Last updated** | 2026-05-31 |
+
+## 1. Summary
+
+This proposal adds first-class Prometheus observability to FMA. It introduces
+three families of metrics — one per component plane (`dual-pods-controller`,
+`launcher-populator`, `launcher`) — surfaced through standard Prometheus
+Operator `ServiceMonitor` / `PodMonitor` CRs and visualized through a single
+self-contained Grafana dashboard. 
+
+## 2. Motivation
+
+Issue #385 captures the same user story repeated from several angles: cluster
+operators and platform engineers cannot currently answer basic questions about
+an FMA deployment without `kubectl exec` into individual pods. Concretely,
+none of these queries have a working answer today:
+
+- *"How much GPU memory is each vLLM instance using right now?"*
+- *"Which launcher pod's instances are sleeping vs. running?"*
+- *"Is the dual-pods reconciler healthy — what's its error rate?"*
+
+Upstream exporters do not close this gap. `dcgm-exporter` reports GPU usage by
+PID and UUID, but knows nothing about FMA's `InferenceServerConfig` (ISC),
+`LauncherConfig` (LCfg), or per-instance identity. `kube-state-metrics`
+reports Pod/Deployment state but is opaque to FMA's `running` / `sleeping`
+distinction. The metrics this proposal introduces fill the FMA-specific gap
+*on top of* whatever upstream exporters the operator chooses to deploy.
+
+## 3. Goals and non-goals
+
+### Goals
+
+1. Expose Prometheus metrics from every FMA component so resource usage is
+   observable at four levels: per vLLM instance, per launcher pod, per node,
+   and cluster-wide.
+2. Ship Prometheus Operator `ServiceMonitor` / `PodMonitor` CRs so
+   kube-prometheus-stack environments work zero-touch.
+3. Ship a self-contained Grafana dashboard that is useful **without**
+   `dcgm-exporter` being installed.
+4. Keep the existing controller architecture untouched.
+5. Stay neutral on environments without Prometheus Operator: `helm install`
+   must succeed even when monitoring CRDs are absent.
+
+### Non-goals
+
+- Replacing `dcgm-exporter` where it is already deployed; the two coexist.
+- Replacing vLLM's own `vllm:*` metrics (a possible federation path is sketched in §6 as an open question).
+- Alerting / `PrometheusRule` CRs.
+- Cluster-wide capacity planning.
+
+## 4. Architecture overview
+
+Three metric sources, two scrape mechanisms, one dashboard delivery channel:
+
+```mermaid
+flowchart LR
+    subgraph Node["Worker node"]
+        direction TB
+        L1["launcher pod\n:9095 /metrics\nfma_launcher_*"]
+        L1V["vLLM child × N\n(localhost, not exposed)"]
+        L1 -. "in-process\npynvml + /proc" .-> L1V
+    end
+
+    subgraph CP["Cluster control plane"]
+        direction TB
+        DP["dual-pods-controller\n:8443 /metrics\nfma_isc_*, fma_provider_pod_*,\nfma_reconcile_*, fma_workqueue_depth"]
+        POP["launcher-populator\n:8080 /metrics\nfma_launcher_pod_*,\nfma_reconcile_*, fma_workqueue_depth"]
+    end
+
+    PROM["Prometheus\n(kube-prometheus-stack)"]
+    GRAF["Grafana\n(sidecar imports ConfigMap)"]
+
+    L1 -- "PodMonitor\n(node label via relabeling)" --> PROM
+    DP -- "ServiceMonitor" --> PROM
+    POP -- "ServiceMonitor" --> PROM
+    PROM --> GRAF
+```
+
+Key choices:
+
+- **Controllers stay hand-rolled.** Each spawns one extra goroutine running
+  `promhttp.HandlerFor(customRegistry)` on its own port. No
+  `controller-runtime` Manager is introduced.
+- **Launcher exposes `/metrics` on a dedicated port** so metrics scraping
+  cannot starve API throughput, and FastAPI API middleware does not affect
+  scraping.
+- **Each `/metrics` endpoint uses its own `prometheus.Registry`** so default
+  process collectors and accidental third-party metrics never leak in.
+- **`monitoring.enabled` is a single chart-level toggle, default off.** The
+  `Service` resources are always rendered so operators without Prometheus
+  Operator can still `kubectl port-forward` to verify `/metrics` manually.
+
+## 5. Design
+
+### 5.1 Three observability planes
+
+| Plane | Source | What it tells you |
+|---|---|---|
+| **Control plane** | `dual-pods-controller`, `launcher-populator` | Reconciler health, queue depth, inventory of ISCs / launcher pods |
+| **Per-instance plane** | `launcher` (`fma_launcher_instance_*`) | One vLLM instance's CPU/memory/GPU footprint and lifecycle state |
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
Regarding Prometheus metrics in particular, I created a Discussion as a place to discuss ideas before getting down to implementation. See #523

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+
+## 2. Motivation
+
+Issue #385 captures the same user story repeated from several angles: cluster
+operators and platform engineers cannot currently answer basic questions about
+an FMA deployment without `kubectl exec` into individual pods. Concretely,
+none of these queries have a working answer today:
+
+- *"How much GPU memory is each vLLM instance using right now?"*
+- *"Which launcher pod's instances are sleeping vs. running?"*
```
MikeSpreitzer, line 25: Without denying the issues around identifying and collecting, I will note that vllm's metrics include on about sleep mode. Following is an example from a recent scrape (the one posted in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/discussions/523#discussioncomment-17160980).

```
# HELP vllm:engine_sleep_state Engine sleep state; awake = 0 means engine is sleeping; awake = 1 means engine is awake; weights_offloaded = 1 means sleep level 1; discard_all = 1 means sleep level 2.

# TYPE vllm:engine_sleep_state gauge

vllm:engine_sleep_state{engine="0",model_name="HuggingFaceTB/SmolLM2-360M-Instruct",sleep_state="awake"} 1.0
vllm:engine_sleep_state{engine="0",model_name="HuggingFaceTB/SmolLM2-360M-Instruct",sleep_state="weights_offloaded"} 0.0
vllm:engine_sleep_state{engine="0",model_name="HuggingFaceTB/SmolLM2-360M-Instruct",sleep_state="discard_all"} 0.0
```

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+## 4. Architecture overview
+
+Three metric sources, two scrape mechanisms, one dashboard delivery channel:
+
+```mermaid
+flowchart LR
+    subgraph Node["Worker node"]
+        direction TB
+        L1["launcher pod\n:9095 /metrics\nfma_launcher_*"]
+        L1V["vLLM child × N\n(localhost, not exposed)"]
```
MikeSpreitzer, line 66: The child is not separate from the launcher Pod. Each child is a subprocess of the main process of the main container in the launcher Pod.

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+## 2. Motivation
+
+Issue #385 captures the same user story repeated from several angles: cluster
+operators and platform engineers cannot currently answer basic questions about
+an FMA deployment without `kubectl exec` into individual pods. Concretely,
+none of these queries have a working answer today:
+
+- *"How much GPU memory is each vLLM instance using right now?"*
+- *"Which launcher pod's instances are sleeping vs. running?"*
+- *"Is the dual-pods reconciler healthy — what's its error rate?"*
```
MikeSpreitzer, line 26: I am not sure what sort of errors are meant here.
User errors reported in the status section of Kubernetes API objects do not reflect on the health of a controller.
Concurrency conflicts in Kube apiserver calls are a normal thing in Kubernetes, and reflect concurrent activity rather than the health of a controller.

I think that perhaps judging health is a matter of interpreting multiple signals. For Prometheus metrics, the question is what are the helpful signals to include.

The worker queue metrics that are built into the workqueue library are one clue to include, and are low-hanging fruit. I _do_ recommend including the Prometheus metrics produced by the Kubernetes libraries.

Another might pervasive instrumentation of all calls to Kube apiservers, producing a histogram of latency with a label that carries the HTTP status code or non-HTTP error id.

Similarly, pervasive instrumentation of the HTTP calls that the controller makes to other components (requester, launcher).

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+
+All metric names use the `fma_` prefix. Histograms follow Prometheus
+conventions (`_seconds`, `_bytes` base units; `_bucket` / `_sum` / `_count`
+emitted automatically).
+
+**`dual-pods-controller` — port 8443**
+
+| Metric | Type | Labels | Purpose |
+|---|---|---|---|
+| `fma_isc_total` | gauge | `launcher_config`, `namespace` | Inventory of `InferenceServerConfig` objects bound to each `LauncherConfig`. Detects misconfigured ISCs that match no LCfg, and lets dashboards size the cluster's intended workload. |
```
MikeSpreitzer, line 123: The metric name suffix `_total` is appropriate for a Prometheus _counter_; the suffix `_count` would be an appropriate suffix for a Prometheus _gauge_.

Neither produces an _inventory_, unless there is a label carrying the ISC name.

I don't know about "Detects misconfigured ISCs". It seems to me that the most obvious way to implement the defined metric is based on looking at what is in the ISCs (each has a name of a LauncherConfig). The dual-pods controller does _not_ have a data structure per existing LauncherConfig that holds an inventory of referencing ISCs.

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+| **Per-instance plane** | `launcher` (`fma_launcher_instance_*`) | One vLLM instance's CPU/memory/GPU footprint and lifecycle state |
+| **GPU plane** | `launcher` (`fma_launcher_gpu_*`) | GPU-wide memory and utilization, usable even when `dcgm-exporter` is absent |
+
+The three planes share a common label vocabulary (`instance_id`, `model`,
+`gpu_uuid`, `node`, `launcher_config`) so dashboards can pivot between them
+without re-keying queries.
+
+### 5.2 Metric catalog
+
+All metric names use the `fma_` prefix. Histograms follow Prometheus
```
MikeSpreitzer, line 115: For the sake of compactness, this prefix could be omitted from the table (with a note saying as much).

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+| `fma_provider_pod_total` | gauge | `launcher_config`, `node`, `state=running\|sleeping` | Live count of provider pods by state and location. Drives the *controller-view* node panel that operators compare against the launcher's *measured* view — the divergence between the two is the primary signal observability is meant to surface. |
+| `fma_reconcile_total` | counter | `controller=dual-pods`, `result=success\|error` | Reconcile error rate is the primary signal of controller health. The `result` label enables direct error-rate dashboards and SLOs without post-processing. |
+| `fma_reconcile_duration_seconds` | histogram | `controller=dual-pods` | Reconcile latency distribution. Sustained P99 growth indicates pressure on the informer cache or downstream API server. |
+| `fma_workqueue_depth` | gauge | `controller=dual-pods`, `queue` | Backpressure signal — a growing queue means events arrive faster than the reconciler completes. Pairs with reconcile duration to distinguish "slow reconciles" from "queue saturated." |
```
MikeSpreitzer, line 127: These (by different names) are two of the metrics produced by the workqueue library.

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+
+### 5.1 Three observability planes
+
+| Plane | Source | What it tells you |
+|---|---|---|
+| **Control plane** | `dual-pods-controller`, `launcher-populator` | Reconciler health, queue depth, inventory of ISCs / launcher pods |
+| **Per-instance plane** | `launcher` (`fma_launcher_instance_*`) | One vLLM instance's CPU/memory/GPU footprint and lifecycle state |
+| **GPU plane** | `launcher` (`fma_launcher_gpu_*`) | GPU-wide memory and utilization, usable even when `dcgm-exporter` is absent |
+
+The three planes share a common label vocabulary (`instance_id`, `model`,
```
MikeSpreitzer, line 109: Note that finer-grained things that "model" are important here. WVA focuses on _variants_, explicitly for the purpose of allowing multiple variants of one model. In FMA the corresponding (more or less?) discriminator is an InferenceServerConfig object.

docs/proposals/resource-observability.md
```diff
@@ -0,0 +1,415 @@
...
+| `fma_launcher_pod_desired` | gauge | `launcher_config`, `node` | Desired count from `LauncherPopulationPolicy`. Compared against `fma_launcher_pod_total{phase="Running"}` to detect populator drift — the populator's whole purpose is keeping these two equal. |
+| `fma_reconcile_total` | counter | `controller=launcher-populator`, `result` | Same role as in dual-pods — reconciler health. |
+| `fma_reconcile_duration_seconds` | histogram | `controller=launcher-populator` | Same role as in dual-pods. |
+| `fma_workqueue_depth` | gauge | `controller=launcher-populator`, `queue` | Same role as in dual-pods. |
+
+**`launcher` — port 9095**
+
+| Metric | Type | Labels | Purpose |
+|---|---|---|---|
+| `fma_launcher_instance_total` | gauge | `state=running\|sleeping\|pending\|failed` | Cheap per-pod aggregate for the cluster instance-count panel; works without label join. |
```
MikeSpreitzer, line 143: The launcher does not know about sleep vs. awake.

docs/proposals/resource-observability.md
MikeSpreitzer, line 109: Note that the "instance ID" values used by the launcher are _only_ meant to distinguish between vllm instances in the same launcher. Instances in different launchers can and do use the same "instance ID".

MikeSpreitzer:
The PID stuff looks really intrusive. I wonder if there is another way to skin this cat. At least when using DCGM metrics, the (launcher) Pod and the GPU are identified. The dual-pods controller can emit information that relates them to what is actually meaningful to the model deployer, which is the server-requesting Pod.

MikeSpreitzer:
This reminds me that overall we would prefer the dual-pods controller to be able to work with quantities of GPU memory rather than counts of instances.

Perhaps rather than making the launcher much more intrusive than it is today there is a better alternative. For example, if vllm were a little more revealing about GPU memory usage --- which it is already logging! The dual-pods controller could do some log parsing. But even better would be if vllm had a method to query that, and the sleep and wake methods reported that in their HTTP response messages.

MikeSpreitzer:
This is still interesting

MikeSpreitzer:
This PR, Issue #385, and Discussion #523 are still interesting.

MikeSpreitzer:
This is still relevant

delavet:
Still active. Plan update:

The instrumentation layer has largely landed since I opened this: `fma_duality` (#569), `fma_actuation_seconds`, `fma_http_latency_seconds`, `fma_launcher_pod_count`, the `fma_dpc_innerqueue_*` series. The PID-based collection I originally proposed here is obsolete; I'll drop it.

What's missing is a consumer. So I'm going to pick up the Grafana dashboard thread from https://github.com/llm-d-incubation/llm-d-fast-model-actuation/discussions/523#discussioncomment-17805145 and build one on top of the existing metrics: GPU attribution (`fma_duality * on(UUID) DCGM_FI_DEV_FB_USED` vs. total per GPU), actuation latency by `path`, and launcher phase counts.

The GPU attribution panel may could also help with the current flake work.

PR #538, opened 2026-06-04
Trim launcher_count_before_reclaim in test-cases.sh

MikeSpreitzer:
The trim is needed so that the later string comparison with the trimmed count after reclaim will succeed.

I do not know how this test case ever passed.


waltforme:
Did you see a failure? And if yes, is the failure on Mac OS or on Linux?

> I do not know how this test case ever passed.

MikeSpreitzer:
The failure was on my Mac.

waltforme:
> The failure was on my Mac.

I think the added `tr` command is a no-op on Linux but load-bearing on macOS.

waltforme (approved):
LGTM

MikeSpreitzer:
@waltforme: that's a plausible hypothesis. Thanks.

PR #541, opened 2026-06-04
Missing authorization for launchers

diegocastanibm:
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/506

PR #543, opened 2026-06-04
Use version 34.1.0 for kube python client in launcher in requirements-sans-vllm.txt

waltforme:
As shown in the title.

This is a follow-up to #541 .


diegocastanibm (approved):
LGTM

Issue #544, opened 2026-06-04
[Feature]: test image tags should have build date in them


MikeSpreitzer:
### Feature Area

Container Image / Build

### Problem Statement

As someone responsible for limiting the accumulation of old test images, I need a way to know how old a given image is.

### Proposed Solution

Change the format of test image tags from `ref-${short_hash}` to `ref-${short_hash}-${day}`, where the day is formatted YYYY-MM-DD.

### Alternatives Considered

Represent the timestamp in seconds-since-epoch. This would make it easier for automation to do math, harder for humans to interpret. So far, the latter has been more important.

### Willingness to Contribute

Yes, I can submit a PR

MikeSpreitzer:
Revised day format to YYYYMMDD

MikeSpreitzer:
This has been resolved by #545 .

For example, in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/27000229755/job/79678823936 we see the following.

> Building images with tag: ref-d4f60e44-20260605

PR #545, opened 2026-06-04
Add build date into tags of test images

MikeSpreitzer:
To address #544

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -301,7 +301,7 @@ jobs:
           GIT_REF: ${{ needs.gate.outputs.pr_head_sha }}
         run: |
           # Use first 8 chars of the git ref (POSIX-compliant)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -301,7 +301,7 @@ jobs:
           GIT_REF: ${{ needs.gate.outputs.pr_head_sha }}
         run: |
           # Use first 8 chars of the git ref (POSIX-compliant)
-          IMAGE_TAG="ref-$(printf '%s' "$GIT_REF" | cut -c1-8)"
+          IMAGE_TAG="ref-$(printf '%s' "$GIT_REF" | cut -c1-8)-$(date -Idate)"
```
diegocastanibm: I suggest to use `date +%Y%m%d` to have this format `20260604`. In that way, we use the dash to separate the concepts `hash` and `date`, instead of `hash-year-month-day`.

diegocastanibm (approved):
LGTM

PR #563, opened 2026-06-15
Remove a stale TODO item

waltforme:
Removes a stale TODO in `ensureUnbound` about launcher-based provider port recovery. Current code persists the launcher server port on the bound provider Pod and recovers it via `recoverInstanceStateFromLauncherPod`.

In other words, the noted restart/unbind concern is already handled (by PR #492).


MikeSpreitzer (approved):
LGTM

PR #569, opened 2026-06-16
Add duality indicator metric

MikeSpreitzer:
As suggested in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/discussions/523#discussioncomment-17196953 . I went with the label names that match up for the join.

This is related to Discussion #523, Issue #385, and PR #532 .

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -170,6 +170,8 @@ var (
 	queueDurationHists *kubemetrics.HistogramVec
 	retriesCounters    *kubemetrics.CounterVec
 	workDurationHists  *kubemetrics.HistogramVec
+
+	dualityGauges *kubemetrics.GaugeVec
 )
 
 // NewController makes a new dual pods controller.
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
Example, using the last E2E-on-OpenShift run.

image

As demonstrated here, it is enough to join `on(UUID)`; aligning the other labels is not necessary in order to get a join result.

The right-hand restriction (`exported_namespace!=""`) is necessary to avoid a problem with DCGM_FI_DEV_FB_USED timeseries that do not identify a Kubernetes Pod that is currently using the GPU.

BTW, that `group_left` is necessary, to get a more informative result.

image

MikeSpreitzer:
The force-push to b88488b is a rebase onto `main`

waltforme (approved):
I saw `ContainerStatusUnknown` again in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/27768076651/job/82161217819. But I believe this sympton is due to recent flakiness of the OpenShift cluster we ran the tests in.

PR #571, opened 2026-06-17
Add more Prometheus metrics

MikeSpreitzer:
Add more Prometheus metrics

- fma_requester_count
- fma_isc_count
- fma_actuation_seconds
- fma_wake_seconds
- fma_launcher_create_seconds
- fma_instance_create_seconds

Also, start documenting the Prometheus metrics.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -28,9 +28,14 @@ Labels are as follows.
 ### fma_dpc_innerqueue_queue_duration_seconds
 
 Vector of classic histograms: Time from unique enqueue to the dequeue.
-Recall that a workqueue item may be enqueued multiple times before
-it is worked on. That first enqueue action is the one referenced here;
-the others are no-ops.
+Recall that a workqueue item may be enqueued multiple times before it
+is worked on. That first of _those_ enqueue actions is the one
+referenced here; the others are no-ops. Note well: the purpose of the
+qualifier here is to distinguish among the possibly-many times that an
+item is enqueued _before it is worked on_. For an item that is worked
+on multiple times, _each_ work is preceded by a "unique" enqueue of
+that item. This is **NOT** about the first time _ever_ that an item is
+enqueued.
 
 Labels are as follows.
 
@@ -113,7 +118,7 @@ become ready. Counted when the controller receives that reply.
 
 Labels are as follows.
 
-- `namespace`
+- `exported_namespace`: Kube namespace of requester and provider
 - `path`: one of "cold", "warm", or "hot"
 - `instancesDeleted`: decimal representation of number of vllm instances
   deleted for any reason in the course of this actuation.
@@ -125,19 +130,26 @@ Vector of histograms: Latency of `/wake_up` call from DPC to vllm.
 
 Labels are as follows.
 
-- `namespace`
+- `exported_namespace`: Kube namespace of requester and provider
 - `isc_name`: name of the relevant InferenceServerConfig object
-- `success`: "true" if no error, "false" otherwise
+- `success`: "true" if no error, "false" otherwise. Here "error" means
+  a failure to construct the HTTP request message, send it, or receive
+  the response message. The HTTP "status" code is not germane.
 
 ### fma_launcher_create_seconds
 
 Vector of histograms: Latency of kube API call to create launcher.
+This is only the time to get the Kube apiserver to create the Pod _API
+object_; the actual construction by of a running Pod, as well as the
+scheduling by the Kube Pod scheduler, are not included here.
 
 Labels are as follows.
 
-- `namespace`
+- `exported_namespace`: Kube namespace of requester and provider
 - `lcfg_name`: name of the relevant LauncherConfig object
-- `success`: "true" if no error, "false" otherwise
+- `success`: "true" if no error, "false" otherwise. Here "error" means
+  a failure to construct the HTTP request message, send it, or receive
+  the response message. The HTTP "status" code is not germane.
 
 ### fma_instance_create_seconds
 
@@ -145,6 +157,8 @@ Vector of histograms: Latency of DPC call to launcher to create vllm instance.
 
 Labels are as follows.
 
-- `namespace`
+- `exported_namespace`: Kube namespace of requester and provider
 - `isc_name`: name of the relevant InferenceServerConfig object
-- `success`: "true" if no error, "false" otherwise
+- `success`: "true" if no error, "false" otherwise. Here "error" means
+  a failure to construct the HTTP request message, send it, or receive
+  the response message. The HTTP "status" code is not germane.
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -70,6 +70,38 @@ Labels are as follows.
 
 - `launcher_config_name`: Name of the relevant LauncherConfig
 
+## FMA requester-provider binding
+
+### fma_duality
+
+Vector of gauges. Value is 1.0 while a server-requesting Pod is bound
+to a vllm instance in a launcher, set to 0.0 when those two get
+unbound.
+
+Labels are as follows.
+
+- `exported_namespace`: Kube API namespace involved
+- `requester_name`: name of the server-requesting Pod
+- `exported_pod`: name of the launcher Pod
+- `exported_container`: name of the container in the launcher Pod
+- `instance_id`: the launcher-local identifier of the vllm-instance
+- `UUID`: of the GPU. Multiple timeseries when multiple GPUs are involved.
+- `node`: name of the Node involved
+
+This metric can be used to effectively do "joins" in PromQL. PromQL
+does not really have joins, and this hack is nowhere near as flexible
+as an SQL join. Following is an example of how this metric can be used
+to associate a DCGM metric about GPUs to the server-requesting Pod.
+
+```
+fma_duality * on(UUID) group_left(exported_namspace,exported_pod) DCGM_FI_DEV_FB_USED{exported_namespace!=""}
+```
+
+The `{exported_namespace!=""}` qualifier filters out the time series
+that DCGM produces when a GPU is not bound to any Pod.
+
+The `group_left` gets more labels into the result.
+
 ## FMA actuation latencies
 
 ### fma_actuation_seconds
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,118 @@
+# Prometheus Metrics
+
+## From libraries
+
+- Go language runtime metrics (`go_...`)
+- Go process metrics (`process_...`)
+- Kubernetes REST client metrics (`rest_client_...`)
+- Kubernetes workqueue metrics (`workqueue_...`)
+
+## Dual-pods controller inner work queue
+
+### fma_dpc_innerqueue_adds_total
+
+Vector of counters: Number of unique adds to the queue.
+
+Labels are as follows.
+
+- `node`: name of Node
+
+### fma_dpc_innerqueue_depth
+
+Vector of gauges: Number of items in the queue.
+
+Labels are as follows.
+
+- `node`: name of Node
+
+### fma_dpc_innerqueue_queue_duration_seconds
+
+Vector of classic histograms: Time from unique enqueue to the dequeue.
+Recall that a workqueue item may be enqueued multiple times before
+it is worked on. That first enqueue action is the one referenced here;
+the others are no-ops.
+
+Labels are as follows.
+
+- `node`: name of Node
+
+### fma_dpc_innerqueue_retries_total
+
+Vector of counters: Total number of retries handled by queue.
+
+Labels are as follows.
+
+- `node`: name of Node
+
+### fma_dpc_innerqueue_work_duration_seconds
+
+Vector of histograms: Time spent syncing (working on one item).
+
+Labels are as follows.
+
+- `node`: name of Node
+
+## FMA workload characterization
+
+### fma_requester_count
+
+Vector of gauges: Number of server-requesting Pods.
+
+Labels are as follows.
+
+- `isc_name`: Name of the relevant InferenceServerConfig
+
+### fma_isc_count
+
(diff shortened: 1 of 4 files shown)


MikeSpreitzer:
The force-push to d490870 is a rebase onto `main`.

MikeSpreitzer:
Following is a graph from the workflow run https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/27768112889

image

MikeSpreitzer:
The force-push to 328304d is a rebase and merge conflict resolution.

docs/metrics.md
```diff
@@ -0,0 +1,118 @@
...
+Vector of gauges: Number of items in the queue.
+
+Labels are as follows.
+
+- `node`: name of Node
+
+### fma_dpc_innerqueue_queue_duration_seconds
+
+Vector of classic histograms: Time from unique enqueue to the dequeue.
+Recall that a workqueue item may be enqueued multiple times before
```
rubambiza: Is there a particular trend we can expect on the first enqueue vs subsequent ones? In other words, how valuable is the signal we get from the first enqueue? I'd conjecture that it's useful iff most of the work queue items are handled at the first enqueue, otherwise it becomes noisy as we get closer to 50/50.

docs/metrics.md
```diff
@@ -0,0 +1,118 @@
...
+
+### fma_wake_seconds
+
+Vector of histograms: Latency of `/wake_up` call from DPC to vllm.
+
+Labels are as follows.
+
+- `namespace`
+- `isc_name`: name of the relevant InferenceServerConfig object
+- `success`: "true" if no error, "false" otherwise
```
rubambiza: What constitutes an error here?

pkg/controller/dual-pods/controller.go
```diff
@@ -617,6 +692,11 @@ func (ctl *controller) OnUpdate(prev, obj any) {
 		}
 	case *fmav1alpha1.InferenceServerConfig:
 		prevTyped, ok := prev.(*fmav1alpha1.InferenceServerConfig)
+		lcfgName := typed.Spec.LauncherConfigName
+		if lcfgName != "" && ok && prevTyped.Spec.LauncherConfigName != lcfgName {
+			iscCountGauges.WithLabelValues(lcfgName).Dec()
+			iscCountGauges.WithLabelValues(prevTyped.Spec.LauncherConfigName).Inc()
```
waltforme: Should the previous count decrease (and current increase)?

pkg/controller/dual-pods/controller.go
```diff
@@ -272,9 +280,52 @@ func (config ControllerConfig) NewController(
...
+			Namespace:      "fma",
+			Name:           "isc_count",
+			Help:           "Number of InferenceServerConfig objects",
+			StabilityLevel: kubemetrics.ALPHA,
+		}, []string{"launcher_config_name"})
+		actuationSecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
+			Namespace:      "fma",
+			Name:           "actuation_seconds",
+			Help:           "Time from requester CreationTimestamp to completion of readiness relay",
+			Buckets:        []float64{0, 1, 3, 5, 7.5, 10, 15, 30, 60, 120, 240, 480, 960},
```
rubambiza: I would suggest going as high as 960 * 2 because I have seen them go as high as ~2000s when it comes to FMA + WVA integration (cc: @aavarghese to confirm). However, that was a case of a potential version regression/bug in WVA. Still, I think having the values go that high could show help us catch when transient cases of things going unexpectedly.

pkg/controller/dual-pods/controller.go
```diff
@@ -272,9 +280,52 @@ func (config ControllerConfig) NewController(
...
+			Name:           "wake_seconds",
+			Help:           "Latency of /wake_up call from DPC to vllm",
+			Buckets:        []float64{0.001, 0.01, 0.1, 1, 2, 3, 5, 7.5, 10, 15, 30},
+			StabilityLevel: kubemetrics.ALPHA,
+		}, []string{"namespace", "isc_name", "success"})
+		launcherCreateSecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
+			Namespace:      "fma",
+			Name:           "launcher_create_seconds",
+			Help:           "Latency of kube API call to create launcher",
+			Buckets:        []float64{0.001, 0.01, 0.1, 1, 2, 3, 4, 5},
```
rubambiza, line 314: Based on early benchmarking numbers, I think the buckets here should be similar to the ones in `instance_create_seconds`. We are able to shave about 5 seconds of time by having a pre-populated launcher. Only caveat is that we haven't replicated those numbers on models of different sizes to assess if the 5s is universal.

pkg/controller/dual-pods/controller.go
```diff
@@ -543,23 +607,34 @@ func (ctl *controller) OnAdd(obj any, isInInitialList bool) {
...
 				}
-			} else if it == infSvrItemRequester && nodeName == "" {
-				ctl.enqueueLogger.V(5).Info("Ignoring add of non-scheduled server-requesting Pod", "name", typed.Name)
-				return
+			case infSvrItemRequester:
+				iscName := typed.Annotations[api.InferenceServerConfigAnnotationName]
+				if iscName != "" {
+					requestersPerISCGauges.WithLabelValues(iscName).Inc()
+				}
+				if nodeName == "" {
```
rubambiza, line 624: *Clarifying question*: The separation of these two `if`s, when combined with the log statement below, imply that we anticipate the case of a requester referencing a particular server config but not scheduled on a node yet. If so, I'd argue that it becomes difficult to parse the value of such increments. Perhaps it makes sense to separate the two such that we get:

- here are the numbers of times requesters referencing a given `isc` were queued without being scheduled
- here are the numbers of times requesters with a given `isc` where queued while already having been scheduled.

What do you think?

pkg/controller/dual-pods/controller.go
```diff
@@ -617,6 +692,11 @@ func (ctl *controller) OnUpdate(prev, obj any) {
 		}
 	case *fmav1alpha1.InferenceServerConfig:
 		prevTyped, ok := prev.(*fmav1alpha1.InferenceServerConfig)
+		lcfgName := typed.Spec.LauncherConfigName
+		if lcfgName != "" && ok && prevTyped.Spec.LauncherConfigName != lcfgName {
+			iscCountGauges.WithLabelValues(lcfgName).Dec()
+			iscCountGauges.WithLabelValues(prevTyped.Spec.LauncherConfigName).Inc()
+		}
```
rubambiza, line 700: I don't follow what's going on here exactly. Perhaps a comment on the case would be helpful.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: Oops, they were switched. Fixed.

pkg/controller/dual-pods/controller.go
```diff
@@ -689,23 +769,34 @@ func (ctl *controller) OnDelete(obj any) {
...
-			} else if it == infSvrItemRequester && nodeName == "" {
-				ctl.enqueueLogger.V(5).Info("Ignoring delete of non-scheduled server-requesting Pod", "name", typed.Name)
-				return
+			case infSvrItemRequester:
+				iscName := typed.Annotations[api.InferenceServerConfigAnnotationName]
+				if iscName != "" {
+					requestersPerISCGauges.WithLabelValues(iscName).Dec()
+				}
+				if nodeName == "" {
+					ctl.enqueueLogger.V(5).Info("Ignoring delete of non-scheduled server-requesting Pod", "name", typed.Name)
```
rubambiza, line 787: Similar question as the above [comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/571/changes#r3438516230). I can imagine a scenario where a server-requesting pod gets deleted because it could not be scheduled for numerous reasons outside's the DPC's control. For the particular ISC whose counters are being decremented, wouldn't it be simpler to increment them only when the requester is actually scheduled (which obviates the need for decrements) *or* go with the idea of separate counters as suggested above so that the signal is actually useful?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -532,6 +538,21 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return err, true
 			}
 			serverDat.ReadinessRelayed = &ready
+			if ready && !serverDat.FirstReadyRelayed {
+				path := "hot"
+				if serverDat.NeededNewLauncher {
+					path = "cold"
+				} else if serverDat.NeededNewInstance {
```
rubambiza, line 545: **Nit**: Since the launcher population policy is trending toward always being on, I think the warm path becomes a more common case than cold. Thus, I'd suggest swapping `if/else` here.

rubambiza (changes requested):
Leaving independent comments with some suggested changes.

docs/metrics.md
MikeSpreitzer: I tried to word this carefully, including an explanation of what is meant by "first". This is considering the possibility of multiple enqueues per dequeue.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: "queuing" is not relevant here. A Pod is added only once, and deleted only once. For a count of Pod objects, it is valid to count at Pod create and delete.

We could consider adding another label to break down by whether the Pod is scheduled or not. Can we do that in a later PR?

pkg/controller/dual-pods/controller.go
MikeSpreitzer: Perhaps the fact that I got it backwards was confusing you. It has been fixed.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: This is not about frequency, it is about logic. If `serverDat.NeededNewLauncher` then it is a cold start, regardless of anything else in `serverDat`.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: This is only the time to get the Kube apiserver to create the Pod _API object_; the actual construction by of a running Pod, as well as the scheduling by the Kube Pod scheduler, are not included here.

https://github.com/kubernetes/community/blob/main/sig-scalability/slos/api_call_latency.md says that the SLO for this is 1 second at the 99th percentile over the last 5 minutes.

Following is what I found in the latest test on OpenShift.

```
# HELP fma_launcher_create_seconds [ALPHA] Latency of kube API call to create launcher
# TYPE fma_launcher_create_seconds histogram
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="0.001"} 0
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="0.01"} 0
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="0.1"} 1
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="1"} 1
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="2"} 1
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="3"} 1
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="4"} 1
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="5"} 1
fma_launcher_create_seconds_bucket{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true",le="+Inf"} 1
fma_launcher_create_seconds_sum{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true"} 0.029627333
fma_launcher_create_seconds_count{lcfg_name="18-20-24-35",namespace="fma-e2e-pr-571",success="true"} 1
```

pkg/controller/dual-pods/controller.go
MikeSpreitzer: Same answer.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: I added some explanation in the doc.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: done

docs/metrics.md
MikeSpreitzer: I added some explanation.

docs/metrics.md
MikeSpreitzer: I revised and extended that remark.

pkg/controller/dual-pods/controller.go
rubambiza: Ah, I see. Thanks for the clarification.

pkg/controller/dual-pods/controller.go
rubambiza: Yes, works for me!

pkg/controller/dual-pods/controller.go
rubambiza: Thank you. I had to get some more context on the method definition, too. I now understand it as an update to the launcherconfig ref/name of an inference server, decrementing the counter for the old launcherconfig and incrementing the counter for the new one. Let me know if there is any misunderstanding, otherwise lgtm.

rubambiza (approved):
Previous comments have been addressed, clarified, or left for a future PR. LGTM.

PR #552, opened 2026-06-08
FMA+WVA integration demo for OCP clusters

dumb0002:
## Summary
This PR adds end-to-end demo scripts for deploying and testing **FMA (Fast Model Actuation)** integrated with **WVA (Workload Variant Autoscaler)** on OpenShift clusters with GPU nodes.

This PRs follow the same pattern as the scripts for FMA + HPA integration [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/tree/main/test/e2e/demo-fma-hpa). 
#### 1. **`demo-fma-wva-ocp.sh`** - Deployment Script
Automated deployment script that installs FMA, WVA, and llm-d components in a single namespace.

**Features:**
- ✅ Deploys FMA CRDs and controllers
- ✅ Deploys WVA controller for workload-aware autoscaling
- ✅ Deploys llm-d EPP (Endpoint Picker) with Gateway API
- ✅ Creates FMA objects (InferenceServerConfig, LauncherConfig, LauncherPopulationPolicy, Deployment)
- ✅ Creates WVA objects (VariantAutoscaling, HPA)
- ✅ Configures PodMonitor for metrics collection
- ✅ Idempotent - safe to run multiple times

#### 2. **`cleanup-fma-wva.sh`** - Cleanup Script
Safe cleanup script with two modes: basic (preserves infrastructure) and full (removes everything).

**Features:**
- ✅ Removes FMA/WVA objects in correct order
- ✅ Strips dual-pods finalizers to prevent hanging pods
- ✅ Supports partial cleanup (preserves namespace/controllers)
- ✅ Supports full cleanup (removes namespace and all components)
- ✅ Handles stuck namespaces automatically

PR #579, opened 2026-06-19
FMA+WVA integration demo for OCP clusters - Continuation of PR # 552

dumb0002:
## Summary
This PR is a continuation of PR https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/552 - it starts with the latest revision addressing all the reviewers comments in PR # 552.

This PR adds end-to-end demo scripts for deploying and testing **FMA (Fast Model Actuation)** integrated with **WVA (Workload Variant Autoscaler)** on OpenShift clusters with GPU nodes.

This PRs follow the same pattern as the scripts for FMA + HPA integration [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/tree/main/test/e2e/demo-fma-hpa). 
#### 1. **`demo-fma-wva-ocp.sh`** - Deployment Script
Automated deployment script that installs FMA, WVA, and llm-d components in a single namespace.

**Features:**
- ✅ Deploys FMA CRDs and controllers
- ✅ Deploys WVA controller for workload-aware autoscaling
- ✅ Deploys llm-d EPP (Endpoint Picker) with Gateway API
- ✅ Creates FMA objects (InferenceServerConfig, LauncherConfig, LauncherPopulationPolicy, Deployment)
- ✅ Creates WVA objects (VariantAutoscaling, HPA)
- ✅ Configures PodMonitor for metrics collection
- ✅ Idempotent - safe to run multiple times

#### 2. **`cleanup-fma-wva.sh`** - Cleanup Script
Safe cleanup script with two modes: basic (preserves infrastructure) and full (removes everything).

**Features:**
- ✅ Removes FMA/WVA objects in correct order
- ✅ Strips dual-pods finalizers to prevent hanging pods
- ✅ Supports partial cleanup (preserves namespace/controllers)
- ✅ Supports full cleanup (removes namespace and all components)
- ✅ Handles stuck namespaces automatically

### DIFF
--- a/test/e2e/demo-fma-wva/README.md
+++ b/test/e2e/demo-fma-wva/README.md
@@ -1,0 +1,200 @@
+# FMA + WVA + KEDA + LLM-D-Router Demo on OpenShift
+
+Two scripts to **deploy and tear down a working end-to-end demo of
+FMA + WVA + KEDA + llm-d-router in a single OpenShift namespace** —
+autoscaling an llm-d workload driven by WVA through KEDA. KEDA must already be installed
+on the cluster (see [Prerequisites](#prerequisites)) — the scripts don't install
+it.
+
+The scripts install the platform components (controllers, CRDs, RBAC,
+llm-d-router) **and** create a concrete workload that exercises them (an
+`InferenceServerConfig`, `LauncherConfig`, `LauncherPopulationPolicy`, requester
+`Deployment`, and a KEDA `ScaledObject` annotated with `llm-d.ai/managed: "true"`
+so WVA discovers it and publishes the `wva_desired_replicas` metric to
+Prometheus). KEDA's Prometheus trigger reads that metric and drives the HPA it
+creates and manages (`wva-keda-hpa-fma-requester`) to scale the deployment.
+
+| Script | Purpose |
+|---|---|
+| `setup-fma-wva.sh` | Deploy FMA (CRDs, ValidatingAdmissionPolicies, RBAC, controllers, sample CRs), WVA, llm-d-router, and the demo requester workload |
+| `cleanup-fma-wva.sh` | Tear it all back down |
+
+The workload-variant-autoscaler (WVA) repo is cloned automatically — no need
+to pre-clone or pass `--wva-repo-path`.
+
+Both scripts use a standard CLI flag interface. Run either with `--help`
+for the full list of options.
+
+## Versioning
+
+> **⚠️ The FMA version deployed is _mixed_.** Images come from the published
+> release passed via `--fma-version` (required — no default); CRDs, VAPs, and
+> the Helm chart come from your **local working tree**. Run this script from a
+> checkout of the **same tag** you pass to `--fma-version`. (No single-release
+> installer exists yet.)
+
+FMA, WVA, GAIE and llm-d release independently, so incompatibilities are
+possible. The defaults pin a known-good combination. If you change one,
+test the others alongside it.
+
+| Component | Flag | Default |
+|---|---|---|
+| All FMA images (dual-pods + launcher-populator controllers, launcher, requester) | `--fma-version` | _(required)_ |
+| WVA (published release tag) | `--wva-version` | `v0.8.0` |
+| GAIE (Gateway API Inference Extension) | `--gaie-version` | `v1.5.0` |
+| llm-d release | `--llm-d-release` | `v0.7.0` |
+
+See the "Pin all components" example below for a fully-pinned invocation.
+
+## Prerequisites
+
+- `oc` authenticated to an OpenShift cluster with NVIDIA GPU nodes
+- **KEDA** (the `keda.sh/v1alpha1` `ScaledObject` CRD + controller) installed
+  and configured on the cluster, with Prometheus access wired up — the deploy
+  script creates a `ScaledObject` but does **not** install or configure KEDA.
+- For **deploy** (`setup-fma-wva.sh`): `helm`, `kubectl`, `jq`, `make`, `git`, `yq` ([mikefarah/yq](https://github.com/mikefarah/yq)) on `$PATH`
+- For **cleanup** (`cleanup-fma-wva.sh`): `helm`, `kubectl`, `jq` on `$PATH` (plus `git` if `--level all` is used)
+
+## Deploy
+
+`--fma-version` is required — pass a published
+[FMA release](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/releases)
+(no default; see [Versioning](#versioning)). Examples use `v0.6.2`.
+
+Deploy (default namespace `fma-wva-demo`):
+
+```shell
+./test/e2e/demo-fma-wva/setup-fma-wva.sh --fma-version v0.6.2
+```
+
+With a custom namespace:
+
+```shell
+./test/e2e/demo-fma-wva/setup-fma-wva.sh --fma-version v0.6.2 --namespace my-fma-demo
+```
+
+Re-running is safe (idempotent — existing components are skipped). On first run
+the WVA repo is cloned to `.wva-checkout/`; pass `--wva-repo-path PATH` to reuse
+an existing checkout.
+
+## Tear down
+
+`cleanup-fma-wva.sh` supports three nested levels via `--level` (default
+`namespace`):
+
+| `--level` | Removes | Leaves | Use case |
+|---|---|---|---|
+| `workload` | The demo workload (ScaledObject + KEDA HPA, requester Deployment, FMA CRs, PodMonitor) | The platform (FMA/WVA/EPP controllers) | Reuse the platform with a different workload |
+| `namespace` *(default)* | `workload` + all **namespaced** platform (FMA controllers, WVA controller, llm-d-router) | Cluster-scoped objects and the namespace | Clear the namespace without disturbing other cluster users |
+| `all` | `namespace` + cluster-scoped RBAC, the `fma-poc=true` node label, and the namespace itself | CRDs only | Full cluster teardown |
+
+```shell
+# Default: clear FMA/WVA/EPP from the namespace, leave cluster objects alone
+./test/e2e/demo-fma-wva/cleanup-fma-wva.sh --namespace my-fma-demo
+
+# Just the workload, keep the platform running
+./test/e2e/demo-fma-wva/cleanup-fma-wva.sh --namespace my-fma-demo --level workload
+
+# Everything (auto-clones the WVA repo to undeploy WVA + llm-d-router)
+./test/e2e/demo-fma-wva/cleanup-fma-wva.sh --namespace my-fma-demo --level all
+```
+
+Even `--level all` is **not** an exhaustive wipe: CRDs (Gateway API, GAIE, FMA,
+WVA) are never removed — they may be shared across namespaces. Delete them by
+hand if you want a complete wipe.
+
+## Common flags
+
+| Flag | Default | Used by |
+|---|---|---|
+| `-n`, `--namespace NAME` | `fma-wva-demo` | both |
+| `--level workload\|namespace\|all` | `namespace` | cleanup |
+| `--wva-repo-path PATH` | `<repo-root>/.wva-checkout` | both |
+| `--wva-repo-url URL` | `https://github.com/llm-d/llm-d-workload-variant-autoscaler` | both |
+| `--wva-version TAG` (published release tag) | `v0.8.0` | both |
+| `--fma-image-registry URL` | `ghcr.io/llm-d-incubation/llm-d-fast-model-actuation` | deploy |
+| `--fma-version VER` | _(required)_ | deploy |
+| `--model NAME` | `Qwen/Qwen3-8B` | deploy |
+| `--gpu-node NODE` | first node with `nvidia.com/gpu.present=true` | deploy |
+| `--hf-token TOKEN` | (unset) | deploy (required for gated models) |
+
+Run `./setup-fma-wva.sh --help` or `./cleanup-fma-wva.sh --help` for the
+complete list. Equivalent environment variables (uppercase, underscored —
+e.g., `NAMESPACE`, `WVA_VERSION`, `WVA_REPO_PATH`) are also accepted, but flags
+take precedence. `--fma-version` is the exception: its env var is `FMA_VERSION`.
+
+## Examples
+
+Pin all components to specific versions. `--fma-version` and `--wva-version`
+must be **published release tags** (each is used as an image tag):
+
+```shell
+./test/e2e/demo-fma-wva/setup-fma-wva.sh \
+  --fma-version v0.6.2 \
+  --wva-version v0.8.0 \
+  --gaie-version v1.5.0 \
+  --llm-d-release v0.7.0
+```
+
+Deploy a gated model (`--hf-token` **and** approved access on the model's HF
+page are both required):
+
+```shell
+./test/e2e/demo-fma-wva/setup-fma-wva.sh \
+  --fma-version v0.6.2 \
+  --model meta-llama/Llama-3.1-8B-Instruct \
+  --hf-token hf_xxx
+```
+
+Reuse an existing WVA checkout instead of the auto-clone:
+
+```shell
+./test/e2e/demo-fma-wva/setup-fma-wva.sh \
+  --fma-version v0.6.2 \
+  --wva-repo-path /path/to/my/wva-checkout
+```
+
+## Troubleshooting
+
+### ScaledObject not scaling / KEDA-managed HPA shows `<unknown>`
+`kubectl get scaledobject fma-requester-scaler -n <ns>` shows the object's
+`READY`/`ACTIVE` status; `kubectl get hpa wva-keda-hpa-fma-requester -n <ns>`
+(the HPA KEDA creates) shows the metric as `<unknown>/1` in the `TARGETS`
+column when KEDA's Prometheus trigger has no value yet. The first 1–2 poll
+cycles after deploy usually look like this and resolve on their own (~30–60s).
+
+If it persists longer, walk the chain from WVA out:
+
+1. **WVA discovered the ScaledObject?** `kubectl logs -n <ns> -l app.kubernetes.io/name=workload-variant-autoscaler`
+   should mention the ScaledObject name; the controller logs each object it
+   picks up via the `llm-d.ai/managed` annotation.
+2. **WVA is publishing the metric?** Query Prometheus directly:
+   `wva_desired_replicas{variant_name="fma-requester", namespace="<ns>"}`.
+   If empty, WVA isn't emitting for this variant.
+3. **KEDA's trigger can read it?** Unlike a plain HPA, KEDA's Prometheus
+   trigger queries Prometheus **directly** (no prometheus-adapter / external
+   metrics API in the path). Check the KEDA operator logs and the ScaledObject
+   status for scaler errors:
+   ```shell
+   kubectl describe scaledobject fma-requester-scaler -n <ns>
+   kubectl logs -n <keda-ns> -l app=keda-operator | grep -i fma-requester
+   ```
+   A `serverAddress`/TLS/auth error here means KEDA can't reach Prometheus —
+   fix the trigger's Prometheus access (out of scope for this demo, which
+   assumes KEDA is already configured).
+4. **Trigger query matches the series?** Compare the ScaledObject trigger's
+   `query` (`spec.triggers[0].metadata.query`) against the labels on the actual
+   `wva_desired_replicas` series in Prometheus — `variant_name` in particular
+   must equal the deployment name (`fma-requester` in this demo).
+
+### Launcher pod missing the `llm-d.ai/variant` label
+The `llm-d.ai/variant` label is applied from
+`InferenceServerConfig.spec.modelServerConfig.labels` when a requester binds
+to a launcher. Unbound (idle) launchers won't carry it. Check the ISC, and
+verify the launcher is bound (has the `dual-pods.llm-d.ai/dual` label set).
+Don't add the label to `LauncherConfig.spec.podTemplate.metadata.labels` —
+it will collide with ISC-applied labels during binding.
+
+### Unknown flag error
+The scripts reject unknown flags. Check spelling and run with `--help` for
+the canonical flag names.
--- a/test/e2e/demo-fma-wva/setup-fma-wva.sh
+++ b/test/e2e/demo-fma-wva/setup-fma-wva.sh
@@ -1,0 +1,660 @@
+#!/usr/bin/env bash
+
+# Deploys FMA + WVA + llm-d components in the same namespace.
+#
+# Idempotent: checks each component before deploying, skips if already present.
+# Can be invoked from any working directory — the script resolves the FMA repo
+# root from its own location.
+#
+# Prerequisites:
+#   - This repo (llm-d-incubation/llm-d-fast-model-actuation) cloned locally
+#   - oc authenticated to an OpenShift cluster with GPU nodes
+#   - helm, kubectl, make, git, jq, yq (mikefarah/yq) on $PATH
+#   - Container images already pushed to registry (see --fma-image-registry)
+#
+# The workload-variant-autoscaler (WVA) repo is auto-cloned to --wva-repo-path
+# if not already present. To use an existing checkout, pass --wva-repo-path
+# /path/to/checkout.
+#
+# Run with --help for the full list of flags.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+
+# ----------------------------------------------------------------------------
+# CLI parsing — flags are the primary interface; matching env vars are honored
+# as a fallback (useful for CI/scripted runs) but flags take precedence.
+# ----------------------------------------------------------------------------
+usage() {
+    cat <<EOF
+Usage: $(basename "$0") [OPTIONS]
+
+Deploys FMA + WVA + llm-d components in the same namespace.
+
+FMA / cluster options:
+  -n, --namespace NAME              Target namespace (default: fma-wva-demo)
+      --fma-image-registry URL      FMA image registry
+                                    (default: ghcr.io/llm-d-incubation/llm-d-fast-model-actuation)
+      --fma-version VER             FMA release version — the image tag applied to
+                                    all FMA images (dual-pods + launcher-populator
+                                    controllers, launcher, requester). REQUIRED;
+                                    must be an already-published release (see
+                                    https://github.com/llm-d-incubation/llm-d-fast-model-actuation/releases).
+      --model NAME                  vLLM model (default: Qwen/Qwen3-8B)
+      --gpu-node NODE               Node for Launcher Population Policy (LPP) (default: first node with
+                                    nvidia.com/gpu.present=true)
+      --hf-token TOKEN              HuggingFace token for gated models (e.g.
+                                    meta-llama/*). Also requires approved access
+                                    on the model's HF page — a token isn't enough.
+      --runtime-class NAME          RuntimeClass for GPU pods (default: nvidia;
+                                    pass an empty string to omit)
+      --wait-requester-timeout DUR  How long to wait for Deployment/fma-requester
+                                    to become Available (default: 600s; covers
+                                    cold-cluster image+model fetch). Accepts
+                                    any kubectl --timeout value (e.g., 90s, 5m).
+
+WVA options:
+      --wva-version TAG             WVA version — used as both the git ref checked
+                                    out and the controller image tag. Must be a
+                                    published release tag (e.g. v0.8.0); branch
+                                    names and commit SHAs won't work (no matching
+                                    image is published). (default: v0.8.0)
+      --wva-repo-path PATH          Path to WVA repo (default: <repo-root>/.wva-checkout)
+      --wva-repo-url URL            WVA git URL
+                                    (default: https://github.com/llm-d/llm-d-workload-variant-autoscaler)
+      --wva-image-repo URL          WVA image repository
+                                    (default: ghcr.io/llm-d/llm-d-workload-variant-autoscaler)
+      --monitoring-namespace NAME   Monitoring namespace
+                                    (default: openshift-user-workload-monitoring)
+      --llm-d-release VER           llm-d release version (default: v0.7.0)
+      --gaie-version VER            GAIE version (default: v1.5.0)
+
+  -h, --help                        Show this help and exit
+
+Every flag above also has a matching environment variable (the flag's
+long-form name in upper snake case — e.g. --namespace → NAMESPACE,
+--wva-repo-path → WVA_REPO_PATH); flags take precedence when both are set.
+The exception is --fma-version, whose env var is FMA_VERSION.
+EOF
+}
+
+# Seed defaults from env vars (so existing callers using env vars still work).
+NAMESPACE="${NAMESPACE:-fma-wva-demo}"
+CONTAINER_IMG_REG="${CONTAINER_IMG_REG:-ghcr.io/llm-d-incubation/llm-d-fast-model-actuation}"
+# --fma-version is required (no default — see the required check below). Env var
+# is FMA_VERSION; IMAGE_TAG is also honored (the name deploy_fma.sh / CI use).
+IMAGE_TAG="${FMA_VERSION:-${IMAGE_TAG:-}}"
+MODEL="${MODEL:-Qwen/Qwen3-8B}"
+GPU_NODE="${GPU_NODE:-}"
+HF_TOKEN="${HF_TOKEN:-}"
+RUNTIME_CLASS_NAME="${RUNTIME_CLASS_NAME-nvidia}"
+WAIT_REQUESTER_TIMEOUT="${WAIT_REQUESTER_TIMEOUT:-600s}"
+
+WVA_REPO_PATH="${WVA_REPO_PATH:-$REPO_ROOT/.wva-checkout}"
+WVA_REPO_URL="${WVA_REPO_URL:-https://github.com/llm-d/llm-d-workload-variant-autoscaler}"
+# Single WVA version knob — used as both the git ref (for the source-based
+# install scripts that the WVA repo provides) and the controller image tag.
+WVA_VERSION="${WVA_VERSION:-v0.8.0}"
+WVA_IMAGE_REPO="${WVA_IMAGE_REPO:-ghcr.io/llm-d/llm-d-workload-variant-autoscaler}"
+MONITORING_NAMESPACE="${MONITORING_NAMESPACE:-openshift-user-workload-monitoring}"
+LLM_D_RELEASE="${LLM_D_RELEASE:-v0.7.0}"
+GAIE_VERSION="${GAIE_VERSION:-v1.5.0}"
+
+# Prometheus + adapter are NOT deployed by this script; OpenShift's built-in
+# user-workload monitoring is assumed. Hard-coded; not user-configurable.
+DEPLOY_PROMETHEUS=false
+DEPLOY_PROMETHEUS_ADAPTER=false
+
+# Helper for required-arg flags
+need_arg() {
+    [[ $# -ge 2 ]] || { echo "ERROR: $1 requires an argument" >&2; exit 2; }
+}
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -n|--namespace)
+            need_arg "$@"; NAMESPACE="$2"; shift 2 ;;
+        --fma-image-registry)
+            need_arg "$@"; CONTAINER_IMG_REG="$2"; shift 2 ;;
+        --fma-version)
+            need_arg "$@"; IMAGE_TAG="$2"; shift 2 ;;
+        --model)
+            need_arg "$@"; MODEL="$2"; shift 2 ;;
+        --gpu-node)
+            need_arg "$@"; GPU_NODE="$2"; shift 2 ;;
+        --hf-token)
+            need_arg "$@"; HF_TOKEN="$2"; shift 2 ;;
+        --runtime-class)
+            need_arg "$@"; RUNTIME_CLASS_NAME="$2"; shift 2 ;;
+        --wait-requester-timeout)
+            need_arg "$@"; WAIT_REQUESTER_TIMEOUT="$2"; shift 2 ;;
+        --wva-version)
+            need_arg "$@"; WVA_VERSION="$2"; shift 2 ;;
+        --wva-repo-path)
+            need_arg "$@"; WVA_REPO_PATH="$2"; shift 2 ;;
+        --wva-repo-url)
+            need_arg "$@"; WVA_REPO_URL="$2"; shift 2 ;;
+        --wva-image-repo)
+            need_arg "$@"; WVA_IMAGE_REPO="$2"; shift 2 ;;
+        --monitoring-namespace)
+            need_arg "$@"; MONITORING_NAMESPACE="$2"; shift 2 ;;
+        --llm-d-release)
+            need_arg "$@"; LLM_D_RELEASE="$2"; shift 2 ;;
+        --gaie-version)
+            need_arg "$@"; GAIE_VERSION="$2"; shift 2 ;;
+        -h|--help)
+            usage; exit 0 ;;
+        *)
+            echo "ERROR: Unknown option: $1" >&2
+            usage >&2
+            exit 2 ;;
+    esac
+done
+
+# --fma-version is required (no default — see the seed comment above).
+if [ -z "$IMAGE_TAG" ]; then
+    echo "ERROR: --fma-version is required (or set FMA_VERSION)." >&2
+    echo "  Pass an already-published FMA release, e.g. --fma-version v0.6.2" >&2
+    echo "  Releases: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/releases" >&2
+    exit 2
+fi
+
+# Derive the launcher/requester image refs from the registry and version.
+LAUNCHER_IMAGE="${CONTAINER_IMG_REG}/launcher:${IMAGE_TAG}"
+REQUESTER_IMAGE="${CONTAINER_IMG_REG}/requester:${IMAGE_TAG}"
+
+# --wva-version is both the git ref and the controller image tag, so it must be
+# a published release tag (branch names / commit SHAs have no matching image).
+WVA_REPO_REF="$WVA_VERSION"
+WVA_IMAGE_TAG="$WVA_VERSION"
+
+if [ ! -d "$WVA_REPO_PATH/.git" ]; then
+    if [ -d "$WVA_REPO_PATH" ] && [ -n "$(ls -A "$WVA_REPO_PATH" 2>/dev/null)" ]; then
+        echo "ERROR: $WVA_REPO_PATH exists but is not a git checkout. Remove it or set WVA_REPO_PATH to a different location." >&2
+        exit 1
+    fi
+    echo "  WVA repo not found at $WVA_REPO_PATH — cloning $WVA_REPO_URL ($WVA_REPO_REF)..."
+    mkdir -p "$(dirname "$WVA_REPO_PATH")"
+    # Shallow clone the release tag.
+    if ! git clone --depth 1 --branch "$WVA_REPO_REF" "$WVA_REPO_URL" "$WVA_REPO_PATH"; then
+        echo "ERROR: failed to clone $WVA_REPO_URL at tag '$WVA_REPO_REF'." >&2
+        echo "  --wva-version must be a published WVA release tag (e.g. v0.8.0)." >&2
+        exit 1
+    fi
+else
+    echo "  Using existing WVA repo at $WVA_REPO_PATH"
+fi
+
+step_num=0
+total_steps=7
+
+step() {
+    step_num=$((step_num + 1))
+    echo ""
+    echo "========================================"
+    echo "  Step ${step_num}/${total_steps}: $*"
+    echo "========================================"
+    echo ""
+}
+
+
+step "Namespace, ServiceAccounts, RBAC"
+
+# Ensure the namespace idempotently (create is not idempotent; apply is).
+kubectl create namespace "$NAMESPACE" --dry-run=client -o yaml \
+    | kubectl apply -f - >/dev/null
+echo "  Ensured namespace $NAMESPACE"
+kubectl label namespace "$NAMESPACE" openshift.io/user-monitoring=true --overwrite >/dev/null
+echo "  Ensured monitoring label openshift.io/user-monitoring=true on $NAMESPACE"
+
+# Ensure the ServiceAccount idempotently.
+kubectl create sa testlauncher -n "$NAMESPACE" --dry-run=client -o yaml \
+    | kubectl apply -f - >/dev/null
+echo "  Ensured ServiceAccount testlauncher"
+
+# Apply the Role idempotently.
+kubectl apply -n "$NAMESPACE" -f - <<'EOF' >/dev/null
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: testlauncher
+rules:
+- apiGroups: [""]
+  resources: [pods]
+  verbs: [get, patch]
+EOF
+echo "  Ensured Role testlauncher"
+
+# Apply the RoleBinding idempotently.
+kubectl apply -n "$NAMESPACE" -f - <<EOF >/dev/null
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: testlauncher
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: testlauncher
+subjects:
+- kind: ServiceAccount
+  name: testlauncher
+  namespace: ${NAMESPACE}
+EOF
+echo "  Ensured RoleBinding testlauncher"
+
+
+step "FMA CRDs and Controllers (via deploy_fma.sh)"
+
+# Helm release (chart instance) name.
+FMA_RELEASE="fma"
+if kubectl get deployment "${FMA_RELEASE}-dual-pods-controller" -n "$NAMESPACE" &>/dev/null; then
+    echo "  FMA controllers already deployed"
+else
+    echo "  Deploying FMA CRDs and controllers via deploy_fma.sh..."
+    (
+        cd "$REPO_ROOT"
+        FMA_NAMESPACE="$NAMESPACE" \
+        FMA_CHART_INSTANCE_NAME="$FMA_RELEASE" \
+        CONTAINER_IMG_REG="$CONTAINER_IMG_REG" \
+        IMAGE_TAG="$IMAGE_TAG" \
+        NODE_VIEW_CLUSTER_ROLE=create/please \
+        RUNTIME_CLASS_NAME="$RUNTIME_CLASS_NAME" \
+        "$SCRIPT_DIR/../deploy_fma.sh"
+    )
+fi
+
+echo "  Verifying FMA CRDs..."
+kubectl get crd | grep -F "fma.llm-d.ai" || echo "  WARNING: FMA CRDs not found"
+
+echo "  Verifying FMA controllers..."
+# `kubectl get -l ...` exits 0 with empty output when nothing matches, so we
+# check for non-empty `-o name` output to detect missing resources reliably.
+if kubectl get deployment -n "$NAMESPACE" -l app.kubernetes.io/part-of=fma -o name 2>/dev/null | grep -q .; then
+    kubectl get deployment -n "$NAMESPACE" -l app.kubernetes.io/part-of=fma
+else
+    echo "  WARNING: FMA controllers not found"
+fi
+
+
+step "WVA Deployment (via make deploy-wva-on-openshift)"
+
+if [ -n "$(kubectl get deployment -n "$NAMESPACE" -l app.kubernetes.io/name=workload-variant-autoscaler -o name 2>/dev/null)" ]; then
+    echo "  WVA controller already deployed"
+else
+    echo "  Deploying WVA to namespace $NAMESPACE..."
+    (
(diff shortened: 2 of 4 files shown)


dumb0002:
@MikeSpreitzer, this is a the refreshed PR for https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/552 - with the latest revision addressing all the existing comments in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/552. It's ready for another round of review. Thanks!

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+Both scripts use a standard CLI flag interface. Run either with `--help`
+for the full list of options.
+
+## Versioning
+
+> **This script consumes _published_ FMA releases only.** Even though it
+> lives inside the FMA repo, it pulls FMA images by tag (e.g.
+> `v0.6.0-alpha.13`) — it does **not** build from the local source tree.
+> Don't change `--fma-image-tag` to a tag that hasn't been pushed yet;
```
MikeSpreitzer: You want to say to not change _the default for_ that flag, right?

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+> doing so couples the FMA repo to itself in a circular way and breaks
+> reproducibility.
+
+FMA, WVA, and GIE/llm-d release independently, so incompatibilities are
+possible. The defaults pin a known-good combination. If you change one,
+test the others alongside it.
+
+| Component | Flag | Default |
+|---|---|---|
+| FMA controllers + launcher/requester images | `--fma-image-tag` | `v0.6.0-alpha.13` |
```
MikeSpreitzer: The latest alpha is 14 and it is substantially better than 13.

MikeSpreitzer: > FMA controllers + launcher/requester images

What about the other parts of FMA that get deployed?

Maybe this should not be defined as an _image_ tag. Why not `--fma-version` or `--fma-release` ?

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+./test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
+```
+
+Pick your own namespace:
+
+```shell
+./test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh --namespace my-fma-demo
+```
+
+The script is idempotent — re-running it skips components that already exist.
```
MikeSpreitzer: What if the script failed or was interrupted in the midst of a previous execution? This is something that really happens and should be properly supported.

What if something else created objects that have the same name but different contents? That is a possibility, and this document should say what happens in that case.

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+On first run it clones the WVA repo to `.wva-checkout/` at the repo root and
+reuses it on subsequent runs. Pass `--wva-repo-path PATH` to use a different
+location (e.g., a shared checkout outside this repo).
+
+`--namespace` and `--wva-repo-path` are just two of many flags. See
+[Common flags](#common-flags) for the most-used ones, or run
+`./test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh --help` for the complete list.
+
+## Tear down
```
MikeSpreitzer, line 80: I find the two levels of teardown here surprising. Is there a specific reason for these two levels? I think that a natural hierarchy would be as follows.

- Delete the workload but not the platform
- Remove the workload and platform from the namespace but do not remove the non-namespaced objects
- Remove everything

Those support some natural use cases:
- Leave the platform so that I can use it with other workload.
- Tear down everything in the namespace (leaving removal of the namespace itself to the discretion of the user) without disturbing other users in the cluster.
- Tear down everything in the cluster.

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+By default cleans up FMA / WVA objects but leaves the namespace, CRDs,
+WVA controller, and EPP/Gateway in place:
+
+```shell
+./test/e2e/demo-fma-wva/cleanup-fma-wva.sh --namespace my-fma-demo
+```
+
+Full cleanup (also removes the namespace, the `fma-poc=true` label that the
```
MikeSpreitzer: This cleanup is not "full", so naming it "full" is confusing.

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Common flags
+
+| Flag | Default | Used by |
+|---|---|---|
+| `-n`, `--namespace NAME` | `fma-wva-demo` | both |
+| `-f`, `--full-cleanup` | (off) | cleanup |
+| `--wva-repo-path PATH` | `<repo-root>/.wva-checkout` | both |
+| `--wva-repo-url URL` | `https://github.com/llm-d/llm-d-workload-variant-autoscaler` | both |
+| `--wva-version REF` | `v0.8.0-rc4` | both |
```
MikeSpreitzer: "REF" suggests that this can be any Git "ref". Is this true? What about pre-built container images? If this script builds container images, to what registry does it push them (and with what credentials)?

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+| `--fma-image-registry URL` | `ghcr.io/llm-d-incubation/llm-d-fast-model-actuation` | deploy |
+| `--fma-image-tag TAG` | `v0.6.0-alpha.13` | deploy |
+| `--model NAME` | `HuggingFaceTB/SmolLM2-360M-Instruct` | deploy |
+| `--gpu-node NODE` | first node with `nvidia.com/gpu.present=true` | deploy |
```
MikeSpreitzer, line 118: We have clusters with a mixed population of types of GPU, including small ones (e.g., L40S). Are all of those adequate and correctly used? Maybe there should be a note here about the requirements on the GPUs.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer, line 1: Shouldn't this file be named something more like "setup-fma-wva.sh", which is clearly complementary to the teardown script name? Saying "demo" suggests that it runs the demo.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
```diff
@@ -0,0 +1,661 @@
...
+
+FMA / cluster options:
+  -n, --namespace NAME              Target namespace (default: fma-wva-demo)
+      --fma-image-registry URL      FMA image registry
+                                    (default: ghcr.io/llm-d-incubation/llm-d-fast-model-actuation)
+      --fma-image-tag TAG           FMA image tag (default: v0.6.0-alpha.13)
+      --fma-launcher-image IMG      Launcher image
+                                    (default: <fma-image-registry>/launcher:<fma-image-tag>)
+      --fma-requester-image IMG     Requester image
+                                    (default: <fma-image-registry>/requester:<fma-image-tag>)
```
MikeSpreitzer: Offering these options seems like unnecessary complexity. What is the use case for this?

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+      --hf-token TOKEN              HuggingFace token (for gated models)
+      --runtime-class NAME          RuntimeClass for GPU pods (default: nvidia;
+                                    pass an empty string to omit)
+      --wait-requester-timeout DUR  How long to wait for Deployment/fma-requester
+                                    to become Available (default: 600s; covers
+                                    cold-cluster image+model fetch). Accepts
+                                    any kubectl --timeout value (e.g., 90s, 5m).
+
+WVA options:
+      --wva-version REF             WVA version. Used as both the git ref to
```
MikeSpreitzer: I do not see how a branch name or commit SHA will work. The listing of image tags in https://github.com/llm-d/llm-d-workload-variant-autoscaler/pkgs/container/llm-d-workload-variant-autoscaler/versions?filters%5Bversion_type%5D=tagged does not have tags that look like a commit SHA or a branch name (with the exception of `main`, but its vintage proves that it does not correspond to what is in the `main` branch now).

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+- kind: ServiceAccount
+  name: testlauncher
+  namespace: ${NAMESPACE}
+EOF
+echo "  Ensured RoleBinding testlauncher"
+
+
+step "FMA CRDs and Controllers (via deploy_fma.sh)"
+
+# Helm release (chart instance) name. The chart itself is `fma-controllers`;
```
MikeSpreitzer: > The chart itself is `fma-controllers`

You mean "The _name of_ the chart itself is `fma-controllers`".

That is a true statement, but I do not see how it is relevant here. This is a comment on an environment variable holding the release name, and the name of the chart itself plays no role in anything in sight.

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+# this is the name passed to `helm install/upgrade --install`, which prefixes
```
MikeSpreitzer: In English, the word "this" refers to the nearest preceding noun phrase. In the current text that would be the name of the chart itself --- which is not what you intend (if I understand correctly).

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+    fi
+fi
+echo "  Using GPU node: $GPU_NODE"
+
+# Label the chosen node for LPP selector
+kubectl label node "$GPU_NODE" fma-poc=true --overwrite=true 2>/dev/null
+echo "  Labeled $GPU_NODE with fma-poc=true"
+
+if kubectl get inferenceserverconfig isc-smol -n "$NAMESPACE" &>/dev/null; then
+    echo "  FMA objects already exist"
```
MikeSpreitzer: But what if they exist with the wrong contents? Why not unconditionally do the `kubectl apply`?

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+apiVersion: fma.llm-d.ai/v1alpha1
+kind: LauncherPopulationPolicy
+metadata:
+  name: lpp-fma
+spec:
+  enhancedNodeSelector:
+    labelSelector:
+      matchLabels:
+        fma-poc: "true"
+        nvidia.com/gpu.present: "true"
```
MikeSpreitzer: Why test `nvidia.com/gpu.present` here? It is redundant with testing fma-poc=true.

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: fma-requester
+  template:
+    metadata:
+      labels:
+        app: fma-requester
+        llm-d.ai/variant: fma-requester
```
MikeSpreitzer: Why is this label here?

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+      annotations:
+        dual-pods.llm-d.ai/admin-port: "8081"
```
MikeSpreitzer: Why is this here? The value given here is the default value.

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+        dual-pods.llm-d.ai/inference-server-config: "isc-smol"
+    spec:
+      ${runtime_class_line}
+      containers:
+        - name: inference-server
+          image: ${REQUESTER_IMAGE}
+          imagePullPolicy: Always
```
MikeSpreitzer: Why not `IfNotPresent`?

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+      llm-d.ai/guide: "optimized-baseline"
+  podMetricsEndpoints:
+  - interval: 30s
+    path: /metrics
+    relabelings:
+    # Only scrape the inference-server container (not state-change-reflector etc.)
+    - action: keep
+      regex: inference-server
+      sourceLabels: [__meta_kubernetes_pod_container_name]
+    # Force address to <pod-ip>:8000 — the launcher's vLLM /metrics port.
```
MikeSpreitzer: No. 8000 is the vllm instance's port, not the launcher's port.

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+      regex: (.+)
+      replacement: \$1:8000
+      sourceLabels: [__meta_kubernetes_pod_ip]
+      targetLabel: __address__
+    # Surface the variant label so WVA can correlate metrics to the VA.
+    - action: replace
+      sourceLabels: [__meta_kubernetes_pod_label_llm_d_ai_variant]
+      targetLabel: llm_d_ai_variant
+EOF
+    echo "  PodMonitor created"
```
MikeSpreitzer: kubectl is also going to print the same information. Why print it twice?

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
```diff
@@ -0,0 +1,661 @@
...
+echo "  Waiting up to ${WAIT_REQUESTER_TIMEOUT} for Deployment/fma-requester to become Available..."
+kubectl wait --for=condition=Available deployment/fma-requester \
+    -n "$NAMESPACE" --timeout="$WAIT_REQUESTER_TIMEOUT" 2>/dev/null || \
+    echo "  Deployment/fma-requester is not Available within ${WAIT_REQUESTER_TIMEOUT} — proceeding anyway"
+
+# Modern WVA uses annotation-based discovery: WVA finds the HPA via the
+# llm-d.ai/managed annotation, computes desired replicas, and publishes
+# wva_desired_replicas{variant_name=<deployment>, exported_namespace=<ns>}
+# to Prometheus. The HPA reads that external metric to scale the deployment.
+# No VariantAutoscaling CR required (it's the legacy path; see config/samples
```
MikeSpreitzer: Here "it's" after a negative statement is confusing, not clear whether you are talking about what _is_ or what _is not_. It would be clearer if the text said something like "using a VariantAutoscaling CR is the legacy...".

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+echo "  --- Monitoring Resources ---"
+kubectl get podmonitor,servicemonitor -n "$NAMESPACE" 2>/dev/null || true
+
+echo ""
+echo "  --- llm-d EPP Resources ---"
+kubectl get inferencepool,gateway,httproute -n "$NAMESPACE" 2>/dev/null || true
+
+echo ""
+echo "========================================"
+echo "  Deployment Complete!"
```
MikeSpreitzer, line 639: But the Validation step doesn't guarantee anything is actually here and in working order.

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+echo "    ✓ FMA objects (ISC, LauncherConfig, LPP, Deployment, PodMonitor)"
+echo "    ✓ WVA-managed HPA (annotation-based, no VariantAutoscaling)"
+echo ""
+echo "  Next steps:"
+echo "    - Check WVA metrics: kubectl get --raw /apis/external.metrics.k8s.io/v1beta1"
+echo "    - Check HPA status: kubectl get hpa fma-requester-hpa -n $NAMESPACE"
+echo "    - Monitor pods: kubectl get pods -n $NAMESPACE -w"
+echo "    - View WVA logs: kubectl logs -n $NAMESPACE -l app.kubernetes.io/name=workload-variant-autoscaler"
+echo "    - View FMA logs: kubectl logs -n $NAMESPACE -l app.kubernetes.io/part-of=fma"
+echo ""
```
MikeSpreitzer: It is best to have a linebreak at the end of a text file.

Image

MikeSpreitzer (commented):

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
```diff
@@ -0,0 +1,661 @@
...
+    cat <<EOF
+Usage: $(basename "$0") [OPTIONS]
+
+Deploys FMA + WVA + llm-d components in the same namespace.
+
+FMA / cluster options:
+  -n, --namespace NAME              Target namespace (default: fma-wva-demo)
+      --fma-image-registry URL      FMA image registry
+                                    (default: ghcr.io/llm-d-incubation/llm-d-fast-model-actuation)
+      --fma-image-tag TAG           FMA image tag (default: v0.6.0-alpha.13)
```
rubambiza: I'd recommend having this at least at `v0.6.1` to align with benchmarking in llm-d.

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+  -n, --namespace NAME              Target namespace (default: fma-wva-demo)
+      --fma-image-registry URL      FMA image registry
+                                    (default: ghcr.io/llm-d-incubation/llm-d-fast-model-actuation)
+      --fma-image-tag TAG           FMA image tag (default: v0.6.0-alpha.13)
+      --fma-launcher-image IMG      Launcher image
+                                    (default: <fma-image-registry>/launcher:<fma-image-tag>)
+      --fma-requester-image IMG     Requester image
+                                    (default: <fma-image-registry>/requester:<fma-image-tag>)
+      --model NAME                  vLLM model (default: HuggingFaceTB/SmolLM2-360M-Instruct)
+      --gpu-node NODE               Node for LPP (default: first node with
```
rubambiza: **nit**: First use of LPP acronym in this file, I'd recommend fully-defining it 👇 
```suggestion
      --gpu-node NODE               Node for Launcher Population Policy (LPP) (default: first node with
```

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+                                    nvidia.com/gpu.present=true)
+      --hf-token TOKEN              HuggingFace token (for gated models)
```
rubambiza: Wrt to gated models, I just wanted to surface [the feedback](https://llm-d.slack.com/files/U09TQ6LV7P0/F0BB09HNNNR/quickstart_feedback.md) we got from @delavet re: quickstart guide. It's specific to the Llama models, but it may be applicable broadly to gated models:
>This is a gated model on HuggingFace — users must request access on the model page and wait for approval before they can download it. The quickstart mentions HF_TOKEN as a prerequisite but does not note the additional gating step, which can block first-time users who have a token but haven’t been granted access to the Llama model family.

Otherwise, I'd suggest switching/standardizing around the un-gated `Qwen/Qwen3-8B` model for now. It is the one we are using in benchmarking and the nightlies.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
```diff
@@ -0,0 +1,661 @@
...
+long-form name in upper snake case — e.g. --namespace → NAMESPACE,
+--wva-repo-path → WVA_REPO_PATH). Env vars are accepted for backward
+compatibility; flags take precedence when both are set.
+EOF
+}
+
+# Seed defaults from env vars (so existing callers using env vars still work).
+NAMESPACE="${NAMESPACE:-fma-wva-demo}"
+CONTAINER_IMG_REG="${CONTAINER_IMG_REG:-ghcr.io/llm-d-incubation/llm-d-fast-model-actuation}"
+IMAGE_TAG="${IMAGE_TAG:-v0.6.0-alpha.13}"
```
rubambiza:
```suggestion
IMAGE_TAG="${IMAGE_TAG:-v0.6.1}"
```

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+
+
+step "FMA-specific objects (ISC, LauncherConfig, LPP, Deployment)"
+
+# Pick a GPU node for the LPP
+if [ -z "$GPU_NODE" ]; then
+    GPU_NODE=$(kubectl get nodes -l nvidia.com/gpu.present=true \
+        -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
+    if [ -z "$GPU_NODE" ]; then
+        echo "  ERROR: No GPU node found. Set GPU_NODE manually." >&2
```
rubambiza: **Nit:** Considering a future where, like the scripts that were re-used from HPA + FMA here, another demo/script leverages this script and perhaps we have generalized beyond NVIDIA GPUs, this makes it clear that only nodes with NVIDIA GPUs were not found. Unlikely, but I figured it's worth flagging.
```suggestion
        echo "  ERROR: No NVIDIA GPU node found. Set GPU_NODE manually." >&2
```

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+kind: InferenceServerConfig
+metadata:
+  name: isc-smol
+spec:
+  modelServerConfig:
+    port: 8000
+    options: "--model ${MODEL} --enable-sleep-mode"
+    env_vars:
+      VLLM_SERVER_DEV_MODE: "1"
+      VLLM_LOGGING_LEVEL: "DEBUG"
```
rubambiza, line 382: To keep up with https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/658, whose descriptions says:
>This PR defaults it to fork (unless the ISC already sets it)
This makes it an explicit, traceable decision.
```suggestion
      VLLM_LOGGING_LEVEL: "DEBUG"
      VLLM_WORKER_MULTIPROC_METHOD: "fork"
```

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+spec:
+  maxInstances: 1
+  podTemplate:
+    spec:
+      ${runtime_class_line}
+      serviceAccountName: testlauncher
+      containers:
+        - name: inference-server
+          image: ${LAUNCHER_IMAGE}
+          imagePullPolicy: Always
```
rubambiza: Given that we are pulling FMA images with particular image tag versions, and in the spirit of idempotency championed in this script, perhaps we should make this `IfNotPresent`? Unless we are using the demo for some form of benchmarking.
```suggestion
          imagePullPolicy: IfNotPresent
```

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+      VLLM_SERVER_DEV_MODE: "1"
+      VLLM_LOGGING_LEVEL: "DEBUG"
+    labels:
+      llm-d.ai/inference-serving: "true"
+      llm-d.ai/guide: "optimized-baseline"
+      llm-d.ai/model: "SmolLM2-360M-Instruct"
+      llm-d.ai/variant: fma-requester
+    annotations:
+      description: "FMA ISC - ${MODEL}"
+  launcherConfigName: lc-fma
```
rubambiza, line 390: **Nit**: Should the launcher config get deployed first since it's being referenced by the ISC?

test/e2e/demo-fma-wva/setup-fma-wva.sh
rubambiza: +1 here. I had drafted [this comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/579/changes#r3605288405) before I saw this one.

test/e2e/demo-fma-wva/setup-fma-wva.sh
```diff
@@ -0,0 +1,661 @@
...
+fi
+
+# Create PodMonitor for vLLM instances in launcher pods
+if kubectl get podmonitor fma-vllm-monitor -n "$NAMESPACE" &>/dev/null; then
+    echo "  PodMonitor for vLLM instances in launcher pods already exists"
+else
+    echo "  Creating PodMonitor for vLLM instances in launcher pods ..."
+
+
+kubectl apply -f - <<EOF
```
rubambiza, line 482: Similar to how we do so for the [requester HPA below](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/579/changes#diff-63c4f3e5511421bcad9a5d42570e490c386cef31f0b96ceb26c86b0f9d00bee3R547), may I suggest that this apply live under the else condition? Otherwise, the if statement is not needed if we always apply.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
```diff
@@ -0,0 +1,661 @@
...
+    # Surface the variant label so WVA can correlate metrics to the VA.
+    - action: replace
+      sourceLabels: [__meta_kubernetes_pod_label_llm_d_ai_variant]
+      targetLabel: llm_d_ai_variant
+EOF
+    echo "  PodMonitor created"
+fi
+
+
+step "WVA Autoscaling (annotation-based HPA, VariantAutoscaling-free)"
```
rubambiza:
```suggestion
step "WVA Autoscaling (annotation-based Horizontal Pod Autoscaler (HPA), VariantAutoscaling-free)"
```

rubambiza: First print in the steps, worth spelling out like we do [below](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/579/changes#diff-63c4f3e5511421bcad9a5d42570e490c386cef31f0b96ceb26c86b0f9d00bee3R548)

test/e2e/demo-fma-wva/README.md
rubambiza: At the time of this writing, v0.6.1 has also been tested with the benchmarking, and v0.6.2 is now released, and they both bring lots of improvements.

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+| FMA controllers + launcher/requester images | `--fma-image-tag` | `v0.6.0-alpha.13` |
+| WVA (git ref + controller image tag, kept in sync) | `--wva-version` | `v0.8.0-rc4` |
+| GIE (Gateway API Inference Extension) | `--gaie-version` | `v1.5.0` |
+| llm-d release (EPP/Gateway) | `--llm-d-release` | `v0.7.0` |
+
+See the "Pin all components" example below for a fully-pinned invocation.
+
+## Prerequisites
+
+- `oc` authenticated to an OpenShift cluster with GPU nodes
```
rubambiza:
```suggestion
- `oc` authenticated to an OpenShift cluster with NVIDIA GPU nodes
```

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
...
+| `-n`, `--namespace NAME` | `fma-wva-demo` | both |
+| `-f`, `--full-cleanup` | (off) | cleanup |
+| `--wva-repo-path PATH` | `<repo-root>/.wva-checkout` | both |
+| `--wva-repo-url URL` | `https://github.com/llm-d/llm-d-workload-variant-autoscaler` | both |
+| `--wva-version REF` | `v0.8.0-rc4` | both |
+| `--fma-image-registry URL` | `ghcr.io/llm-d-incubation/llm-d-fast-model-actuation` | deploy |
+| `--fma-image-tag TAG` | `v0.6.0-alpha.13` | deploy |
+| `--model NAME` | `HuggingFaceTB/SmolLM2-360M-Instruct` | deploy |
+| `--gpu-node NODE` | first node with `nvidia.com/gpu.present=true` | deploy |
+| `--hf-token TOKEN` | (unset) | deploy (if model is gated) |
```
rubambiza: Perhaps this might be a good spot to place a note regarding [this comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/579/changes#r3605075814).

rubambiza (commented):
Leaving more comments. Most of them are suggestions that can be applied immediately. Otherwise LGTM when all comments are resolved.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
MikeSpreitzer: The latest is currently 0.6.2. I strongly recommend that now.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: I strongly agree.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: We will have a lot of places to change when we generalize beyond nvidia, but no harm in being precise here.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: In other words: unless and until #658 merges, you _definitely_ want to specify this here.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: I definitely think that the image pull policy should be "IfNotPresent". The only reason that I know to use "Always" is when doing a rapid modify-build-test cycle using the same tag.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: As is typical in the Kubernetes world, there is no synchronization required. But nothing good happens until all the relevant objects are defined.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: If the point of this script is to put the system into a known state, then skipping the `kubectl` apply when an object is found with the right name but possibly the wrong spec can fail to accomplish the mission.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
```diff
@@ -0,0 +1,661 @@
...
+  metrics:
+  - type: External
+    external:
+      metric:
+        name: wva_desired_replicas
+        selector:
+          matchLabels:
+            # WVA labels its emitted metric by the scale target's name and namespace.
+            variant_name: fma-requester-hpa
+            exported_namespace: ${NAMESPACE}
```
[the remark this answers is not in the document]
dumb0002: N/A - Moved WVA to use KEDA objects

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,201 @@
+# FMA + WVA Demo on OpenShift
+
+Two scripts to deploy and tear down a working end-to-end demo of FMA + WVA + llm-d in a
+single OpenShift namespace. The scripts install the platform components
+(controllers, CRDs, RBAC, EPP/Gateway) **and** create a concrete workload
+that exercises them (an `InferenceServerConfig`, `LauncherConfig`,
+`LauncherPopulationPolicy`, requester `Deployment`, and a
+`HorizontalPodAutoscaler` annotated with `llm-d.ai/managed: "true"` so WVA
+discovers it and publishes the `wva_desired_replicas` external metric the
+HPA consumes to scale the deployment).
```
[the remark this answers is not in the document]
dumb0002: N/A - Moved WVA to use KEDA objects

test/e2e/demo-fma-wva/README.md
dumb0002: Yes! Added the word `default` to make it clear.

test/e2e/demo-fma-wva/README.md
dumb0002: Updated to FMA latest release `v0.6.2`

dumb0002: Yes, I updated it to `--fma-version` instead.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
dumb0002: Yes, I updated it to `0.6.2`

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,207 @@
+# FMA + WVA Demo on OpenShift
+
+Two scripts to deploy and tear down a working end-to-end demo of FMA + WVA + llm-d in a
```
MikeSpreitzer: I recommend explicitly mentioning KEDA in the first sentence. Also: although WVA may imply it, I would also mention HPA. Perhaps simply expand the enumeration of components to "FMA + WVA + HPA + KEDA + llm-d".

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,207 @@
...
+that exercises them (an `InferenceServerConfig`, `LauncherConfig`,
+`LauncherPopulationPolicy`, requester `Deployment`, and a KEDA `ScaledObject`
+annotated with `llm-d.ai/managed: "true"` so WVA discovers it and publishes the
+`wva_desired_replicas` metric to Prometheus). KEDA's Prometheus trigger reads
+that metric and drives the HPA it creates and manages
+(`wva-keda-hpa-fma-requester`) to scale the deployment.
+
+| Script | Purpose |
+|---|---|
+| `demo-fma-wva-ocp.sh` | Deploy FMA (CRDs, ValidatingAdmissionPolicies, RBAC, controllers, sample CRs), WVA, EPP/Gateway, and the demo requester workload |
```
MikeSpreitzer: Include HPA and KEDA in the list here too.
Why is llm-d included in the list up front but not here?

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,207 @@
...
+for the full list of options.
+
+## Versioning
+
+> **This script consumes _published_ FMA releases only.** Even though it
+> lives inside the FMA repo, it pulls FMA images by tag (e.g.
+> `v0.6.2`) — it does **not** build from the local source tree.
+> Don't change the `--fma-version` default to a tag that hasn't been pushed
+> yet; doing so couples the FMA repo to itself in a circular way and breaks
+> reproducibility.
```
MikeSpreitzer: Having a default at all is the problem. It teaches users to not specify the FMA version. Then, once there is a new FMA release and they start using this script from that release, they are silently getting the older FMA release. This creates the pressure to create a circular reference.

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+FMA, WVA, and GIE/llm-d release independently, so incompatibilities are
```
MikeSpreitzer: HPA and KEDA belong in this list.

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+The workload-variant-autoscaler (WVA) repo is cloned automatically — no need
+to pre-clone or pass `--wva-repo-path`.
+
+Both scripts use a standard CLI flag interface. Run either with `--help`
+for the full list of options.
+
+## Versioning
+
+> **This script consumes _published_ FMA releases only.** Even though it
```
MikeSpreitzer: This is not true. The version of FMA used by this script is **MIXED**! For images, this script uses the pre-built images of an FMA release. For everything else --- and there is a lot else --- it uses whatever version the user has locally in their current working directory.

- This is surprising complexity that the user must be made clearly aware of. Right from the start of any discussion of FMA version.
- This unpleasantness is driven by the fact that FMA does not have a script that installs a given release of FMA. We should fix that.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Done! Renamed file to `setup-fma-wva.sh`

test/e2e/demo-fma-wva/README.md
dumb0002: The default model is very small `HuggingFaceTB/SmolLM2-360M-Instruct` and it can be deployed without issues in any of our clusters GPU. Any additional GPU requirement will depend if the user changes the default model (`--model`).

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
dumb0002: Agree, they're unnecessary - removed them.

test/e2e/demo-fma-wva/README.md
MikeSpreitzer: Oh, reading on further I see that this script assumes that KEDA is already installed.
What about HPA?

test/e2e/demo-fma-wva/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+FMA, WVA, and GIE/llm-d release independently, so incompatibilities are
+possible. The defaults pin a known-good combination. If you change one,
+test the others alongside it.
+
+| Component | Flag | Default |
+|---|---|---|
+| All FMA images (dual-pods + launcher-populator controllers, launcher, requester) | `--fma-version` | `v0.6.2` |
+| WVA (git ref + controller image tag, kept in sync) | `--wva-version` | `v0.8.0-rc4` |
+| GIE (Gateway API Inference Extension) | `--gaie-version` | `v1.5.0` |
```
MikeSpreitzer: Is it true that "GIE" is the overwhelmingly common abbreviation for this? If so then make the flag `--gie-version`.

test/e2e/demo-fma-wva/README.md
dumb0002: I don't clearly understand this comment. "This script does not build container image by design".

test/e2e/demo-fma-wva/README.md
dumb0002: Replaced with ` --deep-cleanup.` instead.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Agreed. Removed that statement.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Fixed!

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Agree! Update to use `IfNotPresent` instead.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
dumb0002: Updated with the latest release for FMA.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Agree! Updated to use `IfNotPresent` instead.

test/e2e/demo-fma-wva/README.md
dumb0002: Done!

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Done!

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Done!

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
dumb0002: No longer applicably - moved to KEDA objects instead.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: That sentence was deleted in the latest commits. Hence, this comment is no longer applicable.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Removed all the additional print statement.

test/e2e/demo-fma-wva/demo-fma-wva-ocp.sh
dumb0002: This comment is no longer valid, because the text commented on no longer exists.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: For clarity I renamed the step from `Validation` to `Deployment Summary` .

dumb0002: The scope of this script is on the deployment on the infra only - making sure that everything is in working order is beyond the scope of the current script.

test/e2e/demo-fma-wva/README.md
dumb0002: KEDA already deploys a managed HPA instance.

test/e2e/demo-fma-wva/README.md
dumb0002: We're are not deploying neither KEDA or HPA in this script. The assumption is that KEDA with its own managed HPA already exists in the target cluster.

test/e2e/demo-fma-wva/README.md
dumb0002: The assumption is that KEDA with its own managed HPA already exists in the target cluster.

test/e2e/demo-fma-wva/README.md
dumb0002: Replaced `GIE` with `GAIE` - I have seen the abbreviation `GAIE` being used in other repos such as WVA.

test/e2e/demo-fma-wva/README.md
dumb0002: Removed default FMA release version - `fma-version` flag is now required,

test/e2e/demo-fma-wva/README.md
dumb0002: I updated the version text to make it clear that the FMA version deployed by the script is mixed.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Updated the model to `Qwen/Qwen3-8B` and also add the following comments regarding gated models: 

``Deploy a different model. The default (Qwen/Qwen3-8B) is un-gated. Gated models (e.g. meta-llama/*) also need approved access on the model's HF page — --hf-token alone is not enough:``

test/e2e/demo-fma-wva/README.md
MikeSpreitzer: Please let the person who opened a comment thread decide if and when it has been resolved.

While the script does not install HPA and KEDA, those are included in what is being demonstrated. It should be made clear at the start of the document that they are part of the demo.

dumb0002: Agree! I updated the README to  mentioning KEDA in the first paragraph. 

I only resolved those comments that are very straightforward and easy fix. But I left open all the comments that would require more discussion - Given the very large number of comments in this PR, it was becoming very hard for me to keep track of all of them and displaying them clearly. However, I'll keep open any new comment added to this PR.

test/e2e/demo-fma-wva/README.md
dumb0002: I added the following note in the examples section: 

``"Deploy a different model. The default (Qwen/Qwen3-8B) is un-gated. Gated models (e.g. meta-llama/*) also need approved access on the model's HF page — --hf-token alone is not enough".
``

because adding such not in the --hf-token cell makes that table row balloon and look bad.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Done!

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Fixed. Removed the `if/else` statement.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Removed the `if/else` statement in favor of `kubectl apply`

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Agree! Removed the redundant nvidia label.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: This label needs to be propagated to the launcher pod.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Removed the redundant annotation.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Yes! The comment implies the vLLM runs inside the launcher pod. However to avoid any further confusion, I removed any reference of a launcher pod in that comment.

test/e2e/demo-fma-wva/README.md
dumb0002: i) The user should just re-run the script and IMO it's an overkill to have to handle that very specific use-case in this script. In the worst case the user should just do a `clean-up` and `re-run` the script. 

ii) This script does not use a shared namespace - it creates its own dedicated namespace. Furthermore, based on the comments below, the `if/else` statement where remove in favor or `kubectl apply` - that would handle this scenario too.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: You're right! I updated that flag as following: 
```
--wva-version TAG             WVA version — used as both the git ref checked
                                    out and the controller image tag. Must be a
                                    published release tag (e.g. v0.8.0); branch
                                    names and commit SHAs won't work (no matching
                                    image is published). (default: v0.8.0)
```

test/e2e/demo-fma-wva/README.md
MikeSpreitzer: That's an improvement, but I still think that is burying the lede. The problem actually starts with the document title: it says `# FMA + WVA Demo on OpenShift` --- but the omission of KEDA is misleading. It is perfectly reasonable and possible to demonstrate FMA + WVA on openShift _without_ KEDA. The inclusion of KEDA in the subject matter of the demo is a deliberate and important choice. It should be included in the document title and the description in the first sentence. IMO it would be OK to omit HPA from the document title, since WVA and KEDA need something to do what HPA does _and_ HPA is the usual something --- but HPA should be mentioned in the first sentence for clarity, since WVA and KEDA can, at least in principle, be used together with something other than HPA to get that particular job done.

test/e2e/demo-fma-wva/README.md
dumb0002: Agree! I implemented the 3 tiers level suggested above:
     1.  **Workload**
     2.  **Workload and platform** - keep namespace. 
     3.  **All**

test/e2e/demo-fma-wva/README.md
dumb0002: Ok, I also added `KEDA` in the title as suggested. The `README` already states that we're using a KEDA managed HPA in the first paragraph. But KEDA uses HPA as part of its own architecture - see [here](https://keda.sh/http-add-on/0.15/concepts/architecture/). Hence, it is not need to say both KEDA + HPA. It's common practice to simply say either "KEDA" or "HPA" as example of autoscalers - for example: https://github.com/llm-d/llm-d/tree/main/guides/workload-autoscaling

test/e2e/demo-fma-wva/README.md
MikeSpreitzer: In commentary elsewhere, @rubambiza is recommending using an 8B model.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: FMA does not propagate labels from the server-requesting Pod to the launcher.

dumb0002: Ok, sorry! I miss understood this comment before. We need this label in the server-requesting pod to do  `Prometheus relabeling` for the scraped metrics as we discussed before.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: Let me clarify my agreement.

I strongly agree on using non-gated models.

I somewhat agree on making all the benchmarking use the same model, so that all the results are directly comparable.

Overall, I think that it is better if we demonstrate and test with a variety of models, lest we inadvertently build in a dependency on a detail of one model.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: @aavarghese has clarified that "fork" is the default (right?). So why do we need to specify this here?

test/e2e/demo-fma-wva/README.md
MikeSpreitzer: I see I was mistaken about flexibility wrt HPA. In fact, KEDA unconditionally works with HPA.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: This script provide a sample default model that is now consistent with what's used in all benchmarking. In addition, the script also provides the flexibility to use other models including gated model. However, IMO it's beyond the scope of this README to demonstrate and test with a variety of models - this is the scope for `benchmarking`.

test/e2e/demo-fma-wva/README.md
MikeSpreitzer: The title has been fixed, but a problem remains in the first paragraph. The first sentence mentions what is being demonstrated. Then the second sentence restates that, with a different list.

> Two scripts to deploy and tear down a working end-to-end demo of FMA + WVA + llm-d in a single OpenShift namespace.
> **What's demonstrated is autoscaling of an FMA workload driven by WVA through KEDA and an HPA.**

In my opinion it would be better to just have one sentence, with the correct list.

> Two scripts to deploy and tear down a working end-to-end demo of FMA + WVA + KEDA + HPA + llm-d in a single OpenShift namespace.

(BTW, the workload is not an "FMA" workload. The workload is an _llm-d_ workload.)

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Yes, indeed! Just removed the redundant `env var`.

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: Please help me remember. Why do we want Prometheus metrics from the server-requesting Pod to be scraped? The vLLM instance is elsewhere (in the launcher Pod).

test/e2e/demo-fma-wva/setup-fma-wva.sh
MikeSpreitzer: > it's beyond the scope of this README to demonstrate and test with a variety of models

Agreed.

test/e2e/demo-fma-wva/README.md
MikeSpreitzer: Now that you have adopted that suggestion, is it still true that the default model can be deployed without issues in any of our clusters GPUs?

dumb0002: The default model fits in the `vllm-d` cluster with `A100` and `L40` or the `pokprod` cluster with `H100` isn't it?

test/e2e/demo-fma-wva/README.md
dumb0002: good catch. Thanks! I updated the first paragraph as suggested but with a minor modification: removed some references to HPA because saying KEDA + HPA is not standard practice.

test/e2e/demo-fma-wva/setup-fma-wva.sh
dumb0002: Ok, I had to revise WVA latest changes in `main`. WVA has now deprecated VA objects CRD.  VA is now only an internal resource abstraction. Hence, the `llm-d.ai/variant` is now stamped in the KEDA ScaledObject and the Scaling Workload. The label is used to enable direct `pod-to-VA` mapping and it's used by the metrics collector to filter pods and associate them with their managing VA object.

With this latest changes, the `llm-d.ai/variant` is no longer needed in the FMA requester pod but it still needed in the FMA launcher pods. Thus, I removed this label from the requester pod in this PR.

dumb0002:
@MikeSpreitzer (cc: @jimcadden), FYI, I'll leave this PR open for now while we shift into the llm-d Well-Lit-Paths instead. However, we should have a **Quickstart** in some form for both `FMA + WVA` and `FMA + KEDA`  integrations. IMO a **QuickStart** is the best way to show value to an eager user that wants to experiment with FMA.

dumb0002:
Closing this PR as we shift the focus to KEDA as the supported autoscaler.

Issue #583, opened 2026-06-21
[Bug]: `Created vLLM instance` log line missing despite a `Successfully relayed the readiness` for the same requester


aavarghese:
## Source
- llm-d-benchmark CI run 27859600526 (FMA pass, runner-20260620-044323-886)
- Image: `ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/launcher:v0.6.0-alpha.13`
- Cluster: pokprod (OpenShift)

## Expected vs observed in `dpc-controller.log`

The DPC controller emits these V(2) messages (relevant lines in `pkg/controller/dual-pods/inference-server.go` at v0.6.0-alpha.13):
- `Created vLLM instance` (line 472) — fires AFTER `lClient.CreateNamedInstance(...)` returns success
- `Successfully relayed the readiness` (line 523) — fires after the readiness HTTP relay to the requester

A relay is downstream of an instance create — the relay can only fire if there's a vLLM instance running on the bound launcher.

This log captures:
- 1 `Successfully relayed the readiness` for `requesterName="fma-requester-qwen-qwe-2d2178f3-wen3-32b-64d9b858f5-nxdkw"` at 04:46:27
- **0** `Created vLLM instance` lines anywhere in the file

The vLLM instance for nxdkw must have been created (otherwise the relay couldn't succeed), but the create-log line was never emitted.

## Hypothesis (likely cause)

`logger.V(2).Info("Created vLLM instance", ...)` at inference-server.go:472 is emitted AFTER the HTTP create response is fully received by the controller. If the call succeeds server-side (the launcher created the instance) but the response is lost or times out before reaching the logger.Info call, the log line is dropped. The instance exists on the launcher; the next reconcile sees `instancePresent=true` and silently falls through to `serverDat.InstanceKnownToExist = true` (line 474) without re-emitting the log.

## Suggested fix
Emit two log lines instead of one:
- BEFORE the HTTP call: `Creating vLLM instance` with `httpCallStartTime`
- AFTER the HTTP call (current): `Created vLLM instance` with `instance_id`, `status`, `httpCallStartTime`

The pre-call line guarantees observability of every create attempt; the post-call line confirms success. This also lets downstream tooling (e.g. PR #1504's `dpc_log_parser.py`'s `_CREATE_INSTANCE_MSG`) compute T_warm even when the response is lost.

[...]

PR #586, opened 2026-06-22
Centralize HTTP call logging and metrics

MikeSpreitzer:
Centralize logging and Prometheus instrumentation for HTTP calls (to requesters, launchers, vllm instances).

This will produce more comprehensive metrics on HTTP calls.

This is hoped to address Issue #583 .

This PR will certainly require changes in the log parsing. This PR removes the timing information from an incomplete collection of ad-hoc logging statements and puts it in a comprehensive and regular logging regimen.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -124,17 +124,20 @@ Labels are as follows.
   deleted for any reason in the course of this actuation.
 - `isc_name`: name of the relevant InferenceServerConfig object
 
-### fma_wake_seconds
+### fma_http_latency_seconds
 
-Vector of histograms: Latency of `/wake_up` call from DPC to vllm.
+Vector of histograms: Latency of HTTP calls. Currently, only calls
+from the dual-pods controller are instrumented.
 
 Labels are as follows.
 
+- `purpose`: short token identifying the purpose of the call
+- `method`: the HTTP method
 - `exported_namespace`: Kube namespace of requester and provider
 - `isc_name`: name of the relevant InferenceServerConfig object
-- `success`: "true" if no error, "false" otherwise. Here "error" means
-  a failure to construct the HTTP request message, send it, or receive
-  the response message. The HTTP "status" code is not germane.
+- `status_code`: if an HTTP response was not received then "0",
+  otherwise the decimal representation of the returned HTTP status
+  code.
 
 ### fma_launcher_create_seconds
 
@@ -147,18 +150,6 @@ Labels are as follows.
 
 - `exported_namespace`: Kube namespace of requester and provider
 - `lcfg_name`: name of the relevant LauncherConfig object
-- `success`: "true" if no error, "false" otherwise. Here "error" means
-  a failure to construct the HTTP request message, send it, or receive
-  the response message. The HTTP "status" code is not germane.
-
-### fma_instance_create_seconds
-
-Vector of histograms: Latency of DPC call to launcher to create vllm instance.
-
-Labels are as follows.
-
-- `exported_namespace`: Kube namespace of requester and provider
-- `isc_name`: name of the relevant InferenceServerConfig object
 - `success`: "true" if no error, "false" otherwise. Here "error" means
   a failure to construct the HTTP request message, send it, or receive
   the response message. The HTTP "status" code is not germane.
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -139,6 +139,23 @@ Labels are as follows.
   otherwise the decimal representation of the returned HTTP status
   code.
 
+The possible values of the `purpose` label are as follows.
+
+- `create_instance`
+- `delete_all_instances`
+- `delete_instance`
+- `get_accel_memory_usage`
+- `get_gpu_uuids`
+- `get_health`
+- `get_instance_state`
+- `list_instances`
+- `list_instance_ids`
+- `query_sleeping`
+- `relay_ready`
+- `relay_unready`
+- `sleep`
+- `wake`
+
 ### fma_launcher_create_seconds
 
 Vector of histograms: Latency of kube API call to create launcher.
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -31,6 +31,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -176,9 +178,11 @@ var (
 	requestersPerISCGauges       *kubemetrics.GaugeVec
 	iscCountGauges               *kubemetrics.GaugeVec
 	actuationSecsHistograms      *kubemetrics.HistogramVec
-	wakeSecsHistograms           *kubemetrics.HistogramVec
 	launcherCreateSecsHistograms *kubemetrics.HistogramVec
-	instanceCreateSecsHistograms *kubemetrics.HistogramVec
+
+	// HistogramVec of HTTP call latencies.
+	// Labels are purpose, method, exported_namespace, isc_name, status_code
+	httpLatencySecsHistograms *kubemetrics.HistogramVec
 )
 
 // NewController makes a new dual pods controller.
@@ -190,49 +194,6 @@ func (config ControllerConfig) NewController(
 	corev1PreInformers corev1preinformers.Interface,
 	fmaInformerFactory fmainformers.SharedInformerFactory,
 ) (*controller, error) {
-	ctl := &controller{
-		enqueueLogger:       logger.WithName(ControllerName),
-		coreclient:          coreClient,
-		namespace:           namespace,
-		podInformer:         corev1PreInformers.Pods().Informer(),
-		podLister:           corev1PreInformers.Pods().Lister(),
-		cmInformer:          corev1PreInformers.ConfigMaps().Informer(),
-		cmLister:            corev1PreInformers.ConfigMaps().Lister(),
-		nodeInformer:        corev1PreInformers.Nodes().Informer(),
-		nodeLister:          corev1PreInformers.Nodes().Lister(),
-		iscInformer:         fmaInformerFactory.Fma().V1alpha1().InferenceServerConfigs().Informer(),
-		iscLister:           fmaInformerFactory.Fma().V1alpha1().InferenceServerConfigs().Lister(),
-		lcInformer:          fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Informer(),
-		lcLister:            fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Lister(),
-		sleeperLimit:        config.SleeperLimit,
-		debugAccelMemory:    config.AcceleratorSleepingMemoryLimitMiB < math.MaxInt32,
-		accelMemoryLimitMiB: config.AcceleratorSleepingMemoryLimitMiB,
-		nodeNameToData:      map[string]*nodeData{},
-	}
-	err := ctl.podInformer.AddIndexers(cache.Indexers{
-		inferenceServerConfigIndexName: inferenceServerConfigIndexFunc,
-		launcherConfigHashIndexName:    launcherConfigHashIndexFunc,
-		requesterIndexName:             requesterIndexFunc,
-		nodeNameIndexName:              nodeNameIndexFunc,
-		nominalHashIndexName:           nominalHashIndexFunc,
-		GPUIndexName:                   GPUIndexFunc})
-	if err != nil { //impossible
-		return nil, err
-	}
-	ctl.KnowsProcessedSync = genctlr.NewKnowsProcessedSync(ControllerName, config.NumWorkers, ctl.process,
-		func(distinguisher int) queueItem {
-			return cmItem{ObjectName: cache.ObjectName{Name: strconv.FormatInt(int64(distinguisher), 10), Namespace: sentinelNS}}
-		},
-		func(qi queueItem) bool {
-			if cm, ok := qi.(cmItem); ok {
-				return cm.Namespace == sentinelNS
-			}
-			return false
-		},
-		func(ctx context.Context) {
-			logger := klog.FromContext(ctx)
-			logger.V(1).Info("All initial items processed")
-		})
 	onceler.Do(func() {
 		addsCounters = kubemetrics.NewCounterVec(&kubemetrics.CounterOpts{
 			Namespace:      "fma",
@@ -300,13 +261,6 @@ func (config ControllerConfig) NewController(
 			Buckets:        []float64{0, 1, 3, 5, 7.5, 10, 15, 30, 60, 120, 240, 480, 960, 1920},
 			StabilityLevel: kubemetrics.ALPHA,
 		}, []string{"exported_namespace", "path", "instancesDeleted", "isc_name"})
-		wakeSecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
-			Namespace:      "fma",
-			Name:           "wake_seconds",
-			Help:           "Latency of /wake_up call from DPC to vllm",
-			Buckets:        []float64{0.001, 0.01, 0.1, 1, 2, 3, 5, 7.5, 10, 15, 30},
-			StabilityLevel: kubemetrics.ALPHA,
-		}, []string{"exported_namespace", "isc_name", "success"})
 		launcherCreateSecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
 			Namespace:      "fma",
 			Name:           "launcher_create_seconds",
@@ -314,21 +268,68 @@ func (config ControllerConfig) NewController(
 			Buckets:        []float64{0.001, 0.01, 0.1, 1, 2, 3, 4, 5},
 			StabilityLevel: kubemetrics.ALPHA,
 		}, []string{"exported_namespace", "lcfg_name", "success"})
-		instanceCreateSecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
+
+		httpLatencySecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
 			Namespace:      "fma",
-			Name:           "instance_create_seconds",
-			Help:           "Latency of DPC call to launcher to create vllm instance",
+			Subsystem:      "http",
+			Name:           "latency_seconds",
+			Help:           "Latency of HTTP call from DPC",
 			Buckets:        []float64{0.001, 0.01, 0.1, 1, 2, 3, 5, 10, 30, 90, 270, 810},
 			StabilityLevel: kubemetrics.ALPHA,
-		}, []string{"exported_namespace", "isc_name", "success"})
+		}, []string{"purpose", "method", "exported_namespace", "isc_name", "status_code"})
 
 		legacyregistry.MustRegister(addsCounters, queueDepthGauges, queueDurationHists,
 			retriesCounters, workDurationHists,
 			dualityGauges,
 			requestersPerISCGauges, iscCountGauges, actuationSecsHistograms,
-			wakeSecsHistograms, launcherCreateSecsHistograms, instanceCreateSecsHistograms,
+			launcherCreateSecsHistograms, httpLatencySecsHistograms,
 		)
 	})
+
+	ctl := &controller{
+		enqueueLogger:             logger.WithName(ControllerName),
+		coreclient:                coreClient,
+		namespace:                 namespace,
+		podInformer:               corev1PreInformers.Pods().Informer(),
+		podLister:                 corev1PreInformers.Pods().Lister(),
+		cmInformer:                corev1PreInformers.ConfigMaps().Informer(),
+		cmLister:                  corev1PreInformers.ConfigMaps().Lister(),
+		nodeInformer:              corev1PreInformers.Nodes().Informer(),
+		nodeLister:                corev1PreInformers.Nodes().Lister(),
+		iscInformer:               fmaInformerFactory.Fma().V1alpha1().InferenceServerConfigs().Informer(),
+		iscLister:                 fmaInformerFactory.Fma().V1alpha1().InferenceServerConfigs().Lister(),
+		lcInformer:                fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Informer(),
+		lcLister:                  fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Lister(),
+		httpLatencySecsHistograms: httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"exported_namespace": namespace}),
+		sleeperLimit:              config.SleeperLimit,
+		debugAccelMemory:          config.AcceleratorSleepingMemoryLimitMiB < math.MaxInt32,
+		accelMemoryLimitMiB:       config.AcceleratorSleepingMemoryLimitMiB,
+		nodeNameToData:            map[string]*nodeData{},
+	}
+	err := ctl.podInformer.AddIndexers(cache.Indexers{
+		inferenceServerConfigIndexName: inferenceServerConfigIndexFunc,
+		launcherConfigHashIndexName:    launcherConfigHashIndexFunc,
+		requesterIndexName:             requesterIndexFunc,
+		nodeNameIndexName:              nodeNameIndexFunc,
+		nominalHashIndexName:           nominalHashIndexFunc,
+		GPUIndexName:                   GPUIndexFunc})
+	if err != nil { //impossible
+		return nil, err
+	}
+	ctl.KnowsProcessedSync = genctlr.NewKnowsProcessedSync(ControllerName, config.NumWorkers, ctl.process,
+		func(distinguisher int) queueItem {
+			return cmItem{ObjectName: cache.ObjectName{Name: strconv.FormatInt(int64(distinguisher), 10), Namespace: sentinelNS}}
+		},
+		func(qi queueItem) bool {
+			if cm, ok := qi.(cmItem); ok {
+				return cm.Namespace == sentinelNS
+			}
+			return false
+		},
+		func(ctx context.Context) {
+			logger := klog.FromContext(ctx)
+			logger.V(1).Info("All initial items processed")
+		})
 
 	_, err = ctl.podInformer.AddEventHandler(ctl)
 	if err != nil {
@@ -364,6 +365,9 @@ type controller struct {
 	lcInformer    cache.SharedIndexInformer
 	lcLister      fmalisters.LauncherConfigLister
 	genctlr.KnowsProcessedSync[queueItem]
+
+	// Histogram vector, labels needing values are: purpose, method, isc_name, status_code
+	httpLatencySecsHistograms prometheus.ObserverVec
 
 	sleeperLimit        int
 	debugAccelMemory    bool
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -149,6 +149,7 @@ func (item unboundLauncherPodItem) process(ctx context.Context, ctl *controller,
 }
 
 func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *nodeData) (error, bool) {
+	// The `requesterName` value is relied upon by the log parser in benchmarking
 	logger := klog.FromContext(urCtx).WithValues("serverUID", item.UID, "requesterName", item.RequesterName)
 	serverDat := ctl.getServerData(nodeDat, item.RequesterName, item.UID)
 	if serverDat.InstanceID != "" {
@@ -159,6 +160,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	providerRV := "(non existent)"
 	var requesterDeletionTimestamp, providerDeletionTimestamp *string
 	var requesterRCS, providerRCS *reducedContainerState
+	var iscName string
 
 	requestingPod, err := ctl.podLister.Pods(ctl.namespace).Get(item.RequesterName)
 	if err != nil {
@@ -172,6 +174,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		requesterRV = requestingPod.ResourceVersion
 		requesterDeletionTimestamp = TimePtrToStringPtr(requestingPod.DeletionTimestamp)
 		requesterRCS = getReducedInferenceContainerState(requestingPod)
+		iscName = requestingPod.Annotations[api.InferenceServerConfigAnnotationName]
 	}
 
 	var providingPod *corev1.Pod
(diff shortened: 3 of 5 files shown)


pkg/controller/dual-pods/inference-server.go
```diff
@@ -530,11 +525,9 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				url += stubapi.BecomeUnreadyPath
 				readiness = "unready"
 			}
-			postStart := time.Now()
-			err = doPost(ctx, url)
-			postStartStr := postStart.Format(time.RFC3339Nano)
+			_, err = doHTTP(ctx, "POST", url, httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"purpose": "relay_readiness", "exported_namespace": ctl.namespace, "iscName": requestingPod.Annotations[api.InferenceServerConfigAnnotationName]}), nil, nil)
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1969,62 @@ func init() {
...
+			decoder := json.NewDecoder(resp.Body)
+			bodyErr := decoder.Decode(resultData)
+			if bodyErr != nil {
+				err = fmt.Errorf("failed to decode response to %s %q: %w", method, url, bodyErr)
+			}
+		}
 	}
-	return body, nil
+	logger.V(3).Info("HTTP %s %q started at %v, took %vs, statusCode=%d, err=%v", method, url, httpCallStartTime.Format(time.RFC3339Nano), latencySecs, statusCode, err)
+	latencyHistogramVec.WithLabelValues(method, strconv.FormatInt(int64(statusCode), 10)).Observe(latencySecs)
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1969,62 @@ func init() {
...
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return nil, fmt.Errorf("http get %q: %w", url, err)
+func doHTTP(ctx context.Context, method, url string, latencyHistogramVec prometheus.ObserverVec, requestData, resultData any) (int, error) {
+	logger := klog.FromContext(ctx)
+	var reqBody io.Reader
+	if requestData != nil {
+		b, err := json.Marshal(requestData)
+		if err != nil {
+			return 0, fmt.Errorf("failed to marshel request body for %s %q (requestData=%#v): %w", method, url, requestData, err)
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1969,62 @@ func init() {
 	podDecoder = codecFactory.UniversalDecoder(corev1.SchemeGroupVersion)
 }
 
-func doGet(ctx context.Context, url string) ([]byte, error) {
-	client := &http.Client{
-		Timeout: 5 * time.Second,
-	}
+var myHTTPClient = &http.Client{
+	Timeout: time.Minute,
+}
```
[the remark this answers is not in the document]
MikeSpreitzer: Yes, intentional change. Revised to 30 sec, to match previous LauncherClient behavior.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -207,47 +201,19 @@ func IsInstanceNotFoundError(err error) bool {
...
-	if err != nil {
-		return err
-	}
-	defer func() { _ = resp.Body.Close() }()
-
-	if resp.StatusCode >= 300 {
-		b, _ := io.ReadAll(resp.Body)
-		return &launcherError{StatusCode: resp.StatusCode, Body: string(b)}
-	}
+	statusCode, err := doHTTP(ctx, method, u.String(), c.latencyHistograms.MustCurryWith(prometheus.Labels{"purpose": purpose}), body, out)
```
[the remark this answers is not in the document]
MikeSpreitzer: This problem predates this PR. But fixed.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -17,31 +17,27 @@ limitations under the License.
...
 type launcherError struct {
 	StatusCode int
-	Body       string
+	Err        error
 }
 
 func (e *launcherError) Error() string {
-	return fmt.Sprintf("launcher error %d: %s", e.StatusCode, e.Body)
+	return fmt.Sprintf("launcher error %d: %s", e.StatusCode, e.Err.Error())
 }
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

pkg/controller/dual-pods/controller.go
```diff
@@ -300,33 +299,28 @@ func (config ControllerConfig) NewController(
...
 		}, []string{"exported_namespace", "lcfg_name", "success"})
-		instanceCreateSecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
+
+		httpLatencySecsHistograms = kubemetrics.NewHistogramVec(&kubemetrics.HistogramOpts{
 			Namespace:      "fma",
-			Name:           "instance_create_seconds",
-			Help:           "Latency of DPC call to launcher to create vllm instance",
+			Subsystem:      "http",
+			Name:           "latency_secs",
+			Help:           "Latency of HTTP call from DPC",
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

docs/metrics.md
```diff
@@ -124,17 +124,19 @@ Labels are as follows.
...
 
-Vector of histograms: Latency of `/wake_up` call from DPC to vllm.
+Vector of histograms: Latency of HTTP calls. Currently, only calls
+from the dual-pods controller are instrumented.
 
 Labels are as follows.
 
+- `method`: the HTTP method
+- `purpse`: short token identifying the purpose of the call
 - `exported_namespace`: Kube namespace of requester and provider
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1969,62 @@ func init() {
...
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return nil, fmt.Errorf("http get %q: %w", url, err)
+func doHTTP(ctx context.Context, method, url string, latencyHistogramVec prometheus.ObserverVec, requestData, resultData any) (int, error) {
+	logger := klog.FromContext(ctx)
+	var reqBody io.Reader
+	if requestData != nil {
+		b, err := json.Marshal(requestData)
+		if err != nil {
+			return 0, fmt.Errorf("failed to marshal request body for %s %q (requestData=%#v): %w", method, url, requestData, err)
```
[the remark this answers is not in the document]
MikeSpreitzer: We are presuming no sensitive data there.

docs/metrics.md
```diff
@@ -124,17 +124,19 @@ Labels are as follows.
...
 
+- `method`: the HTTP method
+- `purpose`: short token identifying the purpose of the call
 - `exported_namespace`: Kube namespace of requester and provider
 - `isc_name`: name of the relevant InferenceServerConfig object
-- `success`: "true" if no error, "false" otherwise. Here "error" means
-  a failure to construct the HTTP request message, send it, or receive
-  the response message. The HTTP "status" code is not germane.
+- `status_code`: if the call failed, "0", otherwise the decimal
+  representation of the returned HTTP status code.
```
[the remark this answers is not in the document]
MikeSpreitzer: Revised the language to be clearer.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -293,7 +295,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		if providingPod.Labels != nil {
 			_, providingPodLauncherBased = providingPod.Labels[ctlrcommon.LauncherConfigNameLabelKey]
 		}
-		err := ctl.ensureUnbound(ctx, serverDat, nodeDat, providingPod, providingPodLauncherBased)
+		err := ctl.ensureUnbound(ctx, serverDat, requestingPod.Annotations[api.InferenceServerConfigAnnotationName], nodeDat, providingPod, providingPodLauncherBased)
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1973,62 @@ func init() {
...
+			decoder := json.NewDecoder(resp.Body)
+			bodyErr := decoder.Decode(resultData)
+			if bodyErr != nil {
+				err = fmt.Errorf("failed to decode response to %s %q: %w", method, url, bodyErr)
+			}
+		}
+	}
+	logger.V(3).Info("HTTP call done", "method", method, "url", url, "httpCallStartTime", httpCallStartTime.Format(time.RFC3339Nano), "latencySecs", latencySecs, "statusCode", statusCode, "err", err)
+	latencyHistogramVec.WithLabelValues(method, strconv.FormatInt(int64(statusCode), 10)).Observe(latencySecs)
+	return statusCode, err
```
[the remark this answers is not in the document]
MikeSpreitzer: Copilot you are wrong, see https://github.com/prometheus/client_golang/blob/v1.22.0/prometheus/observer.go#L47

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1582,30 +1564,23 @@ func (ctl *controller) ensureUnbound(ctx context.Context, serverDat *serverData,
...
-	}
-	logger.V(4).Info("Got vLLM instance state", "instanceID", serverDat.InstanceID, "httpCallStartTime", getStartStr)
-	iscName := instState.Annotations[VllmConfigISCNameAnnotationKey]
-	if iscName == "" {
-		logger.V(4).Info("Cannot check instance obsolescence: no ISC name annotation on instance", "instanceID", serverDat.InstanceID)
+		logger.V(4).Info("Cannot check instance obsolescence: failed to get instance state", "instanceID", serverDat.InstanceID, "err", err)
 		return false
 	}
+	logger.V(4).Info("Got vLLM instance state", "instanceID", serverDat.InstanceID)
 	currentISC, err := ctl.iscLister.InferenceServerConfigs(ctl.namespace).Get(iscName)
```
waltforme, line 1597: This should probably keep using the ISC name from `instState.Annotations[VllmConfigISCNameAnnotationKey]`, as the old code did. `ensureUnbound` can be called when `requestingPod == nil`, and in that case this PR passes `iscName == ""`; then this lookup becomes `Get("")`, `maybeDeleteObsoleteInstance` returns false, and an obsolete launcher instance is put to sleep instead of being deleted. The passed ISC name is good for metrics, but the obsolescence check still needs the instance annotation because it survives after the requester pod is gone.

MikeSpreitzer: Revised to use that if it is available.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1987,72 @@ func init() {
...
+var myHTTPClient = &http.Client{
+	Timeout: 30 * time.Second,
+}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return nil, fmt.Errorf("http get %q: %w", url, err)
+// ObserverCube is the needed fragment of prometheus.ObserverVec,
+// semantically (but not in the type system) specialized to
+// needing values for three lables: purpose, method, status_code.
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

docs/metrics.md
```diff
@@ -124,17 +124,20 @@ Labels are as follows.
...
 Labels are as follows.
 
+- `purpose`: short token identifying the purpose of the call
+- `method`: the HTTP method
 - `exported_namespace`: Kube namespace of requester and provider
 - `isc_name`: name of the relevant InferenceServerConfig object
-- `success`: "true" if no error, "false" otherwise. Here "error" means
-  a failure to construct the HTTP request message, send it, or receive
-  the response message. The HTTP "status" code is not germane.
+- `status_code`: if an HTTP response was not received then "0",
```
rubambiza, line 138: Just confirming here, the `otherwise` part covers both successful (e.g., 200s) and erroneous HTTP responses (picking 400s for the sake of discussion), right?

I ask because I've come to learn 0 means all went well, anything else means something bad happened. Ofc, this case is a bit more complex because the set of status codes covers both successes and failures. In that line of thinking, perhaps it makes sense to have -1 instead of 0 for when a response was not received.

docs/metrics.md
```diff
@@ -150,15 +153,3 @@ Labels are as follows.
...
-
-Vector of histograms: Latency of DPC call to launcher to create vllm instance.
-
-Labels are as follows.
-
-- `exported_namespace`: Kube namespace of requester and provider
-- `isc_name`: name of the relevant InferenceServerConfig object
-- `success`: "true" if no error, "false" otherwise. Here "error" means
-  a failure to construct the HTTP request message, send it, or receive
-  the response message. The HTTP "status" code is not germane.
```
rubambiza, line 164: Were all these lines removed in favor of the centralized calls? I appreciated having these because they matched up with the metrics that eventually get produced by the parsers downstream.

rubambiza: It's likely my question may get answered as I review more files further down.

pkg/controller/dual-pods/controller.go
```diff
@@ -300,36 +261,76 @@ func (config ControllerConfig) NewController(
...
+		cmInformer:                corev1PreInformers.ConfigMaps().Informer(),
+		cmLister:                  corev1PreInformers.ConfigMaps().Lister(),
+		nodeInformer:              corev1PreInformers.Nodes().Informer(),
+		nodeLister:                corev1PreInformers.Nodes().Lister(),
+		iscInformer:               fmaInformerFactory.Fma().V1alpha1().InferenceServerConfigs().Informer(),
+		iscLister:                 fmaInformerFactory.Fma().V1alpha1().InferenceServerConfigs().Lister(),
+		lcInformer:                fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Informer(),
+		lcLister:                  fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Lister(),
+		httpLatencySecsHistograms: httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"exported_namespace": namespace}),
+		sleeperLimit:              config.SleeperLimit,
```
rubambiza, line 304: This is separate and beyond scope, but are we at a point where we can now remove the references to `sleeperLimit` in favor of `maxInstances`?

I'd be willing to open an issue to make the necessary changes. Just wanted to flag in case there are nasty downstream effects.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -469,30 +475,23 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				// We just synced, so we know the instance is not on the launcher — create directly.
 				serverDat.NeededNewInstance = true
 				launcherBaseURL := fmt.Sprintf("http://%s:%d", providingPod.Status.PodIP, ctlrcommon.LauncherServicePort)
-				lClient, err := NewLauncherClient(launcherBaseURL)
+				lClient, err := NewLauncherClient(launcherBaseURL, ctl.httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"isc_name": requestingPod.Annotations[api.InferenceServerConfigAnnotationName]}))
```
rubambiza: For consistency, I'd like to suggest initializing the `iscName` variable like we did in the previous function, unless we anticipate it to change in the course of the processing. Perhaps somewhere above line 354 in the new diff.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -207,47 +224,57 @@ func IsInstanceNotFoundError(err error) bool {
...
+	u := c.baseURL.ResolveReference(parsed)
+	var oc ObserverCube
+	switch {
+	case c.latencyHistograms != nil:
+		oc = c.latencyHistograms
+	case complete != nil:
+		oc = completeThenCurry{c, complete}
+	case true:
+		c.latencyHistograms = c.isclessLatencyHistograms.MustCurryWith(prometheus.Labels{"isc_name": ""})
+		oc = c.latencyHistograms
```
diegocastanibm, line 262: What is the `default` here?

diegocastanibm: ouh, I've seen on Claude that `case true` is equivalent to the `default`. OK, nevermind

diegocastanibm (dismissed):
LGTM

pkg/controller/dual-pods/inference-server.go
```diff
@@ -469,30 +475,23 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
-					Observe(createEnd.Sub(createStart).Seconds())
-				createStartStr := createStart.Format(time.RFC3339Nano)
 				if err != nil {
-					return fmt.Errorf("failed to create vLLM instance %q (started %s): %w", serverDat.InstanceID, createStartStr, err), true
+					return fmt.Errorf("failed to create vLLM instance %q: %w", serverDat.InstanceID, err), true
 				}
 				serverDat.InstanceKnownToExist = true
 				launcherDat := ctl.getLauncherData(nodeDat, providingPod.Name)
 				launcherDat.Instances[serverDat.InstanceID] = time.Now()
-				logger.V(2).Info("Created vLLM instance", "instance_id", result.InstanceID, "status", result.Status, "httpCallStartTime", createStartStr)
```
rubambiza, line 490: I anticipate that the move of these `httpCallStartTime` will introduce breaking changes. Will confirm and follow-up with an AI-assisted review.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1987,72 @@ func init() {
...
+		} else if resultData != nil {
+			decoder := json.NewDecoder(resp.Body)
+			bodyErr := decoder.Decode(resultData)
+			if bodyErr != nil {
+				err = fmt.Errorf("failed to decode response to %s %q: %w", method, url, bodyErr)
+			}
+		}
+	}
+	logger := klog.FromContext(ctx)
+	logger.V(3).Info("HTTP call done", "purpose", purpose, "method", method, "url", url, "httpCallStartTime", httpCallStartTime.Format(time.RFC3339Nano), "latencySecs", latencySecs, "statusCode", statusCode, "err", err)
```
rubambiza, line 2045: I like this centralization because it provides a single point of contract for consumers.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -53,17 +55,29 @@ const (
...
+// latencyHistograms needs values for labels purpose, method, status_code
+func NewLauncherClient(baseURL string, latencyHistograms prometheus.ObserverVec) (*LauncherClient, error) {
+	return newLauncherClient(baseURL, nil, latencyHistograms)
+}
+
+// newLauncherClient is a more general form of NewLauncherClient.
+// isclessLatencyHistograms  needs values for labels purpose, method, isc_name, and status_code.
+// latencyHistograms needs values for labels purpose, method, status_code.
+// latencyHistograms may be nil if isclessLatencyHistograms is not;
+// in this case the first method call will set latencyHistograms by currying isclessLatencyHistograms
```
rubambiza: There are two mentions of  "the first method call" in this sentence, but it's unclear what each is referring to because of what sounds like a chain of calls. At a high-level, I understand that `latencyHistograms` is getting some values from `isclessLatencyHistograms` based on the call of another method _beyond_ newLauncherClient, but I don't see any new code for that invocation nor a check for nil on `latencyHistograms`

docs/metrics.md
rubambiza: I don't think I got a satisfactory answer to the question.

rubambiza (changes requested):
Left some related comments with follow-on AI-assisted comments incoming. The tracking issue for the `llm-d-benchmark` changes has also been created based on this review 👇 

https://github.com/llm-d/llm-d-benchmark/issues/1547

pkg/controller/dual-pods/inference-server.go
rubambiza, line 2045: Centralizing the HTTP-call timing here is a nice cleanup — one line with `purpose`/`latencySecs`/`statusCode` beats the scattered per-call timestamps. One heads-up for a downstream consumer: the llm-d-benchmark DPC log parser derives FMA actuation latencies by reading `httpCallStartTime` off three milestone lines this PR strips it from — `Woke inference server` (now :1298), `Created vLLM instance` (:489), and `Successfully relayed the readiness` (:553).

The good news is the data isn't lost — it moves here. Because `process()` seeds the context logger with `requesterName` (inference-server.go:145), this line should inherit `requesterName`, and the `purpose` token already distinguishes the paths the parser needs: `"wake"` (hot), `"create_instance"` (warm), and `"relay_readiness"` (the shared end-time). If you can confirm `requesterName` + `purpose` are guaranteed on this line, the parser can adopt it as the new contract. A one-line code comment noting this line is consumed for benchmark timing would help prevent an accidental future regression.

_(AI-assisted review comment.)_

rubambiza, line 2045: This line is at `V(3)`, while the milestone messages it absorbs the timing from stay at `V(2)` (`Created vLLM instance`:489, `Successfully relayed the readiness`:553, `Woke inference server`:1298). The FMA chart runs the DPC at `-v=5` today, so it's emitted in our benchmark runs — but that's a fragile coupling: anyone running the controller at V(2) (a reasonable default that still emits the actuation milestones) would silently lose all per-request HTTP timing. Would you consider emitting `HTTP call done` at `V(2)` to match the milestones it now carries timing for, or documenting >= V(3) as the required verbosity for HTTP-call observability? (Tradeoff: V(2) raises default volume, since every HTTP call — including GC/sync chatter — would log.)

_(AI-assisted review comment.)_

pkg/controller/dual-pods/inference-server.go
```diff
@@ -553,7 +550,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 					requestingPod.Annotations[api.InferenceServerConfigAnnotationName],
 				).Observe(actuationSecs)
 			}
-			logger.V(2).Info("Successfully relayed the readiness", "name", providingPod.Name, "readiness", readiness, "url", url, "httpCallStartTime", postStartStr)
+			logger.V(2).Info("Successfully relayed the readiness", "name", providingPod.Name, "readiness", readiness, "url", url)
```
rubambiza, line 552: Flagging this one specifically because it's the shared *end-time* anchor for all three actuation latencies (hot/warm/cold-launcher) in the benchmark parser — it loses `httpCallStartTime` here. The `readiness="ready"` field the parser also keys on does survive on this line, so correlation still works once the parser switches to the centralized `HTTP call done` line (`purpose="relay_readiness"`) for the timestamp. No change requested beyond what's in the comment on the `HTTP call done` line — just noting that this single removal is what breaks all three metrics if the centralized line isn't adopted/guaranteed.

_(AI-assisted review comment.)_

docs/metrics.md
```diff
@@ -124,17 +124,20 @@ Labels are as follows.
...
-### fma_wake_seconds
+### fma_http_latency_seconds
 
-Vector of histograms: Latency of `/wake_up` call from DPC to vllm.
+Vector of histograms: Latency of HTTP calls. Currently, only calls
+from the dual-pods controller are instrumented.
 
 Labels are as follows.
 
+- `purpose`: short token identifying the purpose of the call
```
rubambiza, line 134: Since this log line is effectively a consumed interface now (benchmark timing, not just the Prometheus histogram), would you enumerate the `purpose` token values here — `wake`, `create_instance`, `relay_readiness`, `sleep`, `query_sleeping`, `get_accel_memory_usage`, `get_gpu_uuids`, plus the launcher-client ones (`list_instances`, `delete_instance`, …)? Giving the downstream parser a documented set of tokens to target (rather than reverse-engineering them from source) makes this a stable contract.

_(AI-assisted review comment.)_

pkg/controller/dual-pods/inference-server.go
```diff
@@ -2046,47 +1987,72 @@ func init() {
...
+// semantically (but not in the type system) specialized to
+// needing values for three labels: purpose, method, status_code.
+type ObserverCube interface {
+	// WithLabelValues takes purpose, method, and status_code
+	WithLabelValues(values ...string) prometheus.Observer
+}
+
+// Do an HTTP call.
+// latencyHistogramVec needs values for labels purpose, method, status_code
+func doHTTP(ctx context.Context, purpose, method, url string, latencyHistogramVec ObserverCube, requestData, resultData any) (int, error) {
```
dumb0002, line 2003: A minor suggestion: Add `httpCallStartTime` to the wrapped error in `doHTTP` so operators reading `dual-pods.llm-d.ai/status` annotations have it without raising verbosity.

pkg/controller/dual-pods/inference-server.go
rubambiza: This has been done. Resolving.

docs/metrics.md
MikeSpreitzer: > Just confirming here, the otherwise part covers both successful (e.g., 200s) and erroneous HTTP responses (picking 400s for the sake of discussion), right?

Correct.

0 is not a valid HTTP status code (do not conflate these with POSIX process exit codes).

docs/metrics.md
MikeSpreitzer: > Were all these lines removed in favor of the centralized calls?

Yes. This PR deliberately removes two purpose-specific HTTP latency metrics and adds one all-purpose HTTP latency metric.

The new metric lines up with the new logging.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: > are we at a point where we can now remove the references to sleeperLimit in favor of maxInstances?

No, that comes with removing milestone 2 behavior. We have not put that in the scope of release 0.6.0.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: (A) Why would readers of `dual-pods.llm-d.ai/status` need that?

(B) If it changes on every attempted call, that requires a status update on every such HTTP call. This sounds like a bad idea.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: V(2) is definitely too high for every HTTP call. Even V(3) is not really appropriate. For comparison, the Kubernetes client libraries start logging every HTTP call at V(6), IIRC. I think that perhaps V(5) would be right here.

Where would be the right place to document this?

MikeSpreitzer:
The force-push above is a rebase onto `main` plus adding a commit with changes in response to review.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: Revised.

pkg/controller/dual-pods/launcherclient.go
MikeSpreitzer: Revised to clarify.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: I added a comment where the `requesterName` logger attribute is set. The need for inclusion of `purpose` is not subtle.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: The latest revision splits `relay_readiness` into `relay_ready` and `relay_unready`.

docs/metrics.md
MikeSpreitzer: done

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -53,17 +55,30 @@ const (
 	InstanceStatusStopped = "stopped"
 )
 
-func NewLauncherClient(baseURL string) (*LauncherClient, error) {
+// Make a new one.
+// latencyHistograms needs values for labels purpose, method, status_code
+func NewLauncherClient(baseURL string, latencyHistograms prometheus.ObserverVec) (*LauncherClient, error) {
+	return newLauncherClient(baseURL, nil, latencyHistograms)
+}
```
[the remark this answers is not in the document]
MikeSpreitzer: There is not a problem here. `fullDo` relies on `complete()` to set `latencyHistograms`.

pkg/controller/dual-pods/launcherclient.go
```diff
@@ -53,17 +55,30 @@ const (
...
+// latencyHistograms may be nil if isclessLatencyHistograms is not;
+// in this case the first method call to this LauncherClient will set latencyHistograms
+// by currying isclessLatencyHistograms
+// with a value for the `isc_name` label, using the returned value in `GetInstanceState`
+// if that is the first method called and it succeeds and otherwise using the empty string.
+func newLauncherClient(baseURL string, isclessLatencyHistograms, latencyHistograms prometheus.ObserverVec) (*LauncherClient, error) {
 	parsedURL, err := url.Parse(baseURL)
 	if err != nil {
 		return nil, fmt.Errorf("invalid base URL: %w", err)
 	}
```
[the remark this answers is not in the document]
MikeSpreitzer: This is not really needed, this is a package-internal matter. The comment on the non-public constructor is good enough.

pkg/controller/dual-pods/inference-server.go
dumb0002: Ok, I see! I missed to take in consideration the constant status update as stated in (B).

docs/metrics.md
rubambiza: Thanks for the clarification.

pkg/controller/dual-pods/launcherclient.go
MikeSpreitzer: I have revised, to say `default` instead of `case true`.

pkg/controller/dual-pods/inference-server.go
rubambiza: V(5) works well since the FMA Helm chart already runs the DPC at -v=5, which works for the benchmark without any changes. Agreed on V(2)/V(3) being too low.

On where to document: Thanks for listing the purpose token values in `docs/metrics.md`, which is what the log consumer needs. The one thing missing there is a note on verbosity. Something in that same section saying HTTP call done is emitted at V(5) (and that this line is now where log-based timing should read HTTP start times, not the individual milestone messages) would close the loop. A one-line pointer comment at the call site referencing that doc would help anyone who later changes the level (and fingers crossed no one deletes `metrics.md` without checking).

rubambiza (approved):

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: The `docs/metrics.md` file is about Prometheus metrics, it is not really the right place to document logging behavior in general. OTOH, it would not be entirely inappropriate to mention in the documentation of a metric that there is associated logging. Since I do not see a better place right now, I think that I will do that.

MikeSpreitzer:
I will make the logging change discussed above in a follow-on PR.

PR #600, opened 2026-06-26
Add release 0.6.1 to menu in bug template

MikeSpreitzer:
That release does not exist yet, but it surely will exist and probably quite soon.


waltforme (approved):
LGTM

PR #608, opened 2026-06-29
Use fixed 5s polling for /is_sleeping instead of exponential backoff

### LINKED ISSUE #455: [Feature]: Use fixed polling interval for vLLM instance readiness
### Feature Area

Other

### Problem Statement

While a vLLM instance starts up, the dual-pods controller polls `/is_sleeping` through the work queue's exponential backoff (5ms growing to 20s). Early retries are wastefully frequent; later ones add unnecessary latency. A fixed ~5s interval would be more appropriate.                                                                                                                      

From https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/443#issuecomment-4310725850 (point 4).

### Proposed Solution

TBD

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

omerap12:
Fixes: #455 

Added a retryAfter `time.Duration` return value to the process function signature.
When `retryAfter > 0`, `processNextWorkItem` uses `AddAfter` (fixed delay) instead of `AddRateLimited` (exponential backoff). The `querySleeping` call site uses this to poll at a steady 5s interval during startup.

For `nodeItem.process`, which aggregates multiple local items, the maximum `retryAfter` across all items is propagated - this avoids the conflict between items wanting different retry policies while keeping the change minimal.

All other call sites return retryAfter: 0, preserving existing exponential backoff behavior.

### DIFF
--- a/pkg/controller/generic/queue-work.go
+++ b/pkg/controller/generic/queue-work.go
@@ -36,18 +36,18 @@ type QueueAndWorkers[Item comparable] struct {
 	ControllerName string
 	Queue          workqueue.TypedRateLimitingInterface[Item]
 	NumWorkers     int
-	Process        func(ctx context.Context, item Item) (err error, retry bool)
+	Process        func(ctx context.Context, item Item) (err error, retry bool, retryAfter time.Duration)
 
 	earlySync func(context.Context, Item) *bool
 }
 
 // NewQueueAndWorkers makes a new QueueAndWorkers.
-// Iff `process` returns `retry==true` then the item will be
-// requeued for retry.
+// Iff `process` returns `retry==true` then the item will be requeued for retry.
+// If retryAfter > 0, the retry uses a fixed delay instead of exponential backoff.
 func NewQueueAndWorkers[Item comparable](
 	controllerName string,
 	numWorkers int,
-	process func(ctx context.Context, item Item) (err error, retry bool),
+	process func(ctx context.Context, item Item) (err error, retry bool, retryAfter time.Duration),
 ) QueueAndWorkers[Item] {
 	return newQueueAndWorkers(controllerName, numWorkers, process, noEarlySync[Item])
 }
@@ -60,7 +60,7 @@ func noEarlySync[Item comparable](context.Context, Item) *bool {
 func newQueueAndWorkers[Item comparable](
 	controllerName string,
 	numWorkers int,
-	process func(ctx context.Context, item Item) (err error, retry bool),
+	process func(ctx context.Context, item Item) (err error, retry bool, retryAfter time.Duration),
 	earlySync func(context.Context, Item) *bool,
 ) QueueAndWorkers[Item] {
 	ans := QueueAndWorkers[Item]{
@@ -116,12 +116,17 @@ func (ctl *QueueAndWorkers[Item]) processNextWorkItem(ctx context.Context) bool 
 	}
 	var err error
 	var retry bool
+	var retryAfter time.Duration
 	defer func() {
 		if err == nil {
 			if retry {
-				// Nothing went wrong but this item needs to be requeued for reprocessing.
-				ctl.Queue.AddRateLimited(objRef)
-				logger.V(4).Info("Processed workqueue item successfully, requeued for follow-up.", "item", objRef)
+				if retryAfter > 0 {
+					ctl.Queue.AddAfter(objRef, retryAfter)
+					logger.V(4).Info("Processed workqueue item successfully, requeued after fixed delay.", "item", objRef, "retryAfter", retryAfter)
+				} else {
+					ctl.Queue.AddRateLimited(objRef)
+					logger.V(4).Info("Processed workqueue item successfully, requeued for follow-up.", "item", objRef)
+				}
 			} else {
 				// If no error occurs we Forget this item so it does not
 				// get queued again until another change happens.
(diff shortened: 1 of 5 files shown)


pkg/controller/generic/queue-work.go
```diff
@@ -36,18 +36,18 @@ type QueueAndWorkers[Item comparable] struct {
...
 // NewQueueAndWorkers makes a new QueueAndWorkers.
-// Iff `process` returns `retry==true` then the item will be
-// requeued for retry.
+// Iff `process` returns `retry==true` then the item will be requeued for retry.
+// If retryAfter > 0, the retry uses a fixed delay instead of exponential backoff.
 func NewQueueAndWorkers[Item comparable](
 	controllerName string,
 	numWorkers int,
-	process func(ctx context.Context, item Item) (err error, retry bool),
+	process func(ctx context.Context, item Item) (err error, retry bool, retryAfter time.Duration),
```
MikeSpreitzer, line 50: In my opinion, it would be more tasteful to define a single value that conveys the information in both `retry bool` and `retryAfter time.Duration`. Perhaps pack them into a struct. Perhaps define a special value for time.Duration that means "use exponential backoff".

omerap12: Good idea. we can then expand this struct as we see fit if needed. will do.

MikeSpreitzer:
The design here has some problems.

Specified non-zero delays should be combined by MIN rather than MAX. Processing an item later than requested is not honoring the shorter request. It can make latency worse, and this project is all about latency. In contrast, processing an item early or an extra time is safe as far as functionality and latency are concerned; the only down side is possibly-unnecessary work, which is an important but secondary concern.

More importantly, the design mixes the exponential backoff items with the specified-delay items. Again, this can delay the exponential backoff items well beyond how they would otherwise be treated --- as well as accelerate them compared to current treatment.

To be fair, the existing design already has a similar flaw. It puts the same timing on all of a given Node's work queue items. The design here continues in that vein. Rather than make this problem worse, perhaps it is time to take a different approach (which solves that problem).

omerap12:
> The design here has some problems.
> 
> Specified non-zero delays should be combined by MIN rather than MAX. Processing an item later than requested is not honoring the shorter request. It can make latency worse, and this project is all about latency. In contrast, processing an item early or an extra time is safe as far as functionality and latency are concerned; the only down side is possibly-unnecessary work, which is an important but secondary concern.
> 
> More importantly, the design mixes the exponential backoff items with the specified-delay items. Again, this can delay the exponential backoff items well beyond how they would otherwise be treated --- as well as accelerate them compared to current treatment.
> 
> To be fair, the existing design already has a similar flaw. It puts the same timing on all of a given Node's work queue items. The design here continues in that vein. Rather than make this problem worse, perhaps it is time to take a different approach (which solves that problem).

Yes. this is the solution I could come up with without changing the all architecture. 
Another option that I thought that we can take (at the moment at least) is to do something like that:
```diff
diff --git a/pkg/controller/dual-pods/inference-server.go b/pkg/controller/dual-pods/inference-server.go
index c6918ad..7ec7822 100644
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -491,6 +491,14 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		}
  		if serverDat.Sleeping == nil {
 			sleeping, err := ctl.querySleeping(ctx, iscName, providingPod, serverPort)
 			if err != nil {
-				return err, true
+				if isConnectionRefused(err) {
+					nodeDat.add(item)
+					ctl.Queue.AddAfter(nodeItem{nodeDat.NodeName}, 5*time.Second)
+					return nil, false
+				}
+				return err, true
 			}
```
We would break the entire flow but that's a quick solution that won't require refactoring.

MikeSpreitzer:
> Another option that I thought that we can take (at the moment at least) is ...

That is really fragile (makes assumption about `err`) and hacky (sneaks retry information into `err`), and has the same problem of giving the same timing to every work queue item of a given Node.

Rather than make a hack and a fragile assumption about `err`, I think that it is better to change the signature of `itemOnNode.process` so that it can request requeuing after a specific period of time. As you noticed, the controller's queue is available for `AddAfter` calls. Also, it is not intolerable to expand the signature of the `process` function of the generic controllers.

To fix the fundamental problem of giving the same timing to all of a Node's work queue items, I think that a fairly localized change would do the trick. It would involve changing `nodeData` so that rather than having a set of `itemOnNode` it would instead have a priority queue of them, ordered by the time to process the item (or maybe just an unordered slice of item + time-to-start-processing, for simplicity; the number of items on a given Node should not grow very large). `nodeItem.process` could process the ones for which the time has come, and use the controller queue's `AddAfter` to schedule the next call to `nodeItem.process`.

omerap12:
I started implementing the approach you described - localQueue as a slice of scheduledItem (item + addTime + processAfter), with takeReadyItems(now) that only returns items whose processAfter <= now, and nodeItem.process using AddAfter to schedule its next wake-up based on the earliest pending item.

One issue I ran into: when nodeItem.process handles re-enqueuing itself via ctl.Queue.AddAfter(ni, delay) and returns (nil, false), the generic controller calls Forget(ni) - (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/main/pkg/controller/generic/queue-work.go#L128), which resets the rate limiter's backoff counter for that nodeItem. This means items that currently rely on exponential backoff (all the return err, true sites in infSvrItem.process) would lose that behavior - every retry would start from the base delay since the counter gets reset each time.

It seems like once we move retry scheduling into the per-item local queue, backoff tracking needs to move there too.
The generic controller's AddRateLimited would no longer be used for nodeItem at all; it just provides the worker pool and queue infrastructure.

MikeSpreitzer:
> It seems like once we move retry scheduling into the per-item local queue, backoff tracking needs to move there too.
> The generic controller's AddRateLimited would no longer be used for nodeItem at all; it just provides the worker pool and queue infrastructure.

Yes, I think that's right. There would be a rate limiter in `nodeData`, and `nodeItem.process` would use it to find the next time to process an `itemOnNode`; `nodeItem.process` would, once it is done with all the current `itemOnNode`s, find the nearest time to do more and requeue the `nodeItem` with `AddAfter`.

omerap12:
I have opened: #626 , PTAL

omerap12:
closed due to #626

PR #618, opened 2026-06-29
Update docs/upstream-versions.md

MikeSpreitzer:
To track recent developments.

Why are we maintaining redundant information, and why by hand?


diegocastanibm (approved):
LGTM

PR #620, opened 2026-06-29
Update Go tools used to 1.26

MikeSpreitzer:
Also, updated the version of golangci-lint used (to the current latest).

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
-          go-version: '1.24'
+          go-version: '1.25'
 
       - name: check go mod tidy
         run: |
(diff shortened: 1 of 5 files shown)


MikeSpreitzer:
Hold for vulnerability soak period for golangci/golangci-lint-action

MikeSpreitzer:
Actually, let's not hold this on up that one action. It can advance further in another PR.

.github/workflows/golangci-lint.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
-          go-version: '1.24'
+          go-version: '1.25'
```
waltforme, line 18: I see this one is the only one that uses 1.25 (others all 1.26). Is the choice intentional?

MikeSpreitzer: Yes. [v9.2.1](https://github.com/golangci/golangci-lint-action/releases/tag/v9.2.1) only supports Go 1.25, and v9.3.0 is only 6 hours old so we should wait a while before using that.

waltforme: I see. Thanks.

waltforme (approved):
LGTM

PR #622, opened 2026-06-30
Add 0.6.2 to release menu in bug template

MikeSpreitzer:
.. because I expect there will be another patch release.


diegocastanibm (approved):
LGTM

Issue #623, opened 2026-06-30
E2E test on OpenShift is flaky wrt Node ephemeral storage pressure


MikeSpreitzer:
I ran the test on commit 2102520fcf122fd62fa5c7b06af8ece23d3c5249 (which I later tagged as v0.6.1), see https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/28406977943 . The first four attempts failed because eventually Pod(s) got evicted due to ephemeral storage pressure on the Node (they all latched onto the same Node); the 5th attempt latched onto a different Node and succeeded.

Perhaps the probe for 2 GPUs should also probe for ephemeral storage availability.

Perhaps our instructions for users should give some guidance on the ephemeral storage resource needs (https://github.com/kubernetes/api/blob/v0.36.2/core/v1/types.go#L7011).

MikeSpreitzer:
I examined a launcher running one vLLM instance, serving https://huggingface.co/HuggingFaceTB/SmolLM2-360M-Instruct .
The leading uses of ephemeral storage were as follows.

- 694 MB in the HF model cache (`/tmp/hub/models--HuggingFaceTB--SmolLM2-360M-Instruct`)
- 22 MB in the torch compile cache (`/tmp/torch_compile_cache`)
- 8.5 MB in compiled CUDA kernels (directories named things like `/tmp/DANMZW4UDRA4KB54WMBGFJ5C5U4UWMIBQMHNPEPRSEZPYOFHUQCA`)

Following is what the launcher reports about that instance.

```
    {
      "status": "running",
      "instance_id": "I5YIweV9wub4llRQEfgFRXM4kHwukoCNDnRj2X65h7Fwi",
      "revision": 1,
      "options": "--model HuggingFaceTB/SmolLM2-360M-Instruct --enable-sleep-mode --port 8005",
      "gpu_uuids": [
        "GPU-344251ca-5d67-4e07-8e35-1aa7b2b4a8e2"
      ],
      "env_vars": {
        "VLLM_LOGGING_LEVEL": "DEBUG",
        "VLLM_SERVER_DEV_MODE": "1",
        "VLLM_USE_V1": "1",
        "CUDA_VISIBLE_DEVICES": "0"
      },
      "annotations": {
        "inference-port": "8005",
        "isc-name": "inference-server-config-smol-25-14-57-37"
      }
    }
```

PR #624, opened 2026-06-30
Care for ephemeral storage, spare GPU memory

MikeSpreitzer:
Make E2E probe Pod test for ephemeral storage capacity.

Make E2E LauncherConfigs specify ephemeral storage usage.

Make E2E InferenceServerConfigs specify --gpu-memory-utilization .

This is part of addressing #623

### DIFF
--- a/test/e2e/mkobjs-openshift.sh
+++ b/test/e2e/mkobjs-openshift.sh
@@ -112,7 +112,10 @@ metadata:
 spec:
   modelServerConfig:
     port: 8005
-    options: "--model HuggingFaceTB/SmolLM2-360M-Instruct --enable-sleep-mode"
+    options: >-
+      --model HuggingFaceTB/SmolLM2-360M-Instruct
+      --enable-sleep-mode
+      --gpu-memory-utilization 0.85
     env_vars:
       VLLM_SERVER_DEV_MODE: "1"
       VLLM_LOGGING_LEVEL: "DEBUG"
@@ -131,7 +134,10 @@ metadata:
 spec:
   modelServerConfig:
     port: 8006
-    options: "--model Qwen/Qwen2.5-0.5B-Instruct --enable-sleep-mode"
+    options: >-
+      --model Qwen/Qwen2.5-0.5B-Instruct
+      --enable-sleep-mode
+      --gpu-memory-utilization 0.85
     env_vars:
       VLLM_SERVER_DEV_MODE: "1"
       VLLM_LOGGING_LEVEL: "DEBUG"
@@ -150,7 +156,10 @@ metadata:
 spec:
   modelServerConfig:
     port: 8007
-    options: "--model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --enable-sleep-mode"
+    options: >-
+      --model TinyLlama/TinyLlama-1.1B-Chat-v1.0
+      --enable-sleep-mode
+      --gpu-memory-utilization 0.85
     env_vars:
       VLLM_SERVER_DEV_MODE: "1"
       VLLM_LOGGING_LEVEL: "DEBUG"
@@ -199,6 +208,9 @@ spec:
             value: "/tmp"
           - name: XDG_CONFIG_HOME
             value: "/tmp"
+          resources:
+            limits:
+              ephemeral-storage: "2Gi"
 ---
 apiVersion: fma.llm-d.ai/v1alpha1
 kind: LauncherPopulationPolicy
--- a/test/e2e/mkobjs-openshift.sh
+++ b/test/e2e/mkobjs-openshift.sh
@@ -210,7 +210,7 @@ spec:
             value: "/tmp"
           resources:
             limits:
-              ephemeral-storage: "2Gi"
+              ephemeral-storage: "4.5Gi"
 ---
 apiVersion: fma.llm-d.ai/v1alpha1
 kind: LauncherPopulationPolicy
--- a/test/e2e/test-cases.sh
+++ b/test/e2e/test-cases.sh
@@ -158,7 +158,7 @@ spec:
     resources:
       limits:
         nvidia.com/gpu: "2"
-        ephemeral-storage: "4Gi"
+        ephemeral-storage: "9Gi"
   terminationGracePeriodSeconds: 0
 PROBE
 
@@ -252,6 +252,9 @@ assigned_gpu_uuids=$(kubectl get pod "$req1" -n "$NS" -o jsonpath='{.metadata.an
 echo "Assigned GPU UUID(s): $assigned_gpu_uuids"
 
 cheer Successful launcher-based pod creation
+
+# vllm inventory:
+# SmolLM2-360M-Instruct awake on $launcher1
 
 # ---------------------------------------------------------------------------
 # Test enforcement of LPP.Spec.CountForLauncher key uniqueness
@@ -388,6 +391,9 @@ else
 
 fi # available_gpus check
 
+# vllm inventory:
+# SmolLM2-360M-Instruct awake on $launcher1
+
 # ---------------------------------------------------------------------------
 # Instance Wake-up Fast Path
 # ---------------------------------------------------------------------------
@@ -448,6 +454,9 @@ check_gpu_pin $req2
 
 cheer Successful instance wake-up fast path
 
+# vllm inventory:
+# SmolLM2-360M-Instruct awake on $launcher1 (AKA $launcher2)
+
 # ---------------------------------------------------------------------------
 # Multiple Instances Share One Launcher
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


MikeSpreitzer:
With this in place, while testing on the vllm-d cluster, and investigating with `nvidia-smi` in a launcher with one vllm instance (HuggingFaceTB/SmolLM2-360M-Instruct), I found

```
+-----------------------------------------+------------------------+----------------------+
|   4  NVIDIA A100-SXM4-80GB          On  |   00000000:0C:05.0 Off |                    0 |
| N/A   33C    P0             89W /  400W |   69579MiB /  81920MiB |      0%      Default |
```

with the instance awake and

```
+-----------------------------------------+------------------------+----------------------+
|   4  NVIDIA A100-SXM4-80GB          On  |   00000000:0C:05.0 Off |                    0 |
| N/A   33C    P0             89W /  400W |    1165MiB /  81920MiB |      0%      Default |
```

with the instance sleeping.

test/e2e/mkobjs-openshift.sh
```diff
@@ -112,7 +112,10 @@ metadata:
 spec:
   modelServerConfig:
     port: 8005
-    options: "--model HuggingFaceTB/SmolLM2-360M-Instruct --enable-sleep-mode"
+    options: >
+      --model HuggingFaceTB/SmolLM2-360M-Instruct
+      --enable-sleep-mode
+      --gpu-memory-utilization 0.85
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

waltforme (dismissed):
LGTM

waltforme:
Actually there was something wrong with a run of 'CI - OpenShift E2E Tests' https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/28477035158. We need to look into that before merging.

MikeSpreitzer:
The relevant launcher crashed during the controller restart test.

```yaml
- apiVersion: v1
  kind: Pod
  metadata:
...
    name: launcher-30-21-34-19-8q8dv
...
  status:
    conditions:
...
    - lastProbeTime: null
      lastTransitionTime: "2026-06-30T21:42:39Z"
      reason: PodFailed
      status: "False"
      type: Ready
...
    containerStatuses:
    - containerID: cri-o://4e4e501058f30e88afd040c2b3dbc78aef3d47d9a7b1b71ebc32d7bae01a4084
...
      lastState: {}
      name: inference-server
      ready: false
      restartCount: 0
      started: false
      state:
        terminated:
          containerID: cri-o://4e4e501058f30e88afd040c2b3dbc78aef3d47d9a7b1b71ebc32d7bae01a4084
          exitCode: 137
          finishedAt: "2026-06-30T21:42:39Z"
          reason: Error
          startedAt: "2026-06-30T21:34:22Z"
```

MikeSpreitzer:
The event listing shows why:

```
2026-06-30T21:43:09.1618506Z 33s         Warning   Evicted             pod/launcher-30-21-34-19-8q8dv                                  Pod ephemeral local storage usage exceeds the total limit of containers 2Gi.
```

MikeSpreitzer:
The "Controller Restart State Recovery" test case starts with all three models in the HF cache in the launcher's filesystem. They are as follows, with size calculated as (number of parameters) X (parameter size).

- SmolLM2-360M-Instruct: 0.8 GB
- Qwen2.5-0.5B-Instruct: 1 GB
- TinyLlama-1.1B-Chat-v1.0: 2 GB

For an ephemeral storage cost accounting that covers more than the model parameters, see https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/623#issuecomment-4850330388

test/e2e/test-cases.sh
```diff
@@ -850,9 +892,12 @@ check_gpu_pin $req_recovered
 
 cheer Successful stopped instance recovery
 
+# vllm inventory:
+# Qwen2.5-0.5B-Instruct awake on $launcher_after_delete
+
 cheer All launcher-based tests passed
 
-echo "Saving a final scrape of the dual-pods controller; expect port-forward and curl noise"
+note "Saving a final scrape of the dual-pods controller; expect port-forward and curl noise"
```
diegocastanibm, line 900: `note` is a no built-in command. What is the reason to use it instead of `echo`?

MikeSpreitzer: Indeed `note` is not a shell built-in. It is a function defined earlier in the script. It adds a timestamp to the start of the output. It could be perhaps used instead of "echo" or "#" in more places, but I will leave that to another PR.

diegocastanibm: Ah, I see. Got it. I opened the file to see the whole picture (not just the changes) and I saw the function. One more thing that I observed is that sometimes you use `note` without quotes. While the unquoted calls happen to produce the same output today (since `note` uses `$*`),I think it would be better to use it as `note "..."` for consistency and also robustness in case you have special characters. The unquoted form only works while the message contains no shell-special characters.

test/e2e/test-cases.sh
```diff
@@ -663,9 +686,10 @@ launcher_instances_after=$(kubectl exec -n "$NS" $launcher1 -- python3 -c 'impor
 echo "Launcher has $launcher_instances_after instances after controller restart"
 [ "$launcher_instances_after" == "$launcher_instances_before" ]
 
-# Now scale up requester - controller should correctly select the launcher with sleeping instance
-# Use isc2 which should have a sleeping instance from before
+# Set requester RS to use isc2, which should have a sleeping instance from before
 kubectl patch rs $rs -n "$NS" --type=json -p='[{"op": "replace", "path": "/spec/template/metadata/annotations/dual-pods.llm-d.ai~1inference-server-config", "value": "'$isc2'"}]'
+
+note Now scale up requester RS - controller should correctly select the launcher with sleeping instance
```
diegocastanibm, line 692: Per my comment [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/624#discussion_r3507022728)
```suggestion
note "Now scale up requester RS - controller should correctly select the launcher with sleeping instance"
```

test/e2e/test-cases.sh
MikeSpreitzer: This seems like unnecessary complexity to me. We are talking about a literal command line that is right in front of the author's eyes, no "maybe" about whether it contains any special characters. Would you have us quote every argument to every command in case it changes in the future?

test/e2e/test-cases.sh
MikeSpreitzer: Per my comment above, I think that this is unnecessary complexity but otherwise harmless.

test/e2e/test-cases.sh
diegocastanibm: Sorry to insist on this, but I think it is healthy to have some debate. 
I think you have a fair point.  I agree we shouldn't quote literals defensively "just in case", and for a plain text message that's visible right here there's no real risk. I'll drop that part.
However, one case I'd still flag (non-blocking) is the calls that interpolate a variable.
For example `note Wait for $collision_req to be ready`. That's not a hypothetical future change, and `shellcheck` (SC2086) flags it (as you can see below). 
```
diego@macbookpro ~/Documents/my_stuff/llm-d-fast-model-actuation (main) $ shellcheck --include=SC2086 test/e2e/test-cases.sh | grep -B1 -A4 'note Wait'
In test/e2e/test-cases.sh line 370:
    note Wait for $collision_req to be ready
                  ^------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
    note Wait for "$collision_req" to be ready


In test/e2e/test-cases.sh line 371:
    kubectl wait --for condition=Ready pod/$collision_req -n "$NS" --timeout=300s
--
In test/e2e/test-cases.sh line 372:
    note Wait for $collision_launcher to be ready
                  ^-----------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
    note Wait for "$collision_launcher" to be ready


In test/e2e/test-cases.sh line 373:
    kubectl wait --for condition=Ready pod/$collision_launcher -n "$NS" --timeout=0s
--
In test/e2e/test-cases.sh line 594:
note Wait for $req_reclaim and $launcher1 to be ready
              ^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                               ^--------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
note Wait for "$req_reclaim" and "$launcher1" to be ready

```
Quoting just those would cover the only spot with some risk. Happy to leave it as a nit either way (and I approve the PR, but I want to make this suggestion here).

diegocastanibm (approved):
Nit comment:
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/624#discussion_r3507356921

test/e2e/test-cases.sh
MikeSpreitzer: We know that Pod names do not contain any shell metacharacters, so that checker is being overly cautious IMO. However, that excess is beyond our contrel --- except by writing something explicit to cope. Using the dquotes instead will be simpler/better than that. I will address this in a follow-on PR.

PR #626, opened 2026-07-01
Independent item scheduling in DPC, 5-second polling of /is_sleeping

omerap12:
- Add per-item retry scheduling with exponential backoff to the node-local work queue
- Replace the yank-all-and-retry model with a time-aware local queue (`takeReadyItems` / `addAfter` / `earliestPending`) that supports per-item exponential backoff via a rate limiter and fixed-delay retries via `retryAfter`. Items that aren't ready yet stay in the queue rather than being re-processed immediately. 
- The process return type is unified into a processOutput struct carrying err, retry, and retryAfter.

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"maps"
 	"math"
 	"reflect"
 	"strconv"
@@ -41,6 +40,7 @@ import (
 	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 	kubemetrics "k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
@@ -121,6 +121,7 @@ const ControllerName = "dual-pods-controller"
 const GPUMapName = "gpu-map"
 
 const GPUIndexName = "gpu"
+const controllerQueuePerItemRetryMaxDelay = 20 * time.Second
 
 func GPUIndexFunc(obj any) ([]string, error) {
 	pod := obj.(*corev1.Pod)
@@ -411,14 +412,32 @@ type nodeData struct {
 	// and for each the time at which it was first injected into this set.
 	// Hold ItemsMutex while accessing this.
 	LocalQueue localQueue
+
+	rateLimiter workqueue.TypedRateLimiter[itemOnNode]
 }
 
-type localQueue map[itemOnNode]time.Time
+type localQueue map[itemOnNode]*scheduledItem
 
 type itemOnNode interface {
 	// process returns (err error, retry bool).
 	// There will be a retry iff `retry`.
-	process(ctx context.Context, ctl *controller, nodeDat *nodeData) (error, bool)
+	process(ctx context.Context, ctl *controller, nodeDat *nodeData) processResult
+}
+
+type scheduledItem struct {
+	addTime      time.Time // when first enqueued (for queue duration metrics)
+	processAfter time.Time // earliest time to process this item
+}
+
+// processResult is returned by itemOnNode.process to signal outcome and retry intent.
+//   - err may be set with or without retry; a non-nil err with retry=false is a permanent failure.
+//   - retry requests that the item be re-queued for later processing.
+//   - retryAfter, if not zero, specifies the delay until the later processing,
+//     with zero meaning to use exponential backoff.
+type processResult struct {
+	err        error
+	retry      bool
+	retryAfter time.Duration
 }
 
 // Internal state about an inference server
@@ -992,9 +1011,13 @@ func (ctl *controller) getNodeData(nodeName string) *nodeData {
 	if ans == nil {
 		ans = &nodeData{
 			NodeName:         nodeName,
-			LocalQueue:       make(localQueue),
+			LocalQueue:       localQueue{},
 			InferenceServers: make(map[apitypes.UID]*serverData),
 			Launchers:        make(map[string]*launcherData),
+			rateLimiter: workqueue.NewTypedWithMaxWaitRateLimiter(
+				workqueue.DefaultTypedControllerRateLimiter[itemOnNode](),
+				controllerQueuePerItemRetryMaxDelay, // 20s, from queue-work.go
+			),
 		}
 		ctl.nodeNameToData[nodeName] = ans
 	}
@@ -1004,22 +1027,61 @@ func (ctl *controller) getNodeData(nodeName string) *nodeData {
 func (nodeDat *nodeData) add(item itemOnNode) {
 	nodeDat.ItemsMutex.Lock()
 	defer nodeDat.ItemsMutex.Unlock()
-	if _, found := nodeDat.LocalQueue[item]; !found {
-		nodeDat.LocalQueue[item] = time.Now()
-		addsCounters.WithLabelValues(nodeDat.NodeName).Inc()
-		queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(float64(len(nodeDat.LocalQueue)))
+	if _, exist := nodeDat.LocalQueue[item]; exist {
+		return // already queued
 	}
+	now := time.Now()
+	nodeDat.LocalQueue[item] = &scheduledItem{
+		addTime: now, processAfter: now,
+	}
+	addsCounters.WithLabelValues(nodeDat.NodeName).Inc()
+	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(float64(len(nodeDat.LocalQueue)))
 }
 
-// yankItems returns the currently queued items and empties the queue.
-// Caller can access the returned value without synchronization.
-func (nodeDat *nodeData) yankItems() localQueue {
+func (nodeDat *nodeData) addAfter(item itemOnNode, after time.Time) {
 	nodeDat.ItemsMutex.Lock()
 	defer nodeDat.ItemsMutex.Unlock()
-	ans := maps.Clone(nodeDat.LocalQueue)
-	nodeDat.LocalQueue = make(localQueue)
-	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(0)
-	return ans
+	if cur, exist := nodeDat.LocalQueue[item]; exist {
+		if after.Before(cur.processAfter) {
+			cur.processAfter = after
+		}
+		return
+	}
+	nodeDat.LocalQueue[item] = &scheduledItem{
+		addTime: time.Now(), processAfter: after,
+	}
+	addsCounters.WithLabelValues(nodeDat.NodeName).Inc()
+	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(float64(len(nodeDat.LocalQueue)))
+}
+
+// takeReadyItems removes and returns items whose processAfter <= now.
+// Items not yet ready remain in the queue.
+func (nodeDat *nodeData) takeReadyItems(now time.Time) localQueue {
+	nodeDat.ItemsMutex.Lock()
+	defer nodeDat.ItemsMutex.Unlock()
+	ready := make(localQueue)
+	for item, scheduledItem := range nodeDat.LocalQueue {
+		if !scheduledItem.processAfter.After(now) {
+			ready[item] = scheduledItem
+			delete(nodeDat.LocalQueue, item)
+		}
+	}
+	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(float64(len(nodeDat.LocalQueue)))
+	return ready
+}
+
+// earliestPending returns the earliest processAfter among remaining items,
+// or zero time if the queue is empty.
+func (nodeDat *nodeData) earliestPending() time.Time {
+	nodeDat.ItemsMutex.Lock()
+	defer nodeDat.ItemsMutex.Unlock()
+	var earliest time.Time
+	for _, si := range nodeDat.LocalQueue {
+		if earliest.IsZero() || si.processAfter.Before(earliest) {
+			earliest = si.processAfter
+		}
+	}
+	return earliest
 }
 
 func (ctl *controller) getServerData(nodeDat *nodeData, reqName string, reqUID apitypes.UID) *serverData {
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -93,36 +93,55 @@ func (ni nodeItem) process(ctx context.Context, ctl *controller) (error, bool) {
 	logger := klog.FromContext(ctx).WithValues("node", ni.NodeName)
 	ctx = klog.NewContext(ctx, logger)
 	nodeDat := ctl.getNodeData(ni.NodeName)
-	items := nodeDat.yankItems()
+	now := time.Now()
+	readyItems := nodeDat.takeReadyItems(now)
 	var retries int
-	logger.V(4).Info("Processing items for node", "count", len(items))
-	for localItem, addTime := range items {
-		logger.V(4).Info("Processing node-local item", "item", localItem, "enqueuedAt", addTime)
+	logger.V(4).Info("Processing items for node", "readyCount", len(readyItems))
+	for item, si := range readyItems {
+		logger.V(4).Info("Processing node-local item", "item", item, "enqueuedAt", si.addTime)
 		processStart := time.Now()
-		queueDurationHists.WithLabelValues(ni.NodeName).Observe(processStart.Sub(addTime).Seconds())
-		err, retry := localItem.process(ctx, ctl, nodeDat)
+		queueDurationHists.WithLabelValues(ni.NodeName).Observe(processStart.Sub(si.addTime).Seconds())
+		output := item.process(ctx, ctl, nodeDat)
+		err, retry, retryAfter := output.err, output.retry, output.retryAfter
 		processFin := time.Now()
 		workDurationHists.WithLabelValues(ni.NodeName).Observe(processFin.Sub(processStart).Seconds())
 		if err != nil {
 			if retry {
-				logger.Info("Processing node local item suffered transient error, will retry", "item", localItem, "err", err)
+				logger.Info("Processing node local item suffered transient error, will retry", "item", item, "err", err, "retryAfter", retryAfter)
 			} else {
-				logger.Error(err, "Processing node local item failed", "item", localItem)
+				logger.Error(err, "Processing node local item failed", "item", item)
 			}
 		} else {
-			logger.V(4).Info("Finished processing node-local item", "item", localItem, "willRetry", retry)
+			logger.V(4).Info("Finished processing node-local item", "item", item, "willRetry", retry, "retryAfter", retryAfter)
 		}
 		if retry {
-			nodeDat.add(localItem)
+			if retryAfter > 0 {
+				// Fixed delay (e.g. querySleeping 5s) — bypass exponential backoff
+				nodeDat.addAfter(item, time.Now().Add(retryAfter))
+				nodeDat.rateLimiter.Forget(item)
+			} else {
+				// Exponential backoff — per-item, via the local rate limiter
+				backoff := nodeDat.rateLimiter.When(item)
+				nodeDat.addAfter(item, time.Now().Add(backoff))
+			}
 			retriesCounters.WithLabelValues(ni.NodeName).Inc()
 			retries++
+		} else {
+			nodeDat.rateLimiter.Forget(item)
 		}
 	}
+	// we check if anything is pending in the local queue.
+	// if so, we call AddAfter so the item re-enter the queue
+	earliest := nodeDat.earliestPending()
+	if !earliest.IsZero() {
+		delay := max(time.Until(earliest), 0)
+		ctl.Queue.AddAfter(ni, delay)
+	}
 	logger.V(4).Info("Done processing items for node", "numToRetry", retries)
-	return nil, retries > 0
+	return nil, false
 }
 
-func (item unboundLauncherPodItem) process(ctx context.Context, ctl *controller, nodeDat *nodeData) (error, bool) {
+func (item unboundLauncherPodItem) process(ctx context.Context, ctl *controller, nodeDat *nodeData) processResult {
 	logger := klog.FromContext(ctx).WithValues("launcherPod", item.LauncherPodName, "node", item.NodeName)
 	ctx = klog.NewContext(ctx, logger)
 
@@ -132,9 +151,9 @@ func (item unboundLauncherPodItem) process(ctx context.Context, ctl *controller,
 			logger.V(2).Info("Launcher pod deleted, cleaning up launcher data")
 			ctl.clearLauncherData(nodeDat, item.LauncherPodName)
 			ctl.enqueueUnboundInfSvrItemsOnNode(ctx, item.NodeName, fmt.Sprintf("launcher pod %s deleted", item.LauncherPodName))
-			return nil, false
+			return processResult{}
 		}
-		return err, true
+		return processResult{err: err, retry: true}
 	}
 
 	// Sync launcher instances to keep internal state fresh and clean up stopped instances.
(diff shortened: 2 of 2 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -405,14 +406,28 @@ type nodeData struct {
...
+	process(ctx context.Context, ctl *controller, nodeDat *nodeData) processOutput
+}
+
+type scheduledItem struct {
+	item         itemOnNode
+	addTime      time.Time // when first enqueued (for queue duration metrics)
+	processAfter time.Time // earliest time to process this item
+}
+
+type processOutput struct {
```
MikeSpreitzer: This definition deserves comments. The precise semantic relationships among the fields should be documented.

pkg/controller/dual-pods/controller.go
```diff
@@ -978,9 +993,13 @@ func (ctl *controller) getNodeData(nodeName string) *nodeData {
 	if ans == nil {
 		ans = &nodeData{
 			NodeName:         nodeName,
-			LocalQueue:       make(localQueue),
+			LocalQueue:       localQueue{},
```
MikeSpreitzer, line 1014: Why is this change needed?

pkg/controller/dual-pods/controller.go
```diff
@@ -405,14 +406,28 @@ type nodeData struct {
 	// and for each the time at which it was first injected into this set.
 	// Hold ItemsMutex while accessing this.
 	LocalQueue localQueue
+
+	rateLimiter workqueue.TypedRateLimiter[itemOnNode]
 }
 
-type localQueue map[itemOnNode]time.Time
+type localQueue []scheduledItem
```
MikeSpreitzer: I see a lot of iterating through this slice looking for a match on the `itemOnNode`. Wouldn't a map be simpler?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -93,36 +93,58 @@ func (ni nodeItem) process(ctx context.Context, ctl *controller) (error, bool) {
...
-		err, retry := localItem.process(ctx, ctl, nodeDat)
+		queueDurationHists.WithLabelValues(ni.NodeName).Observe(processStart.Sub(si.addTime).Seconds())
+		output := si.item.process(ctx, ctl, nodeDat)
+		err, retry, retryAfter := output.err, output.retry, output.retryAfter
 		processFin := time.Now()
 		workDurationHists.WithLabelValues(ni.NodeName).Observe(processFin.Sub(processStart).Seconds())
 		if err != nil {
 			if retry {
-				logger.Info("Processing node local item suffered transient error, will retry", "item", localItem, "err", err)
+				logger.Info("Processing node local item suffered transient error, will retry", "item", si.item, "err", err)
```
MikeSpreitzer: This and the other log statement should include `retryAfter` along with `retry`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -93,36 +93,58 @@ func (ni nodeItem) process(ctx context.Context, ctl *controller) (error, bool) {
...
+	if !earliest.IsZero() {
+		delay := time.Until(earliest)
+		if delay <= 0 {
+			delay = 0
 		}
+		ctl.Queue.AddAfter(ni, delay)
 	}
-	logger.V(4).Info("Done processing items for node", "numToRetry", retries)
-	return nil, retries > 0
+	logger.V(4).Info("Done processing items for node", "numRetried", retries)
```
MikeSpreitzer: The future tense of "num _to_ retry" is still accurate. This is a counter of items that will be reprocessed in the future.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -93,36 +93,58 @@ func (ni nodeItem) process(ctx context.Context, ctl *controller) (error, bool) {
...
+		} else {
+			nodeDat.rateLimiter.Forget(si.item)
+		}
+	}
+	// we check if anything is pending in the local queue.
+	// if so, we call AddAfter so the item re-enter the queue
+	earliest := nodeDat.earliestPending()
+	if !earliest.IsZero() {
+		delay := time.Until(earliest)
+		if delay <= 0 {
```
MikeSpreitzer: I do not think that we need to change `delay` to zero if it is already zero.
BTW, Go's `max` is now generic.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -197,7 +219,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 			pod := podAny.(*corev1.Pod)
 			return pod.Name, nil
 		})
-		return fmt.Errorf("found multiple bound server-running Pods: %v", providerNames), false
+		return processOutput{err: fmt.Errorf("found multiple bound server-running Pods: %v", providerNames), retry: false}
```
MikeSpreitzer: `retry` defaults to `false`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -254,19 +276,19 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 		changed, err := ctl.removeProviderFinalizer(ctx, providingPod)
 		if err != nil {
-			return err, true
+			return processOutput{err: err, retry: true}
 		}
 		if !changed {
 			logger.V(5).Info("Finalizer is absent from server-providing Pod, waiting for deletions to finish")
 		}
-		return nil, false
+		return processOutput{err: err, retry: false}
```
MikeSpreitzer: This could be simplified to `return processOutput{}`, right?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -300,42 +322,45 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		}
 		err := ctl.ensureUnbound(ctx, serverDat, iscName, nodeDat, providingPod, providingPodLauncherBased)
 		if err != nil {
-			return err, true
+			return processOutput{err: err, retry: true}
 		}
 		if requestingPod != nil {
-			return ctl.ensureReqState(ctx, requestingPod, serverDat, false, true)
+			err, retry := ctl.ensureReqState(ctx, requestingPod, serverDat, false, true)
```
MikeSpreitzer: `ctl.ensureReqState` was designed to return the same thing as `itemOnNode.process`; I think it may be simpler to maintain that.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -581,29 +612,31 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 		if err != nil {
-			return ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to construct the nominal server-providing Pod: %s", err.Error()))
+			err, retry := ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to construct the nominal server-providing Pod: %s", err.Error()))
+			return processOutput{err: err, retry: retry}
 		}
 
 		sleepingAnys, err := ctl.podInformer.GetIndexer().ByIndex(nominalHashIndexName, nominalHash)
 		if err != nil { // impossible
-			return err, false
+			return processOutput{err: err, retry: false}
```
MikeSpreitzer: `retry` defaults to `false`

pkg/controller/dual-pods/inference-server.go
```diff
@@ -678,18 +715,19 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
-			return nil, true
+			return processOutput{retry: true}
 		}
 		if launcherPod != nil {
 			// Bind first, then rely on informer notification to trigger re-reconciliation.
 			// The "bound provider" path will handle instance creation/waking.
 			// This ensures the invariant: vllm awake implies provider Pod is bound.
 			logger.V(5).Info("Selected launcher Pod, binding first", "name", launcherPod.Name, "hasSleepingInstance", hasSleepingInstance)
-			return ctl.bind(ctx, serverDat, requestingPod, launcherPod, desiredInstanceState, true)
+			err, retry := ctl.bind(ctx, serverDat, requestingPod, launcherPod, desiredInstanceState, true)
```
MikeSpreitzer: `ctl.bind` was designed to return the same thing as `itemOnNode.process`. It might be simpler to maintain that.

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
```diff
@@ -492,23 +523,23 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		if serverDat.Sleeping == nil {
 			sleeping, err := ctl.querySleeping(ctx, iscName, providingPod, serverPort)
 			if err != nil {
-				return err, true
+				return processOutput{err: err, retry: true, retryAfter: 5 * time.Second}
```
MikeSpreitzer: The title and description of this PR should mention this change too.

pkg/controller/dual-pods/controller.go
omerap12: Yeah, but because I am changing a lot in this PR I thought that might be better to do as a follow up. I can do it in this PR if you'd like.

pkg/controller/dual-pods/inference-server.go
omerap12: So I should change `ensureReqState` as well?

pkg/controller/dual-pods/inference-server.go
omerap12: Same as https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626#discussion_r3513860829

pkg/controller/dual-pods/inference-server.go
omerap12: 😫 
Fixed in: [a6196ce](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/a6196ce10dbaa9c894afefe46fcfc6528d2a2819)

pkg/controller/dual-pods/inference-server.go
omerap12: correct again 😆 
Fixed in [aca029d](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/aca029dbe08c7fc831ee8bd3971d5acf5e2613c3)

pkg/controller/dual-pods/inference-server.go
omerap12: Yeah good call on the `max` function :) 
Fixed in: [f810a48](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/f810a48c5b8c1f21364bf928b641149012f44d44)

pkg/controller/dual-pods/inference-server.go
omerap12: 👍 
Fixed.

pkg/controller/dual-pods/inference-server.go
omerap12: fixed

pkg/controller/dual-pods/inference-server.go
omerap12: fixed

pkg/controller/dual-pods/controller.go
omerap12: Previously, `localQueue` was a `map`, so I could simply write:
```go
LocalQueue: make(localQueue),
```
Now that `localQueue` is a slice, `make` requires a length, so it has to be:
```go
LocalQueue: make(localQueue, 0),
```

I could certainly write it that way, but I found the current version a bit easier to read. 
Anyway I will change this to a map so it doesn't matter.

pkg/controller/dual-pods/controller.go
omerap12: Fixed

pkg/controller/dual-pods/inference-server.go
omerap12: once you approved the PR will update that as well :)

pkg/controller/dual-pods/controller.go
omerap12: Actually lemme fix it right now

omerap12: Done in [migrate list to map](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/f97c5797a9018045dc47ebfe56b3b4a44d8589e9)

pkg/controller/dual-pods/inference-server.go
omerap12: Updated

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: That's what I am suggesting.
For `ensureReqStatus` and `bind` too.

MikeSpreitzer:
I shortened the title because long ones are problematic.

The introductory comment is outdated; I would update it.

pkg/controller/dual-pods/controller.go
```diff
@@ -405,14 +406,33 @@ type nodeData struct {
...
 
 type itemOnNode interface {
 	// process returns (err error, retry bool).
 	// There will be a retry iff `retry`.
-	process(ctx context.Context, ctl *controller, nodeDat *nodeData) (error, bool)
+	process(ctx context.Context, ctl *controller, nodeDat *nodeData) processOutput
+}
+
+type scheduledItem struct {
+	item         itemOnNode
```
MikeSpreitzer: This seem like unnecessary duplication of the associated map key.

pkg/controller/dual-pods/controller.go
```diff
@@ -990,22 +1014,62 @@ func (ctl *controller) getNodeData(nodeName string) *nodeData {
...
-	ans := maps.Clone(nodeDat.LocalQueue)
-	nodeDat.LocalQueue = make(localQueue)
-	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(0)
-	return ans
+	if cur, exist := nodeDat.LocalQueue[item]; exist {
+		if after.Before(cur.processAfter) {
+			cur.processAfter = after
+		}
+		return
+
```
MikeSpreitzer: Is this blank line intentional?

pkg/controller/dual-pods/controller.go
```diff
@@ -990,22 +1014,62 @@ func (ctl *controller) getNodeData(nodeName string) *nodeData {
...
+	nodeDat.LocalQueue[item] = &scheduledItem{
+		item: item, addTime: time.Now(), processAfter: after,
+	}
+	addsCounters.WithLabelValues(nodeDat.NodeName).Inc()
+	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(float64(len(nodeDat.LocalQueue)))
+}
+
+// takeReadyItems removes and returns items whose processAfter <= now.
+// Items not yet ready remain in the queue.
+func (nodeDat *nodeData) takeReadyItems(now time.Time) []*scheduledItem {
```
MikeSpreitzer: If this returned a map whose domain is `itemOnNode` then you would not be inclined to make the range of `nodeDat.LocalQueue` duplicate its domain. In other words, it would be simpler to make the return type of this func be `localQueue` and take the duplication out of `localQueue`.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: https://github.com/omerap12/llm-d-fast-model-actuation/blob/a6196ce10dbaa9c894afefe46fcfc6528d2a2819/pkg/controller/dual-pods/inference-server.go#L636 still says `retry: false`.

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
omerap12: Fixed in: [3fb900f](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/3fb900f8874d0b86185d6103af93e3e02e231c77)

pkg/controller/dual-pods/inference-server.go
omerap12: Thanks. fixed in [df4db04](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/df4db046d5d106c48a386a84a2067ccac77aaa54)

pkg/controller/dual-pods/controller.go
omerap12: Fixed in: [5d6ea04](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/5d6ea04dc7fcb0f10c5d91fb594f27bd5e40ba87)

pkg/controller/dual-pods/controller.go
omerap12: Thanks. fixed in [be3d3c4](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/be3d3c4dd46187fc89f1eabdaf48cefceaecc764)

pkg/controller/dual-pods/controller.go
omerap12: Fixed in: [5d6ea04](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/5d6ea04dc7fcb0f10c5d91fb594f27bd5e40ba87)

pkg/controller/dual-pods/controller.go
```diff
@@ -405,14 +406,32 @@ type nodeData struct {
...
+type localQueue map[itemOnNode]*scheduledItem
 
 type itemOnNode interface {
 	// process returns (err error, retry bool).
 	// There will be a retry iff `retry`.
-	process(ctx context.Context, ctl *controller, nodeDat *nodeData) (error, bool)
+	process(ctx context.Context, ctl *controller, nodeDat *nodeData) processOutput
+}
+
+type scheduledItem struct {
```
MikeSpreitzer, line 427: Perhaps the name of this type should be adjusted. It is not really an "item", it just two times relevant to the item.

pkg/controller/dual-pods/controller.go
```diff
@@ -405,14 +406,32 @@ type nodeData struct {
...
+	addTime      time.Time // when first enqueued (for queue duration metrics)
+	processAfter time.Time // earliest time to process this item
+}
+
+// processOutput is returned by itemOnNode.process to signal outcome and retry intent.
+//   - err may be set with or without retry; a non-nil err with retry=false is a permanent failure.
+//   - retry requests re-enqueue with exponential backoff via the per-item rate limiter.
+//   - retryAfter, when positive, overrides exponential backoff with a fixed delay and
```
MikeSpreitzer: Rather than say something and then take it back, I think that it would be better to just say what is true. Perhaps something like the following.

- retry requests that the item be re-queued for later processing
- retryAfter, if not zero, specifies the delay until the later processing, with zero meaning to use exponential backoff.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -356,11 +375,8 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		gpuUUIDs, err := getGPUUUIDs(ctx, ctl.httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"isc_name": iscName}), url)
 		if err != nil {
 			queryErr := fmt.Errorf("GET %q fails: %s", url, err.Error())
-			updateErr, _ := ctl.ensureReqStatus(ctx, requestingPod, serverDat, queryErr.Error())
-			if updateErr == nil {
-				return queryErr, true
-			}
-			return errors.Join(queryErr, updateErr), true
+			ctl.ensureReqStatus(ctx, requestingPod, serverDat, queryErr.Error())
```
MikeSpreitzer: This change discards the error that may be returned from `ctl.ensureReqStatus`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -492,23 +508,23 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 			}
 		}
 		if err := ctl.ensureSleepingLabel(ctx, providingPod, *(serverDat.Sleeping)); err != nil {
-			return err, true
+			return processOutput{err: err, retry: true}
 		}
-		err, _ = ctl.ensureReqState(ctx, requestingPod, serverDat, shouldAddRequesterFinalizer, false)
-		if err != nil {
-			return err, true
+		reqState := ctl.ensureReqState(ctx, requestingPod, serverDat, shouldAddRequesterFinalizer, false)
```
MikeSpreitzer: "reqState" is an odd name for this `processOutput`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -633,11 +649,11 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 				return ctl.ensureReqStatus(ctx, requestingPod, serverDat, "the nominal server-providing "+errMsg)
 			}
 			wrappedErr := fmt.Errorf("failed to create server-providing Pod (started %s): %w", createStartStr, err)
-			innerErr, _ := ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to create server-providing Pod: %s", errMsg))
-			if innerErr != nil {
-				return errors.Join(wrappedErr, innerErr), true
+			reqStatus := ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to create server-providing Pod: %s", errMsg))
```
MikeSpreitzer: "reqStatus" is an odd name for this `processOutput`.
In my opinion, "innerOutput" would be better. Or perhaps "reqStatusOutput".

BTW: in my opinion, the name "process _output_" is not wrong but could be improved by being a little more specific about the fact that what this type describes is the _result_ or _return value_ from `process`. Perhaps a name like `processResult` or `processReturn`.

With that renaming, the variable here could be called "innerReturn" or "reqStatusResult".

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
omerap12: [bd301ef](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/bd301efc6681cd52a700526d4395882402873980)

pkg/controller/dual-pods/controller.go
omerap12: [b4056f9](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/b4056f9552d4d64d9b871602ef2f6c39d7c27bf9)

pkg/controller/dual-pods/inference-server.go
omerap12: [ea15c77](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/626/commits/ea15c77d869726821be0747c93def2dac5cb9a58)

pkg/controller/dual-pods/inference-server.go
```diff
@@ -356,11 +375,8 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 		gpuUUIDs, err := getGPUUUIDs(ctx, ctl.httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"isc_name": iscName}), url)
 		if err != nil {
 			queryErr := fmt.Errorf("GET %q fails: %s", url, err.Error())
-			updateErr, _ := ctl.ensureReqStatus(ctx, requestingPod, serverDat, queryErr.Error())
-			if updateErr == nil {
-				return queryErr, true
-			}
-			return errors.Join(queryErr, updateErr), true
+			out := ctl.ensureReqStatus(ctx, requestingPod, serverDat, queryErr.Error())
+			return processResult{err: errors.Join(queryErr, out.err), retry: true}
```
MikeSpreitzer, line 379: +1

pkg/controller/dual-pods/inference-server.go
```diff
@@ -633,11 +649,11 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 			wrappedErr := fmt.Errorf("failed to create server-providing Pod (started %s): %w", createStartStr, err)
-			innerErr, _ := ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to create server-providing Pod: %s", errMsg))
-			if innerErr != nil {
-				return errors.Join(wrappedErr, innerErr), true
+			innerResult := ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to create server-providing Pod: %s", errMsg))
+			if innerResult.err != nil {
+				return processResult{err: errors.Join(wrappedErr, innerResult.err), retry: true}
 			}
-			return wrappedErr, true
+			return processResult{err: wrappedErr, retry: true}
```
MikeSpreitzer: FYI, we could use the nil-discarding behavior of `errors.Join` to simplify this code too.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -724,11 +740,11 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
 		wrappedErr := fmt.Errorf("failed to create launcher-based server-providing Pod (started %s): %w", createStartStr, err)
-		innerErr, _ := ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to create launcher-based server-providing Pod: %s", errMsg))
-		if innerErr != nil {
-			return errors.Join(wrappedErr, innerErr), true
+		innerResult := ctl.ensureReqStatus(ctx, requestingPod, serverDat, fmt.Sprintf("failed to create launcher-based server-providing Pod: %s", errMsg))
+		if innerResult.err != nil {
+			return processResult{err: errors.Join(wrappedErr, innerResult.err), retry: true}
 		}
-		return wrappedErr, true
+		return processResult{err: wrappedErr, retry: true}
```
MikeSpreitzer: ... and here too.

MikeSpreitzer (commented):
You need to rebase this, but other than that it looks good enough to merge. I left a few suggestions for improvement.

MikeSpreitzer (approved):
LGTM
Thanks!

PR #630, opened 2026-07-01
Add some dquotes to shut up shellcheck

MikeSpreitzer:
This addresses the quoting concern that @diegocastanibm's raised in his review of #624 .


diegocastanibm (approved):
LGTM

PR #631, opened 2026-07-01
Update vllm to version 0.23.0 to match WVA/benchmarking etc

aavarghese:
The bump also requires an OpenShift compatibility fix in the launcher image (below) — without it the OpenShift E2E fails. Also, vLLM 0.23.0 moved `cli_env_setup` out of the (now-deleted) v`llm.entrypoints.utils` module.

**Symptom**

On v0.23.0 the OpenShift E2E Run E2E tests step fails (exit 99): launcher inference-server containers `CrashLoopBackOff` (1/2, only the sidecar up) and never reach Ready.

```
launcher-…  1/2  CrashLoopBackOff  6 (…)  12m
Did not become true …: [ … Ready … -ge 2 ]
##[error]Process completed with exit code 99
```

**Root cause**

`launcher.py` imports vLLM in-process. On import, the torch 2.x bundled with 0.23.0 builds a per-user inductor cache dir via `getpass.getuser()`, which falls through to `pwd.getpwuid(os.getuid())`. Under OpenShift's restricted-v2 SCC the pod runs as an arbitrary UID with no /`etc/passwd` entry, so it raises `KeyError` at import — before any vLLM arg parsing:

```
torch/_inductor/runtime/cache_dir_utils.py:23  default_cache_dir()
 → getpass.getuser() → pwd.getpwuid(os.getuid())
KeyError: 'getpwuid(): uid not found: 1003520000'
```

**Fix**

Add to both launcher Dockerfiles (Dockerfile.launcher.benchmark, Dockerfile.launcher.cpu):

```
ENV USER=vllm \
    HOME=/tmp
```

- getpass.getuser() checks $LOGNAME/$USER/$LNAME/$USERNAME before getpwuid(), so USER short-circuits the failing lookup.
- HOME=/tmp covers the sibling os.path.expanduser("~") → getpwuid path and points home at a dir writable under restricted-v2.

This is the canonical llm-d remediation — `deploy-llm-d` troubleshooting doc, "vLLM crashes on OpenShift with KeyError / getpwuid (vLLM v0.22+)": 
https://github.com/llm-d-incubation/llm-d-skills/blob/main/skills/deploy-llm-d/references/troubleshooting.md (prescribes USER=vllm). 
Upstream context: https://github.com/vllm-project/vllm/issues/44548.

### DIFF
--- a/config/examples/inject-launcher/replicaset.yaml
+++ b/config/examples/inject-launcher/replicaset.yaml
@@ -33,7 +33,7 @@ spec:
             name: launcher-code
       containers:
         - name: inference-server
-          image: vllm/vllm-openai:v0.20.2
+          image: docker.io/vllm/vllm-openai:v0.23.0
           ports:
             - containerPort: 8001
           # Install dependencies and run the app
--- a/dockerfiles/Dockerfile.launcher.benchmark
+++ b/dockerfiles/Dockerfile.launcher.benchmark
@@ -1,7 +1,17 @@
-ARG BASE_IMAGE=vllm/vllm-openai:v0.20.2
+ARG BASE_IMAGE=docker.io/vllm/vllm-openai:v0.23.0
 FROM ${BASE_IMAGE}
 
 WORKDIR /app
+
+# vLLM 0.22+ (torch 2.x) computes a per-user inductor cache dir at import via
+# torch's default_cache_dir() -> getpass.getuser() -> pwd.getpwuid(). Under
+# OpenShift's restricted-v2 SCC the arbitrary UID has no /etc/passwd entry, so
+# this raises KeyError. No cache-dir env var avoids it: codecache.py calls
+# default_cache_dir() directly for a deliberately-global header cache (bypassing
+# TORCHINDUCTOR_CACHE_DIR), and default_cache_dir() has no env override. Seeding
+# USER makes getpass.getuser() read the env instead of /etc/passwd.
+# See https://github.com/vllm-project/vllm/issues/44548
+ENV USER=vllm
 
 COPY inference_server/launcher/launcher.py inference_server/launcher/gputranslator.py inference_server/launcher/launcher_pod_notifier.py /app/
 
--- a/inference_server/benchmark/deploy/server-request-generic-model.yaml
+++ b/inference_server/benchmark/deploy/server-request-generic-model.yaml
@@ -22,7 +22,7 @@ spec:
           spec:
             containers:
             - name: inference-server
-              image: docker.io/vllm/vllm-openai:v0.20.2
+              image: docker.io/vllm/vllm-openai:v0.23.0
               command:
               - vllm
               - serve
--- a/inference_server/launcher/launcher.py
+++ b/inference_server/launcher/launcher.py
@@ -38,7 +38,7 @@ from gputranslator import GpuTranslator
 from pydantic import BaseModel
 from vllm.entrypoints.openai.api_server import run_server
 from vllm.entrypoints.openai.cli_args import make_arg_parser, validate_parsed_serve_args
-from vllm.entrypoints.utils import cli_env_setup
+from vllm.entrypoints.serve.utils.api_utils import cli_env_setup
 from vllm.utils.argparse_utils import FlexibleArgumentParser
 
 MAX_LOG_RESPONSE_BYTES = 1 * 1024 * 1024  # 1 MB default for API response
@@ -186,8 +186,8 @@ class VllmInstance:
                 config.env_vars = {}
             config.env_vars["CUDA_VISIBLE_DEVICES"] = ",".join(cuda_indices)
             logger.info(
-                f"Set CUDA_VISIBLE_DEVICES to \
-                    {config.env_vars['CUDA_VISIBLE_DEVICES']} based on UUIDs."
+                "Set CUDA_VISIBLE_DEVICES to %s based on UUIDs.",
+                config.env_vars["CUDA_VISIBLE_DEVICES"],
             )
 
         # Initialize instance variables
--- a/cmd/requester/README.md
+++ b/cmd/requester/README.md
@@ -82,7 +82,7 @@ spec:
           spec:
             containers:
             - name: inference-server
-              image: docker.io/vllm/vllm-openai:v0.20.2
+              image: docker.io/vllm/vllm-openai:v0.23.0
               imagePullPolicy: IfNotPresent
               command:
               - vllm
@@ -156,7 +156,7 @@ spec:
           spec:
             containers:
             - name: inference-server
-              image: docker.io/vllm/vllm-openai:v0.20.2
+              image: docker.io/vllm/vllm-openai:v0.23.0
               imagePullPolicy: IfNotPresent
               command:
               - vllm
@@ -258,7 +258,7 @@ command:
 env:
   - name: CUDA_VISIBLE_DEVICES
     value: "0"
-image: docker.io/vllm/vllm-openai:v0.20.2
+image: docker.io/vllm/vllm-openai:v0.23.0
 imagePullPolicy: IfNotPresent
 name: inference-server
 ports:
--- a/dockerfiles/Dockerfile.launcher.cpu
+++ b/dockerfiles/Dockerfile.launcher.cpu
@@ -2,7 +2,7 @@
 # Supports both arm64 and amd64 architectures
 
 ARG TARGETARCH
-ARG VLLM_VERSION=v0.20.2
+ARG VLLM_VERSION=v0.23.0
 
 # Define base images for each architecture
 FROM public.ecr.aws/q9t5s3a7/vllm-arm64-cpu-release-repo:${VLLM_VERSION} AS base-arm64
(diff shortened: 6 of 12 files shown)


diegocastanibm (commented):
For reference, I like to use more the explicit image name (`docker.io/vllm/vllm-openai:v0.23.0`) than the short version (`vllm/vllm-openai:v0.23.0`). However, the short version is more common in llm-d:
https://github.com/llm-d/llm-d-router/blob/c7319a78d9eb49b49ba713b550e1023e32e718df/config/manifests/vllm/gpu-prefix-cache-deployment.yaml#L18

I suggest to keep it as it is in this PR, but I do not like to have different styles. I will approve it as soon as the tests are passed

config/examples/inject-launcher/replicaset.yaml
```diff
@@ -33,7 +33,7 @@ spec:
             name: launcher-code
       containers:
         - name: inference-server
-          image: vllm/vllm-openai:v0.20.2
+          image: vllm/vllm-openai:v0.23.0
```
diegocastanibm: Format 1

inference_server/benchmark/deploy/server-request-generic-model.yaml
```diff
@@ -22,7 +22,7 @@ spec:
           spec:
             containers:
             - name: inference-server
-              image: docker.io/vllm/vllm-openai:v0.20.2
+              image: docker.io/vllm/vllm-openai:v0.23.0
```
diegocastanibm, line 25: format 2

diegocastanibm (commented):
We should be consistent (at least in our repo). Sometimes you use `docker.io/vllm/vllm-openai:v0.23.0`, and sometimes `vllm/vllm-openai:v0.23.0`

aavarghese:
FYI: The logline reformatting in this PR is due to different Black versions — the hook's 26.5.1 and CI's 24.4.2.
This reformatting now leaves the file unchanged, and isort/flake8 pass.

- .pre-commit-config.yaml pins Black 26.5.1 → unwrapped the line on every commit
- .github/workflows/python-code-quality.yml pins Black 24.4.2 → rejected the unwrapped form

We need to use the same version (different PR)...

dockerfiles/Dockerfile.launcher.benchmark
```diff
@@ -1,8 +1,17 @@
...
 WORKDIR /app
 
+# vLLM 0.22+ (torch 2.x) builds a per-user inductor cache dir via
+# getpass.getuser()/expanduser("~") -> pwd.getpwuid(); under OpenShift's
+# restricted-v2 SCC the arbitrary UID has no /etc/passwd entry, so vLLM crashes
+# at import with KeyError. USER short-circuits getpass.getuser() and HOME
+# short-circuits expanduser; /tmp is writable under restricted-v2.
+# See https://github.com/vllm-project/vllm/issues/44548
+ENV USER=vllm \
+    HOME=/tmp
```
MikeSpreitzer: This is very hacky. In every other case, the relevant library defines an optional environment variable that directs where the cache should be. For example, see the several settings that we are already using at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/428374588cbea903b0ce33cd4233e4fdd4ef446c/test/e2e/mkobjs-openshift.sh#L199-L210 . Can the location of the inductor cache be directed similarly?

inference_server/launcher/launcher.py
```diff
@@ -186,8 +186,8 @@ def __init__(
                 config.env_vars = {}
             config.env_vars["CUDA_VISIBLE_DEVICES"] = ",".join(cuda_indices)
             logger.info(
-                f"Set CUDA_VISIBLE_DEVICES to \
-                    {config.env_vars['CUDA_VISIBLE_DEVICES']} based on UUIDs."
+                f"Set CUDA_VISIBLE_DEVICES to "
+                f"{config.env_vars['CUDA_VISIBLE_DEVICES']} based on UUIDs."
```
MikeSpreitzer: +1

inference_server/benchmark/deploy/server-request-generic-model.yaml
aavarghese: Converted to format 2. Thanks!

diegocastanibm:
> FYI: The logline reformatting in this PR is due to different Black versions — the hook's 26.5.1 and CI's 24.4.2. This reformatting now leaves the file unchanged, and isort/flake8 pass.
> 
>     * .pre-commit-config.yaml pins Black 26.5.1 → unwrapped the line on every commit
> 
>     * .github/workflows/python-code-quality.yml pins Black 24.4.2 → rejected the unwrapped form
> 
> 
> We need to use the same version (different PR)...

Working on it here:
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/632

dockerfiles/Dockerfile.launcher.benchmark
aavarghese: Agreed. Looking into it.
Fyi: had to do the same fix in benchmarking https://github.com/llm-d/llm-d-benchmark/pull/1573/changes

aavarghese: @MikeSpreitzer we don't need to set HOME in the Dockerfile. I can remove that. 
Regarding setting USER, Claude comments this 
```
FMA builds its own launcher image with vllm, and its launcher podTemplates are scattered across 
the e2e scripts, examples, and the FMA template in llm-d-benchmark with no shared macro. 
Baking `USER=vllm` into the launcher image is the centralized equivalent — every launcher inherits it, 
including LauncherConfigs in other repos. Putting it in one pod spec would fix only that one and 
re-introduce the per-template drift.
 ```
 Wdyt? Do you prefer specifying it in every pod-spec or keeping it one place in the Dockerfiles?

File: inference_server/launcher/launcher.py (hunk 4 above)
Comment (diegocastanibm) [reply inference_server/launcher/launcher.py]: I prefer something like this:
```
logger.info(
    "Set CUDA_VISIBLE_DEVICES to %s based on UUIDs.",
    config.env_vars["CUDA_VISIBLE_DEVICES"],
)

dockerfiles/Dockerfile.launcher.benchmark
MikeSpreitzer: I have problems with the solution being proposed.

1. Setting USER to something that is not an actual username runs counter to a widespread and strong expectation. Even if it works in the cases tested so far, it is risky because it is so unusual. It may confuse some software included in some future usage, and will be confusing to anyone who reads it.
2. Caching is an important opportunity for reducing latency, and our launcher image must not preclude that.

The settings that I exhibited above are outside of the image precisely so that users can exploit the relevant caching opportunities.

We have dealt with the inductor before, IIRC it is the source of one of the settings that we now have. I am asking not whether the current hack works, but whether there is another environment variable setting, or command line option, that gets the cache location defined without needing to lie about the username.

aavarghese: I have been reading through https://github.com/llm-d-incubation/llm-d-skills/blob/main/skills/deploy-llm-d/references/troubleshooting.md to help with this issue. It doesn't prescribe an alternate option right now afaik but this might need some further experimentation?

MikeSpreitzer: Can you go to the source of the problem, the inductor caching? Look at its documentation, at its code if necessary, to find out if there is a way to direct it's cache location without lying about the username.

dockerfiles/Dockerfile.launcher.benchmark
```diff
@@ -1,8 +1,18 @@
...
 WORKDIR /app
 
+# vLLM 0.22+ (torch 2.x) computes a per-user inductor cache dir at import via
+# torch's default_cache_dir() -> getpass.getuser() -> pwd.getpwuid(). Under
+# OpenShift's restricted-v2 SCC the arbitrary UID has no /etc/passwd entry, so
+# this raises KeyError. No cache-dir env var avoids it: codecache.py calls
+# default_cache_dir() directly for a deliberately-global header cache (bypassing
+# TORCHINDUCTOR_CACHE_DIR), and default_cache_dir() has no env override. Seeding
+# USER makes getpass.getuser() read the env instead of /etc/passwd.
+# See https://github.com/vllm-project/vllm/issues/44548
```
MikeSpreitzer, line 13: Why is that Issue saying "fixed" and "merged May 2026" by a PR that was closed without merging?

Image

I see that https://github.com/vllm-project/vllm/pull/46734 is merged, but it advertises itself as an interim fix and cites that closed-without-merging PR as "the upstream fix".

MikeSpreitzer: I am trying to understand when we can expect better behavior in a vllm release.

inference_server/launcher/launcher.py
MikeSpreitzer: Oh, right. Since `logger.info` takes its first argument as a format string, it is dangerous to put any value from an external source into there.

aavarghese: Fixed

dockerfiles/Dockerfile.launcher.benchmark
aavarghese: @MikeSpreitzer since vllm has an issue open and investigation ongoing for this issue with so many linked open PRs, I will keep up-to-date with their findings there. The last comment in the main PR suggests to use the USER workaround temporarily

aavarghese:
Adding to https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/631#discussion_r3538036392, since we don't want to be blocked on this vllm upgrade for benchmarking plus blog post's need of metrics reasons, should we keep this PR as-is and merge, while using latest findings from vllm repo for this issue in a following PR? cc @MikeSpreitzer @waltforme @diegocastanibm

MikeSpreitzer (approved):
Regarding the pain wrt the inductor cache in release 0.23.0 of vllm: for now we have to live with the hacky solution, but we should prepare to switch to a better solution when a future release of vllm supports it. I created #643 to track this.

LGTM

PR #632, opened 2026-07-01
`pre-commit` alignment

diegocastanibm:
## Motivation

`.pre-commit-config.yaml` and the GitHub Actions workflows were checking similar
things but with **different tool versions and different implementations**, so
"pre-commit passes locally" did not imply "CI passes" (and vice versa). The main
divergences were:

- **Python tool versions were out of sync**: black `26.5.1` vs CI `24.4.2`,
  isort `8.0.1` vs `5.13.2`, flake8 `7.3.0` vs `7.0.0`. black 24 and 26 format
  differently, so pre-commit could reformat code one way while CI (`--check`)
  rejected it, or vice versa.
- **CI checks missing from pre-commit**: `golangci-lint`, `go mod tidy`, and
  `hack/verify-idl-consumption.sh` had no pre-commit hooks at all.
- **Checks reimplemented differently**: CI did trailing-whitespace / end-of-file /
  large-file checks with ad-hoc `grep`/`find` over a fixed extension list, while
  pre-commit used the standard hooks.

## Approach

Make `.pre-commit-config.yaml` the **single source of truth** for static checks
and have CI **invoke pre-commit** instead of duplicating commands. Tests
(`go test ./...`, `pytest`) remain separate CI steps, since they are not lint
hooks. Workflow and job names are preserved so required status checks / branch
protection are not disturbed.

### `.pre-commit-config.yaml`
- Kept the newer Python tool versions (black 26.5.1, isort 8.0.1, flake8 7.3.0) as
  canonical; the tree was normalized once so everything is already compliant.
- Set `check-added-large-files` to `--maxkb=600` to match CI.
- Added a `repo: local` block with system hooks for Go / codegen:
  - `go-mod-tidy` (commit stage)
  - `golangci-lint` (pre-push stage — linting the whole module is slow per commit;
    requires golangci-lint v2.12.2 on PATH)
  - `verify-idl-consumption` (manual stage — the script deletes/regenerates files
    and needs a clean tree, so it must not run on every commit)

### CI workflows
- **`python-code-quality.yml`**: replaced the manual black/isort/flake8/
  whitespace/eof/large-file steps with a single `pre-commit run --all-files`
  (skipping the Go/codegen hooks, which run in their own jobs). Kept the pytest
  step.
- **`golangci-lint.yml`**: installs pre-commit + golangci-lint v2.12.2 and runs
  the `go-mod-tidy` and `golangci-lint` hooks via pre-commit (one hook id per
  invocation).
- **`verify-idl-consumption.yml`**: pins Go to `go-version-file: go.mod` and runs
  `hack/verify-idl-consumption.sh` directly (the same script the pre-commit hook
  wraps, so the source of truth is shared).

### `CONTRIBUTING.md`
- Added a "Local Checks Before Opening a PR" section documenting how to install
  and run pre-commit, the Go-hook prerequisites, and that CI runs the same config
  (so green locally ⇒ green in CI for these checks). Notes that tests are run
  separately.

### Generated code
- Regenerated `pkg/generated` with the go.mod Go version (1.24.2, built with
  `GOTOOLCHAIN=local`). The committed generated code had been produced with an
  older Go whose formatter did not emit a blank line after the license header;
  current toolchains do. Pinning Go in the verify-idl workflow and regenerating
  makes the check **deterministic** instead of passing/failing based on which
  runner-image Go version the job happened to land on. The diff is formatting only
  (one blank line per file), no structural changes.

## Verification

- `pre-commit run --all-files` passes with no reformatting.
- CI checks `code-quality`, `lint`, and `verify` are green.
- The one remaining red check (OpenShift GPU e2e) is an unrelated vLLM-startup
  timeout — the test logic and runtime code are unchanged by this PR.

## Notes / out of scope
- Go versions differ across jobs (`go.mod` 1.24.2, golangci-lint 1.25, unit-tests
  1.26); unifying them is a separate cleanup.
- Image-build and e2e/kind/OpenShift workflows cannot be reproduced in pre-commit
  (they need a cluster/registry) and remain CI-only.


.github/workflows/golangci-lint.yml
```diff
@@ -16,16 +16,19 @@ jobs:
       - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
         with:
           go-version: '1.25'
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.12.11"
```
MikeSpreitzer: Why is this here? The ubuntu-latest runner includes python3. See https://github.com/actions/runner-images/blob/ubuntu24/20260622.220/images/ubuntu/Ubuntu2404-Readme.md

docs/llm-d-fma-arch-m2.drawio.svg
MikeSpreitzer, line 1: What is the change here? It is not visible, at least to my eyes.
Same question for the other SVG file changed.

CONTRIBUTING.md
```diff
@@ -63,6 +63,52 @@ For changes that fix broken code or add small changes within a component:
 
 The current testing documentation can be found within the respective components of the [docs folder](docs/).
 
+## Local Checks Before Opening a PR
+
+We use [`pre-commit`](https://pre-commit.com) to run our lint, formatting, and
+code-generation checks. **CI runs the exact same `.pre-commit-config.yaml`**, so if
```
MikeSpreitzer, line 69: This is awkward. CI is _post_ commit. Using _pre_ commit checking there seems contrary. Is pre-commit using some underlying things that CI could instead use directly?

CONTRIBUTING.md
```diff
@@ -63,6 +63,52 @@ For changes that fix broken code or add small changes within a component:
...
+the hooks pass locally they will pass in GitHub Actions — please run them before
+opening a PR.
+
+### One-time setup
```
MikeSpreitzer, line 73: Putting requirements on a user's global environment is rude. If two projects do this, a contributor to both may be stuck. Particularly if the two projects stipulate different versions of something.

.github/workflows/verify-idl-consumption.yml
```diff
@@ -25,5 +25,17 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
+      # Pin Go to the version declared in go.mod so code generation is
```
MikeSpreitzer: This is bad, because go.mod's _language version_ is not the _tools version_ that we want to use.
Consistently using the same tool version to generate code is fine, just let it be specified separately.

MikeSpreitzer:
CF #634

docs/llm-d-fma-arch-m2.drawio.svg
diegocastanibm: new line at the end of the file. Just format

CONTRIBUTING.md
diegocastanibm: `pre-commit` is just a runner. The checks that actually execute are the underlying tools (black, isort, flake8, typos, and the pre-commit-hooks scripts), each in an environment pinned by `.pre-commit-config.yaml`. CI could absolutely call those tools directly.
However, the reason CI invokes `pre-commit run --all-files` here is to keep **a single source of truth** for versions and configuration. If CI called each tool directly, the versions and flags would be declared in two places (the workflow and the pre-commit config) and drift apart, which is exactly the problem this PR is fixing (the problem that @aavarghese faced in [her PR](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/631#issuecomment-4859326994)). 

On the naming: `pre-commit` refers to the git hook stage the tool can install into, but the tool itself is a general hook runner, not limited to the pre-commit stage. Running it in CI to enforce the same pinned checks is an intentional, first-class use and it's a mainstream pattern: e.g. vLLM runs its checks this way via the official pre-commit/action  (as you can see [here](https://github.com/vllm-project/vllm/blob/main/.github/workflows/pre-commit.yml)).

Personally, I prefer this solution rather than accepting version duplication across pre-commit and CI, ideally guarded by a check that they stay in sync.

CONTRIBUTING.md
diegocastanibm: Fair point. I can see that requiring a specific globally-installed golangci-lint is unfriendly.
I've changed it so `pre-commit` manages `golangci-lint` itself. The hook now uses the official `golangci-lint` `pre-commit` hook, pinned by rev in `.pre-commit-config.yaml`. `pre-commit` builds it in its own isolated, version-pinned environment on first run.In that way, nothing needs to be on the contributor's global PATH, and the version lives in one place, just like black/isort/flake8. I also dropped the manual curl install from the CI job and removed the global-install requirement from `CONTRIBUTING.md`.

.github/workflows/verify-idl-consumption.yml
diegocastanibm: Agreed. I've pinned it explicitly instead.

CONTRIBUTING.md
MikeSpreitzer: So, in short, "pre-commit" is simply misnamed.

.github/workflows/golangci-lint.yml
diegocastanibm: `setup-python` is here because this job installs `pre-commit` (to run the go mod tidy and golangci-lint hooks from the shared config), and it gives a reliable way to do that. The python3 that ships in the runner image is the system interpreter, which on Ubuntu 24.04 is externally-managed (PEP 668), so a plain `pip install pre-commit` into it fails. It's also the conventional pattern for running pre-commit in CI ( vLLM's [pre-commit.yml](https://github.com/vllm-project/vllm/blob/d29125c0852eb61efef060d55675010d1abf51fe/.github/workflows/pre-commit.yml#L52) uses actions/setup-python for exactly this ).

MikeSpreitzer: https://peps.python.org/pep-0668/ says

> Python-specific package management tools ... should instead guide the end user towards using a virtual environment

I think that we should take the hint. Making a venv is more local and cheap than a full Python install.

.github/workflows/golangci-lint.yml
```diff
@@ -16,16 +16,17 @@ jobs:
...
 
-      - name: check go mod tidy
-        run: |
-             go mod tidy
-             if [ -n "$(git status --porcelain=v2)" ]; then
-                 echo 'You need to run `go mod tidy`' >&2
-                 exit 1
-             fi
+      - name: Install pre-commit
+        run: python -m pip install --upgrade pip pre-commit
```
MikeSpreitzer: Shouldn't we pin the version of pre-commit in this workflow and the others (no problem here with a general developer's activities), as we regularly do for dependencies?

.github/workflows/python-code-quality.yml
```diff
@@ -87,9 +52,7 @@ jobs:
         if: always()
         run: |
           echo "Code quality checks completed!"
-          echo "If any checks failed, please run the following locally to fix issues:"
-          echo "  black ."
-          echo "  isort --profile black ."
-          echo "  flake8 --max-line-length=88 --extend-ignore=E203,W503 ."
+          echo "If any checks failed, run pre-commit locally to fix issues:"
+          echo "  pre-commit run --all-files"
```
MikeSpreitzer: This is missing the SKIP.

Do we really want to keep separating the code quality checks into two different workflows? It seems unnecessarily painful to me.

.github/workflows/verify-idl-consumption.yml
```diff
@@ -25,5 +25,19 @@ jobs:
...
+      # deterministic: the generated code is sensitive to the Go toolchain's
+      # formatter, so without pinning it passes or fails depending on which
+      # runner-image Go version the job lands on. This is the *tools* version
+      # used to generate code, chosen on purpose and specified separately from
+      # go.mod's language version; setup-go sets GOTOOLCHAIN=local, so the
+      # generators are built with exactly this version. If you bump it, also
+      # regenerate the committed code (hack/verify-idl-consumption.sh).
+      - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
+        with:
+          go-version: "1.24.2"
```
MikeSpreitzer: We do not want to use Go 1.24 tools. We should use the latest that is reasonably mature, which is currently 1.26.4.

What I do not see yet is this repo using the same tools version for all code generation, both in workflows and by contributors.

hack/boilerplate.go.txt
```diff
@@ -12,4 +12,4 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
-*/
\ No newline at end of file
+*/
```
MikeSpreitzer, line 15: Git is not showing us here, but the delta here is adding the missing linebreak at the end of the last line.

docs/llm-d-fma-arch-m2.drawio.svg
MikeSpreitzer: You answered with a noun, not a difference.
Upon investigation I see that the difference is _adding_ that newline.

Note that this file was written by an application, draw.io. Presumably that application will continue to write SVG in the same way in the future --- and hopefully it can read files that have had that newline added. Wouldn't it be better to except these SVG files from that troublesome lint check?

.pre-commit-config.yaml
```diff
@@ -41,3 +42,34 @@ repos:
     rev: 37bb98842b0d8c4ffebdb75301a13db0267cef89  # frozen: v1.47.2
     hooks:
       - id: typos
+
+  # Go linting. Managed by pre-commit in an isolated, version-pinned
+  # environment, so it does not require a specific golangci-lint on the
+  # contributor's global PATH. Runs on push, not every commit, because
+  # linting the whole module is slow.
+  - repo: https://github.com/golangci/golangci-lint
```
MikeSpreitzer, line 50: Is there an option here to specify the Go toolchain version? If so, should we do that?

.pre-commit-config.yaml
```diff
@@ -41,3 +42,34 @@ repos:
...
+    rev: c0d3ddc9cf3faa61a4e378e879ece580256d76e5  # frozen: v2.12.2
+    hooks:
+      - id: golangci-lint-full
+        stages: [pre-push]
+
+  # Go mod tidy and codegen verification. These use the project's own Go
+  # toolchain (governed by go.mod), not an extra pinned global tool.
```
MikeSpreitzer: Currently `go.mod` does _not_ specify the toolchain version, it only specifies the _language_ version. These are two different things. See https://medium.com/@alif_biswas/why-your-go-mod-has-two-versions-go-and-toolchain-go-996f8065482e, for example, for an exposition.

Do we actually need to pin the Go _toolchain_ version? What goes wrong if we do not?

hack/boilerplate.go.txt
MikeSpreitzer: Why, oh why, is _this_ delta displayed differently than the one for DCO?

image

.github/workflows/golangci-lint.yml
diegocastanibm: OK then, I'll use a venv instead the interpreter.

docs/llm-d-fma-arch-m2.drawio.svg
diegocastanibm: Sure, NP. I will exclude those files

.github/workflows/golangci-lint.yml
diegocastanibm: I agree. I'll pin the `pre-commit`

.github/workflows/python-code-quality.yml
diegocastanibm: aha! good point about the SKIP. Easy to fix.
About the workflows, while I thik it is a good idea, I do not want to extend this PR into workflow consolidation. I'll open an issue to address the workflows consolidation in another PR.

.github/workflows/verify-idl-consumption.yml
diegocastanibm: `go.mod` says 1.24.2 (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/0faa4223d9f13d30a35bb98325a93cf620da816f/go.mod#L3)

But I see your point. However, the issue is not only the specific version in this workflow, but that we do not yet have a single clearly shared Go toolchain version for code generation across CI and local contributor workflows. I can change the workflow to the Go version that you propose, but the general solution is out of the scope of this PR. Could you create an issue to address it in a different PR?

hack/boilerplate.go.txt
diegocastanibm: 😂

.github/workflows/verify-idl-consumption.yml
MikeSpreitzer: The Go language version and the Go toolchain version are two different things. See https://medium.com/@alif_biswas/why-your-go-mod-has-two-versions-go-and-toolchain-go-996f8065482e, for example, for an exposition.

Do we actually need to use the same the Go _toolchain_ version everwhere? What goes wrong if we do not?

.pre-commit-config.yaml
diegocastanibm: As far as I can read, there is a way to do so fo r all the hooks adding:
```
default_language_version:
  golang: "x.x.x" # any required semver version
```
at the root of the file of the `.pre-commit-config.yaml`. I have added `golanf: "1.26.4"` that will affect `golangci-lint`. However, this does not cover the language: system hooks, which is intentional (the `go-mod-tidy` and `verify-idl-consumption` hooks use the system Go)

.pre-commit-config.yaml
diegocastanibm: Right, the comment is not precise. According to the AI, the `go-mod-tidy` hook is robust to toolchain variation, but `verify-idl-consumption` is sensitive to it since codegen output can vary across toolchain versions. The CI workflow for that hook now pins `1.26.4`, so contributors running the same hook locally should ideally use the same version. I'll update the comment to reflect that accurately.

.github/workflows/verify-idl-consumption.yml
diegocastanibm: I just read the medium article when you shared it in the comment below. About your question, not sure is we "need" to use the same go toolchain, but unless there is a good reason for not to, I suggest to use the same Go toolchain everywhere in the repo for uniformity reasons

.github/workflows/golangci-lint.yml
```diff
@@ -16,16 +16,16 @@ jobs:
       - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
         with:
           go-version: '1.25'
+      - name: Create Python virtualenv
+        run: python3 -m venv .venv
```
MikeSpreitzer, line 20: Don't you want this in `.github/workflows/python-code-quality.yml` too?

.github/workflows/python-code-quality.yml
```diff
@@ -87,9 +52,7 @@ jobs:
         if: always()
         run: |
           echo "Code quality checks completed!"
-          echo "If any checks failed, please run the following locally to fix issues:"
-          echo "  black ."
-          echo "  isort --profile black ."
-          echo "  flake8 --max-line-length=88 --extend-ignore=E203,W503 ."
+          echo "If any checks failed, run pre-commit locally to fix issues:"
```
MikeSpreitzer: I see checks in this job whose failure will not be fixed by `pre-commit`, such as the `python -m pytest tests/test_...` commands.

.github/workflows/verify-idl-consumption.yml
```diff
@@ -25,5 +25,19 @@ jobs:
...
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
+      # Pin the code-generation toolchain explicitly so this check is
+      # deterministic: the generated code is sensitive to the Go toolchain's
+      # formatter, so without pinning it passes or fails depending on which
+      # runner-image Go version the job lands on. This is the *tools* version
+      # used to generate code, chosen on purpose and specified separately from
+      # go.mod's language version; setup-go sets GOTOOLCHAIN=local, so the
+      # generators are built with exactly this version. If you bump it, also
+      # regenerate the committed code (hack/verify-idl-consumption.sh).
```
MikeSpreitzer: This comment assumes that there is a problem and fixes only part of it. The problem --- and I am not sure whether it is actually a problem --- is that different versions of the toolchain may generate different code. If that is true, then the unfixed part of the problem is all the other invocations of code generation are not pinned to the same toolchain version.

.pre-commit-config.yaml
MikeSpreitzer: Well, that declares a _langage_ version. As a default for a pre-commit "hook". And in Go there is a distinction between _language_ version and _toolchain_ version. So this really just gets back to questions about the behavior of the pre-commit hook used here ("golangci-lint-full" from github.com/golangci/golangci-lint). I tried looking into that. I found https://golangci-lint.run/docs/configuration/file/ , and it refers to https://github.com/golangci/golangci-lint/blob/HEAD/.golangci.reference.yml . The latter file, as far as I can tell, (1) does not say that this hook has a sensitivity to _toolchain_ version, and (2) does seem to suggest (https://github.com/golangci/golangci-lint/blob/HEAD/.golangci.reference.yml#L4909-L4911) that the _language_ version is determined in the usual way (from `go.mod` in this and imported modules). So, I think that the answers to my original questions ("Is there an option here to specify the Go toolchain version? If so, should we do that?") are negative. I think that the addition of `default_language_version: {golang: "1.26.4"}` is ineffective and thus confusing.

MikeSpreitzer (commented):

.github/workflows/python-code-quality.yml
MikeSpreitzer: Yes, OK with me to defer the unification to a later PR, but I _do_ recommend that it eventually be done.

.github/workflows/verify-idl-consumption.yml
MikeSpreitzer: I am not loving taking "the AI" as the authority. The code generation is done by application-level code from Kubernetes. And we _are_ pinning the version of that.

Have we observed the generated code _actually varying_ with Go toolchain vesion? I tried the latest 1.24.X and the latest 1.25.X and got no difference from what is in Git.

[try24.log](https://github.com/user-attachments/files/29719307/try24.log)

[try25.log](https://github.com/user-attachments/files/29719308/try25.log)

It _is_ good to pin the version of the Go toolchain here, since it is a dependency. I am not convinced that we have to think or work any harder than that.

.github/workflows/python-code-quality.yml
diegocastanibm: Sure. I have assigned the task to me:
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/640

.github/workflows/golangci-lint.yml
diegocastanibm: Not needed there. The two jobs get their Python differently, so they handle PEP 668 in different (appropriate) ways.

- `golangci-lint.yml` doesn't set up Python; it uses the runner's system python3, which on Ubuntu 24.04 is externally-managed (PEP 668), so a direct pip install fails. Hence the `venv` is needed it .
- `python-code-quality.yml` uses actions/setup-python, which provides a managed interpreter that is not externally-managed, so pip install works there without a venv (and it's the natural fit for that job, which also `pip installs` the `pytest deps`).

So both jobs already avoid the PEP 668 problem. The venv is only necessary in the golangci job because that's the one relying on the system interpreter.

.github/workflows/python-code-quality.yml
diegocastanibm: Yes, right. I'll be more specific. The `pytest` suite is a separate step that `pre-commit` does not fix. I've reworded the summary.

MikeSpreitzer:
The Go unit test failure is due to Issue #641

.github/workflows/verify-idl-consumption.yml
diegocastanibm: Thanks for actually testing it. It seems that you are right. CI is green with this job pinned to 1.26.4 while the committed generated code was produced with 1.24.2, so 1.24, 1.25 and 1.26 all produce identical output. SO this confirms also your tests. 
I've kept the explicit pin  but rewrote the comment. On the "other codegen invocations aren't pinned to the same toolchain" point: agreed that's the remaining gap, but it's the repo-wide uniformity question that we're deferring out of this PR (tracked in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/640).

.pre-commit-config.yaml
diegocastanibm: Thanks for the research on this matter. If it's ineffective and confusing, then I'll remove it. 
I also fixed the comment on the local hooks block that still claimed `verify-idl-consumption` was toolchain-sensitive. That was my interpretation after Google it, but I was wrong.

.github/workflows/golangci-lint.yml
MikeSpreitzer: Both workflows do `pip install`.

Why shouldn't `python-code-quality.yml` use the runner's Python and a venv, like `golangci-lint.yml` does?

MikeSpreitzer (commented):
One comment thread not yet concluded, plus rebase needed after #642 merges.

.github/workflows/golangci-lint.yml
diegocastanibm: They both end up running pip install, but the two jobs relate to Python differently, which is why I provisioned it differently:

- python-code-quality.yml is a Python-centric job: besides running the Python linters via pre-commit, it runs the launcher pytest suite with real dependencies (requirements-sans-vllm.txt). actions/setup-python is the idiomatic, first-class way to provision Python for that. It pins the interpreter (3.12.11) for reproducible test runs, integrates pip caching, and it's what this job already used on main.
- golangci-lint.yml is a Go job where Python is incidental. It only needs Python to run pre-commit. Pulling in setup-python there felt heavier than warranted, so the runner's Python + a venv is enough.

A few points in favor of setup-python:

- It's the standard. actions/setup-python is the officially recommended, first-class way to set up Python in GitHub Actions, and it's exactly what a Python-focused job like this one is meant to use.
- No risk to the OS Python. It doesn't install into or modify the system interpreter. It provisions a separate, self-contained Python in the runner's tool cache (its own site-packages), so pip install there never touches the apt-managed system packages. That's precisely the isolation PEP 668 is trying to protect, achieved by not using the system interpreter at all. So there's nothing bad about it.
- Widely used by major projects. For example, vLLM sets up Python with actions/setup-python in its own pre-commit workflow: .github/workflows/pre-commit.yml#L52.

So it's not that one job is right and the other wrong: both avoid PEP 668, and each uses the mechanism that fits how central Python is to it. The pinned interpreter + caching are worth a bit more in the job that actually runs the test suite.

That said, if you'd still rather have one uniform pattern, I'm happy to switch python-code-quality.yml to the runner's Python + venv as well. It's a low-cost change but my recommendation is to keep setup-python. Just let me know.

MikeSpreitzer: OK, you have explained why actions/setup-python is better.

I do not think that we need to tweak this more in this PR, since you plan to unify the two workflows in a future PR.

MikeSpreitzer (approved):
LGTM

PR #633, opened 2026-07-01
Make all action references include tag in a comment

MikeSpreitzer:
Add the tag in a comment, where missing in GitHub Actions workflow `use` lines.


waltforme (approved):
LGTM

diegocastanibm:
I think we need to adopt (as a team) the idea of pushing to a branch in llm-d-fma instead of using the fork in the PR everytime we modify the workflows. That way we can be sure that the modifications are not breaking the CI process

Issue #585, opened 2026-06-21
[Bug]: launcher-populator counts permanently failed launcher pods toward `current`, holding a node's slot indefinitely


aavarghese:
## Source
- llm-d-benchmark CI run 27892467961 (FMA pass, runner-20260621-040919-620)
- Image: `ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/launcher:v0.6.0-alpha.13`
- Cluster: pokprod (OpenShift) — node `pokprod-b93r39s1` has a degraded GPU device, causing launcher pods to fail with `nvidia-container-runtime-hook` (exit code 1)

## Observed

In `launcher-populator.log`, look at node `pokprod-b93r39s1`:

```
I0621 04:06:57.280064  ... node="pokprod-b93r39s1"  current=0 stale=0 desired=1 diff=1
I0621 04:06:57.809448  ... "Created launcher pod" pod="launcher-fma-...-6xvzv" node="pokprod-b93r39s1"
I0621 04:06:57.809475  ... "Created launchers"   node="pokprod-b93r39s1" count=1
I0621 04:06:57.809518  ... node="pokprod-b93r39s1"  current=1 stale=0 desired=1 diff=0
... ~200 reconcile passes later, all with current=1 desired=1 diff=0 ...
```

But `launcher-fma-...-6xvzv` (the only launcher pod the populator created on this node) is in `phase=Pending` with container in `waiting` state, having failed:

```
40m  Warning  Failed  pod/launcher-fma-qwen-qwe-2d2178f3-wen3-32b-6xvzv
     Error: container create failed: error executing hook
     `/usr/local/nvidia/toolkit/nvidia-container-runtime-hook` (exit code: 1)
```

Per `resources.yaml`, this pod NEVER reached Running, NEVER served any requester. It will never recover (the hook will keep failing as long as the node's GPU device-plugin state is broken).

Yet populator considers `current=1` for that node and reports `diff=0`, so it never:
- Retries by creating a fresh launcher
- Falls through to mark the slot failed
- Surfaces a metric or event the operator can act on

The bad slot is held forever. As HPA drives requester scale-up, requesters can land on this node (because GPU is still advertised), get bound to the broken launcher, and fail; ReplicaSet retries, accumulating ContainerStatusUnknown / Failed pods (~700 in this run's `pods.txt`).

## Suggested investigation
[...]

PR #635, opened 2026-07-02
Add Prometheus metrics fma_launcher_pod_count to launcher-populator

delavet:
This is the first part of the PR aimed at addressing the issue raised in #585.

The launcher-populator previously exposed no metrics. This PR wires pkg/observability into it (Prometheus /metrics + Go /debug/pprof) and adds a gauge:

fma_launcher_pod_count{lcfg_name, phase}

phase is one of bound, unbound, stuck, or stale. A launcher is stuck when it uses the current template, is unbound, has been scheduled longer than a configurable threshold, and is not yet Ready — surfacing launchers that fail to start

### DIFF
--- a/cmd/launcher-populator/main.go
+++ b/cmd/launcher-populator/main.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/common"
 	launcherpopulator "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/launcher-populator"
+	fmaobs "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/observability"
 	"github.com/spf13/pflag"
 
 	kubeinformers "k8s.io/client-go/informers"
@@ -42,11 +43,18 @@ func main() {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	overrides := &clientcmd.ConfigOverrides{}
 
+	obsOpts := fmaobs.DefaultOptions()
+
 	klog.InitFlags(flag.CommandLine)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	common.AddKubernetesClientFlags(*pflag.CommandLine, loadingRules, overrides)
 	expectationTimeout := pflag.Duration("expectation-timeout", launcherpopulator.DefaultExpectationTimeout,
 		"How long to wait for the informer cache to reflect pending Pod mutations before falling back to a direct apiserver query")
+	stuckThreshold := pflag.Duration("stuck-threshold", launcherpopulator.DefaultStuckThreshold,
+		"Minimum age (since scheduling) after which a not-yet-Ready launcher is reported in the \"stuck\" phase of fma_launcher_count")
+	metricsResyncPeriod := pflag.Duration("metrics-resync-period", launcherpopulator.DefaultMetricsResyncPeriod,
+		"How often to recompute the fma_launcher_count gauge from the informer cache")
+	obsOpts.AddToFlagSet(pflag.CommandLine)
 	pflag.Parse()
 
 	// Create a context with cancellation signal
@@ -96,6 +104,8 @@ func main() {
 		kubePreInformers.Core().V1(),
 		fmaPreInformers,
 		*expectationTimeout,
+		*stuckThreshold,
+		*metricsResyncPeriod,
 	)
 	if err != nil {
 		klog.Fatal(err)
@@ -104,6 +114,9 @@ func main() {
 	// Start informers
 	kubePreInformers.Start(ctx.Done())
 	fmaPreInformers.Start(ctx.Done())
+
+	// Serve Prometheus /metrics and Go /debug/pprof.
+	obsOpts.Start(ctx)
 
 	// Start the controller in a goroutine
 	go func() {
--- a/cmd/launcher-populator/main.go
+++ b/cmd/launcher-populator/main.go
@@ -50,10 +50,10 @@ func main() {
 	common.AddKubernetesClientFlags(*pflag.CommandLine, loadingRules, overrides)
 	expectationTimeout := pflag.Duration("expectation-timeout", launcherpopulator.DefaultExpectationTimeout,
 		"How long to wait for the informer cache to reflect pending Pod mutations before falling back to a direct apiserver query")
-	pendingThreshold := pflag.Duration("pending-threshold", launcherpopulator.DefaultPendingThreshold,
-		"Minimum age (since creation) after which an unscheduled, not-yet-Ready launcher is reported in the \"pending\" phase of fma_launcher_pod_count")
-	stuckThreshold := pflag.Duration("stuck-threshold", launcherpopulator.DefaultStuckThreshold,
-		"Minimum age (since scheduling) after which a scheduled, not-yet-Ready launcher is reported in the \"stuck\" phase of fma_launcher_pod_count")
+	stuckSchedulingThreshold := pflag.Duration("stuck-scheduling-threshold", launcherpopulator.DefaultStuckSchedulingThreshold,
+		"Minimum age (since creation) after which an unscheduled launcher is reported in the \"stuck_scheduling\" phase of fma_launcher_pod_count")
+	stuckStartingThreshold := pflag.Duration("stuck-starting-threshold", launcherpopulator.DefaultStuckStartingThreshold,
+		"Minimum age (since scheduling) after which a scheduled, not-yet-Ready launcher is reported in the \"stuck_starting\" phase of fma_launcher_pod_count")
 	obsOpts.AddToFlagSet(pflag.CommandLine)
 	pflag.Parse()
 
@@ -104,8 +104,8 @@ func main() {
 		kubePreInformers.Core().V1(),
 		fmaPreInformers,
 		*expectationTimeout,
-		*pendingThreshold,
-		*stuckThreshold,
+		*stuckSchedulingThreshold,
+		*stuckStartingThreshold,
 	)
 	if err != nil {
 		klog.Fatal(err)
--- a/cmd/launcher-populator/main.go
+++ b/cmd/launcher-populator/main.go
@@ -51,9 +51,7 @@ func main() {
 	expectationTimeout := pflag.Duration("expectation-timeout", launcherpopulator.DefaultExpectationTimeout,
 		"How long to wait for the informer cache to reflect pending Pod mutations before falling back to a direct apiserver query")
 	stuckThreshold := pflag.Duration("stuck-threshold", launcherpopulator.DefaultStuckThreshold,
-		"Minimum age (since scheduling) after which a not-yet-Ready launcher is reported in the \"stuck\" phase of fma_launcher_count")
-	metricsResyncPeriod := pflag.Duration("metrics-resync-period", launcherpopulator.DefaultMetricsResyncPeriod,
-		"How often to recompute the fma_launcher_count gauge from the informer cache")
+		"Minimum age (since scheduling, or since creation when not scheduled) after which a not-yet-Ready launcher is reported in the \"stuck\" phase of fma_launcher_pod_count")
 	obsOpts.AddToFlagSet(pflag.CommandLine)
 	pflag.Parse()
 
@@ -105,7 +103,6 @@ func main() {
 		fmaPreInformers,
 		*expectationTimeout,
 		*stuckThreshold,
-		*metricsResyncPeriod,
 	)
 	if err != nil {
 		klog.Fatal(err)
--- a/cmd/launcher-populator/main.go
+++ b/cmd/launcher-populator/main.go
@@ -50,8 +50,10 @@ func main() {
 	common.AddKubernetesClientFlags(*pflag.CommandLine, loadingRules, overrides)
 	expectationTimeout := pflag.Duration("expectation-timeout", launcherpopulator.DefaultExpectationTimeout,
 		"How long to wait for the informer cache to reflect pending Pod mutations before falling back to a direct apiserver query")
+	pendingThreshold := pflag.Duration("pending-threshold", launcherpopulator.DefaultPendingThreshold,
+		"Minimum age (since creation) after which an unscheduled, not-yet-Ready launcher is reported in the \"pending\" phase of fma_launcher_pod_count")
 	stuckThreshold := pflag.Duration("stuck-threshold", launcherpopulator.DefaultStuckThreshold,
-		"Minimum age (since scheduling, or since creation when not scheduled) after which a not-yet-Ready launcher is reported in the \"stuck\" phase of fma_launcher_pod_count")
+		"Minimum age (since scheduling) after which a scheduled, not-yet-Ready launcher is reported in the \"stuck\" phase of fma_launcher_pod_count")
 	obsOpts.AddToFlagSet(pflag.CommandLine)
 	pflag.Parse()
 
@@ -102,6 +104,7 @@ func main() {
 		kubePreInformers.Core().V1(),
 		fmaPreInformers,
 		*expectationTimeout,
+		*pendingThreshold,
 		*stuckThreshold,
 	)
 	if err != nil {
--- a/pkg/controller/launcher-populator/metrics.go
+++ b/pkg/controller/launcher-populator/metrics.go
@@ -1,0 +1,182 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package launcherpopulator
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	kubemetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/klog/v2"
+
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/common"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
+)
+
+// DefaultStuckThreshold is the default minimum age (measured from the time a
+// launcher Pod was scheduled) after which a not-yet-Ready launcher is reported
+// as "stuck". 7.5 minutes leaves room for slow image pulls while remaining
+// operationally relevant.
+const DefaultStuckThreshold = 7*time.Minute + 30*time.Second
+
+// DefaultMetricsResyncPeriod is how often the launcher-count gauge is recomputed
+// from the Pod informer cache. The "stuck" phase is a function of elapsed time,
+// so it must be recomputed on a timer rather than only on Pod events.
+const DefaultMetricsResyncPeriod = 30 * time.Second
+
+// launcherPhase is the value of the "phase" label on fma_launcher_count.
+type launcherPhase string
+
+const (
+	// phaseBound: the launcher is assigned to a server-requesting Pod.
+	phaseBound launcherPhase = "bound"
+	// phaseUnbound: the launcher uses the current template, is unbound, and is
+	// either Ready or still within the stuck threshold.
+	phaseUnbound launcherPhase = "unbound"
+	// phaseStuck: the launcher uses the current template, is unbound, has been
+	// scheduled longer than the stuck threshold, and is not Ready.
+	phaseStuck launcherPhase = "stuck"
+	// phaseStale: the launcher was built from a superseded template.
+	phaseStale launcherPhase = "stale"
+)
+
+// allLauncherPhases lists every phase value so the gauge can emit an explicit
+// zero for phases with no members. This keeps series from disappearing (which
+// would otherwise make count-based alerts flap between "0" and "no data").
+var allLauncherPhases = []launcherPhase{phaseBound, phaseUnbound, phaseStuck, phaseStale}
+
+const lcfgNameLabel = "lcfg_name"
+const phaseLabel = "phase"
+
+var (
+	metricsOnce sync.Once
+
+	// launcherCountGauge is a GaugeVec of the number of launcher Pods, labeled
+	// by LauncherConfig name and phase. It deliberately carries no Node label:
+	// a Node label would make cardinality grow with cluster size.
+	launcherCountGauge *kubemetrics.GaugeVec
+)
+
+// registerMetrics registers the launcher-populator metrics with the k8s legacy
+// registry exactly once. Safe to call from multiple NewController invocations.
+func registerMetrics() {
+	metricsOnce.Do(func() {
+		launcherCountGauge = kubemetrics.NewGaugeVec(&kubemetrics.GaugeOpts{
+			Namespace:      "fma",
+			Name:           "launcher_pod_count",
+			Help:           "Number of launcher Pods by LauncherConfig and phase (bound, unbound, stuck, stale)",
+			StabilityLevel: kubemetrics.ALPHA,
+		}, []string{lcfgNameLabel, phaseLabel})
+		legacyregistry.MustRegister(launcherCountGauge)
+	})
+}
+
+// launcherPhaseOf classifies a single (non-terminating) launcher Pod into a
+// phase. It is a pure function of the arguments so it can be unit-tested in
+// isolation. currentTemplateHash is the template hash the digest currently
+// wants for this Pod's (node, LC) key; an empty value (LC gone / not yet
+// digested) makes any templated Pod classify as stale.
+func (ctl *controller) launcherPhaseOf(pod *corev1.Pod, currentTemplateHash string, stuckThreshold time.Duration, now time.Time) launcherPhase {
+	if bound, _ := ctl.isLauncherBoundToServerRequestingPod(pod); bound {
+		return phaseBound
+	}
+	if pod.Annotations[common.LauncherTemplateHashAnnotationKey] != currentTemplateHash {
+		return phaseStale
+	}
+	if !utils.IsPodReady(pod) && now.Sub(stuckReferenceTime(pod)) > stuckThreshold {
+		return phaseStuck
+	}
+	return phaseUnbound
+}
+
+// stuckReferenceTime returns the instant from which a launcher's "stuck" age is
+// measured: the time it was scheduled, when known, else its creation time.
+// Measuring from scheduling (rather than creation) avoids blaming a launcher
+// for time spent waiting in the scheduler.
+func stuckReferenceTime(pod *corev1.Pod) time.Time {
+	if cond := utils.GetPodCondition(pod, corev1.PodScheduled); cond != nil && cond.Status == corev1.ConditionTrue {
+		return cond.LastTransitionTime.Time
+	}
+	return pod.CreationTimestamp.Time
+}
+
+// runMetricsLoop recomputes the launcher-count gauge every metricsResyncPeriod
+// until ctx is cancelled. It runs one computation immediately so metrics are
+// populated without waiting a full period.
+func (ctl *controller) runMetricsLoop(ctx context.Context) {
+	logger := klog.FromContext(ctx).WithName("metrics")
+	ticker := time.NewTicker(ctl.metricsResyncPeriod)
+	defer ticker.Stop()
+	for {
+		ctl.updateLauncherCounts(logger)
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		}
+	}
+}
+
+// updateLauncherCounts sweeps the launcher Pods in the informer cache, tallies
+// them per (LauncherConfig, phase), and republishes launcherCountGauge.
+func (ctl *controller) updateLauncherCounts(logger klog.Logger) {
+	selector := labels.SelectorFromSet(labels.Set{
+		common.ComponentLabelKey: common.LauncherComponentLabelValue,
+	})
+	pods, err := ctl.podLister.List(selector)
+	if err != nil {
+		logger.Error(err, "Failed to list launcher Pods for metrics")
+		return
+	}
+
+	now := time.Now()
+	// counts[lcName][phase] = number of launcher Pods.
+	counts := map[string]map[launcherPhase]int{}
+	for _, pod := range pods {
+		if pod.DeletionTimestamp != nil {
+			continue
+		}
+		lcName := pod.Labels[common.LauncherConfigNameLabelKey]
+		if lcName == "" {
+			continue
+		}
+		key := NodeLauncherKey{
+			NodeName:           pod.Labels[common.NodeNameLabelKey],
+			LauncherConfigName: lcName,
+		}
+		currentTemplateHash := ctl.policy.snapshotForKey(key).templateHash
+		phase := ctl.launcherPhaseOf(pod, currentTemplateHash, ctl.stuckThreshold, now)
+		if counts[lcName] == nil {
+			counts[lcName] = map[launcherPhase]int{}
+		}
+		counts[lcName][phase]++
+	}
+
+	// Reset drops series for LauncherConfigs that no longer have any launcher
+	// Pods. Re-emit every phase (including zero) for each LC still present so
+	// count-based alerts see an explicit 0 rather than missing data.
+	launcherCountGauge.Reset()
+	for lcName, phases := range counts {
+		for _, phase := range allLauncherPhases {
+			launcherCountGauge.WithLabelValues(lcName, string(phase)).Set(float64(phases[phase]))
+		}
+	}
+}
--- a/pkg/controller/launcher-populator/metrics.go
+++ b/pkg/controller/launcher-populator/metrics.go
@@ -20,7 +20,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	kubemetrics "k8s.io/component-base/metrics"
@@ -64,12 +63,6 @@ const (
 	phaseStale launcherPhase = "stale"
 )
 
-// allLauncherPhases lists every phase value so that, for every LauncherConfig
-// whose object exists or that has launcher Pods, the gauge always carries an
-// explicit value (including zero) for each phase. This keeps count-based alerts
-// from having to distinguish "0" from "no data".
-var allLauncherPhases = []launcherPhase{phaseBound, phaseUnbound, phaseStuckScheduling, phaseStuckStarting, phaseStale}
-
 const lcfgNameLabel = "lcfg_name"
 const phaseLabel = "phase"
 
@@ -96,18 +89,56 @@ func registerMetrics() {
 	})
 }
 
-// phaseCounts tallies launcher Pods by phase.
-type phaseCounts map[launcherPhase]int
+// phaseCounts tallies launcher Pods by phase. Keeping each phase as a field
+// makes unsupported phases unrepresentable in metricsState.
+type phaseCounts struct {
+	bound           int
+	unbound         int
+	stuckScheduling int
+	stuckStarting   int
+	stale           int
+}
 
 // phaseCountsByNode stores phase tallies indexed by Node name.
 type phaseCountsByNode map[string]phaseCounts
 
 func (pc phaseCounts) total() int {
-	sum := 0
-	for _, n := range pc {
-		sum += n
+	return pc.bound + pc.unbound + pc.stuckScheduling + pc.stuckStarting + pc.stale
+}
+
+func (pc *phaseCounts) addDelta(counts, oldCounts phaseCounts) {
+	pc.bound += counts.bound - oldCounts.bound
+	pc.unbound += counts.unbound - oldCounts.unbound
+	pc.stuckScheduling += counts.stuckScheduling - oldCounts.stuckScheduling
+	pc.stuckStarting += counts.stuckStarting - oldCounts.stuckStarting
+	pc.stale += counts.stale - oldCounts.stale
+}
+
+func (pc *phaseCounts) increment(phase launcherPhase) {
+	switch phase {
+	case phaseBound:
+		pc.bound++
+	case phaseUnbound:
+		pc.unbound++
+	case phaseStuckScheduling:
+		pc.stuckScheduling++
+	case phaseStuckStarting:
+		pc.stuckStarting++
+	case phaseStale:
+		pc.stale++
+	default:
+		panic("unsupported launcher phase: " + string(phase))
 	}
-	return sum
+}
+
+// forEach calls yield once for every supported metric phase, including phases
+// whose count is zero.
+func (pc phaseCounts) forEach(yield func(launcherPhase, int)) {
+	yield(phaseBound, pc.bound)
+	yield(phaseUnbound, pc.unbound)
+	yield(phaseStuckScheduling, pc.stuckScheduling)
+	yield(phaseStuckStarting, pc.stuckStarting)
+	yield(phaseStale, pc.stale)
 }
 
 // metricsState is the source of truth backing launcherPodCountGauge: it holds
@@ -145,13 +176,8 @@ func (ms *metricsState) publish(key NodeLauncherKey, counts phaseCounts) {
 	// Fold the change into the running aggregate by its delta.
 	if counts.total() != 0 || oldCounts.total() != 0 {
 		agg := ms.agg[lcfg]
-		if agg == nil {
-			agg = phaseCounts{}
-			ms.agg[lcfg] = agg
-		}
-		for _, phase := range allLauncherPhases {
-			agg[phase] += counts[phase] - oldCounts[phase]
-		}
+		agg.addDelta(counts, oldCounts)
+		ms.agg[lcfg] = agg
 	}
 
 	// Update the per-Node store, dropping empty Nodes (and the empty LC map).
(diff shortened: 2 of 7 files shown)


pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+// DefaultStuckThreshold is the default minimum age (measured from the time a
+// launcher Pod was scheduled) after which a not-yet-Ready launcher is reported
+// as "stuck". 7.5 minutes leaves room for slow image pulls while remaining
+// operationally relevant.
+const DefaultStuckThreshold = 7*time.Minute + 30*time.Second
+
+// DefaultMetricsResyncPeriod is how often the launcher-count gauge is recomputed
+// from the Pod informer cache. The "stuck" phase is a function of elapsed time,
+// so it must be recomputed on a timer rather than only on Pod events.
+const DefaultMetricsResyncPeriod = 30 * time.Second
```
MikeSpreitzer: There is a less expensive solution. When processing a launcher Pod that will become "stuck" in the future if nothing else changes, enqueue a reference to the Pod to be processed at that future time (using the queue's `AddAfter` method). Cost is `O(rate of launcher creation)`, compared to `O(number of launchers/DefaultMetricsResyncPeriod)`.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+type launcherPhase string
+
+const (
+	// phaseBound: the launcher is assigned to a server-requesting Pod.
+	phaseBound launcherPhase = "bound"
+	// phaseUnbound: the launcher uses the current template, is unbound, and is
+	// either Ready or still within the stuck threshold.
+	phaseUnbound launcherPhase = "unbound"
+	// phaseStuck: the launcher uses the current template, is unbound, has been
+	// scheduled longer than the stuck threshold, and is not Ready.
```
MikeSpreitzer: "has been scheduled longer than the stuck threshold" does not match the implementation below (which will also classify as stuck a Pod that has been unscheduled for at least the threshold amount of time).

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to list launcher Pods for metrics")
+		return
+	}
+
+	now := time.Now()
+	// counts[lcName][phase] = number of launcher Pods.
+	counts := map[string]map[launcherPhase]int{}
+	for _, pod := range pods {
+		if pod.DeletionTimestamp != nil {
```
MikeSpreitzer: Since this metric is about counting Pod objects that exist, I think that it would be proper to include ones that are in the process of being deleted.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	// phaseStuck: the launcher uses the current template, is unbound, has been
+	// scheduled longer than the stuck threshold, and is not Ready.
+	phaseStuck launcherPhase = "stuck"
+	// phaseStale: the launcher was built from a superseded template.
+	phaseStale launcherPhase = "stale"
+)
+
+// allLauncherPhases lists every phase value so the gauge can emit an explicit
+// zero for phases with no members. This keeps series from disappearing (which
+// would otherwise make count-based alerts flap between "0" and "no data").
```
MikeSpreitzer: The flapping is cause by calling the GaugeVec's Reset method every period. This also creates a timing splinter: a scrape done between that Reset and the following Set will improperly show the absence of a time series that is not really absent.

I think that a better way to handle these issues is as follows. Introduce a new thing in `digestedPolicy` or the controller: a map from lcfg name to the set of launcher Pods that use that lcfg (removing map entries when that set becomes empty). Update it when processing new and deleted launcher Pods. When the map gains a new entry, call Set(0) for all of the phases. When an entry is deleted from that map, call DeletePartialMatch for the lcfg name. The periodic sweep then only needs to call Set, for the cross product of (the domain of that new map) X allLauncherPhases.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,153 @@
...
+func TestLauncherPhaseOf(t *testing.T) {
+	ctl := &controller{}
+	const threshold = 7*time.Minute + 30*time.Second
+	now := time.Date(2026, 6, 21, 12, 0, 0, 0, time.UTC)
+	long := now.Add(-10 * time.Minute) // older than threshold
+	short := now.Add(-1 * time.Minute) // younger than threshold
+
+	cases := []struct {
+		name string
+		pod  *corev1.Pod
```
MikeSpreitzer: The test cases would be shorter to write if this held the builder instead of the Pod.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,153 @@
...
+		want launcherPhase
+	}{
+		{
+			name: "bound wins over everything",
```
MikeSpreitzer: But this isn't checking every possible value for the other attributes.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,153 @@
...
+	const threshold = 7*time.Minute + 30*time.Second
+	now := time.Date(2026, 6, 21, 12, 0, 0, 0, time.UTC)
+	long := now.Add(-10 * time.Minute) // older than threshold
+	short := now.Add(-1 * time.Minute) // younger than threshold
+
+	cases := []struct {
+		name string
+		pod  *corev1.Pod
+		want launcherPhase
+	}{
```
MikeSpreitzer: A complete set of test cases would cover the cross product of the following.

- bound or unbound
- hash matches, hash exists and differs, hash does not exist
- ready or not
- (age (time since creation) less than threshold, age greater than threshold and not scheduled, age greater than threshold and scheduled long ago, age greater than threshold and scheduled recently)

MikeSpreitzer (commented):

pkg/controller/launcher-populator/metrics.go
delavet: Indeed, the comment was wrong, not the code. The behavior is intentional: I want a launcher that can never get scheduled (stays Pending, no node ever fits) to eventually surface as stuck too, since "can't even be scheduled" is just as much a stuck-launcher problem as "scheduled but won't start" IMO.

pkg/controller/launcher-populator/metrics.go
delavet: Agreed — done. I removed the `DeletionTimestamp` skip in the tally loop, so terminating Pods are now counted like any other existing Pod object. `DeletionTimestamp` now has exactly one remaining use: I skip scheduling an AddAfter for a terminating Pod. A Pod on its way out doesn't need a future "will it become stuck?" wake-up, I believe.

pkg/controller/launcher-populator/metrics.go
delavet: Adopted — no more `Reset`, so the flapping/splinter is gone. Two small deviations: I keep `map[lcfg] -> map[NodeLauncherKey] -> phaseCounts` instead of a Pod set — this matches the per-key reconcile granularity, so each reconcile just recomputes and overwrites that key's tally (no incremental set maintenance, and it reuses the existing `NodeLauncherKey`). And since component-base's GaugeVec has no `DeletePartialMatch`, I loop `Delete` over the known allLauncherPhases when an lcfg's last Pod goes away (equivalent, since the phase domain is fixed).

pkg/controller/launcher-populator/metrics_test.go
delavet: done

pkg/controller/launcher-populator/metrics_test.go
delavet: done

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/common"
+	"github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/controller/utils"
+)
+
+// DefaultStuckThreshold is the default minimum age after which a not-yet-Ready
+// launcher is reported in the "stuck" phase. Age is measured from the time the
+// launcher was scheduled, or from its creation when it has not been scheduled.
+// 7.5 minutes leaves room for slow image pulls while remaining operationally
+// relevant.
```
MikeSpreitzer: Since there are two phases to get stuck in, perhaps two thresholds are in order. Scheduling does not involve a big image pull.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,218 @@
...
+const (
+	// phaseBound: the launcher is assigned to a server-requesting Pod.
+	phaseBound launcherPhase = "bound"
+	// phaseUnbound: the launcher uses the current template, is unbound, and is
+	// either Ready or has not yet existed past the stuck threshold.
+	phaseUnbound launcherPhase = "unbound"
+	// phaseStuck: the launcher uses the current template, is unbound, is not
+	// Ready, and has existed past the stuck threshold (measured from when it was
+	// scheduled, or from creation when it has not been scheduled).
+	phaseStuck launcherPhase = "stuck"
```
MikeSpreitzer: Since there are two sorts of "stuck", perhaps they deserve distinct labels.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+	for _, phase := range allLauncherPhases {
+		launcherPodCountGauge.WithLabelValues(lcfg, string(phase)).Set(float64(agg[phase]))
+	}
+}
+
+// launcherPhaseOf classifies a single launcher Pod into a phase.
+// currentTemplateHash is the template hash the digest currently wants for this
+// Pod's (node, LC) key; an empty value (LC gone / not yet digested) makes any
```
MikeSpreitzer: The Node is not relevant to this hash. The is the node-independent hash.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,218 @@
...
+// threshold (zero Time if none). Pods being deleted are counted (the metric
+// counts Pod objects that exist) but never schedule a future transition.
+func (ctl *controller) computeKeyPhases(pods []*corev1.Pod, templateHash string, now time.Time) (phaseCounts, time.Time) {
+	counts := phaseCounts{}
+	var earliestStuck time.Time
+	for _, pod := range pods {
+		phase := ctl.launcherPhaseOf(pod, templateHash, ctl.stuckThreshold, now)
+		counts[phase]++
+		if pod.DeletionTimestamp == nil && phase == phaseUnbound && !utils.IsPodReady(pod) {
+			becomesStuckAt := stuckReferenceTime(pod).Add(ctl.stuckThreshold)
```
MikeSpreitzer: The logic here is very closely intertwined with the logic in `launcherPhaseOf`. It would be better to refactor that method so that it exposes this additional result.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+
+const testTemplateHash = "hash-current"
+
+// launcherPodBuilder builds a launcher Pod for classification tests. All
+// timestamps are supplied by tests so classification is deterministic.
+type launcherPodBuilder struct {
+	pod corev1.Pod
+}
+
+func newLauncherPod() *launcherPodBuilder {
```
MikeSpreitzer: Maybe call this func `newBuilder`?

And this thing does not "build", it merely "holds" or "wraps".

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+	ageOldUnscheduled                      // older than threshold, never scheduled
+	ageOldScheduledLongAgo                 // older than threshold, scheduled long ago
+	ageOldScheduledRecently                // created long ago but scheduled recently
+)
+
+var (
+	hashKindNames = map[hashKind]string{hashMatches: "hashMatches", hashDiffers: "hashDiffers", hashAbsent: "hashAbsent"}
+	ageKindNames  = map[ageKind]string{ageYoung: "young", ageOldUnscheduled: "oldUnscheduled", ageOldScheduledLongAgo: "oldSchedLongAgo", ageOldScheduledRecently: "oldSchedRecently"}
+	allHashKinds  = []hashKind{hashMatches, hashDiffers, hashAbsent}
+	allAgeKinds   = []ageKind{ageYoung, ageOldUnscheduled, ageOldScheduledLongAgo, ageOldScheduledRecently}
```
MikeSpreitzer: BTW, you could DRY this out using `sets.KeySet` (and `.Unsortedlist` or `sets.List`, if you like) or `map.Keys` (and `slices.Collect`, if you like).

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+	// Block 4 (4): unbound + matching hash + not Ready -> depends on age.
+	wantByAge := map[ageKind]launcherPhase{
+		ageYoung:                phaseUnbound,
+		ageOldUnscheduled:       phaseStuck,
+		ageOldScheduledLongAgo:  phaseStuck,
+		ageOldScheduledRecently: phaseUnbound,
+	}
+	for _, age := range allAgeKinds {
+		run(t, false, hashMatches, false, age, wantByAge[age])
+	}
```
MikeSpreitzer: It would be shorter to write four calls to `run` here.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+func TestLauncherPhaseOfEmptyCurrentHash(t *testing.T) {
+	ctl := &controller{}
+	now := time.Now()
+	pod := newLauncherPod().scheduledAt(now).build()
+	if got := ctl.launcherPhaseOf(pod, "", DefaultStuckThreshold, now); got != phaseStale {
+		t.Errorf("launcherPhaseOf() with empty current hash = %q, want %q", got, phaseStale)
+	}
+}
+
+// gatheredPhases returns the fma_launcher_pod_count series currently present
```
MikeSpreitzer: The bare word "series" here leaves this reader wondering what sort of series is meant. After reading the body of the function, I think that perhaps what is meant is a collection of time series. But actually this function does not return a collection of time series. Remember that a time series is a _series_ of readings. What this func returns is the current reading from each of a few time series. So I would describe it something like the following.

> gatheredPhases returns a map from phase to the value currently reported by the `fma_launcher_pod_count` metric for the given LauncherConfig name.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+			}
+			if name == lcfg {
+				out[phase] = m.GetGauge().GetValue()
+			}
+		}
+	}
+	return out
+}
+
+// seriesCountForLcfg returns how many fma_launcher_pod_count series exist for
```
MikeSpreitzer: _time_ series

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+// the given LauncherConfig.
+func seriesCountForLcfg(t *testing.T, lcfg string) int {
+	t.Helper()
+	return len(gatheredPhases(t, lcfg))
+}
+
+// assertLcfgPhases asserts the exact set of phase series (values and presence)
```
MikeSpreitzer: "phase series" is not a thing.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+	// The steps below share ms/gauge state and must run in order.
+	t.Run("aggregate across nodes", func(t *testing.T) {
+		// All four phases must be present (count == 4 proves bound/stale exist as
+		// explicit zeros, not just that a value read returns 0).
+		ms.publish(keyA, phaseCounts{phaseUnbound: 2})
+		ms.publish(keyB, phaseCounts{phaseUnbound: 1, phaseStuck: 1})
+		if n := seriesCountForLcfg(t, lcfg); n != 4 {
+			t.Errorf("series count = %d, want 4 (all phases present incl. explicit zeros)", n)
+		}
+		assertLcfgPhases(t, lcfg, 0, 3, 1, 0)
```
MikeSpreitzer: Doesn't this cover the test of `seriesCountForLcfg`?
(Same question applies to other test funcs.)

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// TestComputeKeyPhases covers the two behaviors the maintainer asked for that
+// live only in the tallying path: terminating Pods are still counted (#3), and
+// the earliest future "stuck" instant is computed for AddAfter scheduling (#1).
+func TestComputeKeyPhases(t *testing.T) {
+	ctl := &controller{stuckThreshold: 7*time.Minute + 30*time.Second}
+	now := time.Date(2026, 6, 21, 12, 0, 0, 0, time.UTC)
+	recent := now.Add(-1 * time.Minute)
+	older := now.Add(-2 * time.Minute)
```
MikeSpreitzer: In English, "older" is used in contrast to "old" or "less old", to mean _more_ "old".

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,619 @@
...
+		ctl.recordLauncherPhases(key, pods, testTemplateHash)
+
+		assertLcfgPhases(t, "lc-rec", 0, 1, 0, 0) // publish happened
+		if len(q.added) != 1 {
+			t.Fatalf("AddAfter calls = %d, want 1", len(q.added))
+		}
+		if q.added[0].item != (keyItem{NodeLauncherKey: key}) {
+			t.Errorf("scheduled item = %+v, want key %+v", q.added[0].item, key)
+		}
+		if d := q.added[0].delay; d <= 0 || d > threshold {
```
MikeSpreitzer: This is why you make the controller hold and use a https://github.com/kubernetes/utils/blob/4c0f3b24339726b3d4a1b610c150919126aad841/clock/clock.go#L24 , using the implementation in https://github.com/kubernetes/utils/blob/4c0f3b24339726b3d4a1b610c150919126aad841/clock/clock.go#L85 in non-test code. Using the implementation in https://github.com/kubernetes/utils/blob/4c0f3b24339726b3d4a1b610c150919126aad841/clock/testing/fake_clock.go#L33 in tests, you can control time precisely.

pkg/controller/launcher-populator/populator.go
```diff
@@ -151,6 +155,16 @@ type controller struct {
 	// It is the single source of truth for per-key reconciliation.
 	// Written only by the digestQueue worker; read by keyQueue workers.
 	policy *digestedPolicy
+
+	// stuckThreshold is the minimum age (from scheduling, or from creation when
+	// not scheduled) after which a not-yet-Ready launcher is reported in the
+	// "stuck" phase of the fma_launcher_pod_count metric. It does not change
```
MikeSpreitzer: Off-by-1 error in this comment. This field holds the maximum age that is _not_ stuck, right?

pkg/controller/launcher-populator/populator.go
```diff
@@ -365,6 +379,20 @@ func (ctl *controller) processKey(ctx context.Context, key NodeLauncherKey) (err
 	logger := klog.FromContext(ctx)
 
 	snap := ctl.policy.snapshotForKey(key)
+
+	// Update the fma_launcher_pod_count metric for this key from the observed
+	// launcher Pods (classifying all of them, including any being deleted). This
+	// runs for every key reconcile — before the desired-state and node-existence
+	// branches below — so a LauncherConfig's series never go stale when a Node is
```
MikeSpreitzer: What do you mean by "series"? What does it mean for a "series" to "go stale"?

pkg/controller/launcher-populator/populator.go
```diff
@@ -365,6 +379,20 @@ func (ctl *controller) processKey(ctx context.Context, key NodeLauncherKey) (err
...
+	// deleted or the key becomes handsOff, and are dropped once no launcher Pods
+	// remain.
+	if pods, err := ctl.listPodsFromCache(key); err != nil {
```
MikeSpreitzer: In the common case, this duplicates the call to `ctl.listPodsFromCache` from `ctl.reconcileKey` (through `ctl.getCurrentLaunchersOnNode`). This is an expensive call. I would not do it twice.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,218 @@
...
+// metricsState is the source of truth backing launcherPodCountGauge. It holds,
+// per LauncherConfig, the per-(node, LC) key phase tallies. The gauge is
+// aggregated across nodes from this map. Keeping our own state (rather than
+// GaugeVec.Reset every period) avoids the scrape-timing splinter that Reset
+// creates and lets us delete a LauncherConfig's series only when it truly has
+// no launcher Pods left.
+type metricsState struct {
+	mu      sync.Mutex
+	perLcfg map[string]map[NodeLauncherKey]phaseCounts
+}
```
MikeSpreitzer: This is a pretty expensive solution. Rather than incrementally tracking changes to Pods, it recomputes everything for a NodeLauncherKey every sync of one.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,218 @@
...
+
+// metricsState is the source of truth backing launcherPodCountGauge. It holds,
+// per LauncherConfig, the per-(node, LC) key phase tallies. The gauge is
+// aggregated across nodes from this map. Keeping our own state (rather than
+// GaugeVec.Reset every period) avoids the scrape-timing splinter that Reset
+// creates and lets us delete a LauncherConfig's series only when it truly has
+// no launcher Pods left.
+type metricsState struct {
+	mu      sync.Mutex
+	perLcfg map[string]map[NodeLauncherKey]phaseCounts
```
MikeSpreitzer: The LauncherConfig name is duplicated in this data structure. It would be simpler if `perLcfg` were a map from lcfg name to Node name to phaseCounts.

pkg/controller/launcher-populator/metrics.go
MikeSpreitzer: I was aiming to have a time series exist if EITHER the LauncherConfig object exists OR a Launcer referencing that LauncherConfig exists.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/metrics_test.go
delavet: Renamed the constructor to `newBuilder` and dropped the "build" verb from the terminal — it's now pod(), which just returns the wrapped Pod. I kept "builder" in the type name (podBuilder) since the fluent setters do incrementally build the Pod up, I think.

cmd/launcher-populator/main.go
```diff
@@ -42,11 +43,18 @@ func main() {
...
 
+	obsOpts := fmaobs.DefaultOptions()
+
 	klog.InitFlags(flag.CommandLine)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	common.AddKubernetesClientFlags(*pflag.CommandLine, loadingRules, overrides)
 	expectationTimeout := pflag.Duration("expectation-timeout", launcherpopulator.DefaultExpectationTimeout,
 		"How long to wait for the informer cache to reflect pending Pod mutations before falling back to a direct apiserver query")
+	pendingThreshold := pflag.Duration("pending-threshold", launcherpopulator.DefaultPendingThreshold,
+		"Minimum age (since creation) after which an unscheduled, not-yet-Ready launcher is reported in the \"pending\" phase of fma_launcher_pod_count")
```
MikeSpreitzer: I would remove the not-yet-Ready qualifier from these words, because unscheduled implies not Ready.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+
+const (
+	// phaseBound: the launcher is assigned to a server-requesting Pod.
+	phaseBound launcherPhase = "bound"
+	// phaseUnbound: the launcher uses the current template, is unbound, and is
+	// either Ready or has not yet existed past its pending (unscheduled) or stuck
+	// (scheduled) threshold.
+	phaseUnbound launcherPhase = "unbound"
+	// phasePending: the launcher uses the current template, is unbound, is not
+	// Ready, has not been scheduled, and has existed (since creation) past the
```
MikeSpreitzer: You do not have to include the words bout "not ready" here, since that is implied by "not been scheduled".

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+	// pending threshold. It is stuck getting scheduled.
+	phasePending launcherPhase = "pending"
```
MikeSpreitzer: The term "pending" here is likely to be confusing because the same term is already well known from the STATUS column of the tabular output from `kubectl get pods` to mean something else.

I do not like the verbosity, but my suggestion is to call this phase "stuck_scheduling" (or "stuck_pending") and the other one "stuck_starting".

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	return sum
+}
+
+// metricsState is the source of truth backing launcherPodCountGauge: it holds
+// the per-LauncherConfig launcher tallies and republishes the gauge as they
+// change.
+type metricsState struct {
+	mu sync.Mutex
+	// perNode is the phase tally of each Node with launchers, per LauncherConfig.
```
MikeSpreitzer: It would be much better to be clear about what each of the two strings is.
Reading the current comment and the field name, I inferred that the first string is the node name and the second is the lcfg name. But the code below suggests otherwise.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// publish records the phase tally for one (node, LC) key and republishes the
+// affected LauncherConfig's gauge series.
+func (ms *metricsState) publish(key NodeLauncherKey, counts phaseCounts) {
+	lcfg, node := key.LauncherConfigName, key.NodeName
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
+
+	nodes := ms.perNode[lcfg]
```
MikeSpreitzer: Since this is not a set of nodes, "nodes" is not a great name for this variable.
Also, since it is indexed by lcfg name rather than Node name, `perNode` is a confusing choice for that field name. I think better names would produce lines more like one of the following.

```go
perNode := ms.perLCFG[lcfg]
byNode := ms.byLCFG[lcfg]
countsPerNode := ms.perLCFG[lcfg]
```

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+// change.
+type metricsState struct {
+	mu sync.Mutex
+	// perNode is the phase tally of each Node with launchers, per LauncherConfig.
+	perNode map[string]map[string]phaseCounts
+	// agg is the per-LauncherConfig sum of perNode across Nodes, kept in sync
+	// incrementally so publish need not re-sum every Node.
+	agg map[string]phaseCounts
+	// lcExists is the set of LauncherConfig names whose object currently exists.
+	lcExists map[string]bool
```
MikeSpreitzer: This is a set of strings. Better to use `sets.Set[string]`.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+
+// republishLocked rewrites one LauncherConfig's gauge series from the current
+// aggregate, or deletes them once neither its object nor any launcher remains.
+// Callers must hold ms.mu.
+func (ms *metricsState) republishLocked(lcfg string) {
+	if !ms.lcExists[lcfg] && len(ms.perNode[lcfg]) == 0 {
+		delete(ms.agg, lcfg)
+		for _, phase := range allLauncherPhases {
+			launcherPodCountGauge.Delete(map[string]string{lcfgNameLabel: lcfg, phaseLabel: string(phase)})
+		}
```
MikeSpreitzer: FYI, see https://github.com/kubernetes/component-base/blob/v0.34.9/metrics/gauge.go#L91 . It means that a Kubernetes `GaugeVec` has all the methods of the embedded Prometheus `GaugeVec`. So you _can_ write `launcherPodCountGauge.DeletePartialMatch`.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+	// launcher still surfaces, as pending).
+	if cond := utils.GetPodCondition(pod, corev1.PodScheduled); cond != nil && cond.Status == corev1.ConditionTrue {
+		return phaseByAge(cond.LastTransitionTime.Time, stuckThreshold, phaseStuck, now)
+	}
+	return phaseByAge(pod.CreationTimestamp.Time, pendingThreshold, phasePending, now)
+}
+
+// phaseByAge returns overduePhase when the launcher has been not-yet-Ready past
+// threshold (measured from ref); otherwise it returns phaseUnbound plus the
+// future instant ref+threshold at which it would become overdue.
```
MikeSpreitzer: Again, boundary wording error in this comment and on computeKeyPhases. The boundary time is the last non-stuck time, rather than the first stuck time.

pkg/controller/launcher-populator/populator.go
```diff
@@ -506,32 +548,28 @@ func (ctl *controller) reconcileKey(ctx context.Context, key NodeLauncherKey, de
 	return nil, false
 }
 
-// getCurrentLaunchersOnNode returns launcher pods for a specific config on a specific node.
-// It reads the informer cache and uses the resulting UID set to reconcile any
-// pending expectations, then returns one of:
+// getCurrentLaunchersOnNode reconciles pending expectations for a specific
```
MikeSpreitzer: The word "reconciles" here is confusing, since that is one of the words used (interchangeably with "sync" and "process") to identify what a controller does per item in a workqueue. I think that perhaps a word like "compares" would be better.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,687 @@
...
+	if err != nil {
+		t.Fatalf("gathering metrics: %v", err)
+	}
+	out := map[string]float64{}
+	for _, mf := range mfs {
+		if mf.GetName() != "fma_launcher_pod_count" {
+			continue
+		}
+		for _, m := range mf.GetMetric() {
+			var name, phase string
```
MikeSpreitzer: IMO "name" is a pretty uninformative variable name. I might use "thisLCFG" or "foundLCFG" or "mLCFG".

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,687 @@
...
+		assertLcfgPhases(t, lcfg, 0, 2, 0, 1, 0)
+	})
+
+	t.Run("removing the last launcher then drops the series", func(t *testing.T) {
+		ms.publish(key, phaseCounts{})
+		assertLcfgAbsent(t, lcfg)
+	})
+}
+
+// TestMetricsStateLCObjectOutlivesLaunchers covers the other order: a launcher's
```
MikeSpreitzer: There are two orders for the upswing as well as two orders for going down. Something should test the case of launchers referencing LauncherConfigs that no longer exist.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,270 @@
...
+// unbound, not-yet-Ready launcher on the current template would cross into
+// pending or stuck (zero Time if none). Pods being deleted are counted (the
+// metric counts Pod objects that exist) but never schedule a future transition.
+func (ctl *controller) computeKeyPhases(pods []*corev1.Pod, templateHash string, now time.Time) (phaseCounts, time.Time) {
+	counts := phaseCounts{}
+	var earliestTransition time.Time
+	for _, pod := range pods {
+		phase, becomesOverdueAt := ctl.launcherPhaseOf(pod, templateHash, ctl.pendingThreshold, ctl.stuckThreshold, now)
+		counts[phase]++
+		if pod.DeletionTimestamp == nil && becomesOverdueAt.After(now) &&
```
MikeSpreitzer, line 280: Testing the strict inequality `becomesOverdueAt > now` here means that when `becomesOverdueAt == now` there will not be a rescheduling to reevaluate the Pod later --- which is wrong, since later reevaluation is needed to detect the Pod becoming stuck.

Note that time passes quickly in Go. The clock has nanosecond precision, and typically does use all of that precision. You do not have to worry about too much work being done without the clock advancing.

OTOH, if the boundary were defined in the other way --- the threshold being the shortest duration that is stuck rather than the longest duration that is not stuck --- then you would not even have to worry about this.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,687 @@
...
+		// Only the not-Ready, counting-down unbound pod drives earliestStuck.
+		want := recent.Add(ctl.stuckThreshold)
+		if !earliest.Equal(want) {
+			t.Errorf("earliestStuck=%v want %v", earliest, want)
+		}
+	})
+}
+
+// recordingQueue is a keyQueue stand-in that captures AddAfter calls. Embedding
+// the interface leaves all other methods nil; recordLauncherPhases only calls
```
MikeSpreitzer: "leaves all the other methods nil"?? Embedding the work queue object means that recordingQueue inherits all the methods of the work queue object except the ones specifically overridden for recordingQueue. What's nil is the work queue _value_ in the construction `recordingQueue{}`.

MikeSpreitzer (commented):

pkg/controller/launcher-populator/populator.go
delavet: I’ve revised the wording— please review it and confirm whether this version is clear?
```
// recordingQueue records AddAfter calls for TestRecordLauncherPhases. Embedding
// TypedRateLimitingInterface supplies the rest of the interface's method set.
// Tests leave the embedded interface nil, which is safe because
// recordLauncherPhases only calls the explicitly defined AddAfter method.
```

pkg/controller/launcher-populator/metrics_test.go
delavet: Add `TestMetricsStateLauncherPrecedesLCObject` to test this.

pkg/controller/launcher-populator/metrics.go
```diff
@@ -0,0 +1,276 @@
...
+
+// launcherPhaseOf classifies a launcher Pod into a phase (see the launcherPhase
+// constants) and, for one still counting down toward stuck-scheduling or
+// stuck-starting, also returns the instant at which it will reach that threshold
+// (zero Time if none).
+//
+// currentTemplateHash is the node-independent template hash the digest wants for
+// this Pod's LauncherConfig; an empty value (LC gone / not yet digested) makes
+// any templated Pod stale.
+func (ctl *controller) launcherPhaseOf(pod *corev1.Pod, currentTemplateHash string, stuckSchedulingThreshold, stuckStartingThreshold time.Duration, now time.Time) (launcherPhase, time.Time) {
```
MikeSpreitzer: It is odd for this method to take both (a) the `controller` object, which has the thresholds as fields, and (b) the thresholds as separate parameters. I recommend refactoring so that the thresholds arrive only one way.

I see that `ctl.isLauncherBoundToServerRequestingPod` is a _method_ of the controller, rather than a free-standing function, but does not currently use the controller object for anything. I see a comment in the body about possibly using the controller in there. One option to consider is changing `isLauncherBoundToServerRequestingPod` to a simple function, which would be viable at present.

pkg/controller/launcher-populator/metrics.go
MikeSpreitzer: I see that you switched the boundary to be define the other way: the threshold is now the youngest age at which the Pod will be stuck.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,758 @@
...
+		{name: "stuck starting", phase: phaseStuckStarting, want: "stuck_starting"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := string(tt.phase); got != tt.want {
+				t.Errorf("phase label = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
```
MikeSpreitzer: I see no need to maintain this test. It only proves something that can be seen by inspection of two lines of code (the definitions of `phaseStuckScheduling` and `phaseStuckStarting`). And CI has already successfully run this case at least once.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,758 @@
...
+	ms := newMetricsState()
+	ms.publish(NodeLauncherKey{NodeName: "n", LauncherConfigName: "never"}, phaseCounts{})
+	if len(ms.perLCFG) != 0 {
+		t.Errorf("perLCFG should stay empty, got %v", ms.perLCFG)
+	}
+	assertLcfgAbsent(t, "never")
+}
+
+// TestMetricsStateDeletesAllLCFGSeries verifies that removing a LauncherConfig
+// deletes every phase series for it, including values unknown to this version's
```
MikeSpreitzer: `ms.setLCExists(lcfg, false)` does NOT necessarily delete every timeseries for that lcfg --- that depends on whether `ms` holds non-zero data for any phase.

This test, as currently written, passes because it is _also_ true that `ms` holds no non-zero count for any phase (at that lcfg).

MikeSpreitzer: The testing in this case wrt an unknown phase value is surprising to me. It does NOT test what `ms` does if given a count for an unknown phase. It DOES test whether `ms` stomps, but just in one particular scenario, on behavior of something else that is writing to its own distinct time series of the same metric. I would not expect something else to be writing to its own distinct time series of the same metric --- but if there were such a thing happening, I would expect `ms` to leave those entirely alone.

BTW, if `phaseCounts` were a struct rather than a map then it would be syntactically impossible to give `ms` an unexpected phase value. It would also make this feature less expensive at runtime.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,758 @@
...
+	const lcfg = "lc-exists"
+	key := NodeLauncherKey{NodeName: "n1", LauncherConfigName: lcfg}
+
+	// Steps share ms/gauge state and must run in order.
+	t.Run("LC object with no launchers reports explicit zeros", func(t *testing.T) {
+		ms.setLCExists(lcfg, true)
+		assertLcfgPhases(t, lcfg, 0, 0, 0, 0, 0)
+	})
+
+	t.Run("launchers add on top of the existing zeros", func(t *testing.T) {
```
MikeSpreitzer: The description here is strange: there is no difference in outcome between adding 0 to 0 vs. replacing 0 with 0.

pkg/controller/launcher-populator/metrics_test.go
MikeSpreitzer: Now there are tests for three of the four possible combinations of the two upswing orders and the two downswing orders. This is not wrong, but having the comment here refer simply to "the other order" is vague/confusing.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -0,0 +1,758 @@
...
+		counts, earliest := ctl.computeKeyPhases(pods, testTemplateHash, now)
+		if counts.total() != len(pods) {
+			t.Errorf("total=%d want %d (every pod must be counted)", counts.total(), len(pods))
+		}
+		for phase, want := range map[launcherPhase]int{phaseBound: 1, phaseStale: 1, phaseStuckStarting: 1, phaseUnbound: 2} {
+			if counts[phase] != want {
+				t.Errorf("counts[%s]=%d want %d", phase, counts[phase], want)
+			}
+		}
+		// Only the not-Ready, counting-down unbound pod drives earliestTransition.
```
MikeSpreitzer, line 656: Since `recent` appears in two of the Pods, this test does not really discriminate which among them mattered.

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: I am confused. I wrote about the comment on `getCurrentLaunchersOnNode` and you responded by writing about the comment on `recordingQueue`.

Both comments are now good.

MikeSpreitzer (commented):
I have completed another round of review.

MikeSpreitzer (dismissed):
This is good enough to merge. There are some open comment threads suggesting improvements, which could be done in this PR or a later one.

delavet:
> This is good enough to merge. There are some open comment threads suggesting improvements, which could be done in this PR or a later one.

I have addressed these new comments within this PR. Could you please review them again? Thank you!

pkg/controller/launcher-populator/metrics.go
```diff
@@ -96,18 +89,56 @@ func registerMetrics() {
...
+		pc.stale++
+	default:
+		panic("unsupported launcher phase: " + string(phase))
 	}
-	return sum
+}
+
+// forEach calls yield once for every supported metric phase, including phases
+// whose count is zero.
+func (pc phaseCounts) forEach(yield func(launcherPhase, int)) {
```
MikeSpreitzer, line 136: FYI, this is _almost_ what is needed to make `forEach` usable as the range expression in a `for` statement; see the "function, 2 values" case in https://go.dev/ref/spec#For_statements.

pkg/controller/launcher-populator/metrics_test.go
```diff
@@ -657,20 +639,19 @@ func TestComputeKeyPhases(t *testing.T) {
...
 		counts, earliest := ctl.computeKeyPhases(pods, testTemplateHash, now)
 		if counts.total() != len(pods) {
 			t.Errorf("total=%d want %d (every pod must be counted)", counts.total(), len(pods))
 		}
-		for phase, want := range map[launcherPhase]int{phaseBound: 1, phaseStale: 1, phaseStuckStarting: 1, phaseUnbound: 2} {
-			if counts[phase] != want {
-				t.Errorf("counts[%s]=%d want %d", phase, counts[phase], want)
-			}
+		wantCounts := phaseCounts{bound: 1, unbound: 2, stuckStarting: 1, stale: 1}
+		if !reflect.DeepEqual(counts, wantCounts) {
```
MikeSpreitzer, line 653: FYI, this can use ordinary `==` now.

MikeSpreitzer (approved):
This is good enough to merge. I left a few suggestions for improvement, which can be done in follow-on PR(s).

PR #644, opened 2026-07-08
Postpone label application from ISC to launchers

### LINKED ISSUE #629: EPP Sending Requests to not ready vLLM instances inside a Launcher Pod
In the FMA path, the dual-pods controller decides when the launcher pod is bound and applies the variant and inferencePool labels, and the launcher pod's readiness reflects the launcher control-plane API, not the just-spawned vLLM instance behind it. So the pod becomes a routing target before vLLM's port is up, and EPP forwards into a closed socket → `503`.
  
The gap it's that the launcher pod's readiness signal doesn't reflect whether the specific vLLM instance just spawned on it is serving.

As result of this issue with see significant failed requests in the WVA+FMA benchmarking experiments.

|  | FMA+WVA run | WVA-only run |
  | --- | --- | --- |
  | Backend container | FMA launcher | vLLM container directly |
  | Startup probe on the inference port (`:8000`)? | No — launcher's readiness was its own control-plane API | Yes — `httpGet /v1/models :8000` |
  | Pod admitted to Service/InferencePool while vLLM port not ready? | Yes (immediately after dual-pods controller bound the requester ↔ launcher and applied the variant and inferencepool labels) | No — pod stays `NotReady` until `/v1/models` returns 200 |
  | Result | EPP routed traffic to launcher → vLLM TCP refused → HTTP 503 | EPP never saw the new pod as an endpoint until vLLM was serving → zero scale-up failures |

**See example below**: 

Second requester pod is not ready yet and any requests sent to the EPP envoy fails. 

```
kubectl -n braulio-wva-fma  get pods  -o wide
NAME                                                              READY   STATUS      RESTARTS   AGE     IP             NODE               NOMINATED NODE   READINESS GATES
access-to-harness-data-workload-pvc                               1/1     Running     0          161m    10.130.4.34    pokprod-b93r43s3   <none>           <none>
download-model-fmtz2                                              0/1     Completed   0          163m    10.130.4.33    pokprod-b93r43s3   <none>           <none>
fma-requester-qwen-qwe-df5cf70e-wen3-32b-86c777fc-j8lrv           0/1     Running     0          99s     10.128.5.56    pokprod-b93r44s2   <none>           <none>
fma-requester-qwen-qwe-df5cf70e-wen3-32b-86c777fc-ntc7l           1/1     Running     0          150m    10.129.7.53    pokprod-b93r39s0   <none>           <none>
inference-perf-89yk1spi                                           1/1     Running     0          4m35s   10.129.4.81    pokprod-b93r39s1   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-2z6wx                     2/2     Running     0          161m    10.129.7.49    pokprod-b93r39s0   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-b9lgk                     2/2     Running     0          161m    10.130.4.35    pokprod-b93r43s3   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-ddtlf                     2/2     Running     0          161m    10.128.11.39   pokprod-b93r38s0   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-fgnhc                     2/2     Running     0          161m    10.131.6.244   pokprod-b93r43s1   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-fsqtc                     2/2     Running     0          161m    10.130.2.169   pokprod-b93r44s0   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-hdmmk                     2/2     Running     0          161m    10.128.5.2     pokprod-b93r44s2   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-jj2mx                     2/2     Running     0          161m    10.131.4.234   pokprod-b93r44s3   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-k2zgx                     2/2     Running     0          161m    10.129.4.69    pokprod-b93r39s1   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-mxs5t                     2/2     Running     0          161m    10.130.8.30    pokprod-b93r43s2   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-rv9vg                     2/2     Running     0          161m    10.129.8.182   pokprod-b93r39s2   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-szrsz                     2/2     Running     0          161m    10.130.7.182   pokprod-b93r43s0   <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-w5zp9                     0/2     Pending     0          161m    <none>         <none>             <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-xmxjq                     0/2     Pending     0          161m    <none>         <none>             <none>           <none>
launcher-fma-qwen-qwe-df5cf70e-wen3-32b-xxrc7                     2/2     Running     0          161m    10.131.3.248   pokprod-b93r44s1   <none>           <none>
qwen-qwe-df5cf70e-wen3-32b-fma-dp-dual-pods-controller-78bm7fz8   1/1     Running     0          161m    10.128.11.37   pokprod-b93r38s0   <none>           <none>
qwen-qwe-df5cf70e-wen3-32b-fma-dp-launcher-populator-85786qpldt   1/1     Running     0          161m    10.128.11.38   pokprod-b93r38s0   <none>           <none>
qwen-qwe-df5cf70e-wen3-32b-router-epp-8c5cb8ffc-prszr             2/2     Running     0          13m     10.128.5.43    pokprod-b93r44s2   <none>           <none>
wva-controller-manager-6dc99f6df5-mgm99                           1/1     Running     0          163m    10.128.11.35   pokprod-b93r38s0   <none>           <none>
```

Snapshot from logs of EPP envoy container: 

```{"time":"2026-07-01T15:17:00.310Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"9fd0006f-d58c-44d0-a6e2-82bdcbc6f75d","duration_ms":3322}
{"time":"2026-07-01T15:17:00.911Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"4986e87e-e508-4ab6-b12a-622abd19d297","duration_ms":2722}
{"time":"2026-07-01T15:17:02.623Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"d04e30e9-69d7-4ac5-91e3-54a97cef53b6","duration_ms":1011}
{"time":"2026-07-01T15:17:03.094Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"89b8c490-4ba2-42b9-868d-5853d9957452","duration_ms":541}
{"time":"2026-07-01T15:17:04.050Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"2a81156f-c13a-4029-b33e-44c9317f72e9","duration_ms":2}
{"time":"2026-07-01T15:17:04.257Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"a5503b03-3047-47f8-8dfb-d182811c7c5c","duration_ms":2}
{"time":"2026-07-01T15:17:04.282Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"d3c3e736-0777-425d-875a-7fcb6e0339d2","duration_ms":2}
{"time":"2026-07-01T15:17:05.989Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"af6974c1-97cc-4caa-b364-1189823fa5fb","duration_ms":2}
{"time":"2026-07-01T15:17:07.356Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"f35914ef-9871-48e9-a010-bea00b69e94b","duration_ms":1188}
{"time":"2026-07-01T15:17:07.360Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"fb1f5fe3-da09-40d3-b538-167a3f75367a","duration_ms":1185}
{"time":"2026-07-01T15:17:07.625Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"21746df4-058c-4968-998d-e1e31a0cd006","duration_ms":921}
{"time":"2026-07-01T15:17:07.631Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"2f9eb918-296b-49f9-a798-4cf2d48b2a17","duration_ms":916}
{"time":"2026-07-01T15:17:07.704Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"e38ef6f9-27ae-43df-aa1c-314b52cd844a","duration_ms":845}
{"time":"2026-07-01T15:17:08.029Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"39960c6a-69f2-4ab9-8aef-4c8fe94991f7","duration_ms":521}
{"time":"2026-07-01T15:17:08.120Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"a8dc127e-360f-4b71-9727-02b255768ad7","duration_ms":430}
{"time":"2026-07-01T15:17:08.346Z","status":503,"flags":"UF","upstream_host":"10.128.5.2:8000","request_id":"542594a3-2a12-497a-b824-d7b598bc6bb5","duration_ms":206}
```

waltforme:
Fixes #629

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -100,12 +100,11 @@ const nominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
 const launcherInstanceIDAnnotationKey = "dual-pods.llm-d.ai/instance-id"
 const launcherServerPortAnnotationKey = "dual-pods.llm-d.ai/server-port"
 const launcherVllmConfigAnnotationKey = "dual-pods.llm-d.ai/vllm-config"
-const iscLabelKeysAnnotationKey = "dual-pods.llm-d.ai/isc-label-keys"
-const iscAnnotationKeysAnnotationKey = "dual-pods.llm-d.ai/isc-annotation-keys"
 
 // launcherISCRoutingMetadataAnnotationKey holds, as JSON, the ISC-provided
 // routing labels and annotations the bound instance was created with, so they
-// can be applied once it is serving and recovered after a restart.
+// can be applied once it is serving and recovered after a restart. Their keys
+// are also what identifies the routing labels/annotations to remove on unbind.
 const launcherISCRoutingMetadataAnnotationKey = "dual-pods.llm-d.ai/isc-routing-metadata"
 
 const providerFinalizer = "dual-pods.llm-d.ai/provider"
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -445,14 +445,24 @@ type serverData struct {
 
 	ProvidingPodName string
 
-	// The next two fields form a snapshot of the bound launcher-based
+	// The next four fields form a snapshot of the bound launcher-based
 	// vLLM instance's ISC-derived state. Each field may be reassigned
-	// over time, but whichever value (pointer) is
+	// over time, but whichever value (pointer/map) is
 	// currently stored is deeply immutable for as long as it is stored.
-	InstanceID     string
-	InstanceConfig *VllmConfig
+	// InstanceISCLabels and InstanceISCAnnotations are cached here so the
+	// controller can apply them to the launcher Pod only after the vLLM
+	// instance is confirmed serving; see LabelsApplied.
+	InstanceID             string
+	InstanceConfig         *VllmConfig
+	InstanceISCLabels      map[string]string
+	InstanceISCAnnotations map[string]string
 
 	InstanceKnownToExist bool // meaningful only for launcher-based providers
+
+	// LabelsApplied is true iff the ISC-provided labels and annotations
+	// have been written to the bound launcher Pod. Meaningful only for
+	// launcher-based providers.
+	LabelsApplied bool
 
 	DualityMetricAsserted bool
 
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -445,17 +445,12 @@ type serverData struct {
 
 	ProvidingPodName string
 
-	// The next four fields form a snapshot of the bound launcher-based
+	// The next two fields form a snapshot of the bound launcher-based
 	// vLLM instance's ISC-derived state. Each field may be reassigned
-	// over time, but whichever value (pointer/map) is
+	// over time, but whichever value (pointer) is
 	// currently stored is deeply immutable for as long as it is stored.
-	// InstanceISCLabels and InstanceISCAnnotations are cached here so the
-	// controller can apply them to the launcher Pod only after the vLLM
-	// instance is confirmed serving; see LabelsApplied.
-	InstanceID             string
-	InstanceConfig         *VllmConfig
-	InstanceISCLabels      map[string]string
-	InstanceISCAnnotations map[string]string
+	InstanceID     string
+	InstanceConfig *VllmConfig
 
 	InstanceKnownToExist bool // meaningful only for launcher-based providers
 
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -103,6 +103,11 @@ const launcherVllmConfigAnnotationKey = "dual-pods.llm-d.ai/vllm-config"
 const iscLabelKeysAnnotationKey = "dual-pods.llm-d.ai/isc-label-keys"
 const iscAnnotationKeysAnnotationKey = "dual-pods.llm-d.ai/isc-annotation-keys"
 
+// launcherISCRoutingMetadataAnnotationKey holds, as JSON, the ISC-provided
+// routing labels and annotations the bound instance was created with, so they
+// can be applied once it is serving and recovered after a restart.
+const launcherISCRoutingMetadataAnnotationKey = "dual-pods.llm-d.ai/isc-routing-metadata"
+
 const providerFinalizer = "dual-pods.llm-d.ai/provider"
 const requesterFinalizer = "dual-pods.llm-d.ai/requester"
 
@@ -445,18 +450,21 @@ type serverData struct {
 
 	ProvidingPodName string
 
-	// The next two fields form a snapshot of the bound launcher-based
-	// vLLM instance's ISC-derived state. Each field may be reassigned
-	// over time, but whichever value (pointer) is
-	// currently stored is deeply immutable for as long as it is stored.
-	InstanceID     string
-	InstanceConfig *VllmConfig
+	// Snapshot of the bound launcher-based instance's ISC-derived state.
+	// A field may be reassigned, but whichever value is stored is deeply
+	// immutable while stored. InstanceISCLabels/InstanceISCAnnotations are the
+	// routing metadata the instance was bound with (applied once it is serving,
+	// see LabelsApplied), reflecting the binding-time commitment rather than the
+	// current InferenceServerConfig.
+	InstanceID             string
+	InstanceConfig         *VllmConfig
+	InstanceISCLabels      map[string]string
+	InstanceISCAnnotations map[string]string
 
 	InstanceKnownToExist bool // meaningful only for launcher-based providers
 
-	// LabelsApplied is true iff the ISC-provided labels and annotations
-	// have been written to the bound launcher Pod. Meaningful only for
-	// launcher-based providers.
+	// LabelsApplied is true iff the ISC-provided labels and annotations have
+	// been written to the bound launcher Pod. Launcher-based providers only.
 	LabelsApplied bool
 
 	DualityMetricAsserted bool
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -101,10 +101,12 @@ const launcherInstanceIDAnnotationKey = "dual-pods.llm-d.ai/instance-id"
 const launcherServerPortAnnotationKey = "dual-pods.llm-d.ai/server-port"
 const launcherVllmConfigAnnotationKey = "dual-pods.llm-d.ai/vllm-config"
 
-// launcherISCRoutingMetadataAnnotationKey holds, as JSON, the ISC-provided
-// routing labels and annotations the bound instance was created with, so they
-// can be applied once it is serving and recovered after a restart. Their keys
-// are also what identifies the routing labels/annotations to remove on unbind.
+// launcherISCRoutingMetadataAnnotationKey holds, as the JSON encoding of an
+// iscRoutingMetadata, the ISC-provided routing labels and annotations the bound
+// instance was created with, so they can be applied once it is serving.
+// Externalizing it onto the Pod conveys it reliably from the binding
+// transaction to the later moment when the controller reacts to the instance
+// reaching readiness.
 const launcherISCRoutingMetadataAnnotationKey = "dual-pods.llm-d.ai/isc-routing-metadata"
 
 const providerFinalizer = "dual-pods.llm-d.ai/provider"
--- a/pkg/controller/dual-pods/inference-server.go
+++ b/pkg/controller/dual-pods/inference-server.go
@@ -497,27 +497,21 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 			logger.V(2).Info("Determined whether provider is sleeping", "isSleeping", sleeping)
 			serverDat.Sleeping = &sleeping
 		}
-		if launcherBased {
-			// Wake if needed (network only), then reconcile the sleeping label and
-			// the deferred ISC routing labels/annotations in a single Pod update.
-			// The routing labels are applied only now that the instance is serving,
-			// so the launcher is not an InferencePool endpoint before it can serve.
-			if *(serverDat.Sleeping) {
-				if err = ctl.wakeUp(ctx, serverDat, requestingPod, providingPod, serverPort, "discovered-bound"); err != nil {
-					return err, true
-				}
+		if *(serverDat.Sleeping) {
+			if err = ctl.wakeUp(ctx, serverDat, requestingPod, providingPod, serverPort, "discovered-bound"); err != nil {
+				return err, true
 			}
+		}
+		// Reconcile the sleeping label; for a launcher this also applies the
+		// deferred ISC routing labels/annotations in the same update, now that
+		// the instance is serving, so the launcher is not an InferencePool
+		// endpoint before it can serve.
+		if launcherBased {
 			var stop bool
 			if providingPod, stop, err = ctl.applyBoundLauncherLabels(ctx, requestingPod, serverDat, providingPod); stop {
 				return err, err != nil
 			}
 		} else {
-			if *(serverDat.Sleeping) {
-				providingPod, err = ctl.wakeSleeper(ctx, serverDat, requestingPod, providingPod, serverPort, "discovered-bound")
-				if err != nil {
-					return err, true
-				}
-			}
 			if providingPod, err = ctl.ensureSleepingLabel(ctx, providingPod, *(serverDat.Sleeping)); err != nil {
 				return err, true
 			}
@@ -1172,14 +1166,14 @@ func (ctl *controller) applyBoundLauncherLabels(
 		if problems := applyISCRoutingMetadataToLauncherPod(updated, serverDat.InstanceISCLabels, serverDat.InstanceISCAnnotations); len(problems) > 0 {
 			// The committed metadata is invalid — not transient (it implies
 			// tampering or a validation change since binding), so retrying would
-			// loop. Unbind by deleting the requester Pod; a fresh bind starts over.
-			logger.Error(nil, "Committed ISC routing metadata is invalid; deleting requester Pod to unbind",
-				"requester", requestingPod.Name, "problems", problems)
+			// loop. Unbind by deleting the requester Pod; a fresh requester
+			// starts over.
+			logger.Error(nil, "Committed ISC routing metadata is invalid; deleting requester Pod to unbind", "problems", problems)
 			delErr := ctl.coreclient.Pods(ctl.namespace).Delete(ctx, requestingPod.Name, metav1.DeleteOptions{
 				PropagationPolicy: ptr.To(metav1.DeletePropagationBackground),
-				Preconditions:     &metav1.Preconditions{UID: &requestingPod.UID, ResourceVersion: &requestingPod.ResourceVersion}})
+				Preconditions:     &metav1.Preconditions{UID: &requestingPod.UID}})
 			if delErr != nil && !apierrors.IsNotFound(delErr) && !apierrors.IsGone(delErr) {
-				return providingPod, true, fmt.Errorf("failed to delete requester Pod %q after invalid ISC routing metadata: %w", requestingPod.Name, delErr)
+				return providingPod, true, fmt.Errorf("failed to delete requester Pod %q (provider %q) after invalid ISC routing metadata: %w", requestingPod.Name, providingPod.Name, delErr)
 			}
 			serverDat.RequesterDeleteRequested = true
 			return providingPod, true, nil
@@ -1274,17 +1268,8 @@ func recoverInstanceStateFromLauncherPod(serverDat *serverData, providingPod *co
 // and annotations are already present with matching values on the Pod. Empty
 // metadata (nothing to apply) counts as present.
 func iscRoutingMetadataPresent(providingPod *corev1.Pod, routing iscRoutingMetadata) bool {
-	for k, v := range routing.Labels {
-		if actual, found := providingPod.Labels[k]; !found || actual != v {
-			return false
-		}
-	}
-	for k, v := range routing.Annotations {
-		if actual, found := providingPod.Annotations[k]; !found || actual != v {
-			return false
-		}
-	}
-	return true
+	return utils.MapIsSubset(routing.Labels, providingPod.Labels) &&
+		utils.MapIsSubset(routing.Annotations, providingPod.Annotations)
 }
 
 func getVLLMInstancePort(inst InstanceState) (int32, error) {
@@ -1347,7 +1332,7 @@ func (ctl *controller) removeISCRoutingLabels(ctx context.Context, providingPod 
 	if err != nil {
 		return providingPod, fmt.Errorf("failed to remove ISC routing labels from launcher Pod %s (started %s): %w", providingPod.Name, updStartStr, err)
 	}
-	klog.FromContext(ctx).V(2).Info("Removed ISC routing labels from launcher Pod before sleeping", "name", echo.Name, "removed", toRemove, "newResourceVersion", echo.ResourceVersion, "k8sCallStartTime", updStartStr)
+	klog.FromContext(ctx).V(2).Info("Removed ISC routing labels from launcher Pod before sleeping", "removed", toRemove, "newResourceVersion", echo.ResourceVersion, "k8sCallStartTime", updStartStr)
 	return echo, nil
 }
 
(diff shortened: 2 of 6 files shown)


pkg/controller/dual-pods/inference-server.go
```diff
@@ -1093,9 +1140,96 @@ func commitInstanceState(serverDat *serverData, state *vllmInstanceState) {
 	serverDat.ServerPort = state.serverPort
 }
 
+// applyDeferredISCRoutingMetadata is called from the bound-provider reconcile
+// path once the bound vLLM instance is confirmed serving, to write the
+// ISC-provided labels and annotations onto the launcher Pod. Deferring these
+// until the instance is serving keeps the Pod from becoming an
+// InferencePool-matching routing target while its data-plane port (:8000) is
```
MikeSpreitzer: The inference-serving port is not necessarily 8000. I think that it is clearer to identify the port in question by the fact that it serves inference requests.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1093,9 +1140,96 @@ func commitInstanceState(serverDat *serverData, state *vllmInstanceState) {
...
+// the requester deletion to be observed).
+//
+// If the InferenceServerConfig was edited between bind and readiness so that
+// its current hash no longer matches the bound instance, that instance is
+// obsolete and can never carry the current routing labels. Leaving it in place
+// would strand a Ready-but-unrouted instance: the instance-GC path only reaps
+// obsolete instances that are sleeping, and this one is awake. Instead the
+// requester Pod is deleted so a fresh one rebinds against the current ISC; the
+// standard unbind flow (maybeDeleteObsoleteInstance) then removes the obsolete
+// instance from the launcher.
```
MikeSpreitzer: I think that this is contrary to our chosen answer A2 to Q4 in Issue #201 : "Each corresponding inference server that is sleeping is deleted". The vllm instance in question was not really sleeping, it was starting up. To be fair, the discussion in #201 did not explicitly cover the startup period. But I think that the intent was clear: delete the ones in which there is no current interest (no current requester Pod). A vllm instance in the process of starting up exists because of interest (a current requester Pod), and should not be discarded. Also, note that the launcher metadata made a commitment to an inference server config at binding time (Yeah, I know that not enough information is currently being saved there; that has to be extended); that should be recovered from the pod and used here, not computed afresh from the current InferenceServerConfig object.

test/e2e/test-cases.sh
```diff
@@ -234,9 +234,13 @@ echo "Launcher Pod is $launcher1"
...
-# Verify ISC labels and annotations were propagated to launcher
-[ "$(kubectl get pod -n "$NS" $launcher1 -o jsonpath='{.metadata.labels.e2e-test\.fma\.llm-d\.ai/isc-label}')" == "test-value" ] || { echo "ERROR: ISC label not propagated to launcher pod $launcher1"; exit 1; }
-[ "$(kubectl get pod -n "$NS" $launcher1 -o jsonpath='{.metadata.annotations.e2e-test\.fma\.llm-d\.ai/isc-annotation}')" == "test-value" ] || { echo "ERROR: ISC annotation not propagated to launcher pod $launcher1"; exit 1; }
+# Verify ISC labels and annotations are eventually propagated to the launcher
+# pod. Since issue #629 the dual-pods controller waits for the vLLM instance
+# to reply positively to its readiness test before writing these, so the
+# labels appear some time after the requester ↔ launcher binding — poll for
+# them rather than checking once.
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.e2e-test\\.fma\\.llm-d\\.ai/isc-label})" == "test-value" ]'
+expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.annotations.e2e-test\\.fma\\.llm-d\\.ai/isc-annotation})" == "test-value" ]'
```
MikeSpreitzer: This change changes the timing of the tests in following lines, delaying when they are done. I think that it would be better to move the one-shot tests to appear after the wait for readiness of the requester Pod.

This comment applies in both places where this issue occurs.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -732,6 +747,7 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 	}
 	serverDat.Sleeping = nil
 	commitInstanceState(serverDat, desiredInstanceState)
+	serverDat.LabelsApplied = false
```
MikeSpreitzer: We statically know that this statement is a no-op, right?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -503,6 +503,17 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
...
+		// not treat it as an endpoint. See issue #629.
+		if launcherBased && !*serverDat.Sleeping && !serverDat.LabelsApplied {
+			stop, err := ctl.applyDeferredISCRoutingMetadata(ctx, requestingPod, serverDat, providingPod, iscName)
+			if stop {
+				return err, err != nil
+			}
+		}
 		if err := ctl.ensureSleepingLabel(ctx, providingPod, *(serverDat.Sleeping)); err != nil {
 			return err, true
 		}
```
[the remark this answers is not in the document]
MikeSpreitzer: I think that this is a valid complaint.

MikeSpreitzer (commented):
Also see https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/629#issuecomment-4921577101

waltforme:
Addressed all the comments except the one made to 629: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/629#issuecomment-4921577101. Will ask @dumb0002 to confirm EPP's behavior against a Pod with `.Metadata.DeletionTimestamp` set. https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/629#issuecomment-4931190153

waltforme:
Addressed the comment made to 629 (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/629#issuecomment-4921577101).

pkg/controller/dual-pods/controller.go
```diff
@@ -103,6 +103,11 @@ const launcherVllmConfigAnnotationKey = "dual-pods.llm-d.ai/vllm-config"
 const iscLabelKeysAnnotationKey = "dual-pods.llm-d.ai/isc-label-keys"
 const iscAnnotationKeysAnnotationKey = "dual-pods.llm-d.ai/isc-annotation-keys"
```
MikeSpreitzer: Are these still needed?

pkg/controller/dual-pods/inference-server.go
```diff
@@ -498,12 +498,22 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 			serverDat.Sleeping = &sleeping
 		}
 		if *(serverDat.Sleeping) {
-			err = ctl.wakeSleeper(ctx, serverDat, requestingPod, providingPod, serverPort, "discovered-bound")
+			providingPod, err = ctl.wakeSleeper(ctx, serverDat, requestingPod, providingPod, serverPort, "discovered-bound")
```
MikeSpreitzer: There was excess runtime cost here, and this PR has made it bigger.
See how this code was followed by `ctl.ensureSleepingLabel`? The idea was to do only one API call to update the providing Pod. Somehow that update got duplicated into `ctl.wakeSleeper`. Now this PR adds another API call to update the provider. It would be better to do all the updates in one API call.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1091,18 +1141,61 @@ func commitInstanceState(serverDat *serverData, state *vllmInstanceState) {
...
+	ctx context.Context,
+	requestingPod *corev1.Pod,
+	serverDat *serverData,
+	providingPod *corev1.Pod,
+) (*corev1.Pod, error) {
+	logger := klog.FromContext(ctx)
+	updated := providingPod.DeepCopy()
+	if problems := applyISCRoutingMetadataToLauncherPod(updated, serverDat.InstanceISCLabels, serverDat.InstanceISCAnnotations); len(problems) > 0 {
+		// Committed metadata is invalid; report to the requester and retry. A
+		// corrected ISC arrives via a fresh bind.
```
MikeSpreitzer: This can only arise if either (A) some other party modified the annotation or (B) the controller was deleted and replaced with a newer version with different validation since binding. Neither case is a transient error, retry is guaranteed to run into the same problem.

The proper response is to, one way or another, unbind and try again. The simplest approach would be to delete the requester Pod, after publishing an Event explaining why. An alternative approach would be to unbind and requeue for retry. That would change an important global fact about the dual-pods technique: a given server-requesting Pod gets bound at most once. I do not have in my head all the things that would have to change if that property were broken. My recommendation is to take the simpler approach for now, and try the bigger change later if the need arises.

BTW, this new annotation should be added to the set that only the FMA ServiceAccount is allowed to write to.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1091,18 +1141,61 @@ func commitInstanceState(serverDat *serverData, state *vllmInstanceState) {
...
+		if statusErr, _ := ctl.ensureReqStatus(ctx, requestingPod, serverDat, problems...); statusErr != nil {
+			return providingPod, statusErr
+		}
+		return providingPod, fmt.Errorf("ISC labels/annotations invalid for deferred application: %s", strings.Join(problems, "; "))
+	}
+	updStart := time.Now()
+	echo, err := ctl.coreclient.Pods(ctl.namespace).Update(ctx, updated, metav1.UpdateOptions{FieldManager: ControllerName})
+	updStartStr := updStart.Format(time.RFC3339Nano)
+	if err != nil {
+		return providingPod, fmt.Errorf("failed to apply deferred ISC labels to launcher Pod %s (started %s): %w", updated.Name, updStartStr, err)
```
MikeSpreitzer: "... labels _and annotations_ ..."

MikeSpreitzer (commented):
I have left some independent comments.

pkg/controller/dual-pods/controller.go
MikeSpreitzer: Doesn't the new annotation carry this information?

waltforme: Consolidated.

pkg/controller/dual-pods/inference-server.go
waltforme: Changed as suggested.

pkg/controller/dual-pods/inference-server.go
waltforme: Addressed this except that I log the reason instead of publishing an Event.

We are not using Events anywhere so far, so I think this deserves more overall thinking (what should be logged, what should be published) before we making changes.

pkg/controller/dual-pods/inference-server.go

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: OK to do this in a follow-on PR, but --- in my opinion --- it has to be done. Events are more accessible to users than are controller logs. I did not mean to suggest _not_ logging. It is good to do both. Events persist for a limited lifetime and are often omitted from problem reports, so logging is needed to support debugging.

pkg/controller/dual-pods/controller.go
```diff
@@ -100,8 +100,12 @@ const nominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
 const launcherInstanceIDAnnotationKey = "dual-pods.llm-d.ai/instance-id"
 const launcherServerPortAnnotationKey = "dual-pods.llm-d.ai/server-port"
 const launcherVllmConfigAnnotationKey = "dual-pods.llm-d.ai/vllm-config"
-const iscLabelKeysAnnotationKey = "dual-pods.llm-d.ai/isc-label-keys"
-const iscAnnotationKeysAnnotationKey = "dual-pods.llm-d.ai/isc-annotation-keys"
+
+// launcherISCRoutingMetadataAnnotationKey holds, as JSON, the ISC-provided
+// routing labels and annotations the bound instance was created with, so they
+// can be applied once it is serving and recovered after a restart. Their keys
```
MikeSpreitzer: The grammar here is not great. The most obvious parse is "... it is (serving and recovered after a restart)". The "recovered after a restart" seems to attach to "it", which refers to "the bound instance" --- but that is not what you mean. I would reword to something like the following.

> launcherISCRoutingMetadataAnnotationKey holds, as the JSON encoding of an iscRoutingMetadata, the ISC-provided routing labels and annotations the bound instance was created with, so they can be applied once it is serving. This is externalized to the Pod object so that it can be reliably conveyed from the binding transaction to the later moment when the controller reacts to the vllm instance reaching readiness.

pkg/controller/dual-pods/controller.go
```diff
@@ -100,8 +100,12 @@ const nominalHashAnnotationKey = "dual-pods.llm-d.ai/nominal"
...
+// are also what identifies the routing labels/annotations to remove on unbind.
```
MikeSpreitzer: In my opinion, the remark "Their keys are also what identifies the routing labels/annotations to remove on unbind" is unnecessary. Of course the keys of the key/value pairs identify them.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -497,15 +497,31 @@ func (item infSvrItem) process(urCtx context.Context, ctl *controller, nodeDat *
 			logger.V(2).Info("Determined whether provider is sleeping", "isSleeping", sleeping)
 			serverDat.Sleeping = &sleeping
 		}
-		if *(serverDat.Sleeping) {
-			err = ctl.wakeSleeper(ctx, serverDat, requestingPod, providingPod, serverPort, "discovered-bound")
-			if err != nil {
```
MikeSpreitzer, line 502: No need to delete this code from here. Just remove the call to `ctl.ensureSleepingLabel` from the body of `ctl.wakeSleeper`. Keeping this here means that we do not need the two copies in the two arms of the following `if` statement. Yes, this means that the _other_ call to `wakeSleeper` (from the body of `bind`) needs to gain a following call to `ensureSleepingLabel`. Still, this would be simpler overall.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1057,54 +1086,131 @@ func applyInstanceStateToLauncherPod(providingPod *corev1.Pod, state *vllmInstan
...
+	if providingPod.Labels[api.SleepingLabelName] == desiredSleeping && !applyLabels {
+		return providingPod, false, nil
+	}
+	updated := providingPod.DeepCopy()
+	updated.Labels = utils.MapSet(updated.Labels, api.SleepingLabelName, desiredSleeping)
+	if applyLabels {
+		if problems := applyISCRoutingMetadataToLauncherPod(updated, serverDat.InstanceISCLabels, serverDat.InstanceISCAnnotations); len(problems) > 0 {
+			// The committed metadata is invalid — not transient (it implies
+			// tampering or a validation change since binding), so retrying would
+			// loop. Unbind by deleting the requester Pod; a fresh bind starts over.
```
MikeSpreitzer: "fresh _bind_" understates what is happening here. "fresh _requester_" is more on point.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1057,54 +1086,131 @@ func applyInstanceStateToLauncherPod(providingPod *corev1.Pod, state *vllmInstan
...
+			logger.Error(nil, "Committed ISC routing metadata is invalid; deleting requester Pod to unbind",
+				"requester", requestingPod.Name, "problems", problems)
```
MikeSpreitzer: The requesting Pod name was already put in the logger at the start of `func (item infSvrItem) process`. Same for the provider, in the case of exactly 1 bound provider.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1057,54 +1086,131 @@ func applyInstanceStateToLauncherPod(providingPod *corev1.Pod, state *vllmInstan
...
+			delErr := ctl.coreclient.Pods(ctl.namespace).Delete(ctx, requestingPod.Name, metav1.DeleteOptions{
+				PropagationPolicy: ptr.To(metav1.DeletePropagationBackground),
+				Preconditions:     &metav1.Preconditions{UID: &requestingPod.UID, ResourceVersion: &requestingPod.ResourceVersion}})
```
MikeSpreitzer: I would not include `ResourceVersion: &requestingPod.ResourceVersion` in the preconditions here. Admitted changes to the requester can not resolve the problem here.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1057,54 +1086,131 @@ func applyInstanceStateToLauncherPod(providingPod *corev1.Pod, state *vllmInstan
...
+			if delErr != nil && !apierrors.IsNotFound(delErr) && !apierrors.IsGone(delErr) {
+				return providingPod, true, fmt.Errorf("failed to delete requester Pod %q after invalid ISC routing metadata: %w", requestingPod.Name, delErr)
```
MikeSpreitzer: I would include the provider name too

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1143,18 +1250,41 @@ func recoverInstanceStateFromLauncherPod(serverDat *serverData, providingPod *co
...
 }
 
-func parseSpaceSeparatedAnnotation(value string) []string {
-	if value == "" {
-		return []string{}
+// iscRoutingMetadataPresent reports whether all of the given ISC routing labels
+// and annotations are already present with matching values on the Pod. Empty
+// metadata (nothing to apply) counts as present.
+func iscRoutingMetadataPresent(providingPod *corev1.Pod, routing iscRoutingMetadata) bool {
+	for k, v := range routing.Labels {
```
MikeSpreitzer: I would define, in pkg/controller/utils/generics.go, `func MapIsSubset[Dom, Rng comparable](sub, super map[Dom]Rng) bool` and then use it twice here.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1168,22 +1298,57 @@ func getVLLMInstancePort(inst InstanceState) (int32, error) {
...
+	for _, k := range toRemove {
+		delete(updatedPod.Labels, k)
+	}
+	updStart := time.Now()
+	echo, err := ctl.coreclient.Pods(ctl.namespace).Update(ctx, updatedPod, metav1.UpdateOptions{FieldManager: ControllerName})
+	updStartStr := updStart.Format(time.RFC3339Nano)
+	if err != nil {
+		return providingPod, fmt.Errorf("failed to remove ISC routing labels from launcher Pod %s (started %s): %w", providingPod.Name, updStartStr, err)
+	}
+	klog.FromContext(ctx).V(2).Info("Removed ISC routing labels from launcher Pod before sleeping", "name", echo.Name, "removed", toRemove, "newResourceVersion", echo.ResourceVersion, "k8sCallStartTime", updStartStr)
```
MikeSpreitzer: The requesting Pod name is already in the logger.

MikeSpreitzer (commented):
I have left some independent comments.

waltforme:
Addressed all the comments made today.

MikeSpreitzer (approved):
LGTM
Thanks!

PR #647, opened 2026-07-10
Workflow test consolidation

### LINKED ISSUE #640: [Feature]: Consolidate CI code-quality workflows
### Feature Area

CI/CD

### Problem Statement

Currently the repository has two separate workflows that both set up pre-commit and run subsets of hooks:

- `.github/workflows/python-code-quality.yml`— runs Python hooks (black, isort, flake8, typos, file checks) with SKIP=go-mod-tidy,golangci-lint,verify-idl-consumption
- `.github/workflows/golangci-lint.yml` — runs the Go hooks (go-mod-tidy, golangci-lint-full) separately, with its own Python venv setup to bootstrap pre-commit

This split made sense as an incremental step, but it has some ongoing costs:

- contributors need to look at two different CI jobs to understand a single category of failure
- pre-commit bootstrap (venv creation, pip install) is duplicated across jobs
- the SKIP list in python-code-quality.yml must be kept in sync with what the Go workflow covers
- adding a new hook requires deciding which workflow owns it

### Proposed Solution

A cleaner end state would be a single pre-commit workflow that:

1. sets up both Go and Python (as setup-go and a venv are both cheap)
2. runs `pre-commit run --all-files` for the commit-stage hooks
3. runs the push-stage hooks (go-mod-tidy, golangci-lint-full) as a separate step or job, since they are intentionally slower

`verify-idl-consumption` is already intentionally isolated in its own workflow (it requires a clean tree and is triggered on different paths), so that one should remain separate.

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

diegocastanibm:
https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/640


docs/README.md
```diff
@@ -14,8 +14,7 @@
 
 # CI
 
-- [Markdown and Python code quality check](../.github/workflows/python-code-quality.yml)
-- [Go code quality check](../.github/workflows/golangci-lint.yml)
+- [Code quality (lint, format, typos, Go, tests)](../.github/workflows/code-quality.yml)
```
MikeSpreitzer: This still checks Markdown and Python quality too, right?

CONTRIBUTING.md
```diff
@@ -89,7 +89,7 @@ you do **not** need to install a specific golangci-lint globally.
 ### Running the checks
 
 ```bash
-# Run every hook against the whole tree (what CI's python job runs)
+# Run every hook against the whole tree (what CI's code-quality job runs)
```
MikeSpreitzer: This won't run the `golangci-lint-full` hook, because it has `stages: [pre-push]`, right?

MikeSpreitzer (commented):

docs/README.md
diegocastanibm: off course. I'll make it more clear in the README.  It runs the Markdown and Python hooks via `pre-commit run --all-files`

CONTRIBUTING.md
diegocastanibm: Right. `golangci-lint-full` has `stages: [pre-push]`, so `pre-commit run --all-files` doesn't run it. My "(what CI's code-quality job runs)" note was misleading, since the job runs this command plus the pre-push step plus the launcher tests. I've reworded the comment to say this runs the commit-stage hooks only, and clarified that CI runs it together with the pre-push hooks and tests.

MikeSpreitzer (approved):
LGTM
Thanks!

PR #658, opened 2026-07-15
Defaulting VLLM_WORKER_MULTIPROC_METHOD=fork in launcher code for faster vLLM startup

aavarghese:
FMA launcher does not set `VLLM_WORKER_MULTIPROC_METHOD`, so the vLLM instances it spawns fell back to vLLM's implicit default for how it launches its EngineCore/worker processes. This PR defaults it to fork (unless the ISC already sets it), injected into `config.env_vars` so it flows through the launcher's existing `set_env_vars()` path — the same way `CUDA_VISIBLE_DEVICES` is handled.

- Overridable: uses setdefault, so an explicit VLLM_WORKER_MULTIPROC_METHOD in the ISC's env_vars still wins.
- Observability: logs the effective value ("Set VLLM_WORKER_MULTIPROC_METHOD to fork."), mirroring the CUDA_VISIBLE_DEVICES log line.
- Low risk: no change when the default was already fork; only affects vLLM versions/paths whose implicit default drifted to spawn.

**Why this improves startup time**

vLLM spawns its EngineCore/worker process(es) using a Python multiprocessing start method:
- `fork` clones the already-warm launcher child (copy-on-write): torch, CUDA bindings, and vLLM modules are already imported, so the worker starts almost immediately.
- `spawn` launches a brand-new Python interpreter per worker that must re-import torch + vLLM and re-initialize from scratch — several seconds of overhead per worker, paid on every instance start (and multiplied under TP>1).

Testing

- [x] Unit: assert the launcher injects the default and that an ISC-provided value overrides it.
- [x] E2E: confirm the launcher log shows the new line and vLLM comes up serving.

🤖 Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/inference_server/launcher/launcher.py
+++ b/inference_server/launcher/launcher.py
@@ -172,6 +172,9 @@ class VllmInstance:
         :param log_dir: Directory for log files (defaults to /tmp)
         """
 
+        if config.env_vars is None:
+            config.env_vars = {}
+
         # Check for CUDA device UUIDs and set CUDA_VISIBLE_DEVICES accordingly
         if config.gpu_uuids:
             cuda_indices = []
@@ -182,13 +185,21 @@ class VllmInstance:
                 f"Translated GPU UUIDs {config.gpu_uuids} to indices {cuda_indices}."
             )
 
-            if config.env_vars is None:
-                config.env_vars = {}
             config.env_vars["CUDA_VISIBLE_DEVICES"] = ",".join(cuda_indices)
             logger.info(
                 "Set CUDA_VISIBLE_DEVICES to %s based on UUIDs.",
                 config.env_vars["CUDA_VISIBLE_DEVICES"],
             )
+
+        # Default vLLM's worker multiprocessing method to "fork" unless the
+        # ISC's env_vars already set it. Injected into config.env_vars so it
+        # flows through set_env_vars() alongside every other vLLM env var
+        # (e.g. CUDA_VISIBLE_DEVICES above) instead of being special-cased.
+        config.env_vars.setdefault("VLLM_WORKER_MULTIPROC_METHOD", "fork")
+        logger.info(
+            "Set VLLM_WORKER_MULTIPROC_METHOD to %s.",
+            config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"],
+        )
 
         # Initialize instance variables
         self.instance_id = instance_id
--- a/inference_server/launcher/tests/test_launcher.py
+++ b/inference_server/launcher/tests/test_launcher.py
@@ -173,6 +173,35 @@ class TestVllmInstance:
         assert instance.config == vllm_config
         assert instance.process is None
 
+    def test_instance_sets_default_multiproc_method(self, gpu_translator, tmp_log_dir):
+        """VLLM_WORKER_MULTIPROC_METHOD defaults to 'fork' when the ISC omits it."""
+        config = VllmConfig(options="--model test-model --port 8000")
+        assert config.env_vars is None  # ISC provided none
+        instance = VllmInstance("test-id", config, gpu_translator, log_dir=tmp_log_dir)
+        assert instance.config.env_vars is not None
+        assert instance.config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"] == "fork"
+
+    def test_instance_multiproc_method_override_wins(self, gpu_translator, tmp_log_dir):
+        """An explicit VLLM_WORKER_MULTIPROC_METHOD in ISC's env is preserved."""
+        config = VllmConfig(
+            options="--model test-model --port 8000",
+            env_vars={"VLLM_WORKER_MULTIPROC_METHOD": "spawn"},
+        )
+        instance = VllmInstance("test-id", config, gpu_translator, log_dir=tmp_log_dir)
+        assert instance.config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"] == "spawn"
+
+    def test_instance_multiproc_method_preserves_other_env_vars(
+        self, gpu_translator, tmp_log_dir
+    ):
+        """The default is added without clobbering the ISC's other env vars."""
+        config = VllmConfig(
+            options="--model test-model --port 8000",
+            env_vars={"CUSTOM_VAR": "custom_value"},
+        )
+        instance = VllmInstance("test-id", config, gpu_translator, log_dir=tmp_log_dir)
+        assert instance.config.env_vars["CUSTOM_VAR"] == "custom_value"
+        assert instance.config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"] == "fork"
+
     @patch("launcher.multiprocessing.Process")
     def test_instance_start(
         self, mock_process_class, vllm_config: VllmConfig, gpu_translator, tmp_log_dir


inference_server/launcher/launcher.py
```diff
@@ -190,6 +190,18 @@ def __init__(
...
+        # ISC's env_vars already set it. Injected into config.env_vars so it
+        # flows through set_env_vars() alongside every other vLLM env var
+        # (e.g. CUDA_VISIBLE_DEVICES above) instead of being special-cased.
+        if config.env_vars is None:
+            config.env_vars = {}
+        config.env_vars.setdefault("VLLM_WORKER_MULTIPROC_METHOD", "fork")
+        logger.info(
+            "Set VLLM_WORKER_MULTIPROC_METHOD to %s.",
+            config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"],
+        )
```
diegocastanibm, line 202: Can we add some tests for this part in `test_launcher.py`?

I would add:
```
def test_instance_sets_default_multiproc_method(self, gpu_translator, tmp_log_dir):
    """VLLM_WORKER_MULTIPROC_METHOD defaults to 'fork' when the ISC omits it."""
    config = VllmConfig(options="--model test-model --port 8000")
    assert config.env_vars is None  # ISC provided none
    instance = VllmInstance("test-id", config, gpu_translator, log_dir=tmp_log_dir)
    assert instance.config.env_vars is not None
    assert instance.config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"] == "fork"

def test_instance_multiproc_method_override_wins(self, gpu_translator, tmp_log_dir):
    """An explicit VLLM_WORKER_MULTIPROC_METHOD in the ISC's env_vars is preserved."""
    config = VllmConfig(
        options="--model test-model --port 8000",
        env_vars={"VLLM_WORKER_MULTIPROC_METHOD": "spawn"},
    )
    instance = VllmInstance("test-id", config, gpu_translator, log_dir=tmp_log_dir)
    assert instance.config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"] == "spawn"

def test_instance_multiproc_method_preserves_other_env_vars(self, gpu_translator, tmp_log_dir):
    """The default is added without clobbering the ISC's other env vars."""
    config = VllmConfig(
        options="--model test-model --port 8000",
        env_vars={"CUSTOM_VAR": "custom_value"},
    )
    instance = VllmInstance("test-id", config, gpu_translator, log_dir=tmp_log_dir)
    assert instance.config.env_vars["CUSTOM_VAR"] == "custom_value"
    assert instance.config.env_vars["VLLM_WORKER_MULTIPROC_METHOD"] == "fork"
```

diegocastanibm (commented):
Thanks Ansu. Add and past python tests and I'll be good with it.

MikeSpreitzer:
Regarding observability: putting something in our controller's log is not really very visible. Kubernetes has access controls, and users of llm-d may not necessarily have the authorization to read our controller's log. And I think that we certainly should not expect them to have the knowledge to usefully do that.

I see vllm itself logging non-default settings for command-line arguments, but not for environment variables.

inference_server/launcher/launcher.py
```diff
@@ -190,6 +190,18 @@ def __init__(
                 config.env_vars["CUDA_VISIBLE_DEVICES"],
             )
 
+        # Default vLLM's worker multiprocessing method to "fork" unless the
+        # ISC's env_vars already set it. Injected into config.env_vars so it
+        # flows through set_env_vars() alongside every other vLLM env var
+        # (e.g. CUDA_VISIBLE_DEVICES above) instead of being special-cased.
+        if config.env_vars is None:
+            config.env_vars = {}
```
MikeSpreitzer: If you move this above the `if config.gpu_uuids` block then this code does not have to be repeated in that block.

aavarghese: Sure. Moved it.

inference_server/launcher/launcher.py
aavarghese: Thank you - added in latest commit!

MikeSpreitzer:
An intervention such as this PR proposes needs to be documented somewhere reasonably findable.

aavarghese:
> An intervention such as this PR proposes needs to be documented somewhere reasonably findable.

@MikeSpreitzer are you suggesting somewhere like the FMA well-lit guide?

diegocastanibm:
> > An intervention such as this PR proposes needs to be documented somewhere reasonably findable.
> 
> @MikeSpreitzer are you suggesting somewhere like the FMA well-lit guide?

I think the right place is the `docs/launcher.md`

MikeSpreitzer:
I am not sure we have a good enough place.

In my opinion, one place that this belongs is https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/v0.6.2/api/fma/v1alpha1/inferenceserverconfig_types.go#L46-L48 . That is the official API documentation.

But, I have realized to my dismay, that not every user reads the API documentation.

aavarghese:
E2E test succeeded and shows in log:
```2026-07-15 18:31:44,514 - __main__ - INFO - Set VLLM_WORKER_MULTIPROC_METHOD to fork.```

**Comment by aavarghese** [comment]
> I am not sure we have a good enough place.
> 
> In my opinion, one place that this belongs is https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/v0.6.2/api/fma/v1alpha1/inferenceserverconfig_types.go#L46-L48 . That is the official API documentation.
> 
> But, I have realized to my dismay, that not every user reads the API documentation.

As our docs exist today, Environment Variables section in`launcher.md` like Diego said might be the best place right next to `CUDA_VISIBLE_DEVICES`..

**Comment by MikeSpreitzer** [comment]
Following are things that I am thinking about.

1. Do we really want to be in the business of changing the user's vllm environment variables? If we get into that business then we are introducing a new thing: FMA's vllm environment variables. One thing that I hope that we are all keeping in mind: we want to be able in the future to generalize beyond vllm to other kinds of inference server. If we build in knowledge that is specific to vllm, that gets harder.

2. What to write in the definition of `InferenceServerConfig`. We should put _something_ there. In some sense of "ideal", it would be precise --- because it is the official API declaration. However, this would also mean that any change is an API change. And perhaps not the sort of API change that Kubernetes can handle properly; it depends on the change: no Kube mechanism can read the user's mind, but then maybe no actual change would require that. Shifting the details out of the `inferenceserverconfig_types.go` file into a doc file would make a change not look like an API change --- according to typical tooling. But it _would_ still actually be an API change.

3. We have lots of shortcomings in our user-facing documentation (see #580 for a large but not necessarily good or comprehensive listing). We do not even have the well-lit path document merged anywhere yet, and even that is only one example --- we can not, and should not try to, make one example do the full job of user-facing documentation.

File: inference_server/launcher/tests/test_launcher.py [hunk 3]
```diff
@@ -173,6 +173,35 @@ def test_instance_creation(self, vllm_config, gpu_translator, tmp_log_dir):
         assert instance.config == vllm_config
         assert instance.process is None
 
+    def test_instance_sets_default_multiproc_method(self, gpu_translator, tmp_log_dir):
+        """VLLM_WORKER_MULTIPROC_METHOD defaults to 'fork' when the ISC omits it."""
+        config = VllmConfig(options="--model test-model --port 8000")
```
rubambiza, line 178: Thinking about maintainability here. It seems like the test file already has a [`vllm_config_no_env()`](https://github.com/aavarghese/llm-d-fast-model-actuation/blob/52fb7be661f5f12fba8dc4c024851c906931f074/inference_server/launcher/tests/test_launcher.py#L74) fixture, but we are repeating the same config in all three new tests (see also [L186](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/658/changes#diff-f228dabe57c5a25be3aa9b5419004345a4b7d2f7ad40b136e85937b4ef85c870R186), and [L197](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/658/changes#diff-f228dabe57c5a25be3aa9b5419004345a4b7d2f7ad40b136e85937b4ef85c870R197)). May I suggest invoking the fixture and simply assigning its `env_vars` where appropriate?

This also ensures the port being used in the tests is standardized (currently the configs say 8000 while the fixture uses 8001).

rubambiza (changes requested):
One requested change for maintainability. I'll be good to approve once it's addressed.

rubambiza:
I drafted my review above before reading @MikeSpreitzer's [comment](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/658#issuecomment-4996119579) above. I think it's related to my review in the sense that, if we eventually want to be agnostic to the inference engine, then the `VllmConfig` class would have to be abstracted away into something like `InferenceEngineConfig`, and we can simply do something like a dictionary dump to pass everything directly. The contract (and the validation/sanitation) of what is in there is left up to the user.

MikeSpreitzer:
Just to be clear, we are dealing with a three-layer process hierarchy.
- At the top, there is FMA's launcher process. It creates child processes to run vllm. The launcher always does this using fork, never spawn.
- In the middle is the top half of vllm. This creates a child process called EngineCore/worker. The way to do this is the variable here.
- At the bottom is the vllm EngineCore/worker process.

Why does vllm default to using spawn to create its EngineCore/worker child? That seems like an inferior choice to me. I wonder if I am missing something important.

aavarghese:
Closing this PR as per Scrum call discussion. Will run an experiment from benchmarking with this envvar set and report to team. Also, documentation on best practices for vllm args will need to be added to FMA soon.

PR #660, opened 2026-07-17
Add proposed well-lit path

MikeSpreitzer:
This PR puts a copy of the proposed well-lit path into this repo. This is copied from the current version of https://github.com/llm-d/llm-d/pull/1779 (commit 3b376562a3423380e342a371b281348699af7645), which we do not expect to be able to merge soon because FMA is only a sandbox project. We expect to merge and iterate on this proposed document in this repo. Being merged, it will be easier to find and cite and display.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -6,6 +6,39 @@
 - [Prometheus Metrics](metrics.md)
 - [Design Rules](../DESIGN_RULES.md)
 
+# Well-lit Path
+
+This will ultimately go into
+https://github.com/llm-d/llm-d/tree/main/guides --- but maybe not for
+some time, because FMA is currently only a sandbox project. In the
+meantime, we home a copy here while we work on developing it.
+
+- [Well-lit Path](../guides/fast-model-actuation/README.md)
+
+As explained in https://github.com/llm-d/llm-d/pull/1779, this is
+maintained using the structured guide mechanism described in
+https://github.com/llm-d/llm-d/pull/1988 . The following shows me
+using the three scripts of that mechanism on the copy in my working
+tree for FMA on my local machine.
+
+```console
+me@mymac llm-d-fast-model-actuation % ../../llm-d/llm-d/scripts/guide-render.py --yaml guides/fast-model-actuation/guide.yaml --readme guides/fast-model-actuation/README.md
+updated guides/fast-model-actuation/README.md
+
+me@mymac llm-d-fast-model-actuation % ../../llm-d/llm-d/scripts/guide-check-yaml.py guides/fast-model-actuation/guide.yaml
+Installed 1 package in 2ms
+guides/fast-model-actuation/guide.yaml: OK
+
+me@mymac llm-d-fast-model-actuation % ../../llm-d/llm-d/scripts/guide-check-readme.py --yaml guides/fast-model-actuation/guide.yaml --readme guides/fast-model-actuation/README.md
+Installed 1 package in 2ms
+guides/fast-model-actuation/README.md: OK
+
+```
+
+To support that, I have
+https://github.com/Vezio/llm-d/tree/structured-guide-implementation
+checked out in `../../llm-d/llm-d/`.
+
 # Dev/test
 
 - [Local dev/test in a `kind` cluster](local-test.md)
--- a/guides/fast-model-actuation/README.md
+++ b/guides/fast-model-actuation/README.md
@@ -2,19 +2,19 @@
 
 ## Overview
 
-Fast Model Actuation (FMA) addresses vllm startup time using two technologies. One is vLLM sleep and wake, which can entirely avoid model loading and CUDA graph compilation in lucky scenarios. The other is running multiple vllm processes as children of a launcher process that has already done the Python module loading. FMA can manage multiple vllm instances bound to one GPU, with one instance awake at a time, to accomplish fast switching of which model that GPU is used for; this generalizes to multiple GPUs of one node.
+Fast Model Actuation (FMA) addresses vllm startup time using two technologies. One is vLLM sleep and wake, which can entirely avoid model loading and CUDA graph compilation in applicable scenarios. The other is running multiple vllm processes as children of a launcher process that has already done the Python module loading. FMA can manage multiple vllm instances bound to one GPU, with one instance awake at a time, to accomplish fast switching of which model that GPU is used for; this generalizes to multiple GPUs of one node.
 
-In Kubernetes, a Pod that requests and is allocated some GPUs keeps that exclusive allocation for the Pod's lifetime. FMA uses the following **dual pod** technique to circumvent this constraint:
+In Kubernetes, once a Pod has been allocated some GPUs it keeps that exclusive allocation for rest of the Pod's lifetime. FMA uses the following **dual pod** technique to circumvent that constraint.
 
-- **Server-requesting Pods** reserve GPU resources via the Kubernetes scheduler but do not run inference themselves.
-- **Launcher Pods** (server-providing) run vLLM without requesting GPUs. They (a) gain access to all GPUs of their Node via special provisions that do not count this access as usage and (b) using `CUDA_VISIBLE_DEVICES` as directed by the FMA controller get vllm to use the specific GPU(s) reserved for the requesting pod.
-- **FMA Controllers** manage the lifecycle: binding requesting pods to launchers, starting vLLM instances, and orchestrating sleep/wake.
+- **Server-requesting Pods** reserve GPU resources via the Kubernetes Pod scheduler and the kubelet but do not run inference themselves.
+- **Launcher Pods** (server-providing) run vLLM without requesting GPUs. They (a) gain access to all GPUs of their Node via special provisions that do not count this access as usage and (b) using `CUDA_VISIBLE_DEVICES` as directed by the FMA controllers get vllm to use the specific GPU(s) reserved for the requesting pod.
+- **FMA Controllers** manage the lifecycle: creating/deleting launchers, binding/unbinding requesting pods to launchers and vLLM instances in them, creating/deleteing those vLLM instances, and orchestrating sleep/wake.
 
-Server-requesting pods are managed through standard Kubernetes controllers such as Deployments and autoscalers. The FMA controller watches these pods and translates scheduler decisions into actions on launcher pods and GPUs.
+Server-requesting pods are managed through standard Kubernetes set objects and controllers such as Deployments and autoscalers. An FMA controller watches the server-requesting pods and translates scheduler decisions into actions on launcher pods and GPUs.
 
-When a requesting pod is deleted, the controller puts the corresponding vLLM instance to sleep (model stays in GPU memory). Although the Kubernetes GPU allocation is released when the requesting pod exits, the launcher pod retains the CUDA context and keeps the model in GPU memory. The GPU remains dedicated to that launcher until it is explicitly unbound or the launcher pod is deleted. When a new requesting pod arrives and gets assigned to the same GPU, the controller wakes the sleeping instance, resuming in seconds instead of cold-starting from scratch.
+When a requesting pod is deleted, the controller puts the corresponding vLLM instance to sleep (model tensors move from GPU to main memory). Although the Kubernetes GPU allocation is released when the requesting pod is deleted, the vLLM instance retains the CUDA context in a small amount of GPU memory---until an FMA controller directs that launcher to delete that vLLM instance (which may happen, to limit memory usage). If and when a new requesting pod arrives, serving the same model with the same command-line parameters and assigned to the same GPU, the controller wakes the sleeping instance---which resumes in seconds. This is "hot start"
 
-FMA also supports instant model switching: if a new requesting pod references a different `InferenceServerConfig`, the FMA controller can direct the bound launcher to swap the loaded model in place, avoiding a full cold start.
+When a new server-requesting Pod arrives and there is no sleeping vLLM instance to wake for it, but a launcher is available, an FMA controller will direct the launcher to create a corresponding new vLLM instance. This takes advantage of the Python module loading already done by the launcher. This is "warm start".
 
 > [!NOTE]
 > Fast wake only occurs if the Kubernetes scheduler assigns the new requesting pod to the same node (and GPU) where the sleeping vLLM instance resides. In a cluster with a single GPU per node, if the scheduler picks the same node, the GPU is necessarily the same one. In a multi-node pool the scheduler may assign the pod to a different node.
--- a/guides/fast-model-actuation/README.md
+++ b/guides/fast-model-actuation/README.md
@@ -2,10 +2,12 @@
 
 ## Overview
 
-Fast Model Actuation (FMA) enables rapid model loading and switching for LLM inference on Kubernetes by exploiting vLLM sleep/wake and model swapping. In Kubernetes, GPUs are bound 1-to-1 to pods: a pod that requests a GPU holds it exclusively for its lifetime. FMA uses the following **dual pod** technique to circumvent this constraint:
+Fast Model Actuation (FMA) addresses vllm startup time using two technologies. One is vLLM sleep and wake, which can entirely avoid model loading and CUDA graph compilation in lucky scenarios. The other is running multiple vllm processes as children of a launcher process that has already done the Python module loading. FMA can manage multiple vllm instances bound to one GPU, with one instance awake at a time, to accomplish fast switching of which model that GPU is used for; this generalizes to multiple GPUs of one node.
+
+In Kubernetes, a Pod that requests and is allocated some GPUs keeps that exclusive allocation for the Pod's lifetime. FMA uses the following **dual pod** technique to circumvent this constraint:
 
 - **Server-requesting Pods** reserve GPU resources via the Kubernetes scheduler but do not run inference themselves.
-- **Launcher Pods** (server-providing) run vLLM without requesting GPUs. They gain access to GPUs via `CUDA_VISIBLE_DEVICES`, directed by the FMA controller to the specific GPU(s) reserved by the requesting pod.
+- **Launcher Pods** (server-providing) run vLLM without requesting GPUs. They (a) gain access to all GPUs of their Node via special provisions that do not count this access as usage and (b) using `CUDA_VISIBLE_DEVICES` as directed by the FMA controller get vllm to use the specific GPU(s) reserved for the requesting pod.
 - **FMA Controllers** manage the lifecycle: binding requesting pods to launchers, starting vLLM instances, and orchestrating sleep/wake.
 
 Server-requesting pods are managed through standard Kubernetes controllers such as Deployments and autoscalers. The FMA controller watches these pods and translates scheduler decisions into actions on launcher pods and GPUs.
--- a/guides/fast-model-actuation/README.md
+++ b/guides/fast-model-actuation/README.md
@@ -29,9 +29,9 @@ FMA also supports instant model switching: if a new requesting pod references a 
 
 ## Prerequisites
 
-This guide assumes you have a Kubernetes cluster with GPU nodes and the [llm-d router](../../guides/recipes/router/README.md) infrastructure available. If you are starting from an existing llm-d deployment, the Gateway API Inference Extension CRDs may already be installed and you can skip that step.
+This guide assumes you have a Kubernetes cluster with GPU nodes and the [llm-d router](https://github.com/llm-d/llm-d/tree/main/guides/recipes/router) infrastructure available. If you are starting from an existing llm-d deployment, the Gateway API Inference Extension CRDs may already be installed and you can skip that step.
 
-- Have the [proper client tools installed on your local system](../../helpers/client-setup/README.md) to use this guide.
+- Have the [proper client tools installed on your local system](https://github.com/llm-d/llm-d/blob/main/helpers/client-setup/README.md) to use this guide.
 
 - Checkout llm-d repo:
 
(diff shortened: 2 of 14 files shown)


docs/README.md
```diff
@@ -6,6 +6,39 @@
 - [Prometheus Metrics](metrics.md)
 - [Design Rules](../DESIGN_RULES.md)
 
+# Well-lit Path
+
+This will ultimately go into
+https://github.com/llm-d/llm-d/tree/main/guides --- but maybe not for
+some time, because FMA is currently only a sandbox project. In the
+meantime, we home a copy here while we work on developing it.
```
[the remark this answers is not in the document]
MikeSpreitzer: In English, any noun can be verbed.
This is not the first time that "home" has been treated this way.

guides/fast-model-actuation/router/fast-model-actuation.values.yaml
```diff
@@ -0,0 +1,30 @@
...
+        plugins:
+        - type: queue-scorer
+        - type: kv-cache-utilization-scorer
+        schedulingProfiles:
+        - name: default
+          plugins:
+          - pluginRef: queue-scorer
+            weight: 1
+          - pluginRef: kv-cache-utilization-scorer
+            weight: 1
```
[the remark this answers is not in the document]
MikeSpreitzer: They look indented to me.

```console
me@mymac fast-model-actuation % yq -o=json '.router.epp.pluginsCustomConfig["fast-model-actuation-plugins.yaml"]' router/fast-model-actuation.values.yaml | jq -r . | yq -o=json .
{
  "apiVersion": "llm-d.ai/v1alpha1",
  "kind": "EndpointPickerConfig",
  "plugins": [
    {
      "type": "queue-scorer"
    },
    {
      "type": "kv-cache-utilization-scorer"
    }
  ],
  "schedulingProfiles": [
    {
      "name": "default",
      "plugins": [
        {
          "pluginRef": "queue-scorer",
          "weight": 1
        },
        {
          "pluginRef": "kv-cache-utilization-scorer",
          "weight": 1
        }
      ]
    }
  ]
}
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,303 @@
...
+| Requesting pod replicas | 2                                                             |
+| Launcher count          | 1 (per matching node)                                         |
+| GPUs per requesting pod | 1                                                             |
+| Router                  | llm-d-router-standalone                                      |
+
+## Prerequisites
+
+This guide assumes you have a Kubernetes cluster with GPU nodes and the [llm-d router](../../guides/recipes/router/README.md) infrastructure available. If you are starting from an existing llm-d deployment, the Gateway API Inference Extension CRDs may already be installed and you can skip that step.
+
+- Have the [proper client tools installed on your local system](../../helpers/client-setup/README.md) to use this guide.
```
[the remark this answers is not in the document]
MikeSpreitzer: Revised to use absolute links to latest version.

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+> [!NOTE]
+> Wake latency depends on the Kubernetes scheduler assigning the new requesting pod to the same node and GPU where the sleeping vLLM instance resides. If a different GPU is assigned, a new vLLM instance starts from scratch (cold start). Sleep/wake is most valuable in multi-GPU-per-node configurations where multiple models share the same GPU pool and can be swapped in and out without cold-starting.
+
+## Benchmarking
+
+This guide uses [`llmdbenchmark`](https://github.com/llm-d/llm-d-benchmark) — the supported standard CLI for llm-d performance benchmarking. It defaults to the `nop` harness (which stands the stack up and validates it end-to-end without driving a synthetic load); the richer, FMA-specific experimentation workflow lives in [`llm-d-benchmark`](https://github.com/llm-d/llm-d-benchmark) itself.
+
+> [!IMPORTANT]
+> The Benchmarking section below contains only the **fast-model-actuation-specific commands** needed to drive the stack you just deployed — for everything else (and especially when something goes wrong), start at [`helpers/benchmark.md`](../../helpers/benchmark.md).
```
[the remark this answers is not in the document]
MikeSpreitzer: Revised to use absolute links to latest version.

docs/README.md
```diff
@@ -6,6 +6,39 @@
...
+
+mspreitz@mjs13 llm-d-fast-model-actuation % ../../llm-d/llm-d/scripts/guide-check-readme.py --yaml guides/fast-model-actuation/guide.yaml --readme guides/fast-model-actuation/README.md
+Installed 1 package in 2ms
+guides/fast-model-actuation/README.md: OK
+
+```
+
+To support that, I have
+https://github.com/Vezio/llm-d/tree/structured-guide-implementation
+checked out in `../../llm-d/llm-d/`.
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
+# Fast Model Actuation
+
+## Overview
+
+Fast Model Actuation (FMA) addresses vllm startup time using two technologies. One is vLLM sleep and wake, which can entirely avoid model loading and CUDA graph compilation in lucky scenarios. The other is running multiple vllm processes as children of a launcher process that has already done the Python module loading. FMA can manage multiple vllm instances bound to one GPU, with one instance awake at a time, to accomplish fast switching of which model that GPU is used for; this generalizes to multiple GPUs of one node.
```
rubambiza: I think using the word `lucky` chips away at the credibility, especially since we're describing how FMA addresses startup time.

```suggestion
Fast Model Actuation (FMA) addresses vllm startup time using two technologies. One is vLLM sleep and wake, which can entirely avoid model loading and CUDA graph compilation. The other is running multiple vLLM processes as children of a launcher process that has already done the Python module loading. FMA can manage multiple vLLM instances bound to one GPU, with one instance awake at a time, to accomplish fast switching of which model that GPU is used for; this generalizes to multiple GPUs of one node.
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+In Kubernetes, a Pod that requests and is allocated some GPUs keeps that exclusive allocation for the Pod's lifetime. FMA uses the following **dual pod** technique to circumvent this constraint:
```
rubambiza: I think allocation implies there was a request, especially since the breakdown below also states it in the name of the pods.

```suggestion
In Kubernetes, a Pod is allocated GPUs keeps that exclusive allocation for the Pod's lifetime. FMA uses the following **dual pod** technique to circumvent this constraint:
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+- **Server-requesting Pods** reserve GPU resources via the Kubernetes scheduler but do not run inference themselves.
+- **Launcher Pods** (server-providing) run vLLM without requesting GPUs. They (a) gain access to all GPUs of their Node via special provisions that do not count this access as usage and (b) using `CUDA_VISIBLE_DEVICES` as directed by the FMA controller get vllm to use the specific GPU(s) reserved for the requesting pod.
```
rubambiza:
```suggestion
- **Server-requesting Pods (aka Launcher Pods)** run vLLM instances without requesting GPUs. They (a) gain access to all GPUs of their Node via special provisions that do not count this access as usage and (b) using `CUDA_VISIBLE_DEVICES` as directed by the FMA controller get vLLM to use the specific GPU(s) reserved for the requesting pod.
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+- **FMA Controllers** manage the lifecycle: binding requesting pods to launchers, starting vLLM instances, and orchestrating sleep/wake.
```
rubambiza: I think qualifying this as `FMA Controllers` is both expansive and restrictive. In the former case, I expected the description to capture the functionality of the `LauncherPopulationPolicy` controller. In the latter case, the description seems to primarily focus on the Dual-Pods Controller. If we intend to restrict it to the latter, then I'd like to suggest renaming to `Dual-Pods Controller`, but I suspect that we actually want the more expansive path.

This comment applies to all variations of `FMA controller(s)` below.

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+Server-requesting pods are managed through standard Kubernetes controllers such as Deployments and autoscalers. The FMA controller watches these pods and translates scheduler decisions into actions on launcher pods and GPUs.
```
rubambiza: `these` could be mistaken for deployments and autoscalers.

```suggestion
Server-requesting pods are managed through standard Kubernetes controllers such as Deployments and autoscalers. The FMA controller watches requesting pods and translates scheduler decisions into actions on launcher pods and GPUs.
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+When a requesting pod is deleted, the controller puts the corresponding vLLM instance to sleep (model stays in GPU memory). Although the Kubernetes GPU allocation is released when the requesting pod exits, the launcher pod retains the CUDA context and keeps the model in GPU memory. The GPU remains dedicated to that launcher until it is explicitly unbound or the launcher pod is deleted. When a new requesting pod arrives and gets assigned to the same GPU, the controller wakes the sleeping instance, resuming in seconds instead of cold-starting from scratch.
```
rubambiza: >When a requesting pod is deleted, the controller puts the corresponding vLLM instance to sleep (model stays in GPU memory)

This is also applicable when the deployment is scaled down to zero (like we do for our benchmarking), right?

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+FMA also supports instant model switching: if a new requesting pod references a different `InferenceServerConfig`, the FMA controller can direct the bound launcher to swap the loaded model in place, avoiding a full cold start.
```
rubambiza: Full may imply there is partial (which does happen). Up to this point, the different actuation paths have not been covered, so I think removing the word full removes any confusion.

```suggestion
FMA also supports instant model switching: if a new requesting pod references a different `InferenceServerConfig`, the FMA controller can direct the bound launcher to swap the loaded model in place, avoiding a cold start.
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+> [!NOTE]
+> Fast wake only occurs if the Kubernetes scheduler assigns the new requesting pod to the same node (and GPU) where the sleeping vLLM instance resides. In a cluster with a single GPU per node, if the scheduler picks the same node, the GPU is necessarily the same one. In a multi-node pool the scheduler may assign the pod to a different node.
+
+## Configuration
+
+| Parameter               | Value                                                         |
+| ----------------------- | ------------------------------------------------------------- |
+| Model                   | [Qwen/Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-0.6B)  |
```
rubambiza, line 26: I think the 8B model is a better model that balances showing off the sub-second hot-start of FMA without needing a big GPU to fit the model. This would apply all instances below of `Qwen/Qwen3-0.6B` -> `Qwen/Qwen3-8B`

```suggestion
| Model                   | [Qwen/Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-8B)  |
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+export MODEL=Qwen/Qwen3-0.6B
+export CURL_TEST_IMAGE=cfmanteiga/alpine-bash-curl-jq:latest
+export BENCHMARK_REF=main
+export HARNESS=nop
+export WORKLOAD=nop.yaml
+export GATEWAY_CLASS=epponly # options: epponly, gke, agentgateway, istio
+```
+<!-- guide:env.static end -->
+
+- Source the common guide environment variables (`GAIE_VERSION`, `ROUTER_CHART_VERSION`, `ROUTER_STANDALONE_CHART`, …):
```
rubambiza, line 67: Playing a naive reader here, and I don't know what I would source these with. In fact, even as an FMA helper, I don't know what I would put for those 😅

guides/fast-model-actuation/rbac/rolebinding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: fma-launcher-pod-reader
+subjects:
+  - kind: ServiceAccount
+    name: default
+roleRef:
```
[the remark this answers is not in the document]
rubambiza: Agreed with Copilot here. I think a good example is [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/579/changes#diff-63c4f3e5511421bcad9a5d42570e490c386cef31f0b96ceb26c86b0f9d00bee3R228)

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+```bash
+kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/modelserver/
+```
+<!-- guide:deploy.fma_resources end -->
+
+> [!NOTE]
+> This guide uses [Qwen/Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-0.6B) which is publicly accessible and does not require a HuggingFace token. For gated models, you would need to mount the token via a different mechanism (FMA's ISC does not support `secretKeyRef`).
+
+> [!NOTE]
+> `HF_HOME` points at `/tmp/hf_cache`, which is ephemeral node storage. This keeps the guide self-contained, but every fresh launcher re-downloads the model weights and re-runs `torch.compile`. For production — or any repeated benchmarking — back `HF_HOME` with a shared, persistent volume (a PVC, e.g. `ReadWriteMany`) so model weights and compiled graphs persist across launcher pods and are not recomputed on each start.
```
rubambiza, line 169:
```suggestion
> `HF_HOME` points at `/tmp/hf_cache`, which is ephemeral node storage. This keeps the guide self-contained, but every fresh launcher re-downloads the model weights and re-runs `torch.compile`. For production — or any repeated benchmarking — back `HF_HOME` with a shared, persistent volume (e.g., a PVC with `ReadWriteMany`) so model weights and compiled graphs persist across launcher pods and are not recomputed on each start.
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+> [!NOTE]
+> The launcher pod does **not** request GPU resources from the Kubernetes scheduler or device plugin. Instead, the FMA controller sets `CUDA_VISIBLE_DEVICES` to point to the GPU reserved by the corresponding requesting pod, giving the launcher direct access to that GPU via the CUDA runtime. The `runtimeClassName: nvidia` is required on platforms (e.g., OpenShift) where GPU driver libraries are injected via the runtime class rather than the device plugin resource request.
```
rubambiza, line 172: IIRC from the troubleshooting discussion threads, this was not broadly applicable to all versions of OpenShift, right?

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+> [!NOTE]
+> `launcherCount` is **per matching node**. Setting `launcherCount: 1` creates one launcher pod on each node that has `nvidia.com/gpu.present: "true"`. Only launchers that get bound to a requesting pod will actually start a vLLM instance.
+
+### 6. Create Requesting Pods
+
+Create the server-requesting pods that reserve GPUs and trigger model loading. A `Deployment` is used here (rather than a bare `ReplicaSet`) so the requesting pods integrate cleanly with autoscalers such as the Workload Variant Autoscaler (WVA):
```
rubambiza, line 179: First mention of WVA in this README, worth a pointer to what it is compared to "standard" autoscalers like HPA

```suggestion
Create the server-requesting pods that reserve GPUs and trigger model loading. A `Deployment` is used here (rather than a bare `ReplicaSet`) so the requesting pods integrate cleanly with autoscalers such as HPA or the [Workload Variant Autoscaler (WVA)](https://github.com/llm-d/llm-d-workload-variant-autoscaler):
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+<!-- guide:deploy.requester start -->
+```bash
+kubectl apply -n ${NAMESPACE} -k ${REPO_ROOT}/guides/${GUIDE_NAME}/requester/
+
+kubectl wait --for=condition=ready pod -l app=fma-requester -n ${NAMESPACE} --timeout=300s
+```
+<!-- guide:deploy.requester end -->
+
+You should see:
+- 2 requesting pods (`fma-requester-*`) in `Ready` state
```
rubambiza, line 190: I was taught later in my English learning journey (circa PhD student) that numbers less than 10 are always spelled out.

```suggestion
- Two requesting pods (`fma-requester-*`) in `Ready` state
```

guides/fast-model-actuation/README.md
```diff
@@ -0,0 +1,305 @@
...
+- Launcher pods in `Running` state (one per GPU node in your cluster)
```
rubambiza, line 191: Heterogeneous clusters are unlikely at this phase of FMA, but worth making it clear to readers that we are targeting NVIDIA GPUs.

```suggestion
- Launcher pods in `Running` state (one for each node with NVIDIA GPUs in your cluster)
```

guides/fast-model-actuation/README.md
rubambiza: This is confirmed further [below](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/660/changes#diff-1418499fa9a557aa5fc5edb141fa01d2d3a65e991926576a57d6daf58594422eR222). 

Perhaps we should say
> When a requester pod deployment is deleted or scaled down to zero, ...`

guides/fast-model-actuation/guide.yaml
```diff
@@ -0,0 +1,129 @@
...
+          deployment "${FMA_CHART_INSTANCE_NAME}-dual-pods-controller" -n ${NAMESPACE}
+        kubectl wait --for=condition=available --timeout=120s \
+          deployment "${FMA_CHART_INSTANCE_NAME}-launcher-populator" -n ${NAMESPACE}
+
+  standalone:
+    - run: |
+        helm install ${GUIDE_NAME} \
+          ${ROUTER_STANDALONE_CHART} \
+          -f ${REPO_ROOT}/guides/recipes/router/base.values.yaml \
+          -f ${REPO_ROOT}/guides/${GUIDE_NAME}/router/${GUIDE_NAME}.values.yaml \
```
rubambiza, line 71: I am assuming these paths will make more sense once this content is placed relative to the llm-d repo?

rubambiza (changes requested):
Leaving a few suggestions that address some ambiguities in addition to some clarifying questions.

guides/fast-model-actuation/README.md
MikeSpreitzer: (A) The launcher Pods are server-providing Pods, _not_ server-requesting Pods.

(B) I tended to specifically use all-lowercase "vllm" when I wanted to put the focus on the binary or the process, because the binary is named all-lowercase "vllm".

guides/fast-model-actuation/README.md
MikeSpreitzer: > When a requesting pod is deleted, the controller puts the corresponding vLLM instance to sleep (model stays in GPU memory).

That was a typo! Or a misunderstanding. You are critiquing words copied from https://github.com/llm-d/llm-d/pull/1779, written by @diegocastanibm (and here _is_ the right place to critique them). In fact, a sleeping vllm instance keeps the model in **CPU** memory.

It does not matter what commanded deletion of the sever-requesting Pod. It does not matter whether this deletion reduced a set size to zero or not.

guides/fast-model-actuation/README.md
MikeSpreitzer: I do not like these words, they seem to imply some "swapping" functionality that is not there. All we have is sleep/wake, and vllm instance create/delete.

guides/fast-model-actuation/README.md
MikeSpreitzer: We have heterogeneous clusters today. Clusters with different models of nvidia GPUs.

guides/fast-model-actuation/README.md
MikeSpreitzer: Revised.

guides/fast-model-actuation/README.md
MikeSpreitzer: Revised.

guides/fast-model-actuation/README.md
MikeSpreitzer: Yes, I intended "FMA controllers" in the expansive sense. I have made some revisions to be consistent about that.

guides/fast-model-actuation/README.md
MikeSpreitzer: reworded

guides/fast-model-actuation/README.md
MikeSpreitzer: I have revised the text in this neighborhood.

guides/fast-model-actuation/README.md
MikeSpreitzer: Significantly revised to be much more clear.

guides/fast-model-actuation/README.md
rubambiza: Re:
(A), I meant to reflect the old one (i.e., Server-providing Pods)

(B), Sounds good.

guides/fast-model-actuation/README.md
rubambiza: By heterogeneous, I meant by NVIDIA, AMD, Intel, etc. This matters because there were a couple of presentations at the llm-d NYC meetup in April on porting llm-d/vLLM to AMD GPUs. I'm thinking of those folks looking at our well-lit guide.

rubambiza (commented):
This review, which focuses at the one commit since my last review, comes with a single non-blocking nit. There are still some unresolved comments, but that may be the result of the below message from GitHub 👇 

Image

If I find another source where the comments have been addressed (e.g., the Conversation tab of the PR), then I can proceed with another review approving the PR.

Update: the comments are indeed still unresolved.

guides/fast-model-actuation/README.md
rubambiza: Thanks for the clarification. Indeed, I made the comments keeping in mind that @diegocastanibm wrote the original prose, and I do appreciate you taking care of them on his behalf.

MikeSpreitzer:
It is my understanding that we are abandoning the idea of putting this guide in this repo for a while, and returning our focus to https://github.com/llm-d/llm-d/pull/1779 .

To that end, I am closing this PR and have opened https://github.com/diegocastanibm/llm-d/pull/1 to contribute the edits that I worked on here.

PR #661, opened 2026-07-17
Publish an Event when unbinding on invalid routing metadata

waltforme:
This PR addresses an open comment https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/644#discussion_r3572785937.

### DIFF
--- a/charts/fma-controllers/templates/rbac/role.yaml
+++ b/charts/fma-controllers/templates/rbac/role.yaml
@@ -21,3 +21,6 @@ rules:
 - apiGroups: [ "" ]
   resources: [ events ]
   verbs: [ create, patch ]
+- apiGroups: [ "events.k8s.io" ]
+  resources: [ events ]
+  verbs: [ create, patch ]
--- a/charts/fma-controllers/templates/rbac/role.yaml
+++ b/charts/fma-controllers/templates/rbac/role.yaml
@@ -18,3 +18,6 @@ rules:
 - apiGroups: [ "" ]
   resources: [ pods ]
   verbs: [ get, list, watch, create, update, patch, delete ]
+- apiGroups: [ "" ]
+  resources: [ events ]
+  verbs: [ create, patch ]
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -37,9 +37,11 @@ import (
 	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	corev1preinformers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
 	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	kubemetrics "k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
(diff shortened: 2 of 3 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -859,6 +867,8 @@ func getProviderNodeName(pod *corev1.Pod) (string, error) {
 }
 
 func (ctl *controller) Start(ctx context.Context) error {
+	ctl.eventBroadcaster.StartRecordingToSink(&coreclient.EventSinkImpl{Interface: ctl.coreclient.Events("")})
```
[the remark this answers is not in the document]
waltforme: This concern about namespace is incorrect: client-go derives the namespace from the Event.

An example that has the same pattern in this PR, from K8s Deployment controller:
https://github.com/kubernetes/kubernetes/blob/24e2b02af5543d7910c2bb074c7264df5a8f0467/pkg/controller/deployment/deployment_controller.go#L174-L177

charts/fma-controllers/templates/rbac/role.yaml
```diff
@@ -18,3 +18,6 @@ rules:
 - apiGroups: [ "" ]
   resources: [ pods ]
   verbs: [ get, list, watch, create, update, patch, delete ]
+- apiGroups: [ "" ]
```
MikeSpreitzer, line 21: Both https://github.com/kubernetes/api/blob/v0.34.9/core/v1/types.go#L7440 and https://github.com/kubernetes/api/blob/v0.34.9/events/v1/types.go#L34 are a thing. I am not clear on the relation between the two, but I would authorize publishing both here.

pkg/controller/dual-pods/controller.go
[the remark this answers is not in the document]
MikeSpreitzer: There's also this:

```console
mspreitz@mjs13 llm-d-fast-model-actuation % kubectl api-resources | egrep NAME\|events
NAME                                SHORTNAMES   APIVERSION                               NAMESPACED   KIND
events                              ev           v1                                       true         Event
events                              ev           events.k8s.io/v1                         true         Event
```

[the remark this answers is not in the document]
MikeSpreitzer: I agree with @waltforme , and the reason is even simpler. An `Event` object, in either api group, is namespaced so the apiserver has to check for authorization to publish in the relevant namespace. I see `ctl.recorder.Eventf(requestingPod` below, and suppose that the Event's namespace is taken from `requestingPod`.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1179,6 +1179,8 @@ func (ctl *controller) applyBoundLauncherLabels(
 			// loop. Unbind by deleting the requester Pod; a fresh requester
 			// starts over.
 			logger.Error(nil, "Committed ISC routing metadata is invalid; deleting requester Pod to unbind", "problems", problems)
+			ctl.recorder.Eventf(requestingPod, corev1.EventTypeWarning, "InvalidRoutingMetadata",
+				"Deleting server-requesting Pod to unbind: committed ISC routing metadata is invalid: %s", strings.Join(problems, "; "))
```
MikeSpreitzer: "to unbind" is not really describing the goal here. I would write something more like the following. The goal is recovering from the commitment to outdated labels.

> Deleting server-requesting Pod because bound launcher is committed to outdated ISC routing metadata.

pkg/controller/dual-pods/controller.go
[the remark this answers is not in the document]
MikeSpreitzer: BTW, the change suggested by Copilot would not be wrong. It commits the event publisher to one namespace, and this controller does indeed work only on workload in that one namespace.

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
waltforme: Reworded.

charts/fma-controllers/templates/rbac/role.yaml
waltforme: Changed as suggested.

MikeSpreitzer: FYI, this is a list. It could have two members.

MikeSpreitzer (approved):
This is good enough to merge. I left one suggestion for simplification.

charts/fma-controllers/templates/rbac/role.yaml
waltforme: Thanks for the heads up. Let me merge this PR for now. Will make use of these lists in the future.

PR #666, opened 2026-07-20
Fix skill frontmatter, emphasize rule markdown structure

MikeSpreitzer:
These changes are motivated by poor performance of Claude with the existing instructions, and are based on suggestions from Claude about how to improve it.

### DIFF
--- a/.agents/skills/our-git-commits/SKILL.md
+++ b/.agents/skills/our-git-commits/SKILL.md
@@ -1,3 +1,4 @@
+---
 name: our-git-commits
 description: Things to do when making a Git commit
 trigger: When the user asks to commit changes, create a commit, make a git commit, amend a commit, revise a commit, reword a commit message, or any other action that results in running a `git commit` command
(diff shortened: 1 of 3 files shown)


MikeSpreitzer:
I just tested this by asking Claude Code to review PR 663.

On the plus side, this time Claude Code correctly found the content of DR-10; previously it used a bogus search for text between "DR-10" and "DR-11".

On the neutral side, this made no difference regarding finding the "pr-security-review" skill. I got the following, which is about the same as before. Note: this is only an annoying bit of verbosity, Claude _does_ eventually find and use the skill.

```
pr-security-review)
Initializing…
Error: Unknown skill: pr-security-review

The pr-security-review skill named in AGENTS.md is not registered/available in this session. Let me check what's under .agents/ to see if there's a corresponding SKILL.md I should follow manually.

  Listed 1 directory

The skill exists as a file even though it isn't registered. Let me read it and follow it manually.
```

waltforme (approved):

PR #669, opened 2026-07-21
Update README.md

MikeSpreitzer:
Be clearer about what "model swapping" really means.

Give correct list of names for the launcher population controller.

Update to current status of development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,22 +19,28 @@ has its model tensors in main (CPU) memory rather than accelerator
 inference requests --- but has freed up accelerator resources for use
 by a different instance. But the sleeping instance is still a running
 process (e.g., it can still serve administrative requests) as far as
-the OS is concerned. The process of waking up the sleeping instance is
-very fast; for example, taking about 3 seconds for a model with 64 GiB
-of tensor data. This behavior is available in vLLM today.
+the OS is concerned. And that process is still the main process of a
+container in a Pod, as far as Kubernetes is concerned. The process of
+waking up the sleeping instance is very fast; for example, taking
+about 3 seconds for a model with 64 GiB of tensor data. This behavior
+is available in vLLM today.
 
-2. Model swapping. In model swapping techniques, there is a persistent
-management process that can run various subsidiary inference server
-processes over time. The management process does basic code loading
-and initialization work of the inference server so that this work does
-not have to be done at the startup of the inference server process,
-reducing that startup latency. The inference servers may be able to
-sleep and wake up.
+2. A vllm launcher process. This is a process that can have multiple
+child processes, each running a distinct vllm instance (which is
+itself the usual parent/child pair of main and engine). The launcher
+loads the Python modules, shaving the time for that off of the startup
+of each launched vllm instance. The launcher has a network API for
+CRUDL of vllm instances.
 
-A process with such flexibility does not easily fit into the
-Kubernetes milieu. The most obvious and natural way in Kubernetes to
-define a desired inference server is to create a `Pod`
-object. However, a `Pod` has a static allocation of accelerator
+By reducing the startup latency of vllm instances, these ideas reduce
+the latency to change which model a given accelerator (or array of
+accelerators) is being used for. We sometimes use the term "model
+swapping" to evoke this idea.
+
+A process with sleep/wake and/or launcher functionality does not
+easily fit into the Kubernetes milieu. The most obvious and natural
+way in Kubernetes to define a desired inference server is to create a
+`Pod` object. However, a `Pod` has a static allocation of accelerator
 resources and a static command line. That is, the obvious way to
 define a `Pod` is such that it serves one fixed model and server
 options, with no resource-freeing hiatus. This repository contains a
@@ -57,10 +63,11 @@ The topics above are realized by the following software components.
   the [inference_server/launcher](inference_server/launcher)
   directory.
 
-- A **launcher-populator** controller, which watches LauncherConfig
-  and LauncherPopulationPolicy custom resources and ensures that the
-  right number of launcher pods exist on each node. This controller is
-  also written in Go.
+- A **launcher-population controller** (also called the **launcher
+  populator** or simply the **populator**), which watches
+  LauncherConfig and LauncherPopulationPolicy objects and ensures that
+  the right number of launcher pods exist on each node. This
+  controller is also written in Go.
 
 These controllers are deployed together via a unified Helm chart at
 [charts/fma-controllers](charts/fma-controllers). The chart also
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,22 +19,28 @@ has its model tensors in main (CPU) memory rather than accelerator
...
+2. A vllm launcher process. This is a process that can have multiple
+child processes, each running a distinct vllm instance (which is
+itself the usual parent/child pair of main and engine). The launcher
+loads the Python modules, shaving the time for that off of the startup
+of each launched vllm instance. The launcher has a network API for
+CRUDL of vllm instances.
+
+By reducing the startup latency of vllm instances, these ideas reduce
+the latency to change which model a given accelerator (or array of
+accelerators) is being used for. We sometimes use the term "model
```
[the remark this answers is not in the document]
MikeSpreitzer: All-lowercase "vllm" is used in some places to emphasize the fact that the referent is the `vllm` binary and/or a process running it.

"CRUD" is well-known and "CRUDL" is closely associated.

rubambiza, line 37: Another usage in the academic/technical circles is "[model hot-swapping](https://dl.acm.org/doi/10.1145/3731599.3767354)"

README.md
```diff
@@ -19,22 +19,28 @@ has its model tensors in main (CPU) memory rather than accelerator
...
-sleep and wake up.
-
-A process with such flexibility does not easily fit into the
-Kubernetes milieu. The most obvious and natural way in Kubernetes to
-define a desired inference server is to create a `Pod`
-object. However, a `Pod` has a static allocation of accelerator
+the OS is concerned. And that process is still the main process of a
+container in a Pod, as far as Kubernetes is concerned. The process of
+waking up the sleeping instance is very fast; for example, taking
+about 3 seconds for a model with 64 GiB of tensor data. This behavior
```
rubambiza, line 25: Is this 3s/64GB documented anywhere? Maybe like a blog? It would be great to point to something concrete. Alternatively, we could eventually modify this text to link to our own upcoming blog which shows it takes about 1 second for a Qwen2.5-32B-Instruct model with ~32GB of tensor data.

README.md
```diff
@@ -57,10 +63,11 @@ The topics above are realized by the following software components.
...
   directory.
 
-- A **launcher-populator** controller, which watches LauncherConfig
-  and LauncherPopulationPolicy custom resources and ensures that the
-  right number of launcher pods exist on each node. This controller is
-  also written in Go.
+- A **launcher-population controller** (also called the **launcher
+  populator** or simply the **populator**), which watches
+  LauncherConfig and LauncherPopulationPolicy objects and ensures that
+  the right number of launcher pods exist on each node. This
```
rubambiza, line 69: At quick scan it's easy to misunderstand that population policies are tied to a particular node.

```suggestion
  the right number of launcher pods exist on each node for each defined population policy object. This
```

README.md
```diff
@@ -80,9 +87,8 @@ types are in [pkg/api](pkg/api).
 
 The development roadmap has three milestones. Milestone 2, which
 introduced vLLM sleep/wake without the launcher, is finished.
-Milestone 3, which adds launcher-based model swapping where a
-persistent launcher process manages vLLM instances on each node, is
-under implementation.
+Milestone 3, which adds the launcher, is essentially finished to a
```
rubambiza: > to a first level of ambition

Which is? I think either a phrase or something here would be nice _or_ we can keep it at finished like the previous sentence on Milestone 2 does.

rubambiza (changes requested):
A few suggestions to further clarify terminology and concrete pointers for claims. LGTM once they are addressed.

README.md
MikeSpreitzer: My purpose here is to alert the reader to terminology that _we_ have been, and I suspect will continue to be, using.

You cited one paper that describes one technique and says it is one example of "model hot swapping". That does not tell me what the rest of the world understands by that term, so I am hesitant to make an assertion relative to _that_ term here. I do not think we need to. I think that, for this repo's README, it is enough to explain what we mean by the terms that we use in this repo; this does not have as high a responsibility as a paper written for a conference or journal.

README.md
MikeSpreitzer: These words were written long ago, based our experience then. Sure, when we have something citable then we can update to cite it.

README.md
MikeSpreitzer: A policy object is _not_ "tied to a particular node". Each LPP object has a selector predicate that says yes or no for any given node. Multiple LPP objects can exist and select the same node. The real deal involves LauncherConfig objects too. And the possibility of even more launchers being required due to a surprising number of server-requesting Pods. The full and complete statement is at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/v0.6.2/api/fma/v1alpha1/launcherpopulationpolicy_types.go#L28-L40 . The goal here is only to identify the issue, not precisely state our formulation.

README.md
```diff
@@ -57,10 +63,11 @@ The topics above are realized by the following software components.
...
 
-- A **launcher-populator** controller, which watches LauncherConfig
-  and LauncherPopulationPolicy custom resources and ensures that the
-  right number of launcher pods exist on each node. This controller is
-  also written in Go.
+- A **launcher-population controller** (also called the **launcher
+  populator** or simply the **populator**), which watches
+  LauncherConfig and LauncherPopulationPolicy objects and ensures that
+  the right number of launcher pods exist on each node. This
+  controller is also written in Go.
```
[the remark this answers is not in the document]
MikeSpreitzer: This is observed to not be a problem for GitHub, which is the Markdown renderer that matters for us.

README.md
MikeSpreitzer: Revised.

README.md
```diff
@@ -78,11 +85,14 @@ The repository defines three Custom Resource Definitions (CRDs):
...
 types are in [pkg/api](pkg/api).
 
-The development roadmap has three milestones. Milestone 2, which
-introduced vLLM sleep/wake without the launcher, is finished.
-Milestone 3, which adds launcher-based model swapping where a
-persistent launcher process manages vLLM instances on each node, is
-under implementation.
+The development roadmap started with three milestones and then shifts
+to more nuanced "releases". Milestone 2, which introduced vLLM
+sleep/wake without the launcher, is finished.  Milestone 3 (AKA minor
```
rubambiza, line 90: Thanks for the more detailed status update.

README.md
rubambiza: Fair enough.

README.md
rubambiza: I have filed a tracking issue: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/670

rubambiza (approved):
LGTM.

PR #673, opened 2026-07-22
launcher-populator: report and retry stuck launchers

delavet:
Second part of #585, following up on #635.

#635 classified launchers as stuck_scheduling / stuck_starting for the fma_launcher_pod_count metric only. This makes the launcher-populator act on that classification, per the guidance in #585:

- Durable state on the Pod — a currently-stuck launcher gets the dual-pods.llm-d.ai/launcher-stuck=true label, so operators can locate it with kubectl get pods -l dual-pods.llm-d.ai/launcher-stuck=true even after the Event expires. 
- Warning Event — a LauncherStuck Event is published on the Pod.
- Single retry, restart-safe — a stuck launcher is retried once by creating a replacement carrying an incremented dual-pods.llm-d.ai/launcher-retry-count annotation; the stuck original is then removed by the next reconcile's excess deletion (which prefers stuck, lowest-generation Pods). Creating the replacement before removing the original keeps the counter on some Pod at all times, so the single-retry cap survives a controller restart or a failed Create. A launcher still stuck after its retry is left in place (still counted, so it doesn't pile up with fresh replacements on a broken node) and reported.

### DIFF
--- a/docs/dual-pods.md
+++ b/docs/dual-pods.md
@@ -583,9 +583,9 @@ you would typically use when the nodes have 8 GPUs.
 
 ### FYI labels and annotations
 
-The FMA controllers put some labels and annotations on Pods purely for
-the sake of observability; these labels and annotations are not
-consumed by the FMA controllers. They are as follows.
+The FMA controllers put some labels and annotations on Pods primarily
+for observability. Except where noted below, the controllers do not
+consume them. They are as follows.
 
 - annotation **dual-pods.llm-d.ai/accelerators**. Put on both
   requester and provider Pods. Value is a comma-separated list of GPU
@@ -606,6 +606,18 @@ consumed by the FMA controllers. They are as follows.
   while it is bound to a launcher. Value is the ID that FMA uses to
   distinguish the relevant vLLM instance from others in the same
   launcher.
+
+- label **dual-pods.llm-d.ai/launcher-stuck**. Maintained on an
+  unbound launcher Pod when the launcher has remained unscheduled or
+  not Ready for at least the corresponding configured stuck
+  threshold. Its value is "true". The launcher population controller
+  consumes this label to avoid publishing a duplicate `LauncherStuck`
+  Event on every reconciliation, and removes it if the launcher
+  recovers. To list currently stuck launchers, use:
+
+  ```shell
+  kubectl get pods -l dual-pods.llm-d.ai/launcher-stuck=true
+  ```
 
 ### Kubernetes Event objects
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -71,21 +71,6 @@ Labels are as follows.
 
 - `isc_name`: Name of the relevant InferenceServerConfig
 
-### fma_launcher_pod_count
-
-Vector of gauges: Number of launcher Pods.
-
-Labels are as follows.
-
-- `lcfg_name`: Name of the relevant LauncherConfig
-
-- `phase`: one of the following:
-    - `bound`: bound to a server-requesting Pod
-    - `unbound`: not bound, but has not been in this state for too long
-    - `stale`: was created from outdated `LauncherConfig` contents
-    - `stuck_scheduling`: not scheduled, and age meets or exceeds the `--stuck-scheduling-threshold` argument of the launcher population controller
-    - `stuck_starting`: scheduled but not yet "ready", and the time since being scheduled meets or exceeds the `--stuck-starting-threshold` argument of the launcher population controller
-
 ### fma_isc_count
 
 Vector of gauges: Number of InferenceServerConfig objects.
@@ -93,6 +78,45 @@ Vector of gauges: Number of InferenceServerConfig objects.
 Labels are as follows.
 
 - `launcher_config_name`: Name of the relevant LauncherConfig
+
+### fma_launcher_pod_count
+
+Vector of gauges: Number of launcher Pods for each LauncherConfig and
+lifecycle phase. This metric deliberately has no Node label, so its
+cardinality does not grow with the number of Nodes.
+
+Labels are as follows.
+
+- `lcfg_name`: name of the relevant LauncherConfig
+- `phase`: one of the following:
+  - `bound`: assigned to a server-requesting Pod
+  - `unbound`: uses the current launcher template, is not bound, and is either
+    Ready or has not reached a stuck threshold
+  - `stuck_scheduling`: uses the current launcher template, is unbound and
+    not scheduled, and has reached the configured scheduling threshold,
+    measured from Pod creation
+  - `stuck_starting`: uses the current launcher template, is unbound,
+    scheduled but not Ready, and has reached the configured starting
+    threshold, measured from scheduling
+  - `stale`: is unbound and was created from a superseded launcher template
+
+The launcher-populator's `--stuck-scheduling-threshold` and
+`--stuck-starting-threshold` flags configure the two stuck thresholds.
+When the controller observes a retained launcher in either stuck phase without
+the stuck label, it publishes a `LauncherStuck` Warning Event and sets the
+`dual-pods.llm-d.ai/launcher-stuck=true` label. The label makes the current
+condition discoverable after the Event expires and suppresses further Events
+while the Pod remains stuck. If the launcher recovers, the controller removes
+the label.
+
+The Event is emitted before the label Patch is attempted. Because they are
+separate Kubernetes API transactions, a controller failure between them or a
+failed Patch can produce a duplicate Event on a later reconcile. This ordering
+deliberately favors a possible duplicate over losing the Event entirely.
+
+A stuck launcher remains in place and continues to count toward the desired
+launcher population. The controller does not automatically replace or retry
+it; users decide how to respond using the metric, Event, and label.
 
 ## FMA requester-provider binding
 
--- a/pkg/controller/common/interface.go
+++ b/pkg/controller/common/interface.go
@@ -35,6 +35,15 @@ const (
 	// LauncherTemplateHashAnnotationKey is the node-independent template hash on a launcher Pod, used for spec-drift detection.
 	LauncherTemplateHashAnnotationKey = "dual-pods.llm-d.ai/launcher-populator-template-hash"
 
+	// LauncherStuckLabelKey is set by the launcher-populator on a launcher Pod
+	// that is currently stuck (existing past a configured threshold without
+	// becoming Ready, or without ever being scheduled). Its value is "true".
+	// It is findable state that outlives the corresponding Kubernetes Event.
+	// The label denotes the *current* condition: the populator removes it again
+	// if the launcher later recovers, so it is never a false positive on a
+	// launcher that has since become Ready.
+	LauncherStuckLabelKey = "dual-pods.llm-d.ai/launcher-stuck"
+
 	// LauncherServicePort is the port number on which the launcher exposes its HTTP service
 	// for the management of vLLM instances.
 	// This is a contract between the controllers and the launcher implementation.
--- a/pkg/controller/launcher-populator/populator.go
+++ b/pkg/controller/launcher-populator/populator.go
@@ -30,9 +30,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	corev1preinformers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
 	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 
@@ -82,6 +84,8 @@ func NewController(
 		clock:                    clock.RealClock{},
 		metrics:                  newMetricsState(),
 	}
+	ctl.eventBroadcaster = record.NewBroadcaster()
+	ctl.recorder = ctl.eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: ControllerName})
 	ctl.policy = newDigestedPolicy()
 
 	// digestQueue carries funcItem (LC/LPP/Node references). Single worker, so
@@ -132,14 +136,20 @@ type controller struct {
 	coreclient    coreclient.CoreV1Interface
 	fmaclient     fmaclientv1alpha1.FmaV1alpha1Interface
 	namespace     string
-	podInformer   cache.SharedIndexInformer
-	podLister     corev1listers.PodLister
-	nodeInformer  cache.SharedIndexInformer
-	nodeLister    corev1listers.NodeLister
-	lppInformer   cache.SharedIndexInformer
-	lppLister     fmalisters.LauncherPopulationPolicyLister
-	lcInformer    cache.SharedIndexInformer
-	lcLister      fmalisters.LauncherConfigLister
+
+	// eventBroadcaster/recorder publish Kubernetes Events on launcher Pods (e.g.
+	// a Warning when a launcher becomes stuck).
+	eventBroadcaster record.EventBroadcaster
+	recorder         record.EventRecorder
+
+	podInformer  cache.SharedIndexInformer
+	podLister    corev1listers.PodLister
+	nodeInformer cache.SharedIndexInformer
+	nodeLister   corev1listers.NodeLister
+	lppInformer  cache.SharedIndexInformer
+	lppLister    fmalisters.LauncherPopulationPolicyLister
+	lcInformer   cache.SharedIndexInformer
+	lcLister     fmalisters.LauncherConfigLister
 
 	// digestQueue processes API-object references (LC/LPP/Node). Single worker;
 	// the SOLE writer of ctl.policy.
@@ -161,14 +171,12 @@ type controller struct {
 	policy *digestedPolicy
 
 	// stuckSchedulingThreshold is the minimum age (since creation) at which an
-	// unscheduled launcher is reported in the "stuck_scheduling" phase of the
-	// fma_launcher_pod_count metric. It does not change reconcile behavior.
+	// unscheduled launcher is classified in the "stuck_scheduling" phase.
 	stuckSchedulingThreshold time.Duration
 
 	// stuckStartingThreshold is the minimum age (since scheduling) at which a
-	// scheduled, not-yet-Ready launcher is reported in the "stuck_starting"
-	// phase of the fma_launcher_pod_count metric. It does not change reconcile
-	// behavior.
+	// scheduled, not-yet-Ready launcher is classified in the "stuck_starting"
+	// phase.
 	stuckStartingThreshold time.Duration
 
 	// clock is the time source for metric classification; real in production,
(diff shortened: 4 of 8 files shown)


pkg/controller/launcher-populator/populator.go
```diff
@@ -45,6 +49,13 @@ import (
 
 const ControllerName = "launcher-populator"
 
+// maxLauncherRetries caps how many times the populator recreates a launcher for
+// a slot in response to it becoming stuck. Per Mike Spreitzer's guidance on
+// issue #585 this is a single retry; a launcher that is still stuck after its
+// retry is left in place and reported (via a label and an Event) rather than
+// recreated again, so that it does not silently pile up on a broken node.
```
MikeSpreitzer: The root cause might be in the LauncherConfig rather than the Node being broken.

pkg/controller/launcher-populator/populator.go
```diff
@@ -161,14 +180,15 @@ type controller struct {
...
+	// reconcileKey, labeled/Evented and retried once.
 	stuckSchedulingThreshold time.Duration
 
 	// stuckStartingThreshold is the minimum age (since scheduling) at which a
-	// scheduled, not-yet-Ready launcher is reported in the "stuck_starting"
-	// phase of the fma_launcher_pod_count metric. It does not change reconcile
-	// behavior.
+	// scheduled, not-yet-Ready launcher is classified as stuck: reported in the
+	// "stuck_starting" phase of the fma_launcher_pod_count metric and, in
+	// reconcileKey, labeled/Evented and retried once.
```
MikeSpreitzer: Duplicating the details of the consequences of being stuck is not great.

The design outline currently sitting in the opening comment of the PR is pretty good. It wouldn't be bad to capture that in a Git-controlled file.

pkg/controller/launcher-populator/stuck_test.go
```diff
@@ -0,0 +1,560 @@
...
+
+// containsSubstr reports whether any event string contains sub.
+func containsSubstr(events []string, sub string) bool {
+	for _, e := range events {
+		if strings.Contains(e, sub) {
+			return true
+		}
+	}
+	return false
+}
```
MikeSpreitzer: FYI, generic functions are great. They can eliminate a lot of helper functions that do simple loops. For example, this one here could be eliminated by defining

```go
func SubstringIsIn(sub string) func(super string) bool {
	return func(super string) bool {
		return strings.Contains(super, sub)
	}
}
```

(which is likely to be useful in multiple places)

and then using `slices.IndexFunc(events, SubstringIsIn(sub)) >= 0` instead of calling this helper function.

MikeSpreitzer: Or even more generally useful, define

```
func Curry2of2[T1, T2, Return any](uncurried func(T1, T2) Return, v2 T2) func(T1) Return {
	return func(v1 T1) Return { return uncurried(v1, v2) }
}
```

and test `slices.IndexFunc(events, Curry2of2(strings.Contains, sub)) >= 0`

MikeSpreitzer: While I am on the subject, consider

```
func SliceFilter[Elt any](slice []Elt, pass func(Elt) bool) []Elt {
	ret := make([]Elt, 0, len(slice))
	for _, elt := range slice {
		if pass(elt) {
			ret = append(ret, elt)
		}
	}
	return ret
}

func Not1[T1 any](base func(T1) bool) func(T1) bool {
	return func(v1 T1) bool { return !base(v1) }
}
```

(both looking generally useful).

That would enable replacing

```
		retained := make([]*corev1.Pod, 0, len(liveUnboundCurrentPods))
		for _, pod := range liveUnboundCurrentPods {
			if !deletedExcess.Has(pod.UID) {
				retained = append(retained, pod)
			}
		}
```

with `retained := SliceFilter(liveUnboundCurrentPods, Not1(deletedExcess.Has))`

pkg/controller/launcher-populator/stuck_test.go
```diff
@@ -0,0 +1,560 @@
...
+
+// TestHandleStuckLaunchersExhaustedLabelsAndReports verifies that a stuck
+// launcher that has used its retry is labeled and reported once, and not
+// recreated.
+func TestHandleStuckLaunchersExhaustedLabelsAndReports(t *testing.T) {
+	pod := scheduledNotReadyAt(stuckLauncherPod("launcher-stuck-2", "1"), stuckTestNow.Add(-10*time.Minute))
+	ctl, cs, rec := newStuckTestController(stuckTestNow, pod)
+
+	created, err := ctl.handleStuckLaunchers(context.Background(), stuckTestKey(), testNode(), []*corev1.Pod{pod}, testTemplateHash, nodeTemplate())
```
MikeSpreitzer: Use `t`'s `Context` instead of `context.Background()`

pkg/controller/launcher-populator/populator.go
```diff
@@ -547,6 +597,134 @@ func (ctl *controller) reconcileKey(ctx context.Context, key NodeLauncherKey, de
...
+			// and is state that outlives it).
+			if labeled {
+				continue
+			}
+			ctl.recorder.Eventf(pod, corev1.EventTypeWarning, "LauncherStuck",
+				"Launcher is stuck (%s) and has exhausted its %d retry; leaving it in place for investigation", phase, maxLauncherRetries)
+			if err := ctl.setStuckLabel(ctx, pod, true); err != nil {
+				return created, err
+			}
+			logger.Info("Reported permanently stuck launcher", "pod", pod.Name, "uid", pod.UID, "phase", phase, "node", key.NodeName)
```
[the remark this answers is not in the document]
MikeSpreitzer: I disagree. The controller can crash between any two Kube api transactions. It is better to emit the event first: duplicate events are better than a dropped event.

MikeSpreitzer:
I am sorry, I fear that I did not think this through well enough earlier. I overlooked two things.

1. For a launcher Pod that is not even scheduled yet, a replacement is going to get the same treatment --- there is no point in making a replacement.
2. A launcher Pod should have a _large_ request=limit for main memory, to accommodate all the model tensors when all of its vLLM instances are sleeping. A launcher Pod that is stuck after being scheduled is counted by the Kubernetes Pod scheduler as using that amount of main memory. Adding a replacement adds another request for a _large_ amount of main memory. I mean a significant fraction of the Node's main memory capacity. With a population of N launchers on a Node, if each one is stuck after scheduling then adding one retry each will double the amount of main memory being requested in total. That could be too much.

So, after further consideration, I think that automatic retries are a bad idea. Let's just do the observability things and let the users react.

delavet:
> I am sorry, I fear that I did not think this through well enough earlier. I overlooked two things.
> 
> 1. For a launcher Pod that is not even scheduled yet, a replacement is going to get the same treatment --- there is no point in making a replacement.
> 2. A launcher Pod should have a _large_ request=limit for main memory, to accommodate all the model tensors when all of its vLLM instances are sleeping. A launcher Pod that is stuck after being scheduled is counted by the Kubernetes Pod scheduler as using that amount of main memory. Adding a replacement adds another request for a _large_ amount of main memory. I mean a significant fraction of the Node's main memory capacity. With a population of N launchers on a Node, if each one is stuck after scheduling then adding one retry each will double the amount of main memory being requested in total. That could be too much.
> 
> So, after further consideration, I think that automatic retries are a bad idea. Let's just do the observability things and let the users react.

Okay, and in most cases, retrying likely won’t help either. I’ll fix the content of the PR.

delavet:
> I am sorry, I fear that I did not think this through well enough earlier. I overlooked two things.
> 
> 1. For a launcher Pod that is not even scheduled yet, a replacement is going to get the same treatment --- there is no point in making a replacement.
> 2. A launcher Pod should have a _large_ request=limit for main memory, to accommodate all the model tensors when all of its vLLM instances are sleeping. A launcher Pod that is stuck after being scheduled is counted by the Kubernetes Pod scheduler as using that amount of main memory. Adding a replacement adds another request for a _large_ amount of main memory. I mean a significant fraction of the Node's main memory capacity. With a population of N launchers on a Node, if each one is stuck after scheduling then adding one retry each will double the amount of main memory being requested in total. That could be too much.
> 
> So, after further consideration, I think that automatic retries are a bad idea. Let's just do the observability things and let the users react.

Implemented this direction. I removes automatic retries and retry metadata; a stuck launcher now remains in place and continues to count toward the desired population. The controller only exposes the condition through `fma_launcher_pod_count`, a LauncherStuck Warning Event, and the durable stuck label.

pkg/controller/launcher-populator/populator.go
```diff
@@ -504,17 +526,24 @@ func (ctl *controller) reconcileKey(ctx context.Context, key NodeLauncherKey, de
...
+			}
 			err := ctl.coreclient.Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{
 				Preconditions: &metav1.Preconditions{UID: &pod.UID, ResourceVersion: &pod.ResourceVersion},
 			})
 			if err != nil {
 				if apierrors.IsNotFound(err) || apierrors.IsGone(err) {
 					ctl.expectations.expectDeletion(key, pod.UID)
+					deletedExcess.Insert(pod.UID)
 					numToDelete--
 					continue
```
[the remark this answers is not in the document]
delavet: Fixed. The reconcile now records every Pod selected for excess deletion in an excessDeletionTargets set before issuing the Delete request.

pkg/controller/launcher-populator/populator.go
delavet: Removed the broken-Node assumption together with the automatic retry logic

pkg/controller/launcher-populator/populator.go
delavet: I noticed that `fma_launcher_pod_count` is not yet documented in `doc/metrics.md`, so I’ve moved these descriptions into that document. Please confirm whether this handling is appropriate.

pkg/controller/launcher-populator/stuck_test.go
delavet: done

pkg/controller/launcher-populator/stuck_test.go
delavet: The containsSubstr helper was removed when the retry-oriented tests were removed, so the `SubstringIsIn` and `Curry2of2` variants no longer have a call site.

I kept the retained-Pod filtering as a direct loop rather than introduce `SliceFilter` and `Not1`. Because the current `SliceFilter` and `Not1` appear to be called only once, we may can focus on addressing such issues in a future PR aimed at code-style improvements?

pkg/controller/launcher-populator/populator.go
MikeSpreitzer: Did you see #684 ?

pkg/controller/launcher-populator/stuck_test.go
MikeSpreitzer: Sure. I suggest them because I am sure that I have seen other code that would be simplified by them.

OTOH, that sort of PR rarely gets done. That's why I implicitly suggested incremental improvements.

docs/metrics.md
MikeSpreitzer, line 1: #684 has merged

pkg/controller/common/interface.go
```diff
@@ -35,6 +35,17 @@ const (
...
+	// LauncherStuckLabelKey is set by the launcher-populator on a launcher Pod
+	// that is currently stuck (existing past a configured threshold without
+	// becoming Ready, or without ever being scheduled). Its value is "true".
+	// It is findable state that outlives the corresponding Kubernetes Event, so
+	// an operator can locate stuck launchers with a label selector:
+	//   kubectl get pods -l dual-pods.llm-d.ai/launcher-stuck=true
+	// The label denotes the *current* condition: the populator removes it again
+	// if the launcher later recovers, so it is never a false positive on a
+	// launcher that has since become Ready.
+	LauncherStuckLabelKey = "dual-pods.llm-d.ai/launcher-stuck"
```
MikeSpreitzer, line 45: This should also be documented in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/main/docs/dual-pods.md#fyi-labels-and-annotations

That is perhaps a better place for exhibiting a `kubectl` query.

pkg/controller/launcher-populator/populator.go
```diff
@@ -547,6 +574,67 @@ func (ctl *controller) reconcileKey(ctx context.Context, key NodeLauncherKey, de
...
+
+		if labeled {
+			continue
+		}
+		ctl.recorder.Eventf(pod, corev1.EventTypeWarning, "LauncherStuck",
+			"Launcher is stuck (%s); leaving it in place for investigation", phase)
+		if err := ctl.setStuckLabel(ctx, pod, true); err != nil {
+			return err
+		}
+		logger.Info("Reported stuck launcher", "pod", pod.Name, "uid", pod.UID, "phase", phase)
```
MikeSpreitzer: This is not wrong, but could be simplified by taking a "currently reported ?= desired report" stance. `continue` if `labeled == isStuck()`, otherwise publish a report and set the label to the desired value.

pkg/controller/launcher-populator/stuck_test.go
```diff
@@ -0,0 +1,452 @@
...
+	healthy := ready(stuckLauncherPod("launcher-healthy"))
+	stuck := scheduledNotReadyAt(stuckLauncherPod("launcher-stuck"), stuckTestNow.Add(-10*time.Minute))
+	ctl, cs, rec := newStuckTestController(stuckTestNow, healthy, stuck)
+	cs.PrependReactor("delete", "pods", func(action k8stesting.Action) (bool, runtime.Object, error) {
+		deleteAction := action.(k8stesting.DeleteAction)
+		if deleteAction.GetName() != stuck.Name {
+			return false, nil, nil
+		}
+		return true, nil, apierrors.NewConflict(
+			schema.GroupResource{Resource: "pods"},
```
MikeSpreitzer: Better to use what the API says about its own API group: `corev1.Resource("pods")`.

docs/metrics.md
delavet: Sorry for the oversight—I’ve rebased onto `main` and removed the redundant metric descriptions.

pkg/controller/common/interface.go
delavet: done

pkg/controller/launcher-populator/populator.go
delavet: done

pkg/controller/launcher-populator/stuck_test.go
delavet: done

pkg/controller/launcher-populator/stuck_test.go
delavet: I agree with your perspective. As a first step, I have already implemented two generic functions using the recommended approach.

docs/dual-pods.md
```diff
@@ -607,6 +607,18 @@ consumed by the FMA controllers. They are as follows.
   distinguish the relevant vLLM instance from others in the same
   launcher.
 
+- label **dual-pods.llm-d.ai/launcher-stuck**. Maintained on an
+  unbound launcher Pod when the launcher has remained unscheduled or
+  not Ready for at least the corresponding configured stuck
+  threshold. Its value is "true". The launcher population controller
+  consumes this label to avoid publishing a duplicate `LauncherStuck`
+  Event on every reconciliation, and removes it if the launcher
```
MikeSpreitzer, line 615: This is not really a significant read of that label. It is just part of maintaining the label and publishing an event when changing the label.

docs/dual-pods.md
```diff
@@ -583,9 +583,9 @@ you would typically use when the nodes have 8 GPUs.
 
 ### FYI labels and annotations
 
-The FMA controllers put some labels and annotations on Pods purely for
-the sake of observability; these labels and annotations are not
-consumed by the FMA controllers. They are as follows.
+The FMA controllers put some labels and annotations on Pods primarily
+for observability. Except where noted below, the controllers do not
+consume them. They are as follows.
```
MikeSpreitzer, line 588: I think that this revision is not correct, as explained in my comment on the stuck label.

docs/metrics.md
MikeSpreitzer: This switches the ordering of fma_isc_count and fma_launcher_pod_count.

I deliberately put the two pod metrics first, since the marquee description here is "dual pods".

docs/metrics.md
```diff
@@ -71,28 +71,52 @@ Labels are as follows.
...
+the stuck label, it publishes a `LauncherStuck` Warning Event and sets the
+`dual-pods.llm-d.ai/launcher-stuck=true` label. The label makes the current
+condition discoverable after the Event expires and suppresses further Events
+while the Pod remains stuck. If the launcher recovers, the controller removes
+the label.
+
+The Event is emitted before the label Patch is attempted. Because they are
+separate Kubernetes API transactions, a controller failure between them or a
+failed Patch can produce a duplicate Event on a later reconcile. This ordering
+deliberately favors a possible duplicate over losing the Event entirely.
```
MikeSpreitzer, line 115: This is a level of detail that is usually omitted from user-facing documentation. Every controller that publishes events faces this issue.

docs/metrics.md
```diff
@@ -71,28 +71,52 @@ Labels are as follows.
...
+
+A stuck launcher remains in place and continues to count toward the desired
+launcher population. The controller does not automatically replace or retry
+it; users decide how to respond using the metric, Event, and label.
```
MikeSpreitzer, line 119: This material applies to more than the metrics. We will need to find a better place for it. Can be done in follow-on PR.

pkg/controller/launcher-populator/populator.go
```diff
@@ -530,7 +542,19 @@ func (ctl *controller) reconcileKey(ctx context.Context, key NodeLauncherKey, de
 		}
 	}
 
-	// If any deletions happened or are in progress, requeue before creating.
+	// Report stuck launchers among only the Pods not targeted for excess
+	// deletion in this reconcile. Skipped in orphan cleanup (nil template).
+	if nodeIndependentLauncherTemplate != nil {
+		retained := utils.SliceFilter(liveUnboundCurrentPods, utils.Not1(func(pod *corev1.Pod) bool {
```
MikeSpreitzer, line 548: It would be simpler to write `!` in the `func` here than apply `utils.Not1` to it.

pkg/controller/launcher-populator/populator.go
```diff
@@ -547,6 +571,64 @@ func (ctl *controller) reconcileKey(ctx context.Context, key NodeLauncherKey, de
...
+				"Launcher is stuck (%s); leaving it in place for investigation", phase)
+		}
+		if err := ctl.setStuckLabel(ctx, pod, stuck); err != nil {
+			return err
+		}
+		if stuck {
+			logger.Info("Reported stuck launcher", "pod", pod.Name, "uid", pod.UID, "phase", phase)
+		} else {
+			logger.Info("Cleared stuck label from recovered launcher", "pod", pod.Name, "uid", pod.UID)
+		}
```
MikeSpreitzer, line 603: As far as I am concerned, it would be fine for there to be one log statement with `, "phase", phase, "stuck", stuck` in its keys-and-values.

MikeSpreitzer (approved):
This is good enough to merge.
There are some open comments with suggestions for improvements.

PR #681, opened 2026-07-27
Draft resource planning section of dual-pods doc

MikeSpreitzer:


docs/dual-pods.md
```diff
@@ -487,16 +487,8 @@ spec:
...
-deliberately leaving room because vLLM's control over its GPU memory
-usage is imprecise. To calculate the right utilization cap for your
-usage, measure the amount of GPU memory that remains in use while the
-GPU has one sleeping vLLM instance and no awake one. Pick a number of
-sleeping vLLM instances that you want to allow to coexist with one
-awake instance. Multiply by the amount of GPU memory that each
-sleeping instance takes, and convert to a fraction of the GPU's total
-memory. Subtract that from 0.95, and you will probably be good.
+For an explanation of `--gpu-memory-utilization 0.85`, see [GPU memory
+planning](#gpu-memory-planning) below.
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

docs/dual-pods.md
```diff
@@ -579,6 +571,67 @@ spec:
...
+uses a two-byte datatype; thus, allow about 2 GB for its persistent
+storage.
+
+#### Main memory planning
+
+For server-requesting Pods, there is no significant usage of
+main memory.
+
+For launcher Pods, multiple the `.spec.maxInstances` of the
+`LauncherConfig` by the main memory need for one vLLM instance.
```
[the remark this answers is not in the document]
MikeSpreitzer: fixed

docs/dual-pods.md
rubambiza, line 476: Should we re-align the model here to `Qwen/Qwen3-8B`, as discussed [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/579#discussion_r3605075814), so that the `dual-pods.md` stays in alignment with the rest of our nightlies/WLP?

The comment is also relevant to the exemplar explanation below: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/681/changes#diff-c504369fe286bf123d47180cec083cb7e4b0460cf65a797b2958d1b442b2ee84R593

docs/dual-pods.md
```diff
@@ -579,6 +573,67 @@ spec:
...
+
+#### Persistent storage planning
+
+For server-requesting Pods, there is no significant usage of
+persistent storage.
+
+For launcher Pods, look at the `.spec.maxInstances` of the
+`LauncherConfig` and multiply that by the persistent storage needed
+for one instance. Additionally, allow a few hundred MB for
+miscellaneous other stuff.
```
rubambiza: Miscellaneous stuff like? I think one example here would be enough to get a sense of the category of stuff to be expected, and perhaps you have observed said stuff.

docs/dual-pods.md
```diff
@@ -579,6 +573,67 @@ spec:
...
+
+The persistent storage needed for one model depends on the technique
+used to load model tensors. If that involves the Hugging Face model
```
rubambiza: I think the two "that" used in this sentence are a bit confusing.

- I can't tell if the first `that` refers to the persistent storage or the technique used.
- Is the second `that` referring to the cache? If so, then I'm further befuddled because I interpreted `the Hugging Face model cache` as the upstream repository.

Moreover, once I read "depends on" and saw "if x then y", I expected there to be another technique mentioned for "if not x"

I don't really know what to suggest in terms of edit because there are a lot of assumptions I am making in my read that could be wrong.

docs/dual-pods.md
```diff
@@ -579,6 +573,67 @@ spec:
...
+of the model tensors. This is the copy made by vLLM's level-1 sleep
+operation.
+
+Depending on the technique used by vLLM to load the model tensors
+there may be an additional copy in main memory. For example, when
+using a technique that reads the model tensors from a file, a copy of
+the model tensors may be left in the OS's page cache [TODO: is this
+really an issue for the vLLM container?].
+
+For launcher population policy, remember that the main memory used by
```
rubambiza: Does this imply there has to be coordination between different launcher population policies that match a particular node?

docs/dual-pods.md
rubambiza: IIRC, we're now targeting `Deployments` instead of `ReplicatSets`. So, this may need an update. I found myself beyond the changes of the commits because I needed to verify some stuff (in this case how a LauncherConfig is defined here) before posting comments.

docs/dual-pods.md
```diff
@@ -579,6 +573,67 @@ spec:
...
+
+#### GPU memory planning
+
+There is an important command-line parameter of vLLM:
+`--gpu-memory-utilization`; it takes a floating-point value. This is a
+cap on the fraction (yes, fraction, not percentage) of the GPU's
+memory that vLLM should plan to use (for model tensors, CUDA graphs
+and other context stuff, and kv cache). The default is 0.9,
+deliberately leaving room because vLLM's control over its GPU memory
+usage is imprecise. To calculate the right utilization cap for your
```
rubambiza, line 635: I am wondering if this could be a gamble. That is, the calculation assumes the model whose cap is being set is the same as the one benchmarked (using the term loosely here) to get the GPU memory remaining in use. I understand this as answering the question "how many more of my kind of model (i.e., maxInstances) can I co-exist with on this GPU"? But it becomes unclear when two different kinds of models could be asleep on that same GPU. Perhaps I am misunderstanding a fundamental assumption in how FMA restricts things.

rubambiza (changes requested):
Leaving two suggested changes and a few clarifying questions.

docs/dual-pods.md
MikeSpreitzer: The cited discussion in PR #579 raises two issues: (1) using a gated model is a pain and (2) "switching/standardizing around" one model.

I strongly agree with (1).

For (2), well, I can see the virtue of having all the benchmarking use the same model: it makes the results more comparable. But I think that making _all_ our examples use the same model is dangerous and an alert reader will pick up on the danger and become suspicious of our work: we may be missing something important if we do not test with various models. BTW, the E2E tests use three different models, and the one currently cited here is one of those.

docs/dual-pods.md
MikeSpreitzer: Significantly revised, hopefully it is clearer now.

docs/dual-pods.md
MikeSpreitzer: This is only an example, and I myself think that diversity in examples is good.

Changed to Deployment.

docs/dual-pods.md
MikeSpreitzer: Revised, hopefully it is clearer now.

docs/dual-pods.md
MikeSpreitzer: Revised for clarity.

docs/dual-pods.md
MikeSpreitzer: I added one example (the largest magnitude one, IIRC).

docs/dual-pods.md
rubambiza: Ah, fair enough. Agreed on the need to test with diverse models.

docs/dual-pods.md
rubambiza: Much clearer indeed. The more you iterated on this, the more I kept wondering if this part of the configuration could benefit from something like [llm-d's BLIS](https://llm-d.ai/blog/blis-evolving-llm-d-at-simulation-speed) (perhaps an extension of it that is specific to what FMA cares about?). Anyway, maybe a thought exercise that could be delegated to Claude. Note that I have not explored BLIS in depth.

rubambiza (approved):
All comments have been satisfactorily addressed. LGTM.

docs/dual-pods.md
MikeSpreitzer: I "unresolved" this because it is an interesting open point.

PR #687, opened 2026-07-29
Make DPC produce coverage data

MikeSpreitzer:
This takes a bite out of Issue #688 .

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -673,17 +673,19 @@ jobs:
           kubectl scale -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dpc-coverdata-inspector" --replicas=1
           kubectl wait -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dpc-coverdata-inspector" --for condition=Available
           insp=$(kubectl get -n "$FMA_NAMESPACE" pods -l app.kubernetes.io/component=dpc-coverdata-inspector -o name | sed s:^pod/::)
-          mkdir dpc-coverdata
+          mkdir -p dpc-cover/data
           kubectl wait -n "$FMA_NAMESPACE" pod "${insp}" --for condition=Ready
-          kubectl cp -n "$FMA_NAMESPACE" "${insp}:var/coverdata" dpc-coverdata
+          kubectl cp -n "$FMA_NAMESPACE" "${insp}:var/coverdata" dpc-cover/data
           kubectl scale -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dual-pods-controller" --replicas=1
+          go tool covdata textfmt -i=dpc-cover/data -o dpc-cover/profile.txt
+          go tool cover -func=dpc-cover/profile.txt -o doc-cover/func.txt
 
       - name: upload coverage data from dual pods controller
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: dpc-coverdata
-          path: dpc-coverdata
+          name: dpc-cover
+          path: dpc-cover
           retention-days: 14
 
       - name: Clean up test objects
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -673,7 +673,12 @@ jobs:
         if: always()
         run: |
           kubectl scale -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dual-pods-controller" --replicas=0
-          kubectl wait -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dual-pods-controller" --for condition=Available
+          let try=1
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -665,6 +665,24 @@ jobs:
...
         run: cat scrape2.metrics || true
 
+      - name: extract coverage data from dual pods controller
+        if: always()
+        run: |
+          kubectl scale -n "$FMA_NAMESPACE" deploy/fma-dual-pods-controller --replicas=0
+          kubectl scale -n "$FMA_NAMESPACE" deploy/fma-dpc-coverdata-inspector --replicas=1
+          insp=$(kubectl get -n "$FMA_NAMESPACE" pods -l app.kubernetes.io/component=dpc-coverdata-inspector -o name | sed s:^pod/::)
+          mkdir dpc-coverdata
+          kubectl cp -n "$FMA_NAMESPACE" "$(insp):var/coverdata" dpc-coverdata
```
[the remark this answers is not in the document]
MikeSpreitzer: Fixed expansion.
kubectl no longer likes a leading slash on the path inside a container.

.github/workflows/launcher-based-e2e-test.yml
```diff
@@ -161,6 +161,23 @@ jobs:
...
         run: cat scrape2.metrics || true
 
+      - name: extract coverage data from dual pods controller
+        if: always()
+        run: |
+          kubectl scale deploy/fma-dual-pods-controller --replicas=0
+          kubectl scale deploy/fma-dpc-coverdata-inspector --replicas=1
+          insp=$(kubectl get pods -l app.kubernetes.io/component=dpc-coverdata-inspector -o name | sed s:^pod/::)
+          mkdir dpc-coverdata
+          kubectl cp "$(insp):var/coverdata" dpc-coverdata
```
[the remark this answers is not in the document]
MikeSpreitzer: Fixed expansion.
kubectl no longer likes a leading slash on the path inside a container.

charts/fma-controllers/templates/dual-pods-controller/deployment.yaml
```diff
@@ -32,9 +44,56 @@ spec:
...
+        persistentVolumeClaim:
+          claimName: "{{ .Release.Name }}-coverdata-dual-pods-controller"
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    {{- include "fma-controllers.labels" . | nindent 4 }}
+    app.kubernetes.io/component: dual-pods-controller
+  name: "{{ .Release.Name }}-dpc-coverdata-inspector"
```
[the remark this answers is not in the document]
MikeSpreitzer: Not broken.
But revised.

MikeSpreitzer:
Here is a legacy function coverage report generated from the data reported on the E2E/OpenShift test https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/30427223501/job/90497287265?pr=687

[func.txt](https://github.com/user-attachments/files/30494376/func.txt)

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -665,6 +665,43 @@ jobs:
...
 
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version: '1.26'
+
+      - name: extract coverage data from dual pods controller
+        if: always()
+        run: |
+          kubectl scale -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dual-pods-controller" --replicas=0
+          kubectl wait -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dual-pods-controller" --for condition=Available
```
waltforme: Just a clarification question. How is 'Available' condition defined for a Deployment that is scaled to 0?

I'm asking because I'm concerned about two possible outcomes. One is this condition never meets, the other is this condition immediately meets.

.github/workflows/launcher-based-e2e-test.yml
```diff
@@ -161,6 +161,38 @@ jobs:
         if: always()
         run: cat scrape2.metrics || true
 
+      - name: extract coverage data from dual pods controller
+        if: always()
+        run: |
+          kubectl scale deploy/fma-dual-pods-controller --replicas=0
+          kubectl wait deploy/fma-dual-pods-controller --for condition=Available
```
waltforme: The same question as https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/687/changes#r3677938750

.github/workflows/ci-e2e-openshift.yaml
MikeSpreitzer: Looking at the definition of the condition (https://github.com/kubernetes/api/blob/v0.34.10/apps/v1/types.go#L537-L539): it immediately meets.

I have revised to use a proper wait for zero Pods from that Deployment to exist. With limited patience, a generous amount in my experience, but limited so that this wait will not hold up the job step indefinitely.

BTW, it seems to me that the Kubernetes scheduler should be responsible for enforcing the restriction of one-Pod-at-a-time access.

I put in the explicit wait because (a) Copilot nagged me about it and (b) I observed Kubernetes NOT enforcing the restriction of one-Pod-at-a-time access. This was in a `kind` cluster. BTW, the volume appeared to have successfully supported two accessing Pods at once. I do not have strong experimental evidence from OpenShift.

.github/workflows/launcher-based-e2e-test.yml
MikeSpreitzer: same answer

waltforme (approved):

PR #694, opened 2026-08-03
Switch order of dual-pods and launcher in docs/README.md

MikeSpreitzer:
The dual-pods doc is about the whole technique, so should come first.


diegocastanibm (approved):
LGTM

PR #696, opened 2026-08-03
dual-pods: make nodeData.Add pull forward delayed queue items

### LINKED ISSUE #667: [Bug]: nodeData.Add is too deferential to existing queue entries
### Contact Details

_No response_

### What happened?

https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/e7c0f7e97c1df0b77229766d69e9dadb90eaec13/pkg/controller/dual-pods/controller.go#L1041 accepts any existing queue entry as "good enough". This is wrong. `add(item)` should behave like `addAfter(item, now)`.

### Version

### Branch name

_No response_

### Commit SHA

e7c0f7e97c1df0b77229766d69e9dadb90eaec13

### Relevant log output

```shell

```

Iceber:
fix: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/667

|Scenario|Old behavior | New behavior|
|---|---|---|
|Item not in queue|Enqueue, processAfter=now| unchange|
|Exists, processAfter > now | keep future time(**bug**)| **Pull to now**|
|Exists, processAfter <= now| unchanged | unchanged|

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -1034,18 +1034,12 @@ func (ctl *controller) getNodeData(nodeName string) *nodeData {
 	return ans
 }
 
+// add enqueues item for immediate processing.
+// For an already-queued item whose processAfter is set to a future time
+// (e.g., due to a prior rate-limited retry via addAfter), a fresh notification
+// means the item should be reconciled now, not at that previously scheduled time.
 func (nodeDat *nodeData) add(item itemOnNode) {
-	nodeDat.ItemsMutex.Lock()
-	defer nodeDat.ItemsMutex.Unlock()
-	if _, exist := nodeDat.LocalQueue[item]; exist {
-		return // already queued
-	}
-	now := time.Now()
-	nodeDat.LocalQueue[item] = &scheduledItem{
-		addTime: now, processAfter: now,
-	}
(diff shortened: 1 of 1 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -1034,18 +1034,12 @@ func (ctl *controller) getNodeData(nodeName string) *nodeData {
 	return ans
 }
 
+// add enqueues item for immediate processing.
+// Pull forward an already-queued item whose processAfter was pushed into the
```
MikeSpreitzer: If this change of time is a "pull" then the reader will suppose that "was pushed" refers to an earlier **change** in scheduled time. But that is not what you meant. I think that it would be better to say something like "was in the future".

MikeSpreitzer (dismissed):
LGTM.
I left one suggestion for improvement, which could be made in this PR or a follow-on.

pkg/controller/dual-pods/controller.go
Iceber: Hi @MikeSpreitzer, The comment has been updated.
PTAL, Thanks!

MikeSpreitzer (commented):
LGTM

MikeSpreitzer (approved):
LGTM

PR #698, opened 2026-08-03
Note hands-off case in LPP meaning

MikeSpreitzer:
Update the API documentation to include the hands-off case: no launchers created or deleted while the LauncherConfig is missing or malformed.

This is not an API change, it is fixing an oversight in the prose describing the API.

### DIFF
--- a/api/fma/v1alpha1/launcherpopulationpolicy_types.go
+++ b/api/fma/v1alpha1/launcherpopulationpolicy_types.go
@@ -26,18 +26,21 @@ import (
 // for a given type of Node.
 // Here we introduce and use a particular definition of "type" for Nodes.
 // All the LauncherPopulationPolicy objects together define a map,
-// from (Node, LauncherConfig) to count.
+// from (Node, LauncherConfig name) to count.
 // Call this map `PopulationPolicy`.
-// When multiple CountForLauncher apply to the same (Node, LauncherConfig) pair
+// When multiple CountForLauncher apply to the same (Node, LauncherConfig name) pair
 // the maximum of their counts is what appears in `PopulationPolicy`.
-// When no CountForLauncher applies to a given (Node, LauncherConfig),
+// When no CountForLauncher applies to a given (Node, LauncherConfig name),
 // `PopulationPolicy` implicitly maps that pair to zero.
 //
 // The collective meaning of all the LauncherPopulationPolicy objects
-// and all the server-requesting Pods is that for a given (Node, LauncherConfig)
+// and all the server-requesting Pods is that for a given (Node, LauncherConfig name)
 // the number of launchers that should exist is the larger of
 // (a) what `PopulationPolicy` says for that pair, and
 // (b) the number needed to satisfy the server-requesting Pods.
+// While there is no LauncherConfig with the given name or that LauncherConfig
+// exists and has a malformed Pod template,
+// the controllers will not create or delete launchers.
 //
 // +genclient
 // +kubebuilder:object:root=true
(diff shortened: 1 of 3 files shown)


api/fma/v1alpha1/launcherpopulationpolicy_types.go
```diff
@@ -26,18 +26,20 @@ import (
...
 // `PopulationPolicy` implicitly maps that pair to zero.
 //
 // The collective meaning of all the LauncherPopulationPolicy objects
-// and all the server-requesting Pods is that for a given (Node, LauncherConfig)
+// and all the server-requesting Pods is that for a given (Node, LauncherConfig name)
 // the number of launchers that should exist is the larger of
 // (a) what `PopulationPolicy` says for that pair, and
 // (b) the number needed to satisfy the server-requesting Pods.
+// While there is no LauncherConfig with the given name or that config has a
+// malformed Pod template the controllers will not create or delete launchers.
```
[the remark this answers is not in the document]
MikeSpreitzer: revised.

docs/dual-pods.md
```diff
@@ -158,20 +158,24 @@ objects call for a proactive population of launchers, as follows.
...
-  LauncherConfig) the number of launchers that should exist is the
-  larger of
+  and all the server-requesting Pods is that for a given (`Node`,
+  `LauncherConfig` name) the number of launchers that should exist
+  is the larger of
     (a) what `PopulationPolicy` says for that pair, and
     (b) the number needed to satisfy the existing server-requesting Pods.
+  However: while there is no `LauncherConfig` with the given name or
+  that config has a malformed Pod template the FMA controllers will
+  not create or delete launchers.
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

waltforme (approved):

api/fma/v1alpha1/launcherpopulationpolicy_types.go
```diff
@@ -26,18 +26,21 @@ import (
...
 // `PopulationPolicy` implicitly maps that pair to zero.
 //
 // The collective meaning of all the LauncherPopulationPolicy objects
-// and all the server-requesting Pods is that for a given (Node, LauncherConfig)
+// and all the server-requesting Pods is that for a given (Node, LauncherConfig name)
 // the number of launchers that should exist is the larger of
 // (a) what `PopulationPolicy` says for that pair, and
 // (b) the number needed to satisfy the server-requesting Pods.
+// While there is no LauncherConfig with the given name or that LauncherConfig
+// exists and has a malformed Pod template,
```
rubambiza, line 42: How do we define/detect a malformed Pod template here? In other words, which component is responsible for doing so? Might be worth a word or two to clarify that condition.

rubambiza: Same comment for https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/698/changes#diff-c504369fe286bf123d47180cec083cb7e4b0460cf65a797b2958d1b442b2ee84R177

rubambiza (commented):
Leaving a couple of related comments. Should be good to merge once they are addressed.

api/fma/v1alpha1/launcherpopulationpolicy_types.go
MikeSpreitzer: > which component is responsible for doing so?

I am a little surprised by the question. Isn't the answer obviously that the primary component responsible is the controller that proactively creates and deletes launchers, integrating clues from elsewhere as it sees fit?

rubambiza: I think when the *primary* qualifier is added and we are solely focus on conditions for launcher creation and deletions, then the answer is very obvious. I think my confusion is much earlier in the process. That is, how a `LauncherConfig` with a malformed Pod template came to exist in the first place. In typical scenarios working with `kubectl`, it will give me warnings or straight up reject my object creation request if it is malformed.

Issue #707, opened 2026-08-07
[Bug]: E2E test case flakes wrt coverage data PVC


MikeSpreitzer:
### What happened?

The E2E test on OpenShift sometimes fails in the "Controller Restart State Recovery" test case due to competition over the one-pod-at-a-time PVC for coverage data. See https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/31146729312/attempts/2?pr=703 for a report of a failure. I am not sure why this case ever succeeds on OpenShift. (On `kind` the exclusivity does not seem to be enforced.)

### Version

non-main branch (please specify below)

### Branch name

intro-installer

### Commit SHA

67c57e61ce48b66fd8aa758cab90c63fdb1fc3b7

### Relevant log output

```shell

```

MikeSpreitzer:
This got fixed as part of PR #645 (so that could pass CI).

PR #708, opened 2026-08-10
dual-pods: return ordered entries from takeReadyItems

### LINKED ISSUE #668: localQueue in dual-pods controller is not a *queue*
Currently

```
type localQueue map[itemOnNode]*scheduledItem
```

This does not preserve ordering between items.

I myself specifically suggested using a `map` rather than a slice during development of this, but I now think that might not have been the best idea.

OTOH, a Kubernetes informer does not preserve ordering between events on different objects, and `nodeItem.Add` is called only from informer event handlers --- so there is no important semantic consequence here.

Iceber:
fix: https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/668

https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/668#issuecomment-5164148211

### DIFF
--- a/pkg/controller/dual-pods/controller.go
+++ b/pkg/controller/dual-pods/controller.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -417,12 +418,10 @@ type nodeData struct {
 	// LocalQueue holds the set of object references of this node that need to be synced,
 	// and for each the time at which it was first injected into this set.
 	// Hold ItemsMutex while accessing this.
-	LocalQueue localQueue
+	LocalQueue map[itemOnNode]*scheduledItem
 
 	rateLimiter workqueue.TypedRateLimiter[itemOnNode]
 }
-
-type localQueue map[itemOnNode]*scheduledItem
 
 type itemOnNode interface {
 	// process returns (err error, retry bool).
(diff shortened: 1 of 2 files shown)


pkg/controller/dual-pods/controller.go
```diff
@@ -1043,32 +1048,36 @@ func (nodeDat *nodeData) add(item itemOnNode) {
...
+		if !si.processAfter.After(now) {
+			ready = append(ready, scheduledEntry{item: item, scheduled: si})
+			delete(nodeDat.localQueue, item)
 		}
 	}
-	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(float64(len(nodeDat.LocalQueue)))
+	queueDepthGauges.WithLabelValues(nodeDat.NodeName).Set(float64(len(nodeDat.localQueue)))
+	sort.Slice(ready, func(i, j int) bool {
+		return ready[i].scheduled.addTime.Before(ready[j].scheduled.addTime)
+	})
```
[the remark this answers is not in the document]
MikeSpreitzer: This has been fixed.

pkg/controller/dual-pods/controller.go
```diff
@@ -414,16 +415,14 @@ type nodeData struct {
 	// ItemsMutex may be acquired while holding controller mutex, not vice-versa.
 	ItemsMutex sync.Mutex
 
-	// LocalQueue holds the set of object references of this node that need to be synced,
+	// localQueue holds the set of object references of this node that need to be synced,
```
MikeSpreitzer: This was deliberately capitalized, like all the other fields, so that debug logging would include it. Type `nodeData` is not public.

MikeSpreitzer (commented):
I left one comment.

pkg/controller/dual-pods/controller.go
Iceber: @MikeSpreitzer updated

MikeSpreitzer:
@Iceber: this failed due to a problem (#707) that has been fixed in the `main` branch. You will need to rebase this PR onto `main`.

MikeSpreitzer (approved):
LGTM

PR #710, opened 2026-08-11
Update vllm to version 0.26.0 to match llm-d

aavarghese:
As per title


rubambiza (approved):
LGTM, but the OpenShift CI is currently failing, as highlighted by Mike on Slack.

PR #711, opened 2026-08-12
Always setup-go in ci-e2e-openshift.yaml

MikeSpreitzer:
Because the go tooling to extract coverage data is always used.

This fixes an oversight in the original coding. This fix prevents the job step that uses the go tooling from failing.

This fix only removes an annoyance; in cases where this makes a difference, there is no coverage data to extract.


MikeSpreitzer:
On second thought, it is good that no empty coverage data is posted. The proper fix would be to do that without technically failing.

MikeSpreitzer:
See #716 instead

PR #719, opened 2026-08-14
Add the requester's TCP proxy, without using it yet

delavet:
Introduce `pkg/server/requester/proxy`: a TCP reverse proxy that blocks until a backend target is delivered through its /v1/proxy/config HTTP handler, then forwards every accepted connection to that target. Add the SPI path and request/response type that handler speaks, plus the `github.com/inetaf/tcpproxy dependency` it is built on.

Nothing wires this into any binary, so no behavior changes. This is the first half of #498, split so that it can merge before parallel development accumulates more conflicts against it. 

Relative to that PR, it also applies the review comments still open against the proxy code, except as noted below:

  - `AddFlags` is renamed `AddToFlagSet`, matching Options.AddToFlagSet in `pkg/observability/prom-and-debug.go`.
  - The PUT body is decoded straight off `r.Body` rather than through `io.ReadAll`, so an oversized body cannot be turned into an OOM.
  - The needless defer `r.Body.Close()` is gone; `net/http` closes it.
  - "proxy shutdown cancelled before configuration" described the wrong end of the lifecycle, and "most recently" described a value that is delivered at most once. Both reworded.
  - `TestConfigure_DifferentConfigReturnsConflict` is folded into `TestConfigure_GetStatus`, together now `TestConfigure_Lifecycle`.
  - The "no trailing byte" assertion waits `testQuietPeriod` (2s) instead of 100ms, which is thin on a loaded CI machine.

### DIFF
--- a/pkg/server/requester/proxy/server.go
+++ b/pkg/server/requester/proxy/server.go
@@ -1,0 +1,217 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proxy
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/inetaf/tcpproxy"
+	"github.com/spf13/pflag"
+
+	"k8s.io/klog/v2"
+
+	stubapi "github.com/llm-d-incubation/llm-d-fast-model-actuation/pkg/spi"
+)
+
+// ProxyConfig holds the proxy server's runtime parameters that are provided
+// at requester startup, via the requester command line. It is distinct from
+// stubapi.ProxyTargetConfig, which holds the runtime parameters provided
+// later by the dual-pods controller (the backend the proxy forwards to).
+type ProxyConfig struct {
+	// Port is the port the proxy listens on.
+	Port uint16
+	// DialTimeout is the timeout for dialing backend connections.
+	DialTimeout time.Duration
+}
+
+// DefaultProxyConfig provides sensible defaults.
+var DefaultProxyConfig = ProxyConfig{
+	Port:        8082,
+	DialTimeout: 10 * time.Second,
+}
+
+// AddToFlagSet registers command-line flags for all proxy configuration fields.
+func (cfg *ProxyConfig) AddToFlagSet(fs *pflag.FlagSet) {
+	fs.Uint16Var(&cfg.Port, "proxy-port", cfg.Port, "port for TCP proxy")
+	fs.DurationVar(&cfg.DialTimeout, "proxy-dial-timeout", cfg.DialTimeout, "timeout for proxy backend dial")
+}
+
+// Server is a TCP reverse proxy that waits for a backend target to be
+// delivered via Configure, then forwards all connections to that target.
+type Server struct {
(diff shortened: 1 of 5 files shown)


pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,217 @@
...
+	logger := klog.FromContext(ctx).WithName("proxy-server")
+	logger.Info("Starting TCP proxy server", "port", s.cfg.Port, "dialTimeout", s.cfg.DialTimeout)
+
+	var tgt stubapi.ProxyTargetConfig
+	select {
+	case tgt = <-s.configCh:
+		logger.V(2).Info("Received proxy target from Configure", "target", tgt)
+	case <-ctx.Done():
+		logger.V(2).Info("Context cancelled before proxy was configured")
+		s.startedCh <- fmt.Errorf("proxy shut down before configuration")
```
MikeSpreitzer, line 100: This is not wrong, but nit-pickers will want to see `errors.New` instead of an `fmt.Errorf` call with no actual formatting work.

pkg/server/requester/proxy/server.go
```diff
@@ -0,0 +1,217 @@
...
+		return waitErr
+	}
+	logger.Info("TCP proxy server stopped")
+	return nil
+}
+
+// Configure handles the proxy configuration HTTP endpoint at ProxyConfigPath.
+//
+// GET returns the configured target (200) or 404 if not yet configured.
+// PUT delivers the backend target. PUT may succeed only once; subsequent PUTs
```
MikeSpreitzer, line 145: This is a problem for clients. Because a TCP connection can break at any time, it is possible that a client sends a PUT request that is successfully handled but the TCP connection used for the PUT breaks before the client gets the response message. The client can not tell whether the server handled the request or not. A _redundant_ PUT should be acceptable.

Such breakage may be unlikely inside a Kubernetes cluster, since the client and server are not separated by a lot. But even so, there are no guarantees.

Can be improved in a follow-on PR.

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,270 @@
...
+			conn, err := ln.Accept()
+			if err != nil {
+				return
+			}
+			connsMu.Lock()
+			conns = append(conns, conn)
+			connsMu.Unlock()
+			go func(c net.Conn) {
+				_, _ = io.Copy(c, c)
+				_ = c.Close()
```
MikeSpreitzer, line 92: FYI, this could use `conn` instead of `c` here, because `conn` is declared in the loop body (so every iteration makes a fresh variable).

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	_, backendPort, backendCloser := startTestEchoServer(t)
+	defer backendCloser()
+
+	proxyPort := findFreePort(t)
+	srv := New(ProxyConfig{Port: proxyPort, DialTimeout: testDialTimeout})
+	stop := startProxy(t, srv)
+	defer stop()
+
+	body := stubapi.ProxyTargetConfig{Address: "127.0.0.1", Port: backendPort}
```
MikeSpreitzer, line 130: Why not use the address from `startTestEchoServer` ?
Maybe make `startTestEchoServer` return a `ProxyTargetConfig` ?

pkg/server/requester/proxy/server_test.go
```diff
@@ -0,0 +1,270 @@
...
+		t.Fatalf("failed to parse response JSON: %v", err)
+	}
+	if got != body {
+		t.Errorf("GET returned %+v, want %+v", got, body)
+	}
+
+	// Reconfigure with a DIFFERENT config — should be rejected, so the 409
+	// actually proves "reconfigure is blocked" rather than "duplicate write
+	// is a no-op".
+	differentBody := stubapi.ProxyTargetConfig{Address: "10.0.0.1", Port: backendPort + 1}
```
MikeSpreitzer, line 232: We are not promised that `backendPort + 1 != 0`.

MikeSpreitzer (approved):
This is good enough to merge.
I have left some independent comments for your consideration.

PR #721, opened 2026-08-18
Use the requester's TCP proxy

delavet:
Second half of #498, on top of the proxy that #719 added.

Wires the proxy into both requester binaries and has the dual-pods controller point it at the bound provider. Requester ports move from the `PROBES_PORT`/`SPI_PORT` env vars to flags, gathered in a new `pkg/server/requester/config`. The example manifest probes readiness through the proxy port, so a requester reports ready only once traffic can reach a serving inference server.

Applies the review comments still open against #498 and #719:

- A PUT of the target already in effect returns 200 rather than 409: a connection that breaks after the server handled the PUT leaves the client unable to tell whether it took effect. A PUT naming a *different* target still conflicts.
- The listener's fate now travels as a `struct{}` channel closed after a `startErr` field is written. Carrying the error in the channel made it readable exactly once, so a second PUT would have read a zero value and called a proxy that never started a success.
- `errors.New` for the `fmt.Errorf` with nothing to format; `startTestEchoServer` returns the `ProxyTargetConfig` both callers wanted; the conflicting-PUT target no longer uses `backendPort + 1`, which wraps to 0 at port 65535 and is rejected as invalid before the conflict can be seen; the echo goroutine closes over `conn` directly; the e2e case states its `$isc` assumption and uses `&> /dev/null`.
- The controller calls the `doHTTP` that main has since grown rather than the `doPut`/`ensureProxyConfigured` helpers #498 carried

**Making its use optional** — the second commit, following the review note about going further and making the proxy available without requiring it.

A server-requesting Pod asks for the proxy by naming its port in a `dual-pods.llm-d.ai/proxy-port` annotation, and the controller configures the proxy only for a Pod that does. There is no default: a Pod that names no port is not asking for one, so the controller leaves its proxy alone and clients reach the inference server the way they did before this proxy existed.

### DIFF
--- a/cmd/requester/README.md
+++ b/cmd/requester/README.md
@@ -8,13 +8,13 @@ The requester executable runs in the requester container of a server-requesting
 Pod. Its purpose is two-fold:
 
 1. **Look to the rest of llm-d like an inference server.** The requester listens
-   on the same set of ports an inference server would and forwards client traffic
-   to the bound vLLM instance via a built-in TCP reverse proxy.
+   on the same set of ports an inference server would and can forward client
+   traffic to the bound vLLM instance via a built-in TCP reverse proxy.
 2. **Relay controller-bound state to and from the dual-pods controller.** The
    requester exposes a small SPI surface so that the dual-pods controller can
    read GPU IDs, push the bound server's address, and update readiness state.
 
-Adding the reverse TCP proxy makes the requester look *more* like an inference
+The reverse TCP proxy can make the requester look *more* like an inference
 server than it otherwise would: client requests can be sent directly to the
 requester Pod's IP and the proxy forwards them to the actual server-providing
 Pod after binding.
@@ -29,10 +29,18 @@ The server-requesting Pod runs three servers:
 
 ### TCP Proxy
 
-The requester includes a **TCP proxy** that forwards traffic to the bound inference
-server. After the dual-pods controller binds the requester to a server-providing Pod,
-it configures the proxy target via a `PUT` to the SPI endpoint at `/v1/proxy/config`.
-You can query the current proxy configuration with `GET /v1/proxy/config`.
+The requester carries a **TCP proxy** that can forward traffic to the bound
+inference server, and using it is optional.
+
+A server-requesting Pod asks for it by naming the port in the
+`dual-pods.llm-d.ai/proxy-port` annotation. The dual-pods controller then PUTs
+the bound server's address to the SPI endpoint at `/v1/proxy/config`, and the
+proxy starts accepting connections on that port. You can query the current
+configuration with `GET /v1/proxy/config`.
+
+Without that annotation the controller never configures the proxy, so it never
+opens a listener, and clients reach the inference server the way they did before
+the proxy existed. Nothing else about the requester changes.
 
 ## Command-line flags
 
@@ -118,6 +126,9 @@ spec:
         app: dp-example
       annotations:
         dual-pods.llm-d.ai/admin-port: "8081"
+        # Asks the controller to point this Pod's proxy at the bound server.
+        # Omit it and the proxy stays idle.
+        dual-pods.llm-d.ai/proxy-port: "8082"
         dual-pods.llm-d.ai/server-patch: |
           metadata:
             labels: {
@@ -163,8 +174,8 @@ spec:
             containerPort: 8082
           readinessProbe:
             httpGet:
-              path: /health
-              port: 8082
+              path: /ready
+              port: 8080
             initialDelaySeconds: 2
             periodSeconds: 5
           resources:
@@ -194,6 +205,9 @@ spec:
         app: dp-example
       annotations:
         dual-pods.llm-d.ai/admin-port: "8081"
+        # Asks the controller to point this Pod's proxy at the bound server.
+        # Omit it and the proxy stays idle.
+        dual-pods.llm-d.ai/proxy-port: "8082"
         dual-pods.llm-d.ai/server-patch: |
           metadata:
             labels: {
@@ -249,8 +263,8 @@ spec:
             containerPort: 8082
           readinessProbe:
             httpGet:
-              path: /health
-              port: 8082
+              path: /ready
+              port: 8080
             initialDelaySeconds: 2
             periodSeconds: 5
           resources:
--- a/cmd/requester/README.md
+++ b/cmd/requester/README.md
@@ -2,6 +2,51 @@ This document shows the steps to exercise the requester and dual-pods controller
 in a local k8s environment with model `ibm-granite/granite-3.3-2b-instruct`
 cached on local PV in the cluster.
 
+## Requester Overview
+
+The requester executable runs in the requester container of a server-requesting
+Pod. Its purpose is two-fold:
+
+1. **Look to the rest of llm-d like an inference server.** The requester listens
+   on the same set of ports an inference server would and forwards client traffic
+   to the bound vLLM instance via a built-in TCP reverse proxy.
+2. **Relay controller-bound state to and from the dual-pods controller.** The
+   requester exposes a small SPI surface so that the dual-pods controller can
+   read GPU IDs, push the bound server's address, and update readiness state.
+
+Adding the reverse TCP proxy makes the requester look *more* like an inference
+server than it otherwise would: client requests can be sent directly to the
+requester Pod's IP and the proxy forwards them to the actual server-providing
+Pod after binding.
+
+The server-requesting Pod runs three servers:
+
+| Server | Default Port | Purpose |
+|--------|-------------|---------|
+| Probes | 8080 | Readiness relay (`/ready` endpoint) — reflects the health status of the bound inference server |
+| SPI | 8081 | Dual-pods controller interface (`/v1/dual-pods/accelerators`, `/v1/proxy/config`) — exposes GPU info and accepts proxy configuration from the controller |
+| Proxy | 8082 | TCP reverse proxy listen port — accepts client connections and forwards them to the bound vLLM instance; has no listener at all until configured |
+
+### TCP Proxy
+
+The requester includes a **TCP proxy** that forwards traffic to the bound inference
+server. After the dual-pods controller binds the requester to a server-providing Pod,
+it configures the proxy target via a `PUT` to the SPI endpoint at `/v1/proxy/config`.
+You can query the current proxy configuration with `GET /v1/proxy/config`.
+
+## Command-line flags
+
+Besides the klog flags, which `--help` lists:
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--probes-port` | `8080` | Port number for readiness/liveness probes |
+| `--spi-port` | `8081` | Port for dual-pods SPI requests |
+| `--proxy-port` | `8082` | TCP proxy listen port for forwarding client connections |
+| `--proxy-dial-timeout` | `10s` | Timeout for dialing the backend inference server |
+
+## Local Testing
+
 Build and push the requester container image (use your favorate
 `CONTAINER_IMG_REG`) with a command like the following. You can omit
 the `TARGETARCH` if the runtime ISA matches your build time ISA.
@@ -114,10 +159,12 @@ spec:
             containerPort: 8080
           - name: spi
             containerPort: 8081
+          - name: proxy
+            containerPort: 8082
           readinessProbe:
             httpGet:
-              path: /ready
-              port: 8080
+              path: /health
+              port: 8082
             initialDelaySeconds: 2
             periodSeconds: 5
           resources:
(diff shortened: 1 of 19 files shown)


cmd/requester/README.md
```diff
@@ -2,6 +2,59 @@ This document shows the steps to exercise the requester and dual-pods controller
 in a local k8s environment with model `ibm-granite/granite-3.3-2b-instruct`
 cached on local PV in the cluster.
 
+## Requester Overview
```
MikeSpreitzer, line 5: This PR expands the scope of this document beyond what is stated in the document's first paragraph, which purports to describe the document overall. Such an expansion requires revising the introductory paragraph.

cmd/requester/README.md
```diff
@@ -2,6 +2,59 @@ This document shows the steps to exercise the requester and dual-pods controller
...
+server than it otherwise would: client requests can be sent directly to the
+requester Pod's IP and the proxy forwards them to the actual server-providing
+Pod after binding.
+
+The server-requesting Pod runs three servers:
+
+| Server | Default Port | Purpose |
+|--------|-------------|---------|
+| Probes | 8080 | Readiness relay (`/ready` endpoint) — reflects the health status of the bound inference server |
+| SPI | 8081 | Dual-pods controller interface (`/v1/dual-pods/accelerators`, `/v1/proxy/config`) — exposes GPU info and accepts proxy configuration from the controller |
```
MikeSpreitzer: The readiness relay belongs in here too.

config/examples/requester.yaml
MikeSpreitzer, line 1: This file is not maintained. It was broken before this PR. This PR should not touch it. A separate PR should remove it.

config/examples/requester.yaml
```diff
@@ -2,18 +2,24 @@ apiVersion: v1
...
+        - --proxy-port=8082
       ports:
         - containerPort: 8080
         - containerPort: 8081
-      env:
-        - name: PROBES_PORT
-          value: "8080"
-        - name: SPI_PORT
-          value: "8081"
+        - containerPort: 8082
```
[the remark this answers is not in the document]
MikeSpreitzer: Huh? This manifest probes readiness on port 8080, which is NOT the relay port.

This file was junk to begin with and is not maintained. No point in arguing over it.

pkg/server/requester/coordination/server.go
```diff
@@ -211,14 +211,17 @@ func newSetLogHandler(logger klog.Logger, logWriter io.Writer) func(w http.Respo
...
 	logger := klog.FromContext(ctx).WithName("spi-server")
 	mux := http.NewServeMux()
 	mux.HandleFunc(strings.Join([]string{"GET", stubapi.AcceleratorQueryPath}, " "), gpuHandler(gpuUUIDs))
 	mux.HandleFunc(strings.Join([]string{"GET", stubapi.AcceleratorMemoryQueryPath}, " "), gpuMemoryHandler(logger))
 	mux.HandleFunc("POST "+stubapi.BecomeReadyPath, newSetReadyHandler(logger, ready, true))
 	mux.HandleFunc("POST "+stubapi.BecomeUnreadyPath, newSetReadyHandler(logger, ready, false))
 	mux.HandleFunc("POST "+stubapi.SetLogPath, newSetLogHandler(logger, logWriter))
+	// Registered without a method, unlike the handlers above: this one serves
+	// both GET and PUT and rejects the rest itself.
+	mux.HandleFunc(stubapi.ProxyConfigPath, proxyConfigHandler)
```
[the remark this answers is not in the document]
MikeSpreitzer: That issue was strongly present already, and the answer has been that we are not concerned with attackers inside the cluster. To be addressed, if at all, separately.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1504,6 +1516,30 @@ func (ctl *controller) wakeUp(ctx context.Context, serverDat *serverData, reques
...
+	if adminPort == "" {
+		adminPort = api.AdminPortDefaultValue
+	}
+	target := stubapi.ProxyTargetConfig{Address: providingPod.Status.PodIP, Port: uint16(serverPort)}
+	proxyURL := fmt.Sprintf("http://%s:%s%s", requestingPod.Status.PodIP, adminPort, stubapi.ProxyConfigPath)
+	_, err := doHTTP(ctx, "configure_proxy", "PUT", proxyURL, ctl.httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"isc_name": requestingPod.Annotations[api.InferenceServerConfigAnnotationName]}), target, nil)
+	if err != nil {
+		return fmt.Errorf("failed to configure proxy of %s to reach %s: %w", requestingPod.Name, target.String(), err)
+	}
+	klog.FromContext(ctx).V(2).Info("Configured requester proxy", "requester", requestingPod.Name, "proxyPort", proxyPort, "target", target.String())
```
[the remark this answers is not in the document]
MikeSpreitzer: Fixing this statement will not meet the criterion of logging in all cases, due to the early out before it. The root cause here is that `doHTTP` is intended to satisfy DR-20 but does not use the right log level. That is what should be fixed. It is independent of this PR and should be fixed separately.

test/e2e/test-cases.sh
```diff
@@ -563,6 +563,49 @@ cheer Successful switching instances in one launcher
...
+intro_case Reverse Proxy Initialization and Forwarding
+
+# This test verifies that the dual-pods controller points the requester Pod's
+# TCP proxy at the bound launcher Pod, and that traffic sent to the proxy port
+# arrives at the inference server.
+#
+# Assumed starting state: $req4 is bound to $launcher1, both Ready. Also
+# assumed: the InferenceServerConfig of $req4 is $isc, which is where the
+# inference server's port is read from below. That holds because the
+# ReplicaSet was patched back to $isc before $req4 was created.
```
MikeSpreitzer: I would omit the "This holds because" statement because it is an unnecessary detail that may change.

test/e2e/test-cases.sh
```diff
@@ -563,6 +563,49 @@ cheer Successful switching instances in one launcher
...
+
+kill $pfpid || true
+pfpid=""
+
+note "Checking that the proxy port forwards to the inference server; expect port-forward and curl noise"
+kubectl port-forward -n "$NS" pod/"$req4" 28092:8082 &
+pfpid=$!
+sleep 5
+
+expect "curl -sf http://localhost:28092/health &> /dev/null"
```
MikeSpreitzer: This is not a very close inspection of expected inference server behavior. Can be improved in follow-on.

test/e2e/mkobjs-openshift.sh
```diff
@@ -247,6 +247,7 @@ spec:
         instance: "$inst"
       annotations:
         dual-pods.llm-d.ai/admin-port: "8081"
+        dual-pods.llm-d.ai/proxy-port: "8082"
```
MikeSpreitzer: Where is the test of behavior when this annotation is absent?

What I meant when I suggested wiring in the TCP proxy but making its "use" optional is _use by users_, not _configuration by the dual-pods controller_. I suggest making the configuration by the controller unconditional. That removes some code complexity and removes a factor of two from the test matrix.

MikeSpreitzer (commented):

pkg/controller/dual-pods/inference-server.go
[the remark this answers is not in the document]
MikeSpreitzer: I created #728 to fix doHTTP.

cmd/requester/README.md
delavet: done

test/e2e/test-cases.sh
delavet: done

test/e2e/test-cases.sh
delavet: I can open a follow-up PR for this.

test/e2e/mkobjs-openshift.sh
delavet: Done by removing the  `dual-pods.llm-d.ai/proxy-port` annotation. Proxy is always configured and Users can choose to use it.

delavet:

I have skipped the newly added TCP Proxy test case in the e2e tests for release v0.6.5-alpha.1, as this feature is not yet implemented in this version and would otherwise cause the e2e tests to fail. Could you please review this change again?

delavet:
@MikeSpreitzer Rebased onto main. One thing I'd like your call on.

The new "Reverse Proxy Initialization and Forwarding" case exercises a feature that isn't in any published release, and since #739 checks out the PR source *after* installing the release, the case runs against release images and fails.
I had originally pinned a skip to `0.6.5-alpha.1`, which is now stale — 0.6.5 shipped without this feature — so the guard now skips whenever `FMA_RELEASE` is non-empty, with a comment saying to delete it once a release carries the feature.

That guard is the part I'm unsure about. `test-cases.sh` has no notion of which release it is testing today, and I don't want to introduce version awareness there if we'd rather keep it out. The alternative is to drop the guard entirely
and accept that a release test run from a feature branch fails on cases newer than the release. 

Which one do we prefer? Or any better suggestions?

cmd/requester/README.md
```diff
@@ -1,6 +1,56 @@
...
+| Server | Default Port | Purpose |
+|--------|-------------|---------|
+| Probes | 8080 | Readiness relay (`/ready` endpoint) — reflects the health status of the bound inference server |
+| SPI | 8081 | Dual-pods controller interface — exposes GPU info, accepts proxy configuration at `/v1/proxy/config`, and relays readiness through `/v1/become-ready` and `/v1/become-unready` |
+| Proxy | 8082 | TCP reverse proxy listen port — accepts client connections and forwards them to the bound vLLM instance; has no listener at all until configured |
+
+### TCP Proxy
+
+The requester carries a **TCP proxy** that can forward traffic to the bound
+inference server. The dual-pods controller always PUTs the bound server's
```
MikeSpreitzer, line 33: I am not sure what "always" is doing here. Considering all the ways that things can go sideways, I myself would not write "always".

cmd/requester/README.md
```diff
@@ -1,6 +1,56 @@
...
+address to the SPI endpoint at `/v1/proxy/config`, and the proxy starts accepting
```
MikeSpreitzer, line 34: "... and _then_ the proxy starts ..."

cmd/requester/README.md
```diff
@@ -1,6 +1,56 @@
...
+connections on its `--proxy-port`. You can query the current configuration with
```
MikeSpreitzer, line 35: s/You/Control clients/

cmd/requester/README.md
```diff
@@ -1,6 +1,56 @@
-This document shows the steps to exercise the requester and dual-pods controller
-in a local k8s environment with model `ibm-granite/granite-3.3-2b-instruct`
-cached on local PV in the cluster.
+This document describes the requester and shows how to exercise it with the
+dual-pods controller in a local k8s environment, using model
+`ibm-granite/granite-3.3-2b-instruct` cached on local PV in the cluster.
+
```
MikeSpreitzer, line 4: Since the example shown here requests a direct provider, and those are disfavored now that we are focusing on launchers, it is worth saying something explicit about that here, lest this paragraph set the wrong expectation. Perhaps add a sentence like the following.

> The example here uses a direct provider, and this functionality is planned to eventually be deprecated because using launchers is now the focus of our attention.

docs/dual-pods.md
```diff
@@ -828,6 +828,52 @@ The relay of readiness goes as follows.
...
 
+### Requester TCP Proxy
+
+The requester container includes a TCP proxy server that can forward
+inference requests to the actual vLLM instance running in the
+server-providing Pod (typically managed by the launcher). This lets
+clients send requests to the server-requesting Pod without needing to
+know the port that vLLM is listening on.
+
+The dual-pods controller always configures the proxy after binding the
```
MikeSpreitzer, line 839: I would omit "always" here, and add a note about the proxy listening only after being configured. More interesting to readers will be the sequencing wrt the labels queried by the EPP.

test/e2e/test-cases.sh
```diff
@@ -15,6 +15,7 @@
 #
 # Optional environment variables:
 #   FMA_CHART_INSTANCE_NAME  - Helm release name prefix (default: fma)
+#   FMA_RELEASE              - FMA release under test, empty for local code
```
MikeSpreitzer, line 18: I would move this to the required section.

test/e2e/test-cases.sh
```diff
@@ -581,6 +582,55 @@ cheer Successful switching instances in one launcher
...
+
+    kill $pfpid || true
+    pfpid=""
+
+    note "Checking that the proxy port forwards to the inference server; expect port-forward and curl noise"
+    kubectl port-forward -n "$NS" pod/"$req4" 28092:8082 &
+    pfpid=$!
+    sleep 5
+
+    expect "curl -sf http://localhost:28092/health &> /dev/null"
```
MikeSpreitzer, line 624: This could be improved to test an inference request. Can be done in a follow-on PR.

MikeSpreitzer:
> That guard is the part I'm unsure about. test-cases.sh has no notion of which release it is testing today, and I don't want to introduce version awareness there if we'd rather keep it out.

I think that we must make `test-cases.sh` aware of the release that it is testing, because we are evolving the behavior.

MikeSpreitzer (approved):
This is good enough to merge, and continue work in follow-on PRs.

PR #733, opened 2026-08-26
Switch unit tests to new-style coverage data

MikeSpreitzer:
Using technique mentioned in https://github.com/golang/go/issues/51430#issuecomment-1344749218 and https://dustinspecker.com/posts/go-combined-unit-integration-code-coverage/#run-unit-tests-to-collect-coverage

### DIFF
--- a/.github/workflows/go-unit-tests.yml
+++ b/.github/workflows/go-unit-tests.yml
@@ -26,13 +26,13 @@ jobs:
         run: make test
 
       - name: analyze coverage by func
-        run: go tool cover -func=cover/unit-profile.txt -o cover/unit-func.txt
+        run: go tool covdata func -i cover/unit > cover/unit-func.txt
 
-      - name: upload unit test coverage profile
+      - name: upload unit test coverage data
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          path: cover/unit-profile.txt
-          archive: false
+          name: cover-unit-data
+          path: cover/unit
           retention-days: 14
 
       - name: upload unit test func analysis
--- a/Makefile
+++ b/Makefile
@@ -96,9 +96,9 @@ build-populator:
 
 .PHONY: test
 test: ## Run unit tests.
-	mkdir -p cover
+	mkdir -p cover/unit
 	@echo "Running unit tests..."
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -96,9 +96,9 @@ build-populator:
 
 .PHONY: test
 test: ## Run unit tests.
-	mkdir -p cover
+	mkdir -p cover/unit
 	@echo "Running unit tests..."
-	go test $$(go list ./pkg/... | grep -Ev '/pkg/(api|generated|spi)(/|$$)') -coverprofile=cover/unit-profile.txt
+	go test $$(go list ./pkg/... | grep -Ev '/pkg/(api|generated|spi)(/|$$)') -cover -args -test.gocoverdir=$$PWD/cover/unit
```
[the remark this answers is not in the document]
MikeSpreitzer: Non-problems. This runs in a fresh runner. The shell _does_ provide `$PWD`. Using something that can be cut-and-paste into a shell with less editing is better than requiring more editing and specialized knowledge.

.github/workflows/go-unit-tests.yml
```diff
@@ -26,13 +26,13 @@ jobs:
...
       - name: analyze coverage by func
-        run: go tool cover -func=cover/unit-profile.txt -o cover/unit-func.txt
+        run: go tool covdata func -i cover/unit > cover/unit-func.txt
 
-      - name: upload unit test coverage profile
+      - name: upload unit test coverage data
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          path: cover/unit-profile.txt
-          archive: false
```
rubambiza, line 35: Does this imply archive is default `true` now?

rubambiza (approved):
One non-blocking question. Otherwise, LGTM.

.github/workflows/go-unit-tests.yml
MikeSpreitzer: Yes, the default for `archive` is `true`, and we want this change. We are changing from an old-style text file to a new-style directory.

MikeSpreitzer:
A note of caution: `test.gocoverdir` is an _internal_ flag, which I am told means that it is

> Something undocumented and/or not meant for public consumption. Something considered unsupported.

Since it has been there since go 1.20 and I was told about it when asked on the Gophers Slack, I am hopeful that it will not disappear soon. I think that we should advocate for it becoming supported and "exposed" (i.e., available as `go test -gocoverdir`).

PR #734, opened 2026-08-26
Rename DPC test coverage files

MikeSpreitzer:
1. Use "cover-dpc" instead of "dpc-cover", to complement "cover-unit".

2. Inject the release (or "local") into the name, so that the two jobs in the one workflow produce artifacts with different names.

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -743,26 +743,25 @@ jobs:
           kubectl scale -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dpc-coverdata-inspector" --replicas=1
           kubectl wait -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dpc-coverdata-inspector" --for condition=Available
           insp=$(kubectl get -n "$FMA_NAMESPACE" pods -l app.kubernetes.io/component=dpc-coverdata-inspector -o name | sed s:^pod/::)
-          mkdir -p dpc-cover/data
+          mkdir -p cover-dpc/data
           kubectl wait -n "$FMA_NAMESPACE" pod "${insp}" --for condition=Ready
-          kubectl cp -n "$FMA_NAMESPACE" "${insp}:var/coverdata" dpc-cover/data
-          ls -l dpc-cover/data
-          go tool covdata textfmt -i=dpc-cover/data -o dpc-cover/profile.txt
-          go tool cover -func=dpc-cover/profile.txt -o dpc-cover-func.txt
+          kubectl cp -n "$FMA_NAMESPACE" "${insp}:var/coverdata" cover-dpc/data
+          ls -l cover-dpc/data
+          go tool covdata textfmt -i=cover-dpc/data -o cover-dpc/profile.txt
+          go tool cover -func=cover-dpc/profile.txt -o cover-dpc-func.txt
           kubectl scale -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dual-pods-controller" --replicas=1
 
       - name: upload coverage data from dual pods controller
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: dpc-cover-data
-          path: dpc-cover/data
+          name: cover-dpc-data
+          path: cover-dpc/data
           retention-days: 14
 
       - name: upload coverage summary from dual pods controller
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: dpc-cover-func.txt
-          path: dpc-cover-func.txt
+          path: cover-dpc-func.txt
           archive: false
           retention-days: 14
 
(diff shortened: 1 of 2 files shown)


MikeSpreitzer:
https://www.githubstatus.com/ shows problems right now

.github/workflows/launcher-based-e2e-test.yml
```diff
@@ -178,25 +180,23 @@ jobs:
...
-          mkdir -p dpc-cover/data
+          mkdir -p ${COVERDIR}/data
           kubectl wait pod "${insp}" --for condition=Ready
-          kubectl cp "${insp}:var/coverdata" dpc-cover/data
-          ls -l dpc-cover/data
-          go tool covdata textfmt -i=dpc-cover/data -o dpc-cover/profile.txt
-          go tool cover -func=dpc-cover/profile.txt -o dpc-cover-func.txt
+          kubectl cp "${insp}:var/coverdata" ${COVERDIR}/data
+          ls -l ${COVERDIR}/data
+          go tool covdata func -i=${COVERDIR}/data > ${COVERDIR}-func.txt
```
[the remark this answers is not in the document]
MikeSpreitzer: There is no downstream consumer yet.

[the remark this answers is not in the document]
rubambiza: What about the comment on quoting? I'm sure it's something that shellcheck would also flag. I can't tell if Copilot is saying it's covered or should be done.

.github/workflows/launcher-based-e2e-test.yml
```diff
@@ -178,25 +180,23 @@ jobs:
...
 
       - name: upload coverage data from dual pods controller
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: dpc-cover-data
-          path: dpc-cover/data
+          name: cover-dpc-${{ matrix.release || 'local' }}-data
```
rubambiza: How come we're not using/replicating the `COVERDIR` env on [L171](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/734/changes#diff-22be0c1ecbb38c0688605bd9f8f5ead89cc2a02d5cc06f5013157c7276992a74R171)?

.github/workflows/launcher-based-e2e-test.yml
```diff
@@ -178,25 +180,23 @@ jobs:
...
+          path: "cover-dpc-${{ matrix.release || 'local' }}/data"
           retention-days: 14
 
       - name: upload coverage summary from dual pods controller
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: dpc-cover-func.txt
-          path: dpc-cover-func.txt
+          path: "cover-dpc-${{ matrix.release || 'local' }}-func.txt"
```
rubambiza: Same comment as above to ensure COVERDIR is the single source of truth/change.

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -743,26 +743,25 @@ jobs:
           kubectl scale -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dpc-coverdata-inspector" --replicas=1
           kubectl wait -n "$FMA_NAMESPACE" "deploy/${FMA_CHART_INSTANCE_NAME}-dpc-coverdata-inspector" --for condition=Available
           insp=$(kubectl get -n "$FMA_NAMESPACE" pods -l app.kubernetes.io/component=dpc-coverdata-inspector -o name | sed s:^pod/::)
```
rubambiza, line 745: **Nit**: I know it goes beyond the scope of the changes, but I noticed the same pattern of `dpc-cover*` here, too. Though it's not as crucial because the label makes sense when read. More of a consistency issue.

rubambiza (changes requested):
Requested changes to make `COVERDIR` the single source of truth and consistency (non-blocking nit)

.github/workflows/launcher-based-e2e-test.yml
MikeSpreitzer: I understand how an `env` entry on a job step affects a shell script in the `run` of that job step. Do `env` entries affect stuff that is _not_ shell scripting?

.github/workflows/ci-e2e-openshift.yaml
MikeSpreitzer: Unfortunately we have two centers of gravity. There is a lot of stuff that is oriented around the Deployment/Pod object. Then there is stuff oriented around coverage. I do not see any easy/good way to avoid having a transition at a boundary.

MikeSpreitzer:
CI shows successful tests.

.github/workflows/launcher-based-e2e-test.yml
MikeSpreitzer: I did a little RTFM and revision.

.github/workflows/launcher-based-e2e-test.yml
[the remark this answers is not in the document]
MikeSpreitzer: Quoting has been added.

.github/workflows/ci-e2e-openshift.yaml
rubambiza: I think divergence based on the two orientations make sense.

rubambiza (approved):
LGTM.

PR #739, opened 2026-08-27
Test install of release more seriously

MikeSpreitzer:
This PR does a few things about installing a release in `ci-e2e-openshift.yaml`.

- Stop building the requester and launcher images. They are amongst those built for a release.
- Do not checkout the repo until after the install. Use curl-to-bash style of invocation of the install script.

This PR adds missing documentation about envars to `test-cases.sh`.

This PR also adds a mitigation and version diagnostic for the new flake in `kubectl wait crd "$crd_name" --for condition=Established`.

This PR also make the CI workflow that tests E2E on OpenShift assign the requester Pods a high PriorityClass if running in the pok-prod cluster.

This PR also lengthens some timeouts, based on observed slowness of pulling controller images from GHCR. Even the longer timeouts were observed to sometimes not be enough. Something odd is going on wrt that.

This PR also adds some guessed mitigation for mysterious crashes: increasing the ephemeral storage ask from 4.5 to 5 GiB per launcher, and decreasing the GPU memory utilization ask from 0.825 to 0.8. These need further investigation.

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -730,7 +730,7 @@ jobs:
         if: always()
         run: |
           for pod in $(kubectl get pods -n "$FMA_NAMESPACE" -o 'jsonpath={.items[*].metadata.name} ') ; do
-            containers=$(kubectl get pod -n "$FMA_NAMESPACE" "$pod" -o 'jsonpath={.spec.containers[*].name}')
+            containers=$(kubectl get pod -n "$FMA_NAMESPACE" "$pod" -o 'jsonpath={.spec.containers[*].name}' || true)
             for container in $containers ; do
               echo ""
               echo "=== Previous log of $pod (container: $container) ==="
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -584,7 +584,7 @@ jobs:
           FMA_RELEASE: ${{ github.event.inputs.fma_release || '' }}
         run: |
           echo "github.repository='${{ github.repository }}'"
-          echo "$GITHUB_HEAD_REF='$GITHUB_HEAD_REF'"
+          echo "GITHUB_HEAD_REF='$GITHUB_HEAD_REF'"
           echo "github.event_name='${{ github.event_name }}'"
           echo "github.ref_type='${{ github.ref_type }}'"
           echo "github.ref_name='${{ github.ref_name }}'"
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -622,7 +622,7 @@ jobs:
           MKOBJS_SCRIPT: ./test/e2e/mkobjs-openshift.sh
           POLL_LIMIT_SECS: "720"
         run: |
-          if [ "${{ github.event.inputs.run_cluster || 'pod-prod' }}" = "pok-prod" ]
+          if [ "${{ github.event.inputs.run_cluster || 'pok-prod' }}" = "pok-prod" ]
           then export REQUESTER_PRIORITY_CLASS=nightly-llm-d-cicd-critical
                echo Setting REQUESTER_PRIORITY_CLASS
           else echo Not setting REQUESTER_PRIORITY_CLASS
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -334,16 +334,16 @@ jobs:
           FMA_RELEASE: ${{ github.event.inputs.fma_release || '' }}
         run: |
           # Use first 8 chars of the git ref (POSIX-compliant)
-          IMAGE_TAG="ref-$(printf '%s' "$GIT_REF" | cut -c1-8)-$(date +%Y%m%d)"
           reg="${{ github.repository }}"
           CONTAINER_IMG_REG="ghcr.io/${reg@L}"
 
-          echo "Building images with tag: $IMAGE_TAG"
           echo "Registry: $CONTAINER_IMG_REG"
           echo "FMA_RELEASE=<$FMA_RELEASE>"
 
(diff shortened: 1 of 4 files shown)


MikeSpreitzer:
For a successful example of `ci-e2e-openshift.yaml` testing a release, see https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/33144126414 .

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -574,6 +577,28 @@ jobs:
...
+          echo "github.event_name='${{ github.event_name }}'"
+          echo "github.ref_type='${{ github.ref_type }}'"
+          echo "github.ref_name='${{ github.ref_name }}'"
+          echo "github.ref='${{ github.ref }}'"
+          subject=(--release "$FMA_RELEASE")
+          bash <(curl https://raw.githubusercontent.com/${{ github.repository }}/${{ github.ref_name }}/scripts/install-fma.sh) \
+            "${subject[@]}" \
+            --chart-set global.produceCoverdata=true \
+            --namespace "$FMA_NAMESPACE" \
+            --ensure-node-view-cluster-role "${FMA_CHART_INSTANCE_NAME}-node-view" \
```
rubambiza, line 597: Do we need to modify this once the PR for https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/741 lands.

test/e2e/test-cases.sh
```diff
@@ -9,6 +9,10 @@
 #   FMA_NAMESPACE           - Kubernetes namespace to run tests in
 #   MKOBJS_SCRIPT           - path to the mkobjs script to call
 #
+# Conditionally required environment variables (if MKOBJS_SCRIPT needs them):
```
rubambiza, line 12: Reading this isolation, I'm curious where the conditionality is expressed in this script.

rubambiza (commented):
Leaving independent comments, but some changes may be required depending on any dependencies on the upcoming PR for #741 It would be great to have both merged before the release is cut.

test/e2e/test-cases.sh
MikeSpreitzer: It's not in this script, its in the MKOBJS_SCRIPT script.

This is one reason why I do not like passing parameters via environment variables.

.github/workflows/ci-e2e-openshift.yaml
aavarghese: Won't need to. Since this is referencing a CR, and the PR is modifying a CRB we should be good..

MikeSpreitzer: No, because the construction of FMA_CHART_INSTANCE_NAME does **_NOT_** rely on the assumption that there is only one test in a given namespace at a time. The gating and concurrency control in this workflow is very complicated, not obvious that it allows only one run in a given namespace at a time.

test/e2e/test-cases.sh
rubambiza: It seems like `MKOBJS_SCRIPT` _always_ needs them 

https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/452bb7d9c5d48e583294a9f8f758cc12c54e8602/test/e2e/mkobjs-openshift.sh#L51

and this script invokes it.

Moreover, they are not being initialized nor used in this script itself so that they can be passed along. I'd like to suggest either rephrasing from `Conditionally required` -> `Required` or drop them entirely from this script.

rubambiza (changes requested):
Suggested change on the env descriptions or drop them.

test/e2e/test-cases.sh
MikeSpreitzer: No. The `MKOBJS_SCRIPT` environment variable holds the name of a script. Sometimes it holds "mkobjs.sh", which does not read `LAUNCHER_IMAGE` nor `REQUESTER_IMAGE`.

Those two variables ARE needed and passed along when MKOBJS_SCRIPT refers to "mkobjs-openshift.sh" and NOT when MKOBJS_SCRIPT refers to "mkobjs.sh".

rubambiza: Ah, I see. Nvm then.

rubambiza (dismissed):
LGTM.

MikeSpreitzer:
Attempt 1 of the E2E-on-OpenShift workflow, https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/33208874246/attempts/1?pr=739 , failed in test case "Controller Restart State Recovery", at the point of testing after the restart, waiting for the new requester to become ready. It does not even get scheduled. Node pokprod-b93r38s1, launcher launcher-28-20-41-12-m4sgk, requester my-request-28-20-41-12-ldxnx. Here is an excerpt from the requester's status:

```yaml
2026-08-28T21:07:57.6216811Z     conditions:
2026-08-28T21:07:57.6216891Z     - lastProbeTime: null
2026-08-28T21:07:57.6216979Z       lastTransitionTime: "2026-08-28T20:55:24Z"
2026-08-28T21:07:57.6217145Z       message: '0/20 nodes are available: 1 Too many pods, 13 node(s) didn''t match
2026-08-28T21:07:57.6217348Z         Pod''s node affinity/selector, 3 node(s) had untolerated taint {app: llm-d-benchmark-harness},
2026-08-28T21:07:57.6217519Z         3 node(s) had untolerated taint {node-role.kubernetes.io/master: }. preemption:
2026-08-28T21:07:57.6217668Z         0/20 nodes are available: 1 No preemption victims found for incoming pod,
2026-08-28T21:07:57.6217774Z         19 Preemption is not helpful for scheduling.'
2026-08-28T21:07:57.6217856Z       reason: Unschedulable
2026-08-28T21:07:57.6217923Z       status: "False"
2026-08-28T21:07:57.6218001Z       type: PodScheduled
```

MikeSpreitzer:
The force-push to commit 777c4a7 is a rebase onto `main`.

diegocastanibm (approved):
LGTM

PR #742, opened 2026-08-28
fix:  node-view CRB collision

### LINKED ISSUE #741: [Bug]: fma-controllers chart: node-view ClusterRoleBinding name collides across namespaces (breaks concurrent/leftover FMA installs)
### Contact Details

_No response_

### Summary
`charts/fma-controllers/templates/rbac/role-binding.yaml` names the cluster-scoped
ClusterRoleBinding `{{ .Release.Name }}-node-view` — the name has no namespace in it.
Because a ClusterRoleBinding is cluster-scoped, any two helm releases that share a
release name but live in different namespaces resolve to the **same** cluster object,
and Helm's `meta.helm.sh/release-namespace` ownership annotation lets only one release
own it. The second install fails to adopt it.

### Impact / evidence
llm-d-benchmark FMA nightly, [run 33136691819 (2026-08-28)](https://github.com/llm-d/llm-d-benchmark/actions/runs/33136691819/job/98738187978), Standup step:

    Error: Unable to continue with install: ClusterRoleBinding "fma-controllers-node-view"
    in namespace "" exists and cannot be imported into the current release: invalid ownership
    metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal
    "llmdbenchcicdfns-ocp": current value is "braulio-fma-keda"

A prior run in `braulio-fma-keda` (release name `fma-controllers`) created the CRB and its
namespace was torn down without `helm uninstall`, orphaning the cluster-scoped CRB. The
nightly (same release name, namespace `llmdbenchcicdfns-ocp`) then can't adopt it. This
recurs on every concurrent run or every leftover.

### Affected consumers (all fixed by the one chart change)
- llm-d-benchmark FMA nightly — release name `fma-controllers` → `fma-controllers-node-view`
- llm-d kustomize guide nightly (`guides/fast-model-actuation[-keda]`) — release name `fma` → `fma-node-view`
- any manual/scratch install reusing those release names

### Fix
Namespace-qualify the cluster-scoped CRB name so each release owns its own object:

    -  name: "{{ .Release.Name }}-node-view"
    +  name: "{{ .Release.Name }}-node-view-{{ .Release.Namespace }}"

This mirrors the pattern the guide already uses for its EPP bindings
(`${GUIDE_NAME}-epp-auth-delegator-${NAMESPACE}`). After this, `helm uninstall` also
removes each release's own CRB cleanly, so teardown ordering no longer leaves a landmine.

### Version

### Branch name

_No response_

### Commit SHA

_No response_

### Relevant log output

```shell

```

aavarghese:
Fixes #741

### DIFF
--- a/docs/cluster-sharing.md
+++ b/docs/cluster-sharing.md
@@ -145,6 +145,16 @@ object.
 - The Helm chart can optionally create a ClusterRoleBinding for a
   ClusterRole with a given name.
 
+- That ClusterRoleBinding is itself cluster-scoped, so the chart names
+  it after both the Helm release *and its namespace*
+  (`<release name>-node-view-<namespace>`). A name derived from the
+  release name alone would collide across concurrent installs that
+  share a release name but live in different namespaces: the binding is
+  a single cluster object, and Helm's `release-namespace` ownership
+  annotation lets only one release own it, so the second install fails
+  to adopt it. Including the namespace gives each install its own
+  binding, independently owned and removed by its own `helm uninstall`.
+
 - The Helm chart does nothing about creating the ClusterRole for
   reading Node objects.
 
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -489,7 +489,7 @@ jobs:
           # Clean up cluster-scoped resources from previous runs
           echo "Cleaning up cluster-scoped resources..."
           kubectl delete clusterrole        "${FMA_CHART_INSTANCE_NAME}-node-view" --ignore-not-found || true
-          kubectl delete clusterrolebinding "${FMA_CHART_INSTANCE_NAME}-node-view" --ignore-not-found || true
+          kubectl delete clusterrolebinding "${FMA_CHART_INSTANCE_NAME}-node-view-${FMA_NAMESPACE}" --ignore-not-found || true
 
           echo "Cleanup complete"
 
@@ -824,7 +824,7 @@ jobs:
 
           # Delete cluster-scoped stuff for reading Node objects
           kubectl delete clusterrole        "${FMA_CHART_INSTANCE_NAME}-node-view" --ignore-not-found || true
-          kubectl delete clusterrolebinding "${FMA_CHART_INSTANCE_NAME}-node-view" --ignore-not-found || true
+          kubectl delete clusterrolebinding "${FMA_CHART_INSTANCE_NAME}-node-view-${FMA_NAMESPACE}" --ignore-not-found || true
 
           echo "Cleanup complete"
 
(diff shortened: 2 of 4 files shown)


rubambiza (dismissed):
LGTM

MikeSpreitzer (changes requested):
More should be changed than just this one line.

aavarghese:
> More should be changed than just this one line.

Thanks @MikeSpreitzer! Added one more commit 5f03c15a5bbff2e82b6474eb66c811ddca4e63f5. Can you let me know if there's anything else that was missed please?

MikeSpreitzer:
Following is the result of my search for mentions of cluster role bindings.

```
me@mymac llm-d-fast-model-actuation % find * .a* .github .[h-z]* \( \( -name "*.svg" -or -path hack/tools -or -name bin \) -prune \) -or \( -type f -exec grep -i clusterrolebinding \{\} \; -print -exec echo \; \)

kind: ClusterRoleBinding
charts/fma-controllers/templates/rbac/role-binding.yaml

- A ClusterRoleBinding that binds the node-reading ClusterRole to an
- The Helm chart can optionally create a ClusterRoleBinding for a
  makes a ClusterRoleBinding for it. The workflow deletes the Helm
  ClusterRoleBinding.
docs/cluster-sharing.md

if kubectl get clusterrolebinding "$CRB_NAME" &>/dev/null; then
    echo "  Creating EPP metrics RBAC (ClusterRole + ClusterRoleBinding)..."
kind: ClusterRoleBinding
test/e2e/demo-fma-hpa/demo-fma-hpa-ocp.sh

kubectl delete clusterrolebinding fma-node-view --ignore-not-found 2>/dev/null
kubectl delete clusterrolebinding "${NAMESPACE}-${NAMESPACE}-epp" --ignore-not-found 2>/dev/null
test/e2e/demo-fma-hpa/demo-fma-hpa-cleanup.sh

          kubectl delete clusterrolebinding "${FMA_CHART_INSTANCE_NAME}-node-view" --ignore-not-found || true
          kubectl delete clusterrolebinding "${FMA_CHART_INSTANCE_NAME}-node-view" --ignore-not-found || true
.github/workflows/ci-e2e-openshift.yaml
```

docs/cluster-sharing.md
```diff
@@ -145,6 +145,16 @@ object.
...
 
+- That ClusterRoleBinding is itself cluster-scoped, so the chart names
+  it after both the Helm release *and its namespace*
+  (`<release name>-node-view-<namespace>`). A name derived from the
+  release name alone would collide across concurrent installs that
+  share a release name but live in different namespaces: the binding is
+  a single cluster object, and Helm's `release-namespace` ownership
+  annotation lets only one release own it, so the second install fails
+  to adopt it. Including the namespace gives each install its own
+  binding, independently owned and removed by its own `helm uninstall`.
```
MikeSpreitzer, line 156: In my opinion we do not need this detailed explanation here because the reader is expected to understand this much about Kubernetes. It is incongruous to go into this level of detail here and not everywhere else.

MikeSpreitzer (approved):
This is good enough to merge, except possibly due to the flake in `kubectl wait crd "$crd_name" --for condition=Established` that started appearing today. I am pursuing that in #739.

I left a suggestion for improvement.

Issue #720, opened 2026-08-14
[Bug]: flaky crash of vLLM instance during startup


MikeSpreitzer:
### What happened?

In the CI testing of #718, at the point where the PR's head branch pointed to commit c20b621e1fa210c283df7032adfd5248cb2edb2b, the CI-on-OpenShift workflow failed, and was retried, and failed again in the same way (but different test cases: "Per-launcher Instance Cap Enforcement" in attempt 1, "Multiple Instances Share One Launcher" in attempt 2). The failure mode is that a freshly-created vLLM instance never gets to the point of answering `/is_sleeping`; instead, a while after creation of the child process, the launcher detects that the child process is stopped.

Here are the log archives from the 2 attempts at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/31727354299 :

- attempt 1: [logs_86073710745.zip](https://github.com/user-attachments/files/31054711/logs_86073710745.zip)
- attempt 2: [logs_86073710745(1).zip](https://github.com/user-attachments/files/31054718/logs_86073710745.1.zip)

### Version

non-main branch (please specify below)

### Branch name

make-coverdata-optional

### Commit SHA

c20b621e1fa210c283df7032adfd5248cb2edb2b

### Relevant log output

```shell

```

rubambiza:
Thanks for pulling the vLLM dumps + controller logs as they helped in tracing possible correlations, though I found that ours are closer to the startup than this one on the wake-up side.

Having chased several "instance never answers /is_sleeping" failures with Claude on my own FMA benchmarking runs, I think stopped is a shared signature, not a single root cause. The launcher reports stopped purely from `process.is_alive()`, so distinct underlying infra failures collapse into the same surface signal and child's own exit reason is masked.

In particular, I've hit the signature from three separate causes, all before the instance ever sleeps:
- state-change-reflector 403 patching its own pod → readiness never relayed (resolved by llm-d-benchmark#1799);
- CUDA OOM at sampler warmup on a memory-tight config (resolved by llm-d-benchmark#1791);
- Discussed in somewhat scattered thoughts on Slack: an incomplete HF-hub download → OSError: [Errno 122] Disk quota exceeded at weight load (a full model-cache PVC)

[...]

Issue #709, opened 2026-08-11
[Bug]: flaky fail to start vLLM


MikeSpreitzer:
### What happened?

In the CI test of E2E on OpenShift for #645, the first attempt succeeded. For irrelevant reasons, I requested the test to re-run. The second attempt failed. See https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/31436123803/attempts/2?pr=645 . The reason is that during "Test case: Unbound Launcher Deletion Cleanup", the new vLLM instance crashed during startup. Repeatedly: the DPC detected each crash, deleted the instance, and made a replacement. I do not know the cause of the first crash. By the time the job recorded vLLM instance logs, the then-current instance reported the following as the cause.

> (EngineCore pid=3310) ERROR 08-11 13:56:14 [v1/engine/core.py:1195] ValueError: Free memory on device cuda:0 (0.41/79.18 GiB) on startup is less than desired GPU memory utilization (0.85, 67.3 GiB). Decrease GPU memory utilization or reduce GPU memory used by other processes.

If I understand correctly, that is saying that the GPU memory is almost all in use at startup. This is not a small miss, this is a conflict due to something else using all the GPU's memory. Maybe the previous instance not gone yet?

This may be related to #125, which also reported mysterious GPU memory conflict.

### Version

non-main branch

### Branch name

e2e-on-pokprod

### Commit SHA

bd3754ed192b012956278db20b4a8001768095c3

### Relevant log output

```shell

```

MikeSpreitzer:
I may have gotten this again, without being lucky enough to get the full log of the latest vLLM startup. It came from testing #739, specifically https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/33204429268/attempts/1?pr=739 . The test case was "Per-launcher Instance Cap Enforcement". The Node was pokprod-b93r43s1, the launcher launcher-28-19-39-31-fhfm4, the requester my-request-28-19-39-31-74vjt. The dual-pods controller log shows several iterations of requesting that launcher to create an instance and then, a few seconds later, discovering it stopped, deleting the instance and requester, and then starting over again in response to the replacement requester. Eventually the launcher wedges in a state where the kubelet detects health failure. Following is the debug dump of the relevant event.

[...]

PR #747, opened 2026-09-02
Add debugging for flakes

MikeSpreitzer:
This PR adds debugging for test flake mysteries, such as Issues #709, #720, and #746.

This PR also corrects the way that 5XX HTTP responses are produced (formerly they were mistakenly using a method that is only appropriate for 4XX).

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -666,7 +666,7 @@ jobs:
       - name: Dump DCGM_FI_DEV_FB_USED on testnode
         if: always()
         run: |
-          set -x
+          echo Dump DCGM_FI_DEV_FB_USED on testnode
           if [ -r testnode ]
           then
             case "${{ github.event.inputs.run_cluster || 'pok-prod' }}" in
@@ -675,6 +675,7 @@ jobs:
               (*) echo "Impossible run_cluster '${{ github.event.inputs.run_cluster }}'"
                   exit 1;;
             esac
+            echo querying "https://prometheus-k8s-openshift-monitoring.apps.${ingdom}/api/v1/query" about node "$(cat testnode)"
             curl -sSkG \
               -H "Authorization: Bearer $(oc whoami -t)" \
               --data-urlencode "query=DCGM_FI_DEV_FB_USED{Hostname=\"$(cat testnode)\"}" \
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -666,6 +666,7 @@ jobs:
       - name: Dump DCGM_FI_DEV_FB_USED on testnode
         if: always()
         run: |
+          set -x
           if [ -r testnode ]
           then
             case "${{ github.event.inputs.run_cluster || 'pok-prod' }}" in
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -611,6 +611,40 @@ jobs:
         if: always()
         run: kubectl get crd launcherpopulationpolicies.fma.llm-d.ai -o yaml
 
+      - name: Dump DCGM_FI_DEV_FB_USED for every node
+        if: always()
+        run: |
+          case "${{ github.event.inputs.run_cluster || 'pok-prod' }}" in
+            (vllm-d)   ingdom=fmaas-vllm-d.fmaas.res.ibm.com;;
+            (pok-prod) ingdom=pokprod001.ete14.res.ibm.com;;
+            (*) echo "Impossible run_cluster '${{ github.event.inputs.run_cluster }}'"
+                exit 1;;
+          esac
+          for node in $(kubectl get node -l nvidia.com/gpu.present=true -o jsonpath='{.items[*].metadata.name}'); do
+            echo querying "https://prometheus-k8s-openshift-monitoring.apps.${ingdom}/api/v1/query" about DCGM_FI_DEV_FB_USED on node "$node"
+            curl -sSkG \
+              -H "Authorization: Bearer $(oc whoami -t)" \
+              --data-urlencode "query=DCGM_FI_DEV_FB_USED{Hostname=\"$node\"}" \
+              https://prometheus-k8s-openshift-monitoring.apps.${ingdom}/api/v1/query \
+            | jq '.data.result[]' > "DCGM_FI_DEV_FB_USED-$node.json"
+          done
+
+      - name: Dump GPU-using Pods by node
+        if: always()
+        run: |
+            rm pods-* || true
+            kubectl get pods -A -o json \
+              | jq -r '.items[]
+                  | select(.status.phase != "Succeeded" and .status.phase != "Failed")
+                  | ([.spec.containers[]?.resources.limits["nvidia.com/gpu"] // "0" | tonumber] | add) as $gpu_total
+                  | select($gpu_total > 0)
+                  | "\(.spec.nodeName) \(.metadata.namespace)/\(.metadata.name) gpu=\($gpu_total) \(.status.phase)"' \
+              | sort \
+              | while read node rest; do
+                touch "pods-$node"
+                echo "$rest" >> "pods-$node"
+              done
+
       - name: Checkout source before run
         if: always() && (github.event.inputs.fma_release || '') != ''
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -666,21 +700,44 @@ jobs:
       - name: Dump DCGM_FI_DEV_FB_USED on testnode
         if: always()
         run: |
-          echo Dump DCGM_FI_DEV_FB_USED on testnode
+          case "${{ github.event.inputs.run_cluster || 'pok-prod' }}" in
+            (vllm-d)   ingdom=fmaas-vllm-d.fmaas.res.ibm.com;;
+            (pok-prod) ingdom=pokprod001.ete14.res.ibm.com;;
+            (*) echo "Impossible run_cluster '${{ github.event.inputs.run_cluster }}'"
+                exit 1;;
+          esac
           if [ -r testnode ]
           then
-            case "${{ github.event.inputs.run_cluster || 'pok-prod' }}" in
-              (vllm-d)   ingdom=fmaas-vllm-d.fmaas.res.ibm.com;;
-              (pok-prod) ingdom=pokprod001.ete14.res.ibm.com;;
-              (*) echo "Impossible run_cluster '${{ github.event.inputs.run_cluster }}'"
-                  exit 1;;
-            esac
-            echo querying "https://prometheus-k8s-openshift-monitoring.apps.${ingdom}/api/v1/query" about node "$(cat testnode)"
+            echo querying "https://prometheus-k8s-openshift-monitoring.apps.${ingdom}/api/v1/query" about DCGM_FI_DEV_FB_USED on node "$(cat testnode)"
             curl -sSkG \
               -H "Authorization: Bearer $(oc whoami -t)" \
               --data-urlencode "query=DCGM_FI_DEV_FB_USED{Hostname=\"$(cat testnode)\"}" \
               https://prometheus-k8s-openshift-monitoring.apps.${ingdom}/api/v1/query \
             | jq '.data.result[]'
+            echo Data points from start of run follow
+            cat "DCGM_FI_DEV_FB_USED-$(cat testnode).json" || true
+          else echo there is no readable testnode file
+          fi
+
+      - name: Dump GPU-using Pods on test node at start
+        if: always()
+        run: cat "pods-$(cat testnode)" || true
+
+      - name: Dump GPU-using Pods that were not scheduled at start
+        if: always()
+        run: cat pods-null || true
+
+      - name: Dump GPU-using Pods on test node now
+        if: always()
+        run: |
+          if [ -r testnode ]; then
+            kubectl get pods -A -o json \
+              | jq -r '.items[]
+                  | select(.status.phase != "Succeeded" and .status.phase != "Failed")
+                  | select(.spec.nodeName == "'$(cat testnode)'")
+                  | ([.spec.containers[]?.resources.limits["nvidia.com/gpu"] // "0" | tonumber] | add) as $gpu_total
+                  | select($gpu_total > 0)
+                  | "\(.spec.nodeName) \(.metadata.namespace)/\(.metadata.name) gpu=\($gpu_total) \(.status.phase)"' | sort
           else echo there is no readable testnode file
           fi
 
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -567,6 +567,7 @@ jobs:
           subject=(--image-tag "$IMAGE_TAG"
                    --oci-registry "${CONTAINER_IMG_REG,,}"
                    --config-dir config
+                   --chart-set dualPodsController.debugAcceleratorMemory=true
                   )
           scripts/install-fma.sh \
             "${subject[@]}" \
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -666,6 +666,7 @@ jobs:
       - name: Dump DCGM_FI_DEV_FB_USED on testnode
         if: always()
         run: |
+          set -x
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

inference_server/launcher/launcher.py
```diff
@@ -718,7 +741,41 @@ async def delete_all_vllm_instances():
...
+
+@app.get("/v2/gpu-debug")
+async def get_gpu_debug():
+    try:
+        cp = subprocess.run(
+            "nvidia-smi --query-gpu=index,uuid,memory.used --format=csv; nvidia-smi",
+            shell=True,
+            capture_output=True,
+            timeout=20,
+        )
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

inference_server/launcher/launcher.py
```diff
@@ -888,11 +953,35 @@ def vllm_kickoff(vllm_config: VllmConfig, log_file_path: str):
...
+            logger.warning("nvidia-smi not found")
+        except subprocess.TimeoutExpired:
+            logger.warning("nvidia-smi timed out")
+        except Exception as exn:
+            logger.warning(f"Running nvidia-smi threw exception {exn}")
+        else:
+            logger.info(
+                f"Ran nvidia-smi; got returncode={cp.returncode}"
+                f", stdout={cp.stdout}, stderr={cp.stderr}"
+            )
```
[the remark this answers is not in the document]
MikeSpreitzer: This was useful in an earlier draft and this PR is only for debugging.

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1487,19 +1491,31 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
...
+	logger := klog.FromContext(ctx)
+	if lClient != nil {
+		debugData := make(map[string]any)
+		debugErr := lClient.do(ctx, "gpu-debug", http.MethodGet, "/v2/gpu-debug", nil, debugData)
+		if debugErr != nil {
+			logger.V(2).Info("Launcher gpu-debug failed", "err", debugErr)
+		} else {
+			logger.V(2).Info("Launcher did gpu-debug", "debugData", debugData)
+		}
+	}
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1487,19 +1491,31 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
 // only) and records it as awake in serverDat. It does not touch the Pod's
 // sleeping label; callers reconcile that separately (folding it into another
 // Pod update where possible).
-func (ctl *controller) wakeUp(ctx context.Context, serverDat *serverData, requestingPod, providingPod *corev1.Pod, serverPort int32, description string) error {
+// The `lClient` arg may be `nil`, if not is used to do GPU memory debugging
```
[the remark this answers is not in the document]
MikeSpreitzer: revised

MikeSpreitzer:
I am using the CI runs of `.github/workflows/ci-e2e-openshift.yaml` to try to reproduce flakes and get more information. So far, almost all of the runs have completed the E2E suite without failure. The only failure in the suite was during attempt 1 (https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/33675523277/attempts/1?pr=747), which produced the syndrome of Issue #746 ; see more information about it there.

test/e2e/test-cases.sh
```diff
@@ -349,7 +349,7 @@ allocated_gpus=$(kubectl get pods --all-namespaces --field-selector spec.nodeNam
 available_gpus=$(( allocatable_gpus - allocated_gpus ))
 echo "Node $testnode: allocatable_gpus=$allocatable_gpus allocated_gpus=$allocated_gpus available_gpus=$available_gpus"
 
-if (( available_gpus < 1 )); then
+if false && (( available_gpus < 1 )); then
```
waltforme, line 352: Would you help me to understand this change?

MikeSpreitzer: This is because of the adoption of a higher-than-default PriorityClass for the Pods used in these tests, on the pok-prod cluster. The check disabled here does not take Pod PriorityClass into account, and was introducing spurious test failures.

waltforme: Thanks!

inference_server/launcher/launcher.py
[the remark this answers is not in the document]
MikeSpreitzer: This will be developed further in follow-on PRs. No need to gild this lily here.

inference_server/launcher/launcher.py
```diff
@@ -154,6 +161,31 @@ def __init__(self, instance_id):
         self.instance_id = instance_id
 
 
+def dump_process_group(subject: str, pgpid: int, when: str) -> None:
+    members = []
+    for proc in psutil.process_iter(attrs=["pid", "ppid", "name"]):
+        pid = proc.pid
+        pname = proc.name()
+        ppid = proc.ppid()
```
MikeSpreitzer, line 169: It would be the tiniest bit better to have these in the same order as their results are displayed.

inference_server/launcher/launcher.py
```diff
@@ -248,12 +289,41 @@ def stop(self, timeout: int = 10) -> dict:
...
             except ProcessLookupError:
-                pass
+                logger.error(
+                    f"In stop({self.instance_id}), tried to os.killpg({vllm_pid}) "
+                    f"but that threw ProcessLookupError"
+                )
+            except Exception:
+                logger.error(
+                    f"In stop({self.instance_id}), tried to os.killpg({vllm_pid}) "
+                    f"but that threw an unexpected Exception"
```
MikeSpreitzer, line 310: Why didn't I include the exception in the log message? Can be improved later.

inference_server/launcher/launcher.py
```diff
@@ -718,7 +803,42 @@ async def delete_all_vllm_instances():
...
+            "nvidia-smi --query-gpu=index,uuid,memory.used --format=csv; nvidia-smi",
+            shell=True,
+            capture_output=True,
+            text=True,
+            timeout=20,
+        )
+    except subprocess.TimeoutExpired:
+        logger.error("Timeout trying to use nvidia-smi")
+        return PlainTextResponse(
+            status_code=HTTPStatus.GATEWAY_TIMEOUT,
```
MikeSpreitzer, line 824: This is a slightly bogus status code. I used it to debug a problem that I have since fixed. In later PRs we should change both the GATEWAY_TIMEOUT and the NOT_IMPLEMENTED here to the symbol for 500.

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -666,14 +667,15 @@ jobs:
       - name: Dump DCGM_FI_DEV_FB_USED on testnode
         if: always()
         run: |
+          case "${{ github.event.inputs.run_cluster || 'pok-prod' }}" in
+            (vllm-d)   ingdom=fmaas-vllm-d.fmaas.res.ibm.com;;
```
rubambiza, line 671: What does `ingdom` mean? 👀 Closest I could think was a misspelling of `kingdom` to indicate where the run is happening.

cmd/dual-pods-controller/main.go
```diff
@@ -41,20 +41,21 @@ import (
...
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	overrides := &clientcmd.ConfigOverrides{}
 
 	klog.InitFlags(flag.CommandLine)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.CommandLine.IntVar(&config.SleeperLimit, "sleeper-limit", config.SleeperLimit, "limit on number of sleeping inference servers per GPU")
 	pflag.CommandLine.IntVar(&config.NumWorkers, "num-workers", config.NumWorkers, "number of queue worker goroutines")
-	pflag.CommandLine.BoolVar(&debugAccelMemory, "debug-accelerator-memory", debugAccelMemory, "whether to check accelerator memory usage before wake-up")
+	pflag.CommandLine.BoolVar(&config.DebugAcceleratorMemory, "debug-accelerator-memory", config.DebugAcceleratorMemory, "whether to check accelerator memory usage before wake-up")
+	pflag.CommandLine.Int64Var(&config.GlobalAcceleratorSleepingMemoryLimitMiB, "global-accelerator-sleeping-memory-limit", config.GlobalAcceleratorSleepingMemoryLimitMiB, "limit on memory usage allowed at /wake_up, for each GPU regardless of workload")
```
rubambiza, line 58: Perhaps this will become clearer once I go over the other files below, but what happens when the memory usage is exceeded during a `/wake_up` call? I'd wager a guess that the DPC takes some form of action?

pkg/controller/dual-pods/controller.go
```diff
@@ -309,8 +310,8 @@ func (config ControllerConfig) NewController(
 		lcLister:                  fmaInformerFactory.Fma().V1alpha1().LauncherConfigs().Lister(),
 		httpLatencySecsHistograms: httpLatencySecsHistograms.MustCurryWith(prometheus.Labels{"exported_namespace": namespace}),
 		sleeperLimit:              config.SleeperLimit,
-		debugAccelMemory:          config.AcceleratorSleepingMemoryLimitMiB < math.MaxInt32,
```
rubambiza, line 312: I'm curious what prompted the move from `MaxInt32` to `MaxInt64` 🤔

pkg/controller/dual-pods/inference-server.go
```diff
@@ -1487,19 +1501,31 @@ func (ctl *controller) bind(ctx context.Context, serverDat *serverData, requesti
...
-func (ctl *controller) wakeUp(ctx context.Context, serverDat *serverData, requestingPod, providingPod *corev1.Pod, serverPort int32, description string) error {
+// The `lClient` arg may be `nil`, if not then is used to do GPU memory debugging
+func (ctl *controller) wakeUp(ctx context.Context, serverDat *serverData, requestingPod, providingPod *corev1.Pod, serverPort int32, description string, lClient *LauncherClient) error {
 	if ctl.debugAccelMemory {
 		if err := ctl.accelMemoryIsLowEnough(ctx, requestingPod, serverDat); err != nil {
 			return err
 		}
 	}
+	logger := klog.FromContext(ctx)
+	if lClient != nil && ctl.debugAccelMemory && false { // false because it is redundant with ctl.accelMemoryIsLowEnough
```
rubambiza, line 1512: I am having a tough time wrapping my head around the different combinations of these booleans. What is `it`  in the comment referring to? Is it `lClient != nil`, `ctl.debugAccelMemory`, a combination of the two by &&,  or the value of `err != nil` as you allude by `redundant with ctl.accelMemoryIsEnough`?

cmd/dual-pods-controller/main.go
rubambiza: I think I have found the answer [here](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/747/changes#diff-e73665f2825bcfa9218cdd7d4cebbe5fe43d2002854ca88dd4b398483f45551dR2018). Ignore this comment.

inference_server/launcher/launcher.py
```diff
@@ -25,15 +25,22 @@
 import re
 import signal
 import stat
+import subprocess
```
rubambiza, line 28: Feel free to ignore the comment if I'm counting hairs here. The python module loading does take some startup time. I learned in my early days of programming in Java to avoid doing `*` style imports because it brings in _everything_ in the module. Similar to how we do with

 `from fastapi.responses import ...`

below, may I suggest importing only the exact functions (run + TimeoutExpired) in subprocess that we're using?

inference_server/launcher/launcher.py
rubambiza: Is there a reason we are not doing so with a new commit here? 👀

inference_server/launcher/launcher.py
```diff
@@ -239,6 +274,12 @@ def stop(self, timeout: int = 10) -> dict:
             self._cleanup_log_file()
             return self._make_state("not_running")
 
+        vllm_pid = self.process.pid
+        assert vllm_pid is not None
```
rubambiza, line 278: This is a first time I am seeing this. Is there a rationale for doing an assertion in the main module code instead of tests?

inference_server/launcher/launcher.py
```diff
@@ -248,12 +289,41 @@ def stop(self, timeout: int = 10) -> dict:
...
+                os.killpg(vllm_pid, signal.SIGKILL)
+                logger.debug(
+                    f"In stop({self.instance_id}), os.killpg({vllm_pid}) "
+                    f"because the first process.join was not enough"
+                )
             except ProcessLookupError:
-                pass
+                logger.error(
+                    f"In stop({self.instance_id}), tried to os.killpg({vllm_pid}) "
+                    f"but that threw ProcessLookupError"
```
rubambiza, line 305: It can be argued that, unlike other exceptions such as [PermissionError](https://docs.python.org/3/library/exceptions.html#PermissionError) that may map to multiple error numbers, `ProcessLookupError` only maps to [one errno](https://docs.python.org/3/library/errno.html#errno.ESRCH). However, in the spirit of the [comment below](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/747/changes#r3959831227) for including the exception in the log message, I think it would be worth being consistent by doing something like:

`except ProcessLookupError as ple:`

```suggestion
                    f"but that threw {ple.__str__())"
```

or simply

```suggestion
                    f"but that threw {ple.str())"
```

per [BaseException guidance](https://docs.python.org/3/library/exceptions.html#BaseException) (which calls `__str__()` underneath anyway)

so that we get more info.

inference_server/launcher/launcher.py
```diff
@@ -888,11 +1016,35 @@ def vllm_kickoff(vllm_config: VllmConfig, log_file_path: str):
...
+    if debug_gpu_memory:
+        try:
+            cp = subprocess.run(
+                "nvidia-smi --query-gpu=index,uuid,memory.used --format=csv"
+                "; nvidia-smi",
+                shell=True,
+                timeout=20,
+            )
+        except FileNotFoundError:
+            logger.warning("nvidia-smi not found")
```
rubambiza, line 1032: Does this really get raised as a `FileNotFoundError` or does the invocation do some format of file operation to be retrieved later? I'm familiar with it being associated with the error being raised on file operations (open, read, etc). Otherwise, we might never hit the logger warning.

rubambiza (changes requested):
Leaving some dispersed comments with suggestions.

inference_server/launcher/launcher.py
MikeSpreitzer: I was hoping that we could merge what's there now.

.github/workflows/ci-e2e-openshift.yaml
MikeSpreitzer: INGress DOMain

pkg/controller/dual-pods/controller.go
MikeSpreitzer: Achieving consistency with the datatype.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: The "it" in the comment refers to the debug logic controlled by the `if` condition.

inference_server/launcher/launcher.py
MikeSpreitzer: I believe that the tests above ensure that `vllm_pid is not None`, but the pylance checking in my IDE does not understand enough to reach that conclusion. I put the `assert` here to kill the false positive in that linter.

inference_server/launcher/launcher.py
MikeSpreitzer: I would rather do the smpler `{exn}`. In a follow-on PR.

inference_server/launcher/launcher.py
MikeSpreitzer: This is dead code that Copilot already called out and I already said I would revisit later. I envision more sophisticated debug info in the future.

.github/workflows/ci-e2e-openshift.yaml
MikeSpreitzer: We can rename it to `ing_dom` in a future revision.

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: The comment can be clarified in a future revision.

inference_server/launcher/launcher.py
MikeSpreitzer: It's a fair question, but applicable to all the imports of the launcher. Perhaps address this, and launcher startup time more broadly (e.g., caching PYCACHE), in separate PR(s).

diegocastanibm (approved):
LGTM + we need this PR to land ASAP to debug

inference_server/launcher/launcher.py
MikeSpreitzer: Removed in #752

inference_server/launcher/launcher.py
MikeSpreitzer: Revised in #752

pkg/controller/dual-pods/inference-server.go
MikeSpreitzer: Revised in #752

.github/workflows/ci-e2e-openshift.yaml
MikeSpreitzer: Revised in #752

inference_server/launcher/launcher.py
MikeSpreitzer: Revised in #752

inference_server/launcher/launcher.py
MikeSpreitzer: Revised in #752

inference_server/launcher/launcher.py
MikeSpreitzer: Revised in #752

inference_server/launcher/launcher.py
rubambiza: Tracking this in https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/765.

PR #750, opened 2026-09-05
fix: resolve named readiness ports from the server container

git-jxj:
A direct inference-server Pod can use a named readiness-probe port such as `http` mapped to container port 8000. The controller currently treats `http` as port 80 and rejects other names, so it contacts the wrong port or fails to configure the server.

Resolve the name from the inference-server container's declared ports. Numeric ports retain their existing behavior, and unresolved names return an error.

Tests cover numeric ports, `http` and custom names, same-named sidecar ports, and missing names. Successful resolutions reach a local HTTP server. Three cases fail on the upstream baseline and pass with the fix.

Validation:
- `go test ./...` and `make test`
- Launcher pytest suites: 85 passed
- `pre-commit run --all-files`
- Pre-push `go-mod-tidy`; golangci-lint v2.12.2 on Go 1.26.6: 0 issues
- `typos` on both changed files

The default pre-commit Go environment installs Go 1.27.1, which causes the pinned linter to panic while analyzing the standard library. The same failure reproduces on the upstream baseline. Linting with v2.12.2 built and run on Go 1.26.6 passes with the repository configuration.

### DIFF
--- a/pkg/controller/utils/pod-helper.go
+++ b/pkg/controller/utils/pod-helper.go
@@ -103,11 +103,12 @@ func GetInferenceServerContainerIndexAndPort(pod *corev1.Pod) (int, int32, error
 	case intstr.Int:
 		return cIdx, portIOS.IntVal, nil
 	case intstr.String:
-		if portIOS.StrVal == "http" || portIOS.StrVal == "HTTP" {
-			return cIdx, 80, nil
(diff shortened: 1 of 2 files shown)


MikeSpreitzer (approved):
LGTM

Issue #712, opened 2026-08-12
[Bug]: Flaky wedge of launcher


MikeSpreitzer:
### What happened?

In testing of PR #703, before being rebased to pick up the switch to testing on pokprod (i.e., the problem happened while testing on vllm-d), the last few attempts at E2E on OpenShift failed due to a launcher mysteriously becoming unresponsive to health polling from the kubelet. See https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/703#issuecomment-5257225799 et seq, about attempts 9 and 10.

### Version

non-main branch (please specify below)

### Branch name

intro-installer

### Commit SHA

1a0eca0ea90509f37ffaf05ec3863cb2ace887e6

### Relevant log output

```shell

```

MikeSpreitzer:
Regarding attempt 9 at https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/703#issuecomment-5257225799 , here is a little more info. The dump of launcher Pod logs shows only two PUTs (which create a vLLM instance), both successful, same launcher Pod (launcher-11-16-10-22-kxhkp), but two successive runs of the "inference-server" container. The earlier run's first logged timestamp is 2026-08-11 16:17:46,654 --- two seconds after the latest probe failure events (16:17:44). There are no events about killing that container.

MikeSpreitzer:
Upon closer inspect of the logs of attempts 9, 10, and 11 in #703 (see recent comments there), I see that the all failed the same way and it is as follows.

The launcher, for some unknown reason, takes a little longer to start (the first time) than is allowed by the configured liveness probe. However, during the brief lifetime of the first run of the launcher, the dual-pods controller commands creation of a vLLM instance on that launcher. However, before vLLM has had enough time to fully start up, the kubelet kills the launcher container and creates a replacement. The dual pods controller sees a failed startup of vLLM, and reflects that back to deletion of the server-requesting Pod (as designed). The test case fails because it is waiting on readiness of a server-requesting Pod that instead gets deleted.

diegocastanibm:
> Upon closer inspect of the logs of attempts 9, 10, and 11 in [#703](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/703) (see recent comments there), I see that the all failed the same way and it is as follows.
> 
[...]

PR #753, opened 2026-09-09
Improve launcher configuration

MikeSpreitzer:
Add a startup probe, because I observed that sometimes it can take a little over a minute.

Lengthen the timeout in the liveness probe, because Diego said that the procedure for stopping a vLLM instance sometimes waits on something with a five second timeout.

Add CPU and memory to the resources specd for the LauncherConfig is the E2E-on-OpenShift test.

This PR is intended to take some simple practical steps to reduce the incidence Issue #712, for the sake of immediate improvement in development velocity. More thorough work can come in follow-on PRs.


test/e2e/mkobjs-openshift.sh
```diff
@@ -218,6 +218,8 @@ spec:
           resources:
             limits:
               ephemeral-storage: "5Gi"
+              memory: "6Gi"
+              cpu: 4
```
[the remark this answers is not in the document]
MikeSpreitzer: Don't be a worry wart. https://github.com/kubernetes/apimachinery/blob/v0.32.13/pkg/api/resource/quantity.go#L739-L741

MikeSpreitzer:
The following are from https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/34442286024/attempts/3 .

The memory usage (RSS) of the launcher that went highest (launcher-10-06-38-03-nnqgz) and the launcher that remained at the end (launcher-10-06-38-03-hd8vl):

image

---

The CPU utilization. All three that went non-trivial were launchers.

image

MikeSpreitzer:
Here is the log archive from attempt 3:

[logs_93302383140(2).zip](https://github.com/user-attachments/files/32047428/logs_93302383140.2.zip)

MikeSpreitzer:
The first two attempts failed with the syndrome of Issues #709 and #720.

MikeSpreitzer:
@lijiaxi2018

test/e2e/run-launcher-based.sh
```diff
@@ -133,6 +133,9 @@ kubectl wait --for=create sa default
 kubectl wait --for condition=Ready node fmatest-control-plane
 kubectl wait --for condition=Ready node fmatest-worker
 
+# Display nodes to confirm capacity
```
rubambiza, line 136: And this is mostly meant to be consumed by visually inspecting the test/CI logs later, yes?

test/e2e/test-cases.sh
```diff
@@ -168,6 +168,9 @@ spec:
       limits:
         nvidia.com/gpu: "2"
         ephemeral-storage: "10Gi"
+      requests:
+        memory: "11Gi"
```
rubambiza, line 172: Is there a reason we are doing 11Gi here versus [10Gi in `mkobjs-openshift.sh`](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/753/changes#diff-3991ea1a8c5d176b438427f72bc52eba0e84317c40f602c76365dbe64e6a866fR222)? CPU request seems to be the same.

test/e2e/test-cases.sh
```diff
@@ -551,6 +554,9 @@ kubectl get pod $launcher1 -n "$NS"
 # Verify launcher is unbound
 expect '[ "$(kubectl get pod -n '"$NS"' $launcher1 -o jsonpath={.metadata.labels.dual-pods\\.llm-d\\.ai/dual})" == "" ]'
 
+# Launcher should still remain
+kubectl get pod $launcher1 -n "$NS"
```
rubambiza, line 558: Same question as [above](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/pull/753/changes#r3982674198).

rubambiza (commented):
Some minor clarifying questions. I can approve once they are cleared.

test/e2e/run-launcher-based.sh
MikeSpreitzer: Correct.

test/e2e/test-cases.sh
MikeSpreitzer: The test suite goes through a state with launchers existing at once. So the probe should ask for enough memory to do that.

When testing in a `kind` cluster in a GitHub workflow, I observed that the total quantity of CPU available on the worker node is 4. Asking for all of it for one Pod is too much.

This PR is not intended to be the final word on this topic, just take some simple steps to reduce flakes.

rubambiza (approved):
LGTM.

Issue #760, opened 2026-09-18
[Bug]: E2E test got corrupted wrt release vs. --debug-gpu-memory


MikeSpreitzer:
### What happened?

PR #747 changed the E2E test suite to always assume that the launcher supports the `--debug-gpu-memory` argument, which is not true (that argument is introduced in that same PR).

### Version

main (please specify commit below)

### Commit SHA

e6cddcdfd6b9caa6c6c4c9f4ba0a418e0f5bb7ad

### Relevant log output

```shell

```

MikeSpreitzer:
This has been resolved by #761, as demonstrated by https://github.com/llm-d-incubation/llm-d-fast-model-actuation/actions/runs/35752577538 .

PR #761, opened 2026-09-18
Correct testing of releases wrt --debug-gpu-memory

MikeSpreitzer:
This PR is intended to resolve Issue #760, which is that PR #747 introduced a new launcher flag and a bug in the testing of releases (i.e., assuming the released launcher supports the new flag).


test/e2e/test-cases.sh
```diff
@@ -27,6 +27,7 @@ if [ "${FMA_DEBUG:-false}" = "true" ]; then
     set -x
 fi
 
+: "${FMA_RELEASE:?FMA_RELEASE is required}"
```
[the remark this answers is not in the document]
MikeSpreitzer: Fixed, in a simpler way.

test/e2e/mkobjs-openshift.sh
```diff
@@ -200,7 +201,9 @@ spec:
           - /app/launcher.py
           - --host=0.0.0.0
           - --log-level=debug
-          - --debug-gpu-memory
+$(if [ -z "$FMA_RELEASE" ]; then echo "
```
rubambiza, line 204: I did not know one could embed a conditional in the construction of an `EOF`-driven output like this. Pretty cool!

rubambiza (approved):
LGTM.

test/e2e/mkobjs-openshift.sh
MikeSpreitzer: It is a corollary of the fact that `<<EOF` does shell expansion of everything before the terminating `EOF`. On a related note, you may notice that sometimes the EOF is quoted. That is defined to suppress that shell expansion, and sometimes script authors specifically do NOT want shell expansion applied.

rubambiza: Thanks for the lesson.

MikeSpreitzer: To be a little more precise, when I wrote "shell expansion" meant the expansion that Bash does for `"`-quoted strings.

